PR #926, opened 2014-11-19
Add a Raster(Tile, Extent) case class. Rename warp to resample, allow for interpolation methods.

lossyrob:
- Make a case class `Raster` that replaces all the places we have `(tile: Tile, extent: Extent)`.
 - Renames the poorly-named `warp` to `resample`
 - Put methods on `Raster` that delegate to methods on `Tile` that require that `Tile`'s extent.


lossyrob:
@echeipesh ready for review after Travis passes.

This finishes off the work I started a while ago to 
- rename warp to resample
- introduce a Raster case class where Raster = a Tile coupled with an Extent
- move the Mosaic code into the core raster library.

echeipesh:
++1

PR #928, opened 2014-11-22
Shape file reader

johanstenberg92:
A shape file reader to replace GeoTools for good. It's tested but very slow so there is some more work needed to be done.


lossyrob:
Replaced by #982

PR #949, opened 2015-01-05
WIP: Spark catalog - params

echeipesh:
This is pushing catalog work further. There are a couple of mechanical things that fell out of AttributeCatalog:
- AttributeCatalog needs to be able to report all LayerIds it has attributes for, so catalog can be browsed
- AttributeCatalog still needs to be cached because loadTile relies on pulling RasterMetaData from there

A surprising thing about "Catalog can only ever have one LayerId"
- Params are useless on load, since there can only be one set of attributes
 - This is mechanically true for Accumulo, which can only store one set of metadata per catalog
 - This is not true for Hadoop, because a path will give you both the metadata and the tiles
- This implies that Hadoop catalog should always be a flat folder, no subdirectories allowed
 - makes lookup trivial
 - preserves the unique LayerId constraint
 - implies that HadoopCatalog does not need params on save

@lossyrob I would say that HadoopCatalog should not use Params at all, keep it simple as possible. What do you think ?

Issues:
- Should HadoopCatalog have loadTile? Is that a thing that Catalog can do in general?


lossyrob:
@echeipesh this has the commits of #948, I'm assuming that PR should be merged in before these changes?

PR #970, opened 2015-01-29
BatchAccumuloInputFormat

echeipesh:
This InputFormat groups the Ranges by tablet servers where they are located. The corresponding RecordReader will use BatchScanner to read them. The overall effect is to reduce the number of splits to the minimum possible and provide some parallelism in reading them. This reduces the spark overhead of processing each task.


lossyrob:
+:100:

PR #975, opened 2015-01-30
Z-Order space filling curve encoding/decoding/search

echeipesh:
geotrellis-index subproject will contain different methods of indexing and querying multi-dimentional data.

Currently it contains z-order curve functions to encode/decode data, generate z-ranges for an arbitrary multi-dimentional query range and z-divide function that can be used to implement binary search algorithm on multi-diemntional tree described here: http://www.vision-tools.com/h-tropf/multidimensionalrangequery.pdf

### DIFF
--- a/index/src/test/scala/geotrellis/index/zcurve/Z2Spec.scala
+++ b/index/src/test/scala/geotrellis/index/zcurve/Z2Spec.scala
@@ -1,0 +1,62 @@
+package geotrellis.index.zcurve
+
+import org.scalatest._
+
+class Z2Spec extends FunSpec with Matchers {
+  describe("Z2 encoding") {
+    it("interlaces bits"){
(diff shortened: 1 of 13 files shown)


lossyrob:
boo-yah :+1:

index/src/test/scala/geotrellis/index/zcurve/Z2Spec.scala
```diff
@@ -0,0 +1,62 @@
...
+      Z2(Int.MaxValue, Int.MaxValue).decode should equal(Int.MaxValue, Int.MaxValue)
+    }
+
+    it("unapply"){
+      val Z2(x,y) = Z2(3,5)
+      x should be (3)
+      y should be (5)
+    }    
+
+    it("replaces example in Herzog apper"){
```
jnh5y: paper?

PR #979, opened 2015-02-02
Added script for publishing to local maven repository.

lossyrob:


lossyrob:
Publishing to a local maven repository is useful for integration with spark notebook.

PR #956, opened 2015-01-15
Review: Tile compressions

johanstenberg92:
Tile compressions implemented, tested and benched. Would like a code review on this. Thanks

PR #981, opened 2015-02-03
Tile compressions

lossyrob:
An update to #956.


lossyrob:
@echeipesh ready for merge.

lossyrob:
The story with this code is that Johan wrote it in the hopes that we could do a per-element map or combine in RDD's, and only uncompress tiles when we're working with them. That didn't work out, but the compression code is still valid, and I think should be merged in for future use.

echeipesh:
+1

PR #982, opened 2015-02-05
Shape file reader

lossyrob:
Optimization to @johanstenberg92's shape file reader. Replaces #928


vector/src/main/scala/geotrellis/vector/io/shape/reader/ShapeFileReader.scala
```diff
@@ -0,0 +1,73 @@
...
+    */
+  lazy val read: ShapeFile = {
+    val p = if (path.endsWith(".shp")) path.substring(0, path.size - 4) else path
+
+    val (spf, sdf) = Await.result({
+
+      val shapePointFileFuture = 
+        Future {
+          ShapePointFileReader(p + ShapePointFileReader.FileExtension).read
+        }
```
lossyrob: Reading these in parallel sped things up quite a bit.

vector/src/main/scala/geotrellis/vector/io/shape/reader/MultiPolygonReader.scala
```diff
@@ -0,0 +1,161 @@
...
+import com.vividsolutions.jts.{geom => jts}
+import com.vividsolutions.jts.algorithm.CGAlgorithms
+import spire.syntax.cfor._
+
+import scala.collection.mutable
+import java.nio.ByteBuffer
+
+import GeomFactory._
+
+object MultiPolygonReader {
```
lossyrob: This code executed 100's of milliseconds faster than the other code. Probably due to wrapper object creation?

lossyrob:
@johanstenberg92 this now beats the GeoTools reader in the benchmark.

Similar optimizations need to be made for the other geometry types besides MultiPolygon.

lossyrob:
@echeipesh hopefully passes Travis this time, ready for review then.

There's work around doing similar optimizations around the other geometry readers, like the optimization I did to the MultiPolygonReader. Merging this PR in now will let other people work off of this code in master.

vector/src/main/scala/geotrellis/vector/io/shape/reader/ShapeDBaseFileReader.scala
```diff
@@ -0,0 +1,237 @@
...
+    if (path.endsWith(FileExtension)) {
+      val charset = CodePageFileReader(path.take(path.length - 4) +
+        CodePageFileReader.FileExtension).read
+      apply(FileSystem.slurp(path), charset)
+    }
+    else throw new MalformedShapeDBaseFileException(
+      s"Bad file ending (must be $FileExtension)."
+    )
+
+  def apply(bytes: Array[Byte], charset: Charset): ShapeDBaseFileReader =
```
echeipesh: Is this the same deal as GeoTiffReader where we're making a reader just to call .read on it?

echeipesh:
+1, we should tweak the API when we work on GeoTiffReader.

lossyrob:
@echeipesh yeah, the API for actually getting geometries out of this is not done at all. There's more optimization and API work to do to this...perhaps would be a good project for the students. Perhaps this should keep out of master as a feature branch?

PR #985, opened 2015-02-07
Modified build file to run tests in serial

st028:
88 tests failed on osx, ubuntu, and windows. Changed the way parallel execution was set in the tests according to http://www.scala-sbt.org/0.12.4/docs/Detailed-Topics/Testing.html#disable-parallel-execution-of-tests


lossyrob:
This ended up not working every time :/

PR #991, opened 2015-02-08
Added sequence unions for lines, points, polygons and multis

### LINKED ISSUE #893: Add sequence unions on lines sequence and point sequence
Add an implicit class to take a `Seq[Line]` and add a `union` method (or any other collection related operations) so we can do:

```scala
val merged = Seq(l1, l2, l3, l4).union
```

where `l1`, `l2`, ... are of type `Line`

kyeah:
Adds implicit classes for Geometry sequences. Fixes #893.

### DIFF
--- a/vector-benchmark/src/main/scala/geotrellis/vector/benchmark/GeneratorBenchmarks.scala
+++ b/vector-benchmark/src/main/scala/geotrellis/vector/benchmark/GeneratorBenchmarks.scala
@@ -18,8 +18,6 @@ package geotrellis.vector.benchmark
 
 import geotrellis.vector.check.jts.Generators
 
-import geotrellis.vector.MultiLine
-
 import org.scalacheck.{Prop,Test,Gen,Arbitrary}
 
 import com.vividsolutions.jts.geom._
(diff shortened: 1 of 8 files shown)


lossyrob:
@kyeah nice. Is there other sequence operations we should support? e.g. `intersection`. Let's think about all the sequence operations we should support and implement them in this PR.

kyeah:
union is the only operation with a default (no-arg) implementation, but we can add reduction methods to the Multi geometries and introduce appropriate Results:

```scala
lines.map { line => line.jtsGeom }.reduce[Geometry] {
    (acc, line) => acc.intersection(line)
}
```

Same for `difference` and `symDifference`.

lossyrob:
Did you check the JTS codebase to make sure that's the case (e.g. how union
uses
https://github.com/lossyrob/jts/blob/master/jts/java/src/com/vividsolutions/jts/operation/union/CascadedPolygonUnion.java)?
Some operations are hidden in JTS in the operations and algorithms. If so,
then yup the reduction methodology should work.

kyeah:
There aren't any other collection operations from what I can tell.

lossyrob:
Cool. Then the reduction operations make sense. You'll want to write some benchmarks for them to make sure it's reasonably fast (see the vector-benchmark subproject).

kyeah:
Output from reduction op benchmarks:
```
[info] Average for MultiLine intersection: 1.6644260599793175
[info] Average for MultiLine difference: 0.2829056111352762
[info] Average for MultiLine symDifference: 0.014963880288957688
```
Intersections have a high amount of variability between tests - between 1-5 seconds.

lossyrob:
Try it with a caliper benchmark (like the ones in the benchmark subproject). Should decrease the variability.

kyeah:
```
[info]              benchmark     ms linear runtime                                                                  
[info]         MultiLineUnion  69.74 ==                                                                              
[info]  MultiLineIntersection   2.78 =                                                                               
[info]    MultiLineDifference  38.23 =                                                                               
[info] MultiLineSymDifference 386.32 ===============
```

lossyrob:
Awesome. Thanks.

lossyrob:
@kyeah sorry I didn't catch this initially....this functinoality should go in geotrellis.vector, not geotrellis.vector.op

Can you do that change?

lossyrob:
@kyeah actually, @moradology is going to take care of it, so disregard (but look at the code after his PR to see what needed to be done).

PR #999, opened 2015-02-10
Scala 2.11

echeipesh:
Also added cross-compilation for Scala 2.10.

ref: https://github.com/geotrellis/geotrellis/issues/946


echeipesh:
This should really have cross-compile to 2.10. I will need to find a different logger that has consistent API between two versions.

PR #1002, opened 2015-02-12
Add class breaks to stats raster rdd methods

### LINKED ISSUE #977: Add classBreaks to spark stats op
Just like `Tile` has a `classBreaks` op (in `geotrellis.raster.stats`), so should RasterRDD have a classBreaks method.

notthatbreezy:
Closes #977


echeipesh:
+1

PR #1003, opened 2015-02-12
Read/Write Indexed GeoJson  FeatureCollection

### LINKED ISSUE #947: Handle the "id" field of a GeoJSON Feature.
The spec specifies an optional "id" field on Features: http://geojson.org/geojson-spec.html#feature-objects

We need to be able to read these. @echeipesh thoughts?

moradology:
Closes #947

### DIFF
--- a/vector-test/src/test/scala/geotrellis/vector/spec/json/GeoJsonSpec.scala
+++ b/vector-test/src/test/scala/geotrellis/vector/spec/json/GeoJsonSpec.scala
@@ -128,11 +128,18 @@ class GeoJsonSpec extends FlatSpec with Matchers {
   it should "parse geojson with IDs on custom data" in {
     case class DataBox(data: Int)
     implicit val boxFormat = jsonFormat1(DataBox)
-    val json = """{"type":"FeatureCollection","features":[{"type":"Feature","geometry":{"type":"Point","coordinates":[2674010.3642432094,264342.94293908775]},"properties":{ "data" : 291 },"id":"jackson5"},{"type":"Feature","geometry":{"type":"Point","coordinates":[2714118.684319839,263231.3878492862]},"properties": { "data": 1273 },"id":"volcano"}]}"""
+    val json = """{
+                 |  "type":"FeatureCollection",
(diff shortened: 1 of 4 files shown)


vector-test/src/test/scala/geotrellis/vector/spec/json/GeoJsonSpec.scala
```diff
@@ -68,6 +68,7 @@ class GeoJsonSpec extends FlatSpec with Matchers {
 
     val json="""{"type":"FeatureCollection","features":[{"type":"Feature","geometry":{"type":"Point","coordinates":[2674010.3642432094,264342.94293908775]},"properties":{"data":291}},{"type":"Feature","geometry":{"type":"Point","coordinates":[2714118.684319839,263231.3878492862]},"properties":{"data":1273}}]}"""
 
+    println(json.parseGeoJson[JsonFeatureCollection].getAllPointFeatures[DataBox])
```
echeipesh: if you use info() instead of println, ScalaTest is going to include that readout along with the test list rather than the spam scroll

vector/src/main/scala/geotrellis/vector/json/JsonFeatureCollectionMap.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  def toJson: JsValue =
+    JsObject(
+      "type" -> JsString("FeatureCollection"),
+      "features" -> JsArray(buffer)
+    )
+
+  // This helper function is called below to grab the ID field for Map keys
+  private def getFeatureID(js: JsValue): String = {
+    Try(js.asJsObject.getFields("id").head.convertTo[String]) match {
```
echeipesh: I'm feeling Try hate. The normal pattern is to match on Seq from .getFields; Seq() => "id" does not exist

lossyrob:
@moradology this actually passes travis, but because of previous merges just needs to be updated to upstream/master before being merged in.

moradology:
Should be good to go now

PR #1013, opened 2015-02-13
Add markdown documentation and reorganize packages

moradology:
ALSO: use cascaded polygon union where possible as the default union operation.


vector/src/main/scala/geotrellis/vector/MultiPolygon.scala
```diff
@@ -127,10 +128,10 @@ case class MultiPolygon(jtsGeom: jts.MultiPolygon) extends MultiGeometry
   def |(ps: MultiPolygon): TwoDimensionsTwoDimensionsUnionResult =
     union(ps)
   def union(ps: MultiPolygon): TwoDimensionsTwoDimensionsUnionResult =
-    Seq(this, ps).unioned
+    (this.polygons ++ ps.polygons).toSeq.cascadeUnion
```
lossyrob: Perhaps this should just be `unionGeometries` to fit with the other methods? The fact that it is a "cascading" union (not even sure what that means) might not be relevant or easy to remember to client code.

vector/src/main/scala/geotrellis/vector/README.md
```diff
@@ -0,0 +1,78 @@
...
+assert(unwrapPoint(selfIntersection) == Some(myPoint))
+assert(unwrapPoint(nonIntersection) == None)
+```
+
+Beyond the methods which come with any `Geometry` object there are
+implicits in many geotrellis modules which will extend Geometry
+capabilities. For instance, after importing `geotrellis.vector.json._`,
+it becomes possible to call the `toGeoJson` method on any `Geometry`:
+```scala
+import geotrellis.vector.json._
```
echeipesh: it's in `geotrellis.vector.io.json._` now right?

lossyrob: Yeah, good point. I'm going to merge this in and make those corrections.

lossyrob:
Already merged this, but just moving the `geotrellis.vector.json` package wasn't enough, we needed to move the package names (which is a lot of renames).

PR #1014, opened 2015-02-16
RasterRDD union

echeipesh:
Provide a union for RasterRDD given that CellType, Layout and CRS match. An envelope of two raster extents is taken.


echeipesh:
This opens up a question as to what RasterRDD actually is:

**Option 1:**
RasterRDD is an RDD, so its possible for there to be duplicate keys and operations that reduce by keys in single RDD make perfect sense. What happens if we try to save this? Nothing good currently, some tiles will be overwritten.

**Option 2:**
RasterRDD represents a distributed raster with some metadata. This implies that keys must be unique. Non-unique keys are possible in intermediate states, since this is how spark works, but care must be taken for RasterRDD given as a result of any user facing operation to always have a unique keys.

In this case RasterRDD should not extend RDD but only encapsulate it.

PR #1015, opened 2015-02-17
Immutable JTS Wrappers

### LINKED ISSUE #890: Mutable operations in JTS wrapped in geotrellis.vector may effect called geometry.
For example, calling

```
val nl = line.normalize
```

Will have side effects on the original `line.jtsGeom`. This should be immutable.

kyeah:
Makes `normalized` and getter operations in Multi geometries return jtsGeom copies. Fixes #890.

I had another commit to clone jtsGeoms when wrapping them, but I'm not sure we want that.


lossyrob:
Does this PR take care of all instances of mutability in `geotrellis.vector`?

kyeah:
Looked it over again to be sure, but the only places where jts geometries are passed without cloning now are in Geometry/Feature wrapper constructors and implicit casts (e.g. `jtsToLine`, `feature2Geom`). Feel free to correct me.

lossyrob:
@st028 can you review this, and see if there is any other places `geotrellis.vector` may mutate? Thanks.

st028:
Sure thing.

st028:
Just looked through the `geotrellis.vector` project. Looks to me like the wrappers either clone its instance of `jts.Geometry` directly or call other methods that create new `jts.Geometry` instances as Kevin says. This doesn't seem to stop anyone from modifying the object from underneath the wrapper (but maybe there's a scala subtlety that I'm missing, or maybe this is by design), so in a sense the immutability breaks down there:

```
// assume we have some instance of geotrellis.vector.Geometry g
g.jtsGeom.normalize
```

As far as testing, the simplest thing that I can think of are test functions that call the methods of a `geotrellis.vector.Geometry` subclass and then a check to see that the original object is still the same. The test cases would have to be designed in such a way that you wouldn't get false positives, but that shouldn't be too difficult.

lossyrob:
That fact that `jstGeom` is publicly exposed is ok. If the user drops down to the JTS level, they should be aware that our immutability guarantee is no longer valid.

So your test functions seem right, except more along these lines:
 - Copy the jtsGeom out of the target geometry (an actual new copy of that object type, so a new intstance)
 - perform the operation
 - check to see if the jtsGeom and the copy of the jtsGeom are the same (by using jtsGeom.equals(copyJtsGeom), which checks if the geometries are the same, not an object equals).

st028:
Sounds good to me. Should I append these tests to the existing test files, or create new test files?

PR #1018, opened 2015-02-18
Speed up spark tests

lossyrob:
The spark tests were running woefully slow. This PR speeds them up. They run in less than a minute now on my machine. This included optimizing some code, and also pulling the graph stuff into it's own subproject...after reviewing the code and performance a bit more, it's not ready for prime time, so we won't be releasing it wit the 0.10 release.


notthatbreezy:
Not sure if you've tried it or not, but it might help to mess around with the number of executors and  increase the default memory used by each executor - probably somewhere around here https://github.com/geotrellis/geotrellis/blob/master/spark/src/test/scala/geotrellis/spark/OnlyIfCanRunSpark.scala#L30

The default is 512 mb, which really isn't a lot.

notthatbreezy:
The setting would be `spark.executor.memory` http://spark.apache.org/docs/1.2.0/configuration.html

lossyrob:
Good suggestions @notthatbreezy...we actually set the number of executors, but don't want to give to much memory to the unit tests, they have to run on Travis and smaller machines. It was mostly slow code and large test rasters that was causing the slowdown.

@echeipesh this is read for review.

notthatbreezy:
Ah ok - yeah I saw the setting the number of executors. Forgot about the travis limitation, but that makes sense then.

lossyrob:
@notthatbreezy actually the Travis thing was false, I forgot that Travis skips the spark tests since it doesn't have spark installed. But I tried setting the executor memory higher and it's still just as fast.

spark/src/test/scala/geotrellis/spark/op/zonal/summary/MeanSpec.scala
```diff
@@ -38,19 +39,10 @@ class MeanSpec extends FunSpec
           totalExtent.xmin + xd / 2,
           totalExtent.ymin + yd / 2
         )
+        val result = inc.zonalMean(quarterExtent.toPolygon)
+        val expected = inc.stitch.zonalMean(totalExtent, quarterExtent.toPolygon)
```
echeipesh: This is hot

echeipesh:
+1

PR #1025, opened 2015-02-25
Fixed issue with reading small ASCII strings as tag values.

lossyrob:
From http://cool.conservation-us.org/bytopic/imaging/std/tiff4.html

```
If the Value fits within 4 bytes, the Offset is interpreted to contain the
Value instead of pointing to the Value, to save a little time and space.  If
the Value is less than 4 bytes, it is left-justified.  Whether or not it fits
within 4 bytes can be determined by looking at the Type and Length of the
field.
```

This was being respected in cases where we were reading things like short arrays, but not when reading strings. So if you had a GDAL NoData value for a byte raster of say, '255', which is 3 characters, it would error on trying to read that string.


echeipesh:
+1

PR #1027, opened 2015-03-02
Immutability test merge conflict fix

### LINKED ISSUE #890: Mutable operations in JTS wrapped in geotrellis.vector may effect called geometry.

st028:
Fixed merge conflicts for #1015 and opened a new PR per @lossyrob's request. Fixes #890.


lossyrob:
+1

PR #1031, opened 2015-03-04
WIP: Affine transformations

### LINKED ISSUE #799: Implement Affine Transformations in geotrellis-feature
com.vividsolutions.jts.geom.util.AffineTransformation et al has a lot of great functionality for rotating, scaling, shearing, reflecting and translating geometries. These should be implemented in our wrapper library.

Related to #742

st028:
Will fix #799.


vector/src/main/scala/geotrellis/vector/affine/package.scala
```diff
@@ -0,0 +1,56 @@
+package geotrellis.vector
+
+import com.vividsolutions.jts.{geom => jts}
+
+package object affine {
+
+  implicit class AffineTransformationPolygon(geom: Polygon) {
+    import com.vividsolutions.jts.geom.util.AffineTransformation._
```
lossyrob: What's the reasoning for putting the import inside of the class, instead of at the top?

vector/src/main/scala/geotrellis/vector/affine/package.scala
```diff
@@ -0,0 +1,56 @@
+package geotrellis.vector
+
+import com.vividsolutions.jts.{geom => jts}
+
+package object affine {
+
+  implicit class AffineTransformationPolygon(geom: Polygon) {
```
lossyrob: This is right on track. However I'd break it out into a separate trait, so the two files would look like:

```scala
package geotrellis.vector

...

package object affine {
  implicit class PolygonTransformations(val geom: Polygon) extends PolygonTransformationMethods
}
```

```scala
package geotrellis.vector.affine

trait PolygonTransformationMethods extends Any {
  val geom: Polygon

  def reflect(x: Double, y: Double): Polygon = { ... }
  ...  
}
```

This allows for better code organization, as you create each of the implicit classes to add methods to the different geometry types

lossyrob:
This is looking good. I'm going to merge it in and we can work with it in master.

PR #1032, opened 2015-03-08
Embedded Cassandra tests

kyeah:
Swaps environment-dependent Cassandra tests with an embedded cluster. Also updates the spark-cassandra-connector version to support Scala 2.11.


lossyrob:
+1

PR #1036, opened 2015-03-20
Remove SprayJsonSupport references from geotrellis.vector

### LINKED ISSUE #1035: SprayJsonSupport is not needed in geotrellis.vector.io.json
In fact `spray-httpx` is not needed as a dependency. SprayJsonSupport is for marshalling to and from types in the HTTP library: https://github.com/spray/spray/blob/master/spray-httpx/src/main/scala/spray/httpx/SprayJsonSupport.scala.

Since we're not by default using the HTTP library in geotrellis.vector, we shouldn't have this dependency.

This is a matter of removing the dependency and removing the references to SprayJsonSupport. Also the other "Support" trait names should probably be changed.

cc: @echeipesh

lossyrob:
Fixes #1035


echeipesh:
+1, good thing

PR #1037, opened 2015-03-24
Cassandra raster driver

kyeah:
Implements the raster driver for Cassandra catalogs + tests.


spark/src/main/scala/geotrellis/spark/io/cassandra/CassandraDriver.scala
```diff
@@ -0,0 +1,54 @@
...
+  def decode(rdd: RDD[(String, ByteBuffer)], metaData: RasterMetaData): RasterRDD[K]
+  def rowId(id: LayerId, key: K): String
+
+  def load(sc: SparkContext, keyspace: String)(
+    id: LayerId, metaData: RasterMetaData, table: String, filters: FilterSet[K]): RasterRDD[K] = {
+
+    val rdd: CassandraRDD[(String, ByteBuffer)] = 
+      sc.cassandraTable[(String, ByteBuffer)](keyspace, table).select("id", "tile")
+
+    val filteredRDD = applyFilter(rdd, id, filters)
```
lossyrob: This is the crux of why building on of these catalogs is hard...

The current code as you have it loads the whole RDD, and then applies a filter. That means we load *every* tile. Unless you are working with the whole RDD, you are not going to want to pay the price of loading each element into memory...that loads each individual tile into memory at some point. Very slow.

So, we need to find a way to filter things on the Cassandra side. In the Accumulo catalog, we send along Accumulo filter information so that the tiles are filtered by Accumulo before being returned. That way we don't load every tile, only the tiles we want.

So, the question is, how can we only load what we want out of Cassandra?

kyeah: Ah, I was under the impression that Spark does not fetch the tiles from Cassandra immediately. From [this guide](https://github.com/datastax/spark-cassandra-connector/blob/master/doc/3_selection.md#filtering-rows), using `where` instead of `filter` in the `applyFilter` method causes filtering to occur on the Cassandra side. I need to look into the connector and the `union` method more, but I believe Spark doesn't query for these tiles until either the union or some later point in time. 

Unfortunately with Cassandra, there's no built-in support for multi-range queries, so performing multiple 'where' queries is the only way to get that functionality.

echeipesh: I may have to eat my words on this one. Looking closer at it `CassandraRDD.where` just appends the clause and `UnionRDD` does not appear to cause a collect. This may actually be doing what we want. It would be great to ingest a large layer and see the difference in the fetch of a small query extent vs a large one. Also it would be interesting to see how it behaves when there are thousands of where clauses with a large enough query extent.

lossyrob: Ok, thanks for the notes Kevin, I didn't know that about the Cassandra RDDs. I'm going to merge this in and hope to see if the where filters shake out upon testing.

PR #1039, opened 2015-03-24
WIP: Accumulo 1.6

echeipesh:
This PR upgrades to accumulo 1.6.2 for reasons discussed in https://github.com/geotrellis/geotrellis/issues/1026

To be clear we are still using accumulo internals as there is no alternative to getting this functionality in any other way. Notable Accumulo API has shifted to using `Job`  instead of `Configuration` as the parameter container for `AccumuloInputFormat` for its configuration and our code is changed to reflect that.

This should not be merged until https://github.com/hectcastro/vagrant-geotrellis-mesos-spark is updated to deploy Accumulo 1.6.2 and a cluster is verified to work.


echeipesh:
This is going to be merged in through feature/catalog-refactor branch.

PR #1041, opened 2015-03-26
Cleaned up Tiler, changes to Mosaic.

lossyrob:


spark/src/main/scala/geotrellis/spark/ingest/Tiler.scala
```diff
@@ -1,78 +1,66 @@
...
-    // Functions for combine step
-    val createTile =  { (tup: (K, Extent, Tile)) => {
-      val (key, extent, tile) = tup
-      //val metaData = bcMetaData.value
-      val tmsTile = ArrayTile.empty(metaData.cellType, metaData.tileLayout.tileCols, metaData.tileLayout.tileRows)
-      tmsTile.merge(metaData.mapTransform(key), extent, tile)
-    }}
+            val outKey = createKey(inKey, spatialComponent)
+            val tmsTile = ArrayTile.empty(cellType, tileLayout.tileCols, tileLayout.tileRows)
+            tmsTile.merge(mapTransform(outKey), extent, tile)
```
echeipesh: This is a lot cleaner looking but does it sacrifice performance? Now it's blowing up the flatMap into a set of actual tiles rather than references to original tile.

spark/src/main/scala/geotrellis/spark/ingest/Tiler.scala
```diff
@@ -1,78 +1,66 @@
...
+  def apply[T, K: SpatialComponent: ClassTag](
+    getExtent: T=> Extent,
+    createKey: (T, SpatialKey) => K,
+    rdd: RDD[(T, Tile)],
+    mapTransform: MapKeyTransform,
+    cellType: CellType,
+    tileLayout: TileLayout
+  ): RDD[(K, Tile)] =
+    cutTiles(getExtent, createKey, rdd, mapTransform, cellType, tileLayout)
+      .reduceByKey { case (tile1: Tile, tile2: Tile) =>
```
echeipesh: The combiners in combineByKey can be re-used, it's safe to return the reference to leftmost argument to mergeCombiners. Using reduceByKey requires allocation of a new object on every merge. Is it enough to make a difference?  I guess it depends on how many layout tiles the source tile intersects with.

lossyrob:
Eugene and I had discussed the two comments and this PR in person. Ready to merge.

PR #1042, opened 2015-03-27
RDD Caching during Ingest and Catalog save

echeipesh:
This PR addresses the issue of forking collects on RDD. This happens during Ingest, when each RDD is required for saving/sink and for generating the next pyramid level. Also this is an issue during Catalog save, where the catalog needs to collect the raster histogram before saving.

In case of pyramid at most two RDDs are kept in memory at a time, the source and destination for each level, creating a rolling persistence effect.

### DIFF
--- a/project/Version.scala
+++ b/project/Version.scala
@@ -21,7 +21,7 @@ object Version {
     Properties.envOrElse(environmentVariable, default)
 
   val geotrellis  = "0.10.0-SNAPSHOT"
-  val scala       = "2.11.5"
+  val scala       = "2.10.4"
   val geotools    = "11.0"
   val akka        = "2.3.9"
   val spray       = "1.3.2"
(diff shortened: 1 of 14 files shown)


project/Version.scala
```diff
@@ -21,7 +21,7 @@ object Version {
     Properties.envOrElse(environmentVariable, default)
 
   val geotrellis  = "0.10.0-SNAPSHOT"
-  val scala       = "2.11.5"
```
lossyrob: Are we downgrading?

echeipesh: Spark 1.2 introduced support for 2.11 but both my local and cluster spark assembly are still built for 2.10. We will have to figure out what that deployment looks like. We still have the cross build, but maybe let's keep 2.10 as default for now.

lossyrob: Ok, that makes sense. Can you leave a comment in the build file talking about why we are using the lower version as the default?

lossyrob:
+1

PR #1047, opened 2015-04-01
CassandraIngestCommand + tweaks for CassandraInstance

kyeah:
Provides CassandraIngestCommand and moves the metadata tablename into a configuration file, similar to AccumuloInstance. Also implements zoomLevelsFor, which was needed for the gt-admin server.

Tested with climate data using the `ingest-cassandra` and `run-cassandra-server` scripts found [here](https://github.com/kyeah/gt-admin).

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/ingest/CassandraIngestCommand.scala
+++ b/spark/src/main/scala/geotrellis/spark/ingest/CassandraIngestCommand.scala
@@ -25,7 +25,7 @@ class CassandraIngestArgs extends IngestArgs with CassandraArgs {
   @Required var table: String = _
 }
 
-object CassandraIngestCommmand extends ArgMain[CassandraIngestArgs] with Logging {
+object CassandraIngestCommand extends ArgMain[CassandraIngestArgs] with Logging {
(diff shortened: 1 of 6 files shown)


spark/src/main/scala/geotrellis/spark/ingest/CassandraIngestCommand.scala
```diff
@@ -0,0 +1,56 @@
...
+object CassandraIngestCommand extends ArgMain[CassandraIngestArgs] with Logging {
+  def main(args: CassandraIngestArgs): Unit = {
+    System.setProperty("com.sun.media.jai.disableMediaLib", "true")
+
+    implicit val sparkContext = SparkUtils.createSparkContext("Ingest")
+
+    val sparkConf = sparkContext.getConf
+    val hadoopConf = sparkContext.hadoopConfiguration
+    sparkConf.set("spark.cassandra.connection.host", args.host)
+    hadoopConf.set("io.map.index.interval", "1")
```
echeipesh: This should not be needed here, just an option for writing HDFS Map Files

PR #1049, opened 2015-04-01
Add check for gdal

notthatbreezy:
This commit adds a check before running tests that
require Java bindings for GDAL. Additionally, this
commit adds the ability to get the path for the
for the Java bindings from an environment variable


lossyrob:
+1

PR #1051, opened 2015-04-03
Modifications to catalog architecture.

lossyrob:
__NOTE:__ This PR requires that you rebuild the test catalog in spark; to do so, run `./sbt "project spark" test:run`


lossyrob:
@echeipesh I think this is now at a point where it's good to review, and to merge in.

Hadoop has some TODO's around filtering, but we are planning on solving that in another feature.

There's work to do around being able to plugin different indices, but that can be done in a separate PR.

What do you think?

spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloRasterCatalog.scala
```diff
@@ -0,0 +1,76 @@
...
+            histogram = Some(rdd.histogram),
+            keyClass = classTag[K].toString,
+            tileTable = tileTable
+          )
+
+        val minKey = rdd.map(_._1).min
+        val maxKey = rdd.map(_._1).max
+
+        metaDataCatalog.write(layerId, md)
+        attributeStore.write[KeyBounds[K]](layerId, "keyBounds", KeyBounds(minKey, maxKey))
```
echeipesh: It feels like AttributeStore and MetaDataCatalog are basically the same thing, a way to store some extra information about a layer. Could they be merged in the future or is there an intention here?

echeipesh:
I like this, going to start coding against it see where it goes. The best fallout is that this allows injection of Reader[K,V]/Writer[K,V] after instantiation, which is a problem I've ran into before testing a service.

Some issues that still need to be addressed:
- is  `DefaultParams` construct is useful. We use it in hadoop but not accumulo catalog. I like it for both.
- error handling for attempting to write two different key types to the same accumulo table

spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloRasterCatalog.scala
lossyrob: Totally. I was planning on doing this. No reason to have two of these things kicking around.

lossyrob:
I like the idea of `DefaultParams` in Accumulo, but don't want to enforce it's required use. Also it's name should change...Params went the way of the Dodo. Any suggestions?

As far as error handling...not sure a way for us to enforce this. I feel like this is client concern.

echeipesh:
Probably `DefaultLocation` for generic and `DefaultPath` and `DefaultTable` for specific implementations.

PR #1054, opened 2015-04-07
Add KeyIndex, use in Accumulo and HDFS.

lossyrob:
These changes were done in preparation for and along the way of having HadoopRasterCatalog be able to filter RDDs.

KeyIndex[K] is a one way indexing which provides a `Long` index for a key type, as well as a sequence of index ranges for KeyBounds.

Accumulo now saves (key, tileBytes) as it's row value instead of just the tile bytes. Using Kryo serialization to serialize/deserialize those values.

Accumulo now leans on the index instead of defaulting to row_col for SpatialKey and Z3 for SpaceTimeKey.

In both AccumuloRasterCatalog and HadoopRasterCatalog there's a hack to provide the key indexes...it calls into the provider to get the index, which is hard coded to one index type. This is a first step, what should happen next is that the key index type is pluggable and decided on writing the RasterRDD, and when we load we should read the index as an attribute.

Fixed up some stuff in the TestFiles, rewrote the logic a bit to simplify and have the design work with SpaceTimeKeys.


lossyrob:
@echeipesh, I think merging this in at this point makes sense. Can you review and merge at will? This sets up the index read/write task, and will let me continue to focus on the Hadoop filter.

lossyrob:
@kyeah this is important to your work, so I'm going to merge this in. @echeipesh heads up, next time you merge in upstream there might be some breaking changes for the S3 catalog. Both the S3 and Cassandra catalogs should start to use this methodology for indexing.

PR #1063, opened 2015-04-20
WIP: S3 Catalog

echeipesh:
- Make TileReader
- Check if Boundable source on spatial raster can be made more efficiently
- Fine grain filtering on spacetime raster
- Consider cropping prefix length if you know the length of max key that will be written


lossyrob:
Does this add more tests that hit S3? I've already been hung up on the one we have/wanted to figure out a way around that (e.g. it's caused me to not be able to run tests at a coffee shop with spotty service).

echeipesh:
Yep, for the moment. It's not great for several reasons. Looking around for alternatives.

spark/src/main/scala/geotrellis/spark/io/s3/S3RasterCatalog.scala
```diff
@@ -0,0 +1,122 @@
...
+        attributeStore.write[S3LayerMetaData](layerId, "metaData", md)
+
+        val rddWriter = implicitly[RasterRDDWriter[K]]
+          .write(catalogConfig.s3client, bucket, layerPath, index, clobber)(layerId, rdd)
+
+        rdd.unpersist(blocking = false)
+      }
+    }
+
+  // def tileReader[K: TileReaderProvider](layerId: LayerId): Reader[K, Tile] = {
```
lossyrob: This needs to be implemented yeah?

spark/src/main/scala/geotrellis/spark/io/s3/S3RasterCatalog.scala
```diff
@@ -0,0 +1,122 @@
...
+            s"${rootPath}/${catalogConfig.layerDataDir(layerId)}"
+
+        val md = S3LayerMetaData(
+            layerId = layerId,
+            keyClass = classTag[K].toString,
+            rasterMetaData = rdd.metaData,
+            bucket = bucket,
+            key = layerPath)
+
+        // Note: this is a waste to do for spatial raster, we can derive them easily from RasterMetaData
```
lossyrob: This is true, so shouldn't this method be implemented by the particular writer (for spatial or spacetime?). I see this writer is generic, but this would be a good use case for following the `SpatialRasterRDDWriterProvider` pattern like the other catalogs.

spark/src/main/scala/geotrellis/spark/io/s3/RasterRDDReader.scala
```diff
@@ -0,0 +1,151 @@
...
+  def setFilters(filterSet: FilterSet[K], kb: KeyBounds[K], ki: KeyIndex[K]): Seq[(Long, Long)]
+
+  def read(
+    s3client: () => S3Client,
+    layerMetaData: S3LayerMetaData,
+    keyBounds: KeyBounds[K],
+    keyIndex: KeyIndex[K])
+  (layerId: LayerId, filterSet: FilterSet[K])
+  (implicit sc: SparkContext): RasterRDD[K] = {
+    val bucket = layerMetaData.bucket
```
echeipesh: @lossyrob I tried out this change we were talking about as a single commit so it can be reverted, but I think it's much clearer at this level.

lossyrob: Which change are you talking about?

echeipesh: https://github.com/echeipesh/geotrellis/commit/c72d5abb6baf687e7cb539b46550f1f84532b9c3

lossyrob: Got it. That works. I guess we should make them all the same? @moradology is that something you could do as part of the refactor PR?

moradology: Not a problem - I'll make accumulo/hadoop consistent with this.

PR #1064, opened 2015-04-20
WIP Catalog refactor... refactor

moradology:


lossyrob:
If filterSet.isEmpty, we should have a simpler code path that doesn’t actually set any unneeded range filters on Accumulo. https://github.com/geotrellis/geotrellis/blob/feature/catalog-refactor/spark/src/main/scala/geotrellis/spark/io/accumulo/spacetime/SpaceTimeRasterRDDReaderProvider.scala#L123
In Hadoop we do a simpler thing if there are no filters: https://github.com/geotrellis/geotrellis/blob/feature/catalog-refactor/spark/src/main/scala/geotrellis/spark/io/hadoop/spacetime/SpaceTimeRasterRDDReaderProvider.scala#L79

moradology:
It seems to me there's some difficulty with extending that pattern to Accumulo - whereas Hadoop's response to empty filters is key-agnostic, Accumulo has the column filter to apply only for SpaceTime keys. I could break this out but it appears to come at the cost of adding an extra (noop in the case of spatialkeys) function for setting default configuration values.

I'm not necessarily opposed to that idea, as it could well belong in an API. Thoughts?

2
```diff
@@ -0,0 +1,162 @@
...
+import org.apache.hadoop.mapreduce.lib.input.SequenceFileInputFormat
+import org.apache.hadoop.mapreduce.lib.output.{MapFileOutputFormat, SequenceFileOutputFormat}
+import org.apache.hadoop.mapreduce.{JobContext, Job}
+import org.apache.spark._
+import org.apache.spark.rdd._
+import org.apache.spark.SparkContext._
+import scala.reflect._
+import spray.json._
+
+case class HadoopRasterCatalogConfig(
```
echeipesh: Shouldn't this file be 2.scala ? :)

PR #1067, opened 2015-04-23
Call sink when pyramid is false

notthatbreezy:
This fixes a bug where the sink was not being called
when pyramiding is not true. Also adds some logging
to help with debugging later on.


lossyrob:
Already fixed in the catalog-refactor branch :grimacing:

PR #1074, opened 2015-04-29
Single loop for minmax in raster catalog

moradology:
This should just about half the time spent on a minor RDD action.


lossyrob:
Why'd you close this?

moradology:
This change messes up serialization, unfortunately. It could  perhaps be  put elsewhere

moradology:
It looks like just the concrete rddwriters

PR #1075, opened 2015-05-04
Feature/catalog refactor single minmax

moradology:
A serializable, single-pass minimum/maximum key method


echeipesh:
Is this for purposes of generating KeyBounds ?

moradology:
Absolutely; perhaps it belongs on the keybounds companion?

echeipesh:
Unfortunately straight up equality does not give you proper key bounds if the tiles are not uniformly distributed. Imagine a grid of points with lower right point missing. Since ordering is basically row-major addressing you're going to either cut off the last row or right column with these KeyBounds.

Even though the bottom right point is missing, that point is still the upper bound. You need to take a min/max of each individual dimension to possible create a point that is not in the dataset but still bounds it.

This is how I started doing it in S3 catalog:
https://github.com/geotrellis/geotrellis/blob/feature/catalog-refactor/spark/src/main/scala/geotrellis/spark/io/s3/S3RasterCatalog.scala#L73

moradology:
Interesting. Sounds like there's a bug on catalog-refactor (naive min/maxing is what this code was aiming to replace)

echeipesh:
Yep, I'll be addressing it today. I didn't realize the minMax was for this purpose when you were talking about or else I would have mentioned it. My bad on that.

moradology:
I think a typeclass is a good way to implement this, given that it will be extensible in case users want to add other types of dimension.

PR #1076, opened 2015-05-04
Accumulo write strategies - WIP

moradology:


spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloRasterCatalog.scala
```diff
@@ -45,8 +45,12 @@ class AccumuloRasterCatalog(
         implicitly[RasterRDDReader[K]].read(instance, metaData, keyBounds, index)(layerId, filterSet)
       }
     }
-  
-  def writer[K: SpatialComponent: RasterRDDWriter: JsonFormat: Ordering: ClassTag](keyIndexMethod: KeyIndexMethod[K], tileTable: String): Writer[LayerId, RasterRDD[K]] = {
+
+  def writer[K: SpatialComponent: RasterRDDWriter: JsonFormat: Ordering: ClassTag](
+    keyIndexMethod: KeyIndexMethod[K],
+    tileTable: String,
+    strategy: AccumuloWriteStrategy
```
echeipesh: We should default a write strategy here to avoid surprise. I think HDFS should be default for now until we resolve the performance problems on the BatchWriter strategy.

spark/src/main/scala/geotrellis/spark/io/accumulo/RasterRDDWriter.scala
```diff
@@ -60,9 +69,55 @@ trait RasterRDDWriter[K] {
...
+
+        try {
+          HdfsUtils.ensurePathExists(failuresPath, conf)
+
+          kvPairs
+            .saveAsNewAPIHadoopFile(
+              outPath.toString,
+              classOf[Key],
+              classOf[Value],
+              classOf[AccumuloOutputFormat],
```
echeipesh: I think this needs to be `classOf[AccumuloFileOutputFormat]`

spark/src/main/scala/geotrellis/spark/io/accumulo/spacetime/SpaceTimeRasterRDDWriter.scala
```diff
@@ -34,7 +35,7 @@ object SpaceTimeRasterRDDWriter extends RasterRDDWriter[SpaceTimeKey] {
       .reduce { boundable.combine }
   }
 
-  /** TODO: What is the rules about the "num" parameter? */
+  /** TODO: What are the rules about the "num" parameter? */
```
echeipesh: 'num' should be at least as large as the number of tablet servers you have. If they are some small multiple (2x, 3x) it is also ok. Number of splits determines the number of empty tablets and number of concurrent connections that can be opened for ingest to an empty table.

lossyrob:
Is this ready to be merged in? If you wanted to refactor the strategy code a bit that could be a separate PR.

echeipesh:
This looks good to me, I think the refactor to move in the strategy code can be another PR

moradology:
Fire away - as you say, some cleanup will be required, but it is currently what we're running benchmarks with.

PR #1070, opened 2015-04-25
refactor index/ into spark/ and add tests for curves

sandan:
So here's a summary of the changes

The index/ subproject was refactored into the spark/ subproject in the following way:
-The Zcurves in Index/ were moved into spark/io/index/zcurve. Similarly for the tests.
-spark/io/index/ now has additional packages hilbert/, zcurve/, and rowmajor/ each of which contain the SpatialKeyIndex and SpaceTimeKeyIndex implementations for those curves.
-This affected import statements and package names of various other classes such as the KeyIndexMethods in spark/io/index/ and the RasterRDDWriterProvider and RasterRDDReaderProvider in spark/io/accumulo.

-Tests for the ZSpatialKeyIndex and ZSpaceTimeKeyIndex are implemented
-Tests for the HilbertSpatialKeyIndex and HilbertSpaceTimeKeyIndex are implemented
-Tests for the RowMajorSpatialKeyIndex are implemented

PR #1083, opened 2015-05-06
refactor index/ into spark/ and add tests for curves

lossyrob:
This PR merges in upstream changes for #1070.


lossyrob:
@sandan see my email about squashing this commit to get rid of the ghost commits.

PR #1089, opened 2015-05-09
Fix/catalog refactor

sandan:
Contains author credentials and tests for curves


lossyrob:
This is against catalog-refactor, but has to be against master. That's fine, I can create a new PR.

PR #1093, opened 2015-05-13
Fixes to S3 catalog.

lossyrob:
S3 Catalog had some issues:
 - reader was not truncating/padding indexes like writer was
 - some indexes were being skipped over by readKeys logic
 - configuration was not being passed to the AmazonS3Client.

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/io/s3/RasterRDDReader.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/s3/RasterRDDReader.scala
@@ -20,7 +20,7 @@ abstract class RasterRDDReader[K: ClassTag] extends LazyLogging {
   def setFilters(filterSet: FilterSet[K], kb: KeyBounds[K], ki: KeyIndex[K]): Seq[(Long, Long)]
 
   /** Converting lower bound of Range to first Key for Marker */
-  val indexToPath: Long => String
+  val indexToPath: (Long, Int) => String
 
   /* Convert key to index to know if we have rached the range end */
   val pathToIndex: String => Long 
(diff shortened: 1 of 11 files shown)


spark/src/main/scala/geotrellis/spark/io/s3/RasterRDDReader.scala
```diff
@@ -38,36 +40,41 @@ abstract class RasterRDDReader[K: ClassTag] extends LazyLogging {
...
-            listKeys(s3client, bucket, dir, range, itp, pti)            
-              .map{ path => 
-                val is = s3client.getObject(bucket, path).getObjectContent
-                KryoSerializer.deserializeStream[(K, Tile)](is)
-              }
-              .filter{ row => filterSet.includeKey(row._1) }
-          }
-      }
+      sc
+        .parallelize(bins, bins.size)
```
notthatbreezy: I wonder if this is a place where we could take advantage of the built-in s3 and binary records support for spark -- maybe something like

```scala
sc.binaryRecords("s3a://file1,s3a://file2...", numPartitions)
                            .map{ case (filename, bytes) => KryoSerializer.deserializeStream[(K, Tile)](bytes) }
                            .filter{ row => filterSet.includeKey(row._1) }
```

notthatbreezy: Actually, that would probably `binaryFiles` I think -- would prevent from worrying about binning etc, deferring that logic to spark as long as we know the list of keys

echeipesh: Would be interesting to try, you would have to list all the files first on the master. I'm not sure if that would have a significant impact.

echeipesh:
+1

spark/src/main/scala/geotrellis/spark/io/s3/RasterRDDReader.scala
notthatbreezy: Oh that's a good point

PR #1096, opened 2015-05-15
Accumulo SocketWriterStrategy

echeipesh:
- Replace use of AccumuloInputFormat with BatchWriter per partition
While these methods are similar the chosen alternative appears to have more consistent performance. and lower per-task memory requirements.
- Fix split generation to account for gaps in indices generated by raster KeyBounds
The new splits on ingest produce balanced tablets
- Generate only one split per tablet server in cluster
- HdfsWriteStrategy cleans up the import directories if no errors have taken place.


lossyrob:
+1

PR #1102, opened 2015-05-21
FilterSet refactor

echeipesh:
This PR replace the problems with `FilterSet` (incorrect filtering when using multiple filters) by replacing it with `RasterRDDQuery`, which encapsulated logic for filtering and splitting the layer `KeyBounds`.
Along the way some errors in Hadoop file format and Hilbert curve index method have been found and fixed.


lossyrob:
:+1:

PR #1103, opened 2015-05-30
Kriging interpolation

vishalanand:
This framework on Kriging Interpolation introduces Simple, Ordinary, Universal and Geostatistical methods of Kriging; along with Spherical, Gaussian, Exponential, Circular, Wave and Linear semivariogram models.


vishalanand:
Notes on adding a library : Include them for each of the modules in the `Build.scala`
N.B. If the raster module uses a library, the raster-test by default *does not include* the same libraries.

IntelliJ IDEA does not refresh its content of library information by default even after running SBT idea. One has to either clear the Cache and rebuild the project or a lesser known better approach is to 
`Ctrl+Shift+A` -> `reimport` -> `reimport Maven dependencies` and Voila! 
(P.S. searching does not yield very direct answers towards this)

project/Dependencies.scala
```diff
@@ -53,6 +56,7 @@ object Dependencies {
   val sprayJson     = "io.spray"        %% "spray-json"    % "1.3.1"
 
   val apacheMath    = "org.apache.commons" % "commons-math3" % "3.2"
+  val scalaBreeze   = "org.scalanlp"  %%  "breeze"  % "0.11.2"
```
lossyrob: What exactly are we using this dependency for? Is there a way to not use it? It's a big enough dependency that I want to be sure we need it to add it to the project.

raster/src/main/scala/geotrellis/raster/interpolation/KrigingInterpolation.scala
```diff
@@ -0,0 +1,62 @@
...
+case object KrigingOrdinary extends KrigingInterpolationMethod
+case object KrigingUniversal extends KrigingInterpolationMethod
+case object KrigingGeo extends KrigingInterpolationMethod
+
+
+object KrigingInterpolationMethod {
+  val DEFAULT = KrigingOrdinary
+}
+
+abstract class KrigingInterpolation(method: KrigingInterpolationMethod, points: Seq[PointFeature[Int]], re: RasterExtent,
```
lossyrob: I see that you're basing your design on classes a lot, which is a very Object Oriented way of doing this. I recommend you try to think about, when does it make sense to just write something as a function, and when does something really benefit from being a class. I'm not asking for a change, but just want you to keep it in mind as you code. Scala in general and GeoTrellis in particular prefers a functional approach, which often eschews stateful instances for a simple idempotent side effect free transformation (e.g. an `object` method). If the design definitely benefits from having classes, or from having inheritance, then we definitely do it when it makes sense. But preferring functions over classes, and composition over inheritance, is a good practice.

project/Dependencies.scala
moradology: GT already has spire which covers similar terrain

lossyrob: I think Breeze has more Linear Algebra type algorithms in it; that's my impression anyway. But we should definitely see if spire fills any requirements and go with that if we can.

raster/src/main/scala/geotrellis/raster/interpolation/KrigingInterpolation.scala
vishalanand: @lossyrob Thanks for the feedback, I will change my code to reflect a more functional oriented approach. I am about to finish one of the Kriging modules(Simple); and then I will switch to spire to do the matrix computations.

vishalanand:
The Breeze dependencies have been removed, and after trying out the jts library, and spire, but seemed not so friendly for matrix operations. Finally used the Apache Math3 (which was present in GT) library for handling matrices in Kriging Interpolation.
Rigorous testing will be done after the various interpolation types have been implemented.
The next thing is to change code structure to a better functional format, following which Ordinary Kriging is to be implemented.

vishalanand:
Now to implement the semivariogram models apart from the one already implemented [here](https://github.com/geotrellis/geotrellis/blob/master/vector/src/main/scala/geotrellis/vector/Semivariogram.scala) and extend the Kriging interpolations using the new semivariogram models developed, while optimizing the implementations.

raster/src/main/scala/geotrellis/raster/VectorToRaster.scala
```diff
@@ -172,4 +172,11 @@ object VectorToRaster {
     }
     IntArrayTile(array, cols, rows)
   }
+/**
+ *Returns the output of countPoints function, since countPoints() is not a intuitive nomenclature
```
lossyrob: Not good to have comments that refer to code that isn't there...that would be confusing to someone who had never seen the `countPoints` name.

I think `convertToRaster` sounds more like Rasterization, which is it's own set of code (which we have in the `rasterize` package). If `countPoints` isn't a good name, we should figure out a better one.

How about `cellCounts`? or `pointCountPerCell`?

vishalanand:
This commit also includes extension of Simple Kriging to other semivariograms (initially it corresponded to the linear semivariogram alone).

vishalanand:
While polishing Universal Kriging, generating raster test-cases for the Kriging functions using QGIS.

vishalanand:
Cholesky decomposition for linear equations is touted to be twice as fast as LU Decomposition. Thus, switching the LU invocations with Cholesky ones.

vishalanand:
To convert Jacobian matrices generation to individual functions corresponding to the semivariograms; and refactor the Semivariogram.scala

vishalanand:
After refactoring the deprecated function calls in non-linear optimization of variogram fitting; exceptions were encountered in a test. The new implementation is in `SemivaroigramNew.scala` Looking into the source of error.

vishalanand:
Seems introducing trait is a good approach for enabling user defined function optimization fittings

vishalanand:
Comments on the data generation and testing in this [gist file](https://gist.github.com/vishalanand/a18047a2b8879a564af3)

vishalanand:
*  GeoKriging has been implemented.
*  The Semivariograms testing has been done. 
*  The failing tests from the previous commit were due to firstly incorrect function definitions; and secondly due to optimization shifting to local minima rather than the global minima (which has been dealt by providing the starting point as an option to the user). I will inform the related communities about the erroneous function (places where the erroneous circular semivariogram function have been mentioned)
*  The tests cover these aspects, so that the user can sense the problem in the way the data that is being interpolated
*  Empirical Semivraiogram generation and fitting operations are separated into different functions
*  Usage of Cobalt concentration data [UPenn : ESE 502](http://www.seas.upenn.edu/~ese502/lab-content/matlab/cobalt_2.TXT) yielded good Kriging resulsts. 
*  The data is divided into 4:1 parts for training and testing. Some of the interpolation points however caused the inversion of matrices to throw exceptions. This handling is underway. 
*  A wrapper function for related input formats for sequence of data samples is to be written as well (rather than the usage of hardcoded data)

vishalanand:
The last check which has failed is [incorrectly](https://travis-ci.org/geotrellis/geotrellis/jobs/72114591#L102) reported (the Travis could not download sbt from its end). 
```
Downloading sbt launcher 0.13.8:
  From  http://typesafe.artifactoryonline.com/typesafe/ivy-releases/org.scala-sbt/sbt-launch/0.13.8/sbt-launch.jar
    To  ./.lib/0.13.8/sbt-launch.jar
Download failed. Obtain the jar manually and place it at ./.lib/0.13.8/sbt-launch.jar
```
The code compiles and the tests pass successfully.

non:
See the comments [here](https://github.com/non/cats/pull/415) and [here](https://github.com/non/cats/pull/416).

vishalanand:
:+1: Thank!

vishalanand:
The Kriging tests are almost completed (with Nickel deposits and Venice groundwater data). After visualizing them (with ArcMap), I will wrap up the testing and proceed with refactoring of the four Kriging methods.

vishalanand:
The Universal and Geostatistical kriging visualization can be viewed [here](http://imgur.com/a/QsPWo)

vishalanand:
@lossyrob , I have made the changes discussed in the August 5th meeting [link](https://github.com/lossyrob/geotrellis/commit/3e711bec3af954b565c0147de2cf4dd1bd1c2d8a).
Is there some more changes I should make to this PR?

vishalanand:
@lossyrob , I have made the changes discussed in the August 5th meeting [(link)](https://github.com/lossyrob/geotrellis/commit/3e711bec3af954b565c0147de2cf4dd1bd1c2d8a).
Is there some more changes I should make to this PR?

P.S. One of the four tests by Travis died off in between (the tests are passing)

raster/src/main/scala/geotrellis/raster/interpolation/Kriging.scala
```diff
@@ -0,0 +1,128 @@
...
+import geotrellis.vector.Point
+import geotrellis.vector.interpolation.Semivariogram
+
+import org.apache.commons.math3.linear._
+
+import spire.syntax.cfor._
+
+import scala.collection.mutable
+
+trait Kriging{
```
lossyrob: format: space

raster/src/main/scala/geotrellis/raster/interpolation/GeoKriging.scala
```diff
@@ -0,0 +1,224 @@
...
+ * @param model           The [[ModelType]] to be used for prediction
+ */
+class GeoKriging(points: Array[PointFeature[Double]],
+                 attrFunc: (Double, Double) => Array[Double],
+                 bandwidth: Double,
+                 model: ModelType) extends Kriging {
+  /**
+   * Overloaded constructor, for default attribute matrix generation
+   */
+  def this(points: Array[PointFeature[Double]], bandwidth: Double, model: ModelType) {
```
lossyrob: Use companion object `apply`

raster/src/main/scala/geotrellis/raster/interpolation/UniversalKriging.scala
```diff
@@ -0,0 +1,187 @@
...
+ * @param model           The [[ModelType]] to be used for prediction
+ */
+class UniversalKriging(points: Array[PointFeature[Double]],
+                       attrFunc: (Double, Double) => Array[Double],
+                       bandwidth: Double,
+                       model: ModelType) extends Kriging {
+  /**
+   * Overloaded constructor, for default attribute matrix generation
+   */
+  def this(points: Array[PointFeature[Double]], bandwidth: Double, model: ModelType) {
```
lossyrob: Use companion object `apply`

lossyrob:
TODO:
- [ ] Merge in changes from https://github.com/lossyrob/geotrellis/tree/vishal-kriging-aug12
- [ ] Create overloads to all type creations that have sensible defaults. Remember that, because this stuff is actually really to use if you don't understand the math behind it, looking at how defaults are created is a good way to learn how to use the code.
- [ ] Create a README.md inside of the `geotrellis.vector.interpolation` package that gives a brief overview of all the different types of Semivariograms and Krigins. The more info the better, but at least show how to use the code. If you want to talk about the math behind it, that would be awesome! but make sure to explain it to those of us who would have trouble grasping the harder math topics first :) (e.g. what's the difference between a linear and non-linear semivariogram, ELI5)

lossyrob:
@vishalanand TODO:
 [ ] - Merge in changes at https://github.com/vishalanand/geotrellis/pull/2 . These change the Methods to be on `Array[PointFeature[Double]]` instead of `Tile`. It also adds some additional implicit conversion support for things like `Traversable[PointFeature[D]]` where `D <% Double`. 
 [ ] - Get rid of the implicit class in `geotrellis.vector.interpolation`'s package object. It's not necessary. Nor are the methods. I'm going to comment on the files explicitly. We have a `predict` function on the `Kriging` classes that takes an array of points, which is more efficient than this method, and I don't think it would be a better API to implicitly add the method to the array of points rather then constructing the Kriging and calling it on the points (whereas the raster case is a better API).

vector/src/main/scala/geotrellis/vector/interpolation/package.scala
```diff
@@ -0,0 +1,26 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+package geotrellis.vector
+
+package object interpolation {
+
+  implicit class InterpolationVectorExtensions(val pointArr: Array[Point])
```
lossyrob: Remove. See general comment on PR. (you can remove the package.scala file)

vector/src/main/scala/geotrellis/vector/interpolation/GeoKrigingVectorMethods.scala
```diff
@@ -0,0 +1,50 @@
+/*
```
lossyrob: Remove. See general comment on PR. (you can remove the package.scala file)

vector/src/main/scala/geotrellis/vector/interpolation/OrdinaryKrigingVectorMethods.scala
```diff
@@ -0,0 +1,38 @@
...
+*/
+
+package geotrellis.vector.interpolation
+
+import geotrellis.vector.{PointFeature, Point}
+
+/** Methods implicitly added to Array[Point] via the package import.
+  * Contains a method for each overloaded way to create a OrdinaryKriging
+  */
+trait OrdinaryKrigingVectorMethods {
```
lossyrob: Remove. See general comment on PR. (you can remove the package.scala file)

vector/src/main/scala/geotrellis/vector/interpolation/SimpleKrigingVectorMethods.scala
```diff
@@ -0,0 +1,38 @@
...
+*/
+
+package geotrellis.vector.interpolation
+
+import geotrellis.vector.{PointFeature, Point}
+
+/** Methods implicitly added to Array[Point] via the package import.
+  * Contains a method for each overloaded way to create a SimpleKriging
+  */
+trait SimpleKrigingVectorMethods {
```
lossyrob: Remove file. See general comment on PR.

vector/src/main/scala/geotrellis/vector/interpolation/UniversalKrigingVectorMethods.scala
```diff
@@ -0,0 +1,50 @@
...
+*/
+
+package geotrellis.vector.interpolation
+
+import geotrellis.vector.{PointFeature, Point}
+
+/** Methods implicitly added to Array[Point] via the package import.
+  * Contains a method for each overloaded way to create a UniversalKriging
+  */
+trait UniversalKrigingVectorMethods {
```
lossyrob: Remove file. See general comment on PR.

raster/src/main/scala/geotrellis/raster/RasterExtent.scala
```diff
@@ -217,7 +217,7 @@ case class RasterExtent(extent: Extent, cellwidth: Double, cellheight: Double, c
     Extent(xmin, ymin, xmax, ymax)
   }
 
-  /** Adjusts a raster extent so that in can encompass the tile layout.
+  /** Adjusts a raster extent so that it can encompass the tile layout.
```
lossyrob: Thanks :)

raster/src/main/scala/geotrellis/raster/interpolation/README.md
```diff
@@ -0,0 +1,140 @@
...
+    
+The `attrFunc` function is the attribute function, which is used for evaluating non-constant spatial trend structures. Unlike the Simple and Ordinary Kriging models which rely only on the residual values for evaluating the spatial structures, the General Spatial Models may be modelled by the user based on the data (viz. evaluating the beta variable to be used for interpolation).
+
+In case the user does not specify an attribute function, by default the function used is a quadratic trend function for Point(s1, s2) :
+
+```mu(s) = beta0 + beta1*s1 + beta2*s2 + beta3*s1*s1 + beta4*s2*s2 + beta5*s1*s2```
+
+General example of a trend function is : 
+
+```mu(s) = beta0 + Sigma[ beta_j * (s1^n_j) * (s2^m_j) ]```
```
lossyrob: Thanks for the explanation! Could you perhaps give a use-case example of when someone would not want to use the default attribute function, and they would perhaps use a different attribute function? Giving a real-world example of it might help clarify when it *would* be used, even if it's the case that if you don't know how you'd use it then you probably shouldn't :)

lossyrob:
Add attribute for images: Smith, T.E., (2014) Notebook on Spatial Data Analysis [online]  http://www.seas.upenn.edu/~ese502/#notebook

lossyrob:
:100: :+1: :clap::metal::bangbang:

PR #1104, opened 2015-06-01
ZCurve index build by pattern support

pomadchin:
A function `timeToGrid: DateTime => Int` have to be in a project scope, otherwise there would be `Unable to find class` error on workers, trying to read tiles stored in a custom way, but without having the definition of the exact ingest job in the result assembly.

P.S. tried to save code style, but it is possible to reduce code by using wildcards.

### DIFF
--- a/spark/src/test/scala/geotrellis/spark/io/index/zcurve/ZSpaceTimeKeyIndexSpec.scala
+++ b/spark/src/test/scala/geotrellis/spark/io/index/zcurve/ZSpaceTimeKeyIndexSpec.scala
@@ -8,7 +8,6 @@ import scala.collection.immutable.TreeSet
 class ZSpaceTimeKeySpec extends FunSpec with Matchers{
   val y2k = new DateTime(2000, 1, 1, 0, 0)
   val upperBound = 8
-  val ymPattern = "YM"
 
   describe("ZSpaceTimeKey test"){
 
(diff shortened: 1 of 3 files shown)


lossyrob:
Thanks @pomadchin! Could you sign a CLA, found here: http://geotrellis.github.com/files/2014_05_20-GeoTrellis-Open-Source-Contributor-Agreement-Individual.pdf?raw=true. You have to print it out, sign it, scan it, and email it to info@azavea.com . Sorry for the pain, but it keeps us IP clean :).

Also, could you include a small unit test? This would serve as a catch for later if something ended up breaking it, as well as give an example of how to use this feature.

pomadchin:
Sure, no problems! Is it is enough, or to add a catalog test (`write` and `read`)?

spark/src/test/scala/geotrellis/spark/io/index/zcurve/ZSpaceTimeKeyIndexSpec.scala
```diff
@@ -8,6 +8,7 @@ import scala.collection.immutable.TreeSet
 class ZSpaceTimeKeySpec extends FunSpec with Matchers{
   val y2k = new DateTime(2000, 1, 1, 0, 0)
   val upperBound = 8
+  val ymPattern = "YM"
```
lossyrob: Usually we have test-specific vals inside the test itself, can you move this into the the `it` block?

lossyrob:
Great! Just had one small nit-pick, but otherwise this PR is :+1: . I'll have to wait for the CLA before merging though.

lossyrob:
:+1:

PR #1109, opened 2015-06-11
Catalog layer attribute caching

echeipesh:
To read/write a catalog Metadata, KeyBounds and KeyIndex are required. This PR adds the `AttributeCaching` trait that keeps a simple in memory cache for these attributes. The catalogs take care to read/write attribute only through the methods on this trait, keeping it locally consistent.

There are no provision to keep the cache consistent across different instances of a catalog at the moment. As a final escape hatch `.clearCache` method is provided as part of the trait.


lossyrob:
:+1:

PR #1110, opened 2015-06-11
Unify catalog exceptions

echeipesh:
This PR upgrade the use of `sys.error` in the catalog package to named exceptions that should make client code cleaner.


lossyrob:
:+1:

PR #1115, opened 2015-06-12
Feature/catalog read

zifeo:
As discussed on Gitter, here is a way to get the whole raster with an overload of `read` (Accumulo, Hadoop, S3 catalogs). Context bound `Boundable` has been added to original read method for overloading purpose.

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloRasterCatalog.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloRasterCatalog.scala
@@ -1,16 +1,18 @@
 package geotrellis.spark.io.accumulo
 
-import com.typesafe.config.ConfigFactory
-import geotrellis.raster._
 import geotrellis.spark._
 import geotrellis.spark.io._
-import geotrellis.spark.io.index._
 import geotrellis.spark.io.json._
+import geotrellis.spark.io.index._
+import geotrellis.raster._
 import org.apache.hadoop.fs.Path
+
 import org.apache.spark.SparkContext
-import spray.json._
+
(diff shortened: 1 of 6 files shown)


spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloRasterCatalog.scala
```diff
@@ -1,19 +1,16 @@
 package geotrellis.spark.io.accumulo
 
+import com.typesafe.config.ConfigFactory
```
lossyrob: Why are these imports reordered?

If your IDE is doing it, please tell it to stop :) We have a method to the madness of imports, geotrellis ones up top, then other libraries, usually java or scala at the bottom.

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopRasterCatalog.scala
```diff
@@ -1,25 +1,16 @@
...
 import geotrellis.spark.io.index._
-import geotrellis.spark.utils._
-import geotrellis.spark.op.stats._
-import geotrellis.raster._
-import org.apache.hadoop.conf.Configuration
+import geotrellis.spark.io.json._
 import org.apache.hadoop.fs.Path
-import org.apache.hadoop.io.SequenceFile
-import org.apache.hadoop.mapreduce.lib.input.SequenceFileInputFormat
-import org.apache.hadoop.mapreduce.lib.output.{MapFileOutputFormat, SequenceFileOutputFormat}
```
lossyrob: Removing unused imports though is a good thing, thanks.

lossyrob:
:+1:

PR #1118, opened 2015-06-18
Defunct: Feature/Add Parquet Catalog (Writer)

notthatbreezy:
Still very much a WIP, but this is working -- need to to a bunch of clean-up.

- Probably should pull out HadoopAttributeStore or rename it
- Reading isn't a focus yet so there's still a bunch uncommented out code


hectcastro:
:clap:

PR #1120, opened 2015-06-19
WIP: MultiBand RasterRDD

echeipesh:
This PR makes RasterRDD generic on Tile implementation such that the Tile can become a MultiBandTile. The catalogs are adjusted accordingly to allow serialization of generic tile payload.

I considered an alternative implementation of Tile being an abstract type member, as to simplify the type signature, but the resulting code was needlessly complex (not to mention possibility for very bizarre errors).

The current usage pattern would involve using the map methods on the RDD pairs to perform local operations on MultiBandTiles. Accordingly, mapping over MultiBand RDD with `f: MultiBandTile => Tile` will produce single band RDD. As all the appropriate operations are done through method extension, they will become available at that point.


spark/src/main/scala/geotrellis/spark/RasterRDD.scala
```diff
@@ -23,103 +23,60 @@ import org.apache.spark.rdd._
 
 import scala.reflect.ClassTag
 
-class RasterRDD[K: ClassTag](val tileRdd: RDD[(K, Tile)], val metaData: RasterMetaData) extends RDD[(K, Tile)](tileRdd) {
+class RasterRDD[K: ClassTag, T: ClassTag](val tileRdd: RDD[(K, T)], val metaData: RasterMetaData) extends RDD[(K, T)](tileRdd) {
+  type Self = RasterRDD[K, T]
```
lossyrob: Is this used?

echeipesh: This is an artifact, remove it out

spark/src/main/scala/geotrellis/spark/RasterRDDQuery.scala
```diff
@@ -37,9 +37,9 @@ class RasterRDDQuery[K: Boundable](
 /**
  * Wrapper for [[RasterRDDQuery]] that binds it to some function that is able to produce a [[RasterRDD]].
  */
-class BoundRasterRDDQuery[K](query: RasterRDDQuery[K], f: RasterRDDQuery[K] => RasterRDD[K]) {
-  def where[F, T](params: RasterRDDFilter.Expression[F, T])(implicit ev: RasterRDDFilter[K,F,T]): BoundRasterRDDQuery[K] =
+class BoundRasterRDDQuery[K,TileType](query: RasterRDDQuery[K], f: RasterRDDQuery[K] => RasterRDD[K, TileType]) {
```
lossyrob: space after comma

spark/src/main/scala/geotrellis/spark/RasterRDD.scala
```diff
@@ -23,103 +23,60 @@ import org.apache.spark.rdd._
 
 import scala.reflect.ClassTag
 
-class RasterRDD[K: ClassTag](val tileRdd: RDD[(K, Tile)], val metaData: RasterMetaData) extends RDD[(K, Tile)](tileRdd) {
+class RasterRDD[K: ClassTag, T: ClassTag](val tileRdd: RDD[(K, T)], val metaData: RasterMetaData) extends RDD[(K, T)](tileRdd) {
```
lossyrob: I think that `RasterRDD` should be the `Tile` implementation, and not change the type signature. So something like 

```scala
class RasterRDD[K] extends GeoRDD[(K, Tile)](tileRdd)
class MultiBandRasterRDD[K] extends GeoRDD[(K, MultiBandTile)](tileRDD)
```
makes sense, it's less type code for the user to type out and won't break the current API.

lossyrob:
This is WIP right?

echeipesh:
Yes, this is back to WIP after our latest discussion.

PR #1121, opened 2015-06-20
Fixed UInt16GeoTiff reader bug and added a test

alasin:
UInt16GeoTIFFs were not being read correctly and needed to be converted to signed Ints. 

The added test only passes after this fix.


lossyrob:
+1

PR #1122, opened 2015-06-25
WIP: Spark TestKit

echeipesh:
Breaking out testing functionality into it's own project. The intent is to provide some functions to create RasterRDDs for test data and an implementation of Spark Catalog without any persistence dependencies that mimics the behavior of existing catalogs.

Hopefully these will be useful to projects that rely on geotrellis and unit testing.

TODO:
- [x] Provide cleaner version of `OnlyIfSparkInstalled` trait
- [x] Expand the `createSpaceTimeRasterRDD` method
- [ ] Move `RasterRDDMatchers`
- [ ] Test that SBT witchery to avoid cyclical dependency produces a well formed library
- [x] Find out why original `createRasterRDD` produces different tiles when using resample vs merge (extents and tiles sizes match perfectly)


echeipesh:
Master is already able to execute spark tests. The remaining portion is to create spark test kit as a PR.

echeipesh:
replaced by https://github.com/geotrellis/geotrellis/pull/1266

PR #1123, opened 2015-06-29
Feature/resample typeclass

moradology:


lossyrob:
What's going on with these travis build fails?

PR #1126, opened 2015-07-02
Fix s3 catalog rootpath

echeipesh:
Fixes issue with S3 catalog where using empty rootPath would result in path of "/layerName/zoom". The leading slash is interpreted as a folder with empty name by S3 browser.


lossyrob:
+1

PR #1131, opened 2015-07-09
Feature/spatial-ops and various masks

zifeo:
Added masking method (for 1 or more `Geometry`) on `Raster`, `Tile` and `RasterRDD`.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/op/zonal/summary/Histogram.scala
+++ b/raster/src/main/scala/geotrellis/raster/op/zonal/summary/Histogram.scala
@@ -10,15 +10,10 @@ object Histogram extends TileIntersectionHandler[Histogram, Histogram] {
     val PartialTileIntersection(tile, _, polygon) = pt
     val rasterExtent = pt.rasterExtent
     val histogram = FastMapHistogram()
-    Rasterizer.foreachCellByGeometry(polygon, rasterExtent)(
-      new Callback {
-        def apply (col: Int, row: Int) {
-          val z = tile.get(col, row)
-          if (isData(z)) histogram.countItem(z, 1)
-        }
-      }
-    )
-
+    Rasterizer.foreachCellByGeometry(polygon, rasterExtent) { (col: Int, row: Int) =>
+      val z = tile.get(col, row)
+      if (isData(z)) histogram.countItem(z, 1)
+    }
     histogram
   }
 
--- a/raster/src/main/scala/geotrellis/raster/rasterize/polygon/PolygonRasterizer.scala
+++ b/raster/src/main/scala/geotrellis/raster/rasterize/polygon/PolygonRasterizer.scala
@@ -27,7 +27,7 @@ object PolygonRasterizer {
    * Apply a function to each raster cell that intersects with a polygon.
    */
   def foreachCellByPolygon(p: Polygon, re: RasterExtent, includeExterior: Boolean=false)(f: Callback): Unit = 
-    if (p.intersects(re.extent)) { 
+    if (p.intersects(re.extent)) {
 
       val (edges, rowMinOrg, rowMaxOrg) = {
         val TestLineSet(lines, rowMin, rowMax) =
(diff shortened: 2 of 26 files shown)


raster/src/main/scala/geotrellis/raster/op/zonal/summary/Histogram.scala
```diff
@@ -10,15 +10,10 @@ object Histogram extends TileIntersectionHandler[Histogram, Histogram] {
     val PartialTileIntersection(tile, _, polygon) = pt
     val rasterExtent = pt.rasterExtent
     val histogram = FastMapHistogram()
-    Rasterizer.foreachCellByGeometry(polygon, rasterExtent)(
-      new Callback {
```
lossyrob: Nice, glad to see this go!

raster/src/main/scala/geotrellis/raster/rasterize/polygon/PolygonRasterizer.scala
```diff
@@ -43,7 +43,7 @@ object PolygonRasterizer {
 
       var activeEdges: List[Intercept] = List[Intercept]()
 
-      if (rowMaxOrg > 0 && rowMinOrg < re.rows) {
+      if (rowMaxOrg >= 0 && rowMinOrg < re.rows) {
```
lossyrob: Good catch!

spark/src/main/scala/geotrellis/spark/op/local/spatial/LocalSpatialRasterRDDMethods.scala
```diff
@@ -0,0 +1,24 @@
...
+  def mask(geom: Geometry): RasterRDD[K] = {
+    mask(Seq(geom))
+  }
+
+  /** Masks this raster by the given Geometry. */
+  def mask(geoms: Iterable[Geometry]): RasterRDD[K] = {
+    val mapTransform = rasterRDD.metaData.mapTransform
+    rasterRDD.mapPairs { case (k, tile) =>
+      val key = k.spatialComponent
+      val result  = tile.mask(mapTransform(key), geoms)
```
lossyrob:
```scala
val tileExtent = rdd.metaData.mapTransform(key)
val tileGeoms = geoms.map { g => g.intersection(tileExtent).toGeometry }.flatten 
val result = tile.mask(tileExtent, tileGeoms)
```

lossyrob:
+1 after the comment is resolved

spark/src/main/scala/geotrellis/spark/op/local/spatial/LocalSpatialRasterRDDMethods.scala
zifeo: Hum, forgot to mention it on Hangouts, but how do you deal with general intersections as it is not defined on all `Geometry` ? Going through `jtsGeom` seems costly.

lossyrob: Let's go a different route....

Let's make mask work with Polygon and MultiPolygon. That's really the only types we'll be masking with, so it's better to be explicit about that. Extent can be transformed into a Polygon so that takes care of that case.

If we wrote a mask for Polygon and MultiPolygon, we could do one of two things: always wrap the polygon to a MultiPolygon, or write two identical looking versions. Depends on how much code duplication that would accumulate, but I'd almost prefer the explicit duplication (because wrapping always leads to worse performance).

I'm wondering if we want to write a mask also for Extent that doesn't even use the Rasterizer, but uses a RasterExtent to determine what cells to include (which would be waaay faster). I think that might be the way to go.

jwalgran:
This pull would be a big help to the OpenTreeMap project.

zifeo:
It is almost finished, we can merge at this point. One more thing that will be coming in the following days is a safer version in case you got some `TopologyException` due to a high precision.

lossyrob:
Great, thanks Teo!

lossyrob:
:+1:

PR #1136, opened 2015-07-14
Travis Build and version bumps

echeipesh:
Fixing travis build to generate non-zero exit code for all subproject build fails. Also bumping versions for spark and geotools.

### DIFF
--- a/buildall.sh
+++ b/buildall.sh
@@ -1,19 +1,18 @@
 #!/bin/bash
 
-./sbt -J-Xmx2G "project index" test 
-./sbt -J-Xmx2G "project proj4" test 
-./sbt -J-Xmx2G "project vector-test" test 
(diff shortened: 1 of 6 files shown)


echeipesh:
https://github.com/geotrellis/geotrellis/blob/master/engine-test/src/test/scala/geotrellis/engine/RasterSourceSpec.scala#L282

New failure in resample test. Currently the output is:
```
    20   200  2000 20000
    40   400  4000 40000
```
Which sounds correct, center point of a 2x2 region should fall on lower right cell as lower and left cell boundaries are exclusive on the other cells.

lossyrob:
Can you fix as part of the PR?

echeipesh:
Yes, just wanted to make sure that it looks like a bad test to somebody else.

PR #1139, opened 2015-07-16
Added JsonReader and JsonWriter for Feature[D]

lossyrob:
Before this, if we had the situation where we'd like to read a GeoJSON feature collection data property into a class but had no interest in ever writing it through GeoJSON, we'd have to do something like this:

```scala
  case class NeighborhoodData(name: String, listName: String)

  implicit object NeighborhoodDataFormat extends JsonFormat[NeighborhoodData] {
    def write(obj: NeighborhoodData): JsValue = ???

    def read(value: JsValue): NeighborhoodData = 
      value.asJsObject.getFields("name", "listname") match {
        case Seq(JsString(name), JsString(listName)) =>
          NeighborhoodData(name, listName)
        case _ => throw new DeserializationException("Couldn't read neighborhood data")
      }
  }

  def fromJson(path: String): Seq[MultiPolygonFeature[String]] =
      Source.fromFile(path)
        .getLines
        .mkString
        .parseGeoJson[JsonFeatureCollection]
        .getAllMultiPolygonFeatures[NeighborhoodData]
        .map { feature => MultiPolygonFeature(feature.geom, feature.data.name) }
```

with this change, we can just implement the reader if we want to read, and the writer if we want to write.

The `JsonFormat` is still defined specifically because spray's implicit marsheller requires a JsonFormat, not a Reader or a Writer.

### DIFF
--- a/buildall.sh
+++ b/buildall.sh
@@ -1,12 +1,12 @@
 #!/bin/bash
 
-./sbt -J-Xmx2G "project spark" test  || { exit 1; }
-./sbt -J-Xmx2G "project index" test || { exit 1; }
-./sbt -J-Xmx2G "project raster-test" test || { exit 1; } 
-./sbt -J-Xmx2G "project vector-test" test || { exit 1; }
-./sbt -J-Xmx2G "project engine-test" test || { exit 1; } 
-./sbt -J-Xmx2G "project benchmark" compile || { exit 1; } 
 ./sbt -J-Xmx2G "project proj4" test || { exit 1; } 
+./sbt -J-Xmx2G "project vector-test" test || { exit 1; }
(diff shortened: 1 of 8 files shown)


buildall.sh
```diff
@@ -1,12 +1,12 @@
 #!/bin/bash
 
-./sbt -J-Xmx2G "project spark" test  || { exit 1; }
-./sbt -J-Xmx2G "project index" test || { exit 1; }
-./sbt -J-Xmx2G "project raster-test" test || { exit 1; } 
+./sbt -J-Xmx2G "project proj4" test || { exit 1; }
```
lossyrob: reordered project to match the dependency graph a bit more.

PR #1140, opened 2015-07-17
Refactored Feature to include Geometry type as type param.

lossyrob:
This PR changes `abstract class Feature[D] { type G <: Geometry }` into `case class Feature[G <: Geometry, D](...)`. This fell out of me wanting to write a simple `mapData` method, and the current type architecture not allowing me to. This mostly affects internal code; the API should remain the same as long as users use things like `PointFeature[D]`, etc.

### DIFF
--- a/vector/src/main/scala/geotrellis/vector/reproject/package.scala
+++ b/vector/src/main/scala/geotrellis/vector/reproject/package.scala
@@ -3,6 +3,8 @@ package geotrellis.vector
 import geotrellis.proj4._
 
 package object reproject {
+  type DI = DummyImplicit
+
   object Reproject {
     def apply(t: (Double, Double), src: CRS, dest: CRS): (Double, Double) =
       apply(t, Transform(src, dest))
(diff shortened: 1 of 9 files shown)


lossyrob:
@echeipesh please review.

vector/src/main/scala/geotrellis/vector/reproject/package.scala
```diff
@@ -3,6 +3,8 @@ package geotrellis.vector
 import geotrellis.proj4._
 
 package object reproject {
+  type DI = DummyImplicit
```
echeipesh: Looks like this was needed when trying an overload but not with renames of apply?

echeipesh:
+1, shame I didn't think to implement maps to see this.

vector/src/main/scala/geotrellis/vector/reproject/package.scala
lossyrob: Good call!

PR #1143, opened 2015-07-22
Set a more precise WebMercator world extent.

lossyrob:
This is what gdal2tiles uses, see here for an explanation: https://github.com/pramsey/gdal2tilesp/blob/master/gdal2tilesp.py#L141


echeipesh:
+1

Issue #1142, opened 2015-07-20
Invalid GeoJSON coming from serializing multiple polygons


lossyrob:
```scala
val p1: Polygon = ???
val p2: Polygon = ???

Seq(p1, p2).toGeoJson
```
produces something like

```javascript
{"type":"FeatureCollection","features":[{"type":"Polygon","coordinates":[[[-9465961.876490302,29588072.039175928],[-9465961.876490302,29613382.62261249],[-9461069.90681594,29613382.62261249],[-9461069.90681594,29588072.039175928],[-9465961.876490302,29588072.039175928]]]},{"type":"Polygon","coordinates":[[[-9470853.552646479,5713820.738373496],[-9470853.552646479,5723604.677993998],[-9461069.613025976,5723604.677993998],[-9461069.613025976,5713820.738373496],[-9470853.552646479,5713820.738373496]]]}]} 
```

Which doesn't pass geojsonlint.com

lossyrob:
Also, `Seq[Feature[D]]` doesn't have an implicit `toGeoJson`

Also, Seq[Extent] should get an implicit `toGeoJson`

lossyrob:
This implicit:

https://github.com/geotrellis/geotrellis/blob/master/vector/src/main/scala/geotrellis/vector/io/json/package.scala#L20

Should actually look more like

```scala
implicit class FeaturesToGeoJson[D](features: Traversable[Feature[D]]) {
    def toGeoJson(implicit valueJsonWriter: JsonWriter[D]): String = {
      JsonFeatureCollection(features).toJson.compactPrint
    }
  }
```

point is as a user I would like to have a clearer error message when I try to do a `Seq(feature1, feature2).toGeoJson` and the reason it's failing is because the data type doesn't have a JsonWriter found.

allixender:
I guess there needs to be a proper differentiation between Feature (geometry with properties) and Geometry (really just geometry):
Polygon is only a Geometry. The single json Geometries generated are correct, but their Sequence would be a GeometryCollection, so the embedding of a sequence of polygon geometries into a FeatureCollection is wrong. Only PolygonFeature (jtsgeom, data) and a Sequence of those thereof should go into a FeatureCollection, a Sequence of Polygon (jtsgeom) should be embedded into a GeometryCollection.

So apart of the coordinates not plotting nicely, this validates:

```json
{
    "type": "GeometryCollection",
    "geometries": [
        {
        "type": "Polygon",
        "coordinates": [
[...]

PR #1144, opened 2015-07-24
#1142 changed to GeometryCollection, fix vector-test

allixender:
#1142 

tested with points and polygons for val json = Seq(p1, p2).toGeoJson and json.parseGeoJson[GeometryCollection] and json.parseGeoJson[GeometryCollection].polygons

### DIFF
--- a/vector/src/main/scala/geotrellis/vector/io/json/JsonFeatureCollection.scala
+++ b/vector/src/main/scala/geotrellis/vector/io/json/JsonFeatureCollection.scala
@@ -99,11 +99,11 @@ object JsonFeatureCollection{
     fc
   }
 
-  def apply(geometries: Traversable[Geometry]) = {
+/*  def apply(geometries: Traversable[Geometry]) = {
     val fc = new JsonFeatureCollection()
     fc ++= geometries.toList
     fc
-  }
+  }*/
 
   def apply(features: Traversable[JsValue])(implicit d: DummyImplicit): JsonFeatureCollection =
     new JsonFeatureCollection(features.toList)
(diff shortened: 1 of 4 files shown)


vector/src/main/scala/geotrellis/vector/io/json/package.scala
```diff
@@ -7,7 +7,8 @@ import spray.json.JsonFormat
 package object json extends GeoJsonSupport {
   implicit class GeometriesToGeoJson(val geoms: Traversable[Geometry]) extends AnyVal {
     def toGeoJson: String = {
-      JsonFeatureCollection(geoms).toJson.compactPrint
+      //JsonFeatureCollection(geoms).toJson.compactPrint
```
lossyrob: No need to keep the comment around.

allixender:
The build only fails because of the sbt-launcher.jar issue ... which seems to be resolved with the current version of the sbt runner (https://github.com/geotrellis/geotrellis/pull/1147)

vector/src/main/scala/geotrellis/vector/io/json/JsonFeatureCollection.scala
```diff
@@ -99,11 +99,11 @@ object JsonFeatureCollection{
     fc
   }
 
-  def apply(geometries: Traversable[Geometry]) = {
+/*  def apply(geometries: Traversable[Geometry]) = {
     val fc = new JsonFeatureCollection()
     fc ++= geometries.toList
     fc
-  }
+  }*/
```
allixender: I think we discussed that lists/seq of pure Geometry types cannot be applied to a FeatureCollection, and therefore, this would need to be removed, right?

lossyrob: True, this can be removed.

allixender: ok, done, also in JsonFeatureCollectionMap. The travis-CI will probably still fail, the sbt update is not in this pull request.

PR #1146, opened 2015-07-27
Updating sbt script to the latest version from paulp

echeipesh:
Old SBT jar URL at Typesafe is redirecting, curl not configured to follow it. Upgrading whole script for good measure.


lossyrob:
Whoops. @allixender did the same thing in #1147, didn't see this one.

PR #1147, opened 2015-07-28
newer version of sbt runner script

allixender:
newer version of sbt runner script to load the sbt-launcher.jar from valid typesafe repo url

(old version placed the launcher in ./lib/, and was not cached in ~/.sbt or ~/.ivy2 as one might expect), new repo URL caused failing download. New sbt runner downloads from current valid repo and seems to store/cache it in ~/.sbt/launchers/0.13.8/sbt-launch.jar 

`sbt` from https://github.com/paulp/sbt-extras/blob/master/sbt

### DIFF
--- a/sbt
+++ b/sbt
@@ -1,63 +1,61 @@
 #!/usr/bin/env bash
 #
 # A more capable sbt runner, coincidentally also called sbt.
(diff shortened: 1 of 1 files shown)


lossyrob:
:+1:

PR #1151, opened 2015-07-29
Hadoop also handle ".tiff" extension

zifeo:
Hadoop filesystem now matches on one or more specified extensions.

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopSparkContextMethods.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopSparkContextMethods.scala
@@ -31,7 +31,9 @@ trait HadoopSparkContextMethods {
   def hadoopGeoTiffRDD(path: Path, tiffExtensions: Seq[String]): RDD[(ProjectedExtent, Tile)] = {
     val searchPath = path.toString match {
       case p if tiffExtensions.exists(p.endsWith) => path
-      case p => new Path(p + "/*" + tiffExtensions.mkString("{", ",", "}"))
+      case p =>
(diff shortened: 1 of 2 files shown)


pomadchin:
There can also be tif extensions in the uppercase, mb to make smth like `toLowerCase` for the path string, or just to add upper case extensions into extensions seq?

zifeo:
Good point, thanks! The uppercase versions make also sens.

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopSparkContextMethods.scala
```diff
@@ -11,18 +11,27 @@ import org.apache.hadoop.mapreduce.Job
...
+
+  def hadoopGeoTiffRDD(path: Path, tiffExtension: String): RDD[(ProjectedExtent, Tile)] =
+    hadoopGeoTiffRDD(path, Seq(tiffExtension))
+
+  def hadoopGeoTiffRDD(path: Path, tiffExtensions: Seq[String]): RDD[(ProjectedExtent, Tile)] = {
     val searchPath = path.toString match {
-      case p if p.contains(tiffExtension) => path
-      case p => new Path(s"$p/*$tiffExtension")
+      case p if tiffExtensions.exists(p.endsWith) => path
+      case p => new Path(p + "/*" + tiffExtensions.mkString("{", ",", "}"))
```
lossyrob: case p =>
  val extensions =  tiffExtensions.mkString("{", ",", "}")
  new Path(s"$p/*$extensions")

lossyrob:
+1

PR #1153, opened 2015-08-06
Avro S3

echeipesh:
This branch introduces Avro encoding for S3 catalog, displacing Kryo which was problematic: https://github.com/geotrellis/geotrellis/issues/1138 

We introduce a very basic type-class interface to Avro with restrictions that we only can encode at granularity "records" and all members of the union must share a meaningful super type. This interface is limited but sufficient for current requirements and can be improved in the future.

Avro writer schema is stored in the attribute store under key "schema" and can be used in the future for schema migration.

Current record schema is `Array[(Key, Tile)]`, where Tile is a union of possible tile types. This supports both spatial and spatial-temporal layers as well as providing support for future meta-tile work. 

Overall this should provide a point for S3 catalogs from which we can refactor without having to re-ingest the layers.


lossyrob:
+1000
 after travis says yay

PR #1155, opened 2015-08-13
Feature/geojson extract 1152, fixes #1152

### LINKED ISSUE #1152: Handle GeoJson to Geometries and Features more gracefully
Can we have something like?

```scala
 val json: String = ???
 
  val p = json.extractGeometries[G <: Geometry]: Seq[G] // May be empty
  val p = json.extractFeatures[G <: Geometry, D]: Seq[Feature[G, D]]
  val p = json.parseGeoJson[PolygonFeature[Int]]: PolygonFeature[Int] // This expects all the types to line up, throws if not.
  val p = json.parseGeojson[JsonFeatureCollection].getAllPolygonFeature[D]().head
 
  p
```

allixender:
Slimmed down extractGeometries, implemented extractFeatures
Kudos to @lossyrob 
Fixes #1152


lossyrob:
:+1: w00t!

PR #1156, opened 2015-08-13
Cloud removal

alasin:
This code enables GeoTrellis to generate a cloud-free image of multiple satellite images of a location, some or most of which might be cloudy. 
Input -> Set of RGB GeoTIFFs.
Output -> Cloud-less tile and an optional PNG (RGB) of the reultant image.

Test file contains the testing programs and might be ignored for this purpose.


lossyrob:
@alasin getting a travis error based on file path `java.io.FileNotFoundException: /home/kamikaze/gsoc/cropped_images/image00/B2_cropped.TIF (No such file or directory)` (check Travis logs).

raster/src/main/scala/geotrellis/raster/imagery/CloudRemoval.scala
```diff
@@ -0,0 +1,127 @@
...
+      val v = if(isData(z))
+        ((z - minVal).toDouble/maxVal)*255
+      else 0
+
+      byteTile.set(col, row, v.toInt)
+    }
+    byteTile
+  }
+
+  def writePng(t: MultiBandTile, filename: String) = {
```
lossyrob: This will need to be generalized (by just removing the `cloudless` nomenclature) and moved to `geotrellis.raster.render`. We should have this as part of an implicit method that is added to MultiBandTile. It should be called `renderPng`, and create a `PNG` object. The user can write the PNG to a file with the `write` call if so desired. Something like https://github.com/geotrellis/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/render/RenderMethods.scala should be created (say `MultiBandRenderMethods.scala`) to hold this method, and it should use this technique https://github.com/geotrellis/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/render/package.scala#L4 to implicitly add the method to MultiBandTile whenever someone does `import geotrellis.raster.render._`

alasin:
@lossyrob Fixed that bug. Added a few unit tests and README. Added the example code in README itself.

lossyrob:
lgtm, @echeipesh @notthatbreezy @moradology could someone else give a once over?

raster/src/main/scala/geotrellis/raster/imagery/CloudRemoval.scala
```diff
@@ -0,0 +1,57 @@
...
+        var count = 0
+        cfor(0)(_ < images.length, _ + 1) { i =>
+          val v = images(i).get(col, row)
+          if(isData(v) && v < threshold) {
+            sum += v
+            count += 1
+          }
+        }
+        
+        result.set(col, row, sum / count)
```
moradology: Since this is just using a central tendency, is there a good reason to use the mean rather than the median? With a small number of photos (or a large proportion of cloud covered images), I imagine this will cause some washing out of actual values.

alasin: Ya, there's no specific reason for choosing mean over median. I used mean because folks at Mapbox had done a similar averaging for their cloudless atlas. Mode is not used because the difference in frequency of pixel intensities is hardly 1 or 2 in 32/16 bit images (which I used for testing). I am yet to test this using median. Could give it a shot. But I doubt its importance in the output image.

echeipesh:
Some test files need to be committed: 

```scala
[info] - Pixel value should be less than original cloudy image *** FAILED ***
[info]   java.io.FileNotFoundException: raster-test/data/geotiff-test-files/cloud_images/red/1.TIF (No such file or directory)

[info] - Pixel value should be less than threshold *** FAILED ***
[info]   java.io.FileNotFoundException: raster-test/data/geotiff-test-files/cloud_images/red/1.TIF (No such file or directory)

[info] - Overloaded functions should give the same result for a specific threshold *** FAILED ***
[info]   java.io.FileNotFoundException: raster-test/data/geotiff-test-files/cloud_images/red/1.TIF (No such file or directory)
```

alasin:
@echeipesh These files have been committed as *geotiff-test-files.zip*.

PR #1158, opened 2015-08-15
RasterRDD render to PNG and GeoTiff

echeipesh:
- Adds `.renderPng` and `.renderGeoTiff` on `RasterRDD[SpatialKey]`
- ETL subproject gains `render` output option that uses above functionality
- `.saveToS3` and `.saveToHadoop` on `RDD[(K,V)]` can be exploited to do other formats easily

Based on tests Hadoop support for S3 appears poor. Neither s3a nor s3n are able to upload a sizable RDD without running out of memory or stalling out. So the s3 support uses similar mechanism that is used in S3 catalog and all URLs with `s3` schema are redirected to use that mechanism.


lossyrob:
:+1:

PR #1159, opened 2015-08-16
Feature/avro-accumulo

zifeo:
Change serializer from Kryo to Avro

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/io/index/hilbert/HilbertSpaceTimeKeyIndex.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/index/hilbert/HilbertSpaceTimeKeyIndex.scala
@@ -45,7 +45,7 @@ class HilbertSpaceTimeKeyIndex(
   val timeWidth = keyBounds.maxKey.temporalKey.time.getMillis - startMillis
   val temporalBinCount = math.pow(2, temporalResolution)
  
-  val chc = {
+  lazy val chc = {
     val dimensionSpec =
       new MultiDimensionalSpec( 
         List(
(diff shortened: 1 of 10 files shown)


spark/src/main/scala/geotrellis/spark/io/index/hilbert/HilbertSpaceTimeKeyIndex.scala
```diff
@@ -45,7 +41,7 @@ class HilbertSpaceTimeKeyIndex(
   val timeWidth = keyBounds.maxKey.temporalKey.time.getMillis - startMillis
   val temporalBinCount = math.pow(2, temporalResolution)
  
-  val chc = {
+  lazy val chc = {
```
echeipesh: Good to know that java serialization does not chase after lazy vals.

zifeo: I believe that `lazy` is exactly doing what you advise me, or at least it once did.

echeipesh: This doesn't quite work. If the lazy val is accessed, I just put a `println(chc)` in the body, then java see will try to follow it (test will fail again with serialization errors). 

It appears that if you make it `@transiet lazy val` we get the behavior we're after.
Relevant SO post: http://stackoverflow.com/questions/27882307/how-does-serialization-of-lazy-fields-work

zifeo: That make sense, I will push this change. Thanks!

echeipesh:
+1

PR #1160, opened 2015-08-18
Add spatial and spacetime accumulo inputs

zifeo:


spark-etl/src/main/scala/geotrellis/spark/etl/accumulo/SpatialAccumuloInput.scala
```diff
@@ -0,0 +1,31 @@
...
+
+class SpatialAccumuloInput extends AccumuloInput {
+
+  def key = classTag[SpatialKey]
+
+  def apply[K](lvl: StorageLevel, crs: CRS, scheme: LayoutScheme, props: Map[String, String])(implicit sc: SparkContext) = {
+
+    val bbox = props.get("bbox").map(Extent.fromString)
+    val chunks = props("layer").split(":")
+    val id = LayerId(chunks(0), chunks(1).toInt)
```
echeipesh: bbox and id parsing should just go into the super class

echeipesh:
This looks about right. It would be good to drive this for chatta-demo, or some other convenient case, to make sure it works as expected and isn't missing some useful option.

lossyrob:
+1

PR #1161, opened 2015-08-18
Support of Native Resolution Layers

echeipesh:
Support for having layers at native resolutions. This is very helpful layers that will be used for computation as it avoids resampling, matching the source raster resolution. It's not clear how these layers will be displayed in the future and they do not support pyramiding at the moment.

Also note it only makes sense to perform operations on layers that share the same `LayoutDefinition`, a check that should be added in the future.


echeipesh:
To clarify a question we had, at the moment on pyramid up each tile can only contribute to one tile in the upper layer. So as long as that tile boundaries line up between the levels there is no reason Floating layout can't be pyramided. Of course we can always pass the bottom layer through a Tiler and that would remove any such restrictions. All that is to say, there is a way forward.

raster/src/main/scala/geotrellis/raster/RasterExtent.scala
```diff
@@ -241,10 +241,10 @@ object RasterExtent {
     RasterExtent(extent, cw, ch, cols, rows)
   }
 
-  def apply(extent: Extent, cellwidth: Double, cellheight: Double): RasterExtent = {
-    val cols = (extent.width / cellwidth).toInt
-    val rows = (extent.height / cellheight).toInt
-    RasterExtent(extent, cellwidth, cellheight, cols, rows)
+  def apply(extent: Extent, cellSize: CellSize): RasterExtent = {
```
lossyrob: Any reason not to have both as overloads? This might break some things, and unless there's a good reason to prevent the overload that takes extent double double I'd prefer to keep it.

echeipesh: Yes, I ran into a surprising problem: When giving two longs (like from TileLayout) it will choose the Double, Double overload rather than Int, Int. It eventually broke because CellSize was 255,255 for LatLng extent, but it was not at all obvious why at first.

lossyrob: Ah, auto conversions. That makes sense.

spark-etl/src/main/scala/geotrellis/spark/etl/accumulo/SpatialAccumuloInput.scala
```diff
@@ -20,7 +20,7 @@ class SpatialAccumuloInput extends AccumuloInput {
       case Some(extent) => catalog.query[SpatialKey](id).where(Intersects(extent)).toRDD
       case None => catalog.read[SpatialKey](id)
     }
-    LayoutLevel(id.zoom, rdd.metaData.tileLayout) -> rdd.asInstanceOf[RasterRDD[K]]
+    id.zoom -> rdd.asInstanceOf[RasterRDD[K]]
```
lossyrob: I've been avoiding using the `->` notation, because the `@inline` usually fails and that's where all the compiler warnings get when we run with the `-optimized` flag.

spark/src/main/scala/geotrellis/spark/ingest/Pyramid.scala
```diff
@@ -15,15 +13,15 @@ object Pyramid extends Logging {
...
-    val nextLevel = layoutScheme.zoomOut(level)
+    val LayoutLevel(nextZoom, nextLevel) = layoutScheme.zoomOut(LayoutLevel(zoom, rdd.metaData.layout))
     val nextMetaData = 
       RasterMetaData(
         metaData.cellType,
-        metaData.extent,
-        metaData.crs,
-        nextLevel.tileLayout
+        nextLevel,
+        metaData.dataExtent, // TODO no way, these have to change
```
lossyrob: What does this TODO mean?

spark/src/main/scala/geotrellis/spark/ingest/Reproject.scala
```diff
@@ -37,8 +37,8 @@ object Reproject {
         ((key, newExtent), newTile)
       }
 
-    val metaData = 
-      RasterMetaData.fromRdd(reprojectedTiles, destCRS, rdd.metaData.tileLayout) { key => key._2 }
+    val metadata =
```
lossyrob: Tough, because `metadata` is right, but `metaData` is consistent. We should choose and make the change, but remain consistent between.

spark/src/main/scala/geotrellis/spark/io/json/package.scala
```diff
@@ -49,19 +50,20 @@ package object json {
     def write(metaData: RasterMetaData) = 
       JsObject(
         "cellType" -> metaData.cellType.toJson,
-        "extent" -> metaData.extent.toJson,
-        "crs" -> metaData.crs.toJson,
-        "tileLayout" -> metaData.tileLayout.toJson
+        "dataExtent" -> metaData.dataExtent.toJson,
+        "layoutExtent" -> metaData.layoutExtent.toJson,
+        "tileLayout" -> metaData.tileLayout.toJson,
```
lossyrob: Did we not end up making a type that is the composition of a `TileLayout` and `Extent`? If that were it's own type, then we wouldn't have to use the `dataExtent` name, which feels kind of awkward for me.

spark-etl/src/main/scala/geotrellis/spark/etl/accumulo/SpatialAccumuloInput.scala
echeipesh: Ah, that's a shame. I liked the syntax, I'll change and avoid it in the future.

spark/src/main/scala/geotrellis/spark/tiling/LayoutDefinition.scala
```diff
@@ -0,0 +1,41 @@
+package geotrellis.spark.tiling
+
+import geotrellis.raster.{RasterExtent, TileLayout}
+import geotrellis.vector.Extent
+
+case class LayoutDefinition(extent: Extent, tileLayout: TileLayout) extends Product {
```
lossyrob: Oh, this is the type mentioned in the comment above.

spark/src/main/scala/geotrellis/spark/ingest/Pyramid.scala
echeipesh: Stale, will remove. Used to hold GridBounds, which would have had to be recollected at this point. Holding the extent avoids the need for recalculation on zoom out/in.

spark/src/main/scala/geotrellis/spark/ingest/Reproject.scala
echeipesh: Should this be the PR for the change?

spark/src/main/scala/geotrellis/spark/RasterMetaData.scala
```diff
@@ -2,26 +2,37 @@ package geotrellis.spark
...
+ * @param dataExtent  Extent covering the source data
+ * @param crs         CRS of the raster projection
+ */
 case class RasterMetaData(
   cellType: CellType,
-  extent: Extent,
-  crs: CRS,
-  tileLayout: TileLayout
+  layout: LayoutDefinition,
+  dataExtent: Extent,
```
lossyrob: Why does this need to be `dataExtent` rather than `extent`? I think with the `LayoutDefinition` (which the member name should probably be the more verbose `layoutDefinition` to not confuse it with `TileLayout`.

spark/src/main/scala/geotrellis/spark/io/json/package.scala
lossyrob: Probably be better to have `LayoutDefinition` have it's own JSON format, so that it matches the object more and then `LayoutDefinition` could have it's JSON used other places.

spark/src/main/scala/geotrellis/spark/RasterMetaData.scala
echeipesh: Even with nested encoding distinction between layout extent (extent covering the tiles) and data extent (extent covering the cells with data) is not obvious, and this is something we missed and got wrong the first time around. Maybe doc-string is all that's required, but I felt hurt enough to give it a special name. Given that reasoning is it still too awkward for the benefit ?

spark/src/main/scala/geotrellis/spark/ingest/Reproject.scala
lossyrob: No, should be separate PR, but we should maintain consistency until we make the switch.

spark/src/main/scala/geotrellis/spark/tiling/ZoomedLayoutScheme.scala
```diff
@@ -33,42 +33,52 @@ class ZoomedLayoutScheme(tileSize: Int) extends LayoutScheme {
   /** TODO: Improve this algorithm. One improvement is to follow the algorithm
     * described in  "Tile-Based Geospatial Information Systems Principles and Practices"
     * by John T. Sample & Elias Ioup, section 3.1.2 */
-  def levelFor(worldExtent: Extent, cellSize: CellSize): LayoutLevel = {
+  def levelFor(projectedExtent: ProjectedExtent, cellSize: CellSize) = {
```
lossyrob: What would passing in the `extent` of the `projectedExtent` do here?

spark/src/test/scala/geotrellis/spark/tiling/ZoomedLayoutSchemeSpec.scala
```diff
@@ -9,7 +9,7 @@ import org.scalatest._
 class ZoomedLayoutSchemeSpec extends FunSpec with Matchers {
   describe("ZoomedLayoutScheme") { 
     it("Cuts up the world in two for lowest zoom level") {
-      val LayoutLevel(_, tileLayout) = ZoomedLayoutScheme().levelFor(1)
+      val LayoutLevel(_, tileLayout) = ZoomedLayoutScheme().levelForZoom(LatLng.worldExtent, 1)
```
lossyrob: Hmmm, I don't want to have to pass in the world extent every time we use these types of functions. Should the CRS or extent be included in the `ZoomedLayoutScheme` definition?

lossyrob: e.g. I usually have a `val layoutScheme = ZoomedLayoutScheme(256)` defined at the top level of code I'm working with (e.g. the WebMercator image tiler for OAM). If I had to do `val layoutScheme = ZoomedLayoutScheme(WebMercator, 256)`, I think that would be better then having to pass in `WebMercator.worldExtent` for every call to `levelForZoom`

spark/src/main/scala/geotrellis/spark/tiling/ZoomedLayoutScheme.scala
echeipesh: Do you mean that the extent is unused here? Basically this allows the "fork" in behavior between FloatingLayout and Zoomed layout:
https://github.com/geotrellis/geotrellis/pull/1161/files#diff-13f7f1feac59e88edb81e3998f049605R15

Zoomed layout uses the CRS (to get world extent), FloatingLayout uses the extent.

This seems like the cleanest way to unify the interface.

spark/src/test/scala/geotrellis/spark/tiling/ZoomedLayoutSchemeSpec.scala
echeipesh: I like this idea, lets see if it works out.

lossyrob: This would address the `ProjectedExtent` overload issue as well, those would go away.

spark/src/main/scala/geotrellis/spark/RasterMetaData.scala
```diff
@@ -2,26 +2,37 @@ package geotrellis.spark
...
+  crs: CRS
 ) {
+  /** Transformations between tiling scheme and map references */
+  def mapTransform = layout.mapTransform
+  /** Layout raster extent */
+  def rasterExtent = layout.rasterExtent
+  /** TileLayout of the layout */
+  def tileLayout = layout.tileLayout
+  /** Full extent of the layout */
+  def layoutExtent = layout.extent
```
lossyrob: Is there a huge need to have these delegation methods? I'm usually opposed to having the shortcut method, because it makes refactoring twice as hard, and there's more explicit ownership when only one type has the method.

echeipesh: These are pretty heavily referred to:
`mapTransform`: 14 usages
`tileLayout`: 36 usages

So there is some thought to minimizing code shuffles and it avoids this train `RasterMetaData.layout.tileLayout` (which feels ugly to me).

Issue #1117, opened 2015-06-15
Bringing Cassandra branch up to current master


allixender:
working on https://github.com/allixender/geotrellis/tree/feature/cassandra-merge-master

feature/cassandra branch has now long lived in parallel with master

In favour of not discarding the commit history and not rebasing, will try to merge master into a copy of feature/cassandra

allixender:
a diff reveals tht even full catch up with master there are some differences in the spark package that I'll have to "over write" which are not related to Cassandra

```diff
diff -ur spark/src/main/scala/geotrellis/spark/Boundable.scala ../../ws-idea-01/geotrellis-git/spark/src/main/scala/geotrellis/spark/Boundable.scala
--- spark/src/main/scala/geotrellis/spark/Boundable.scala	2015-06-15 17:59:26.901952890 +0000
+++ ../../ws-idea-01/geotrellis-git/spark/src/main/scala/geotrellis/spark/Boundable.scala	2015-06-15 18:50:28.015325145 +0000
@@ -1,7 +1,5 @@
 package geotrellis.spark
 
-import scala.reflect.ClassTag
-
 /**
  * This type class marks K as point that can be bounded in space.
  * It is used to construct bounding hypercube for a set of Ks.
@@ -31,13 +29,6 @@
       minBound(b1.minKey, b2.minKey),
       maxBound(b1.maxKey, b2.maxKey))
   }
-}
-
-object Boundable {
-  def getKeyBounds[K: ClassTag](rdd: RasterRDD[K])(implicit boundable: Boundable[K]): KeyBounds[K] =
-    rdd
-      .map { case (key, _) => KeyBounds(key, key) }
-      .reduce(boundable.combine)
 
   def intersect(b1: KeyBounds[K], b2: KeyBounds[K]): Option[KeyBounds[K]] = {
     val kb = KeyBounds(
@@ -56,4 +47,4 @@
   }
 
   def getKeyBounds(rdd: RasterRDD[K]): KeyBounds[K]
-}
+}
\ No newline at end of file
Only in spark/src/main/scala/geotrellis/spark/cmd/args: CassandraArgs.scala
Only in spark/src/main/scala/geotrellis/spark/ingest: CassandraIngestCommand.scala
[...]

PR #1163, opened 2015-08-21
Feature/cassandra

allixender:
fixes to #1117 to date.

- has spark-etl Input/Output (to be tested though)
- has AvroCodecs
- has lots of merges from geotrellis/master


lossyrob:
Nice, thanks @allixender ! I'm hoping with the catalog refactors that @echeipesh is doing, it'll be more clear what Readers/Writers we support in Cassandra, and we can perhaps get this code into the 0.10 release.

PR #1164, opened 2015-08-23
extractGeometries, JsonFeatureCollection.getAll[G], et al  …

### LINKED ISSUE #1162: extractGeometries, JsonFeatureCollection.getAll[G], et al swallow exceptions that they shouldn't
Right now, the `geotrellis.vector.io.json` package logic for `extractGeometries` and `extractFeatures`, as well as the `JsonFeatureCollection.getAll` and `JsonFeatureCollectionMap.getAll` logic swallows exceptions that it should not. The solution is to throw any exceptions during the `Try` matching that are not `DeserializationException`s.

e.g.

```scala
    def extractGeometries[G <: Geometry : JsonReader: TypeTag]: Seq[G] =
      Try(s.parseJson.convertTo[G]) match {
        case Success(g) => Seq(g)
        case Failure(_: spray.json.DeserializationException) =>
          Try(s.parseGeoJson[JsonFeatureCollection]) match {
            case Success(featureCollection) =>
              featureCollection.getAll[G]
            case Failure(_: spray.json.DeserializationException) =>
              Try(s.parseGeoJson[GeometryCollection]) match {
                case Success(gc) => gc.getAll[G]
                case Failure(_: spray.json.DeserializationException) => Seq()
                case Failure(e) => throw e
              }
            case Failure(e) =>
              throw e
          }
        case Failure(e) =>
          throw e
      }
```

Thanks to @jwalgran for finding, and @allixender for figuring out.

allixender:
…now bubble up exceptions if not DeserialisationExceptions, fixes  #1162

### DIFF
--- a/vector-test/src/test/scala/spec/geotrellis/vector/io/json/GeoJsonSpec.scala
+++ b/vector-test/src/test/scala/spec/geotrellis/vector/io/json/GeoJsonSpec.scala
@@ -217,5 +217,11 @@ class GeoJsonSpec extends FlatSpec with Matchers {
     val t12 = jsonFeatCol.extractFeatures[LineFeature[SomeData]]
     t12 should be (Seq(linefeature2))
 
+    // catch java.lang.AssertionError from Polygon.scala "Empty Geometry"
+    intercept[java.lang.AssertionError] {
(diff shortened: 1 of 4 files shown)


lossyrob:
+1

PR #1167, opened 2015-09-01
Bintray publishing and new style SBT configuration

echeipesh:
Preparing to integrate publishing from Travis to Bintray on every PR merge.

The pattern will be to publish artifacts tagged with the first 7 letters of the commit SHA to bintray in order to allow developers to work against a recent release without doing publish-local. (ex: https://bintray.com/azavea/geotrellis)

The build will take the contents of the `GEOTRELLIS_VERSION_PREFIX` environmental variable and append it to the version. If unset the version will be prefixed with "-SNAPSHOT", allowing the usual publish-local centered development.

The move to new style SBT configuration is incidental but provides more clarity in the build. 
 - SBT will merge  subproject `build.sbt` with the main `build.sbt`, so a lot of project specific tweaks can be moved out and modularized. (ex: `spark/build.sbt`)
 - Only val/def declarations are allowed in `build.sbt` limiting the amount of possible chaos.
 - Gave a chance to clean up some redundant resolver definitions


hectcastro:
Awesome. [Here](https://bintray.com/azavea/geotrellis) is a link to the Bintray repository.

spark/build.sbt
```diff
@@ -0,0 +1,9 @@
+test in assembly := {}
+
+assemblyMergeStrategy in assembly := {
+  case "reference.conf" => MergeStrategy.concat
+  case "application.conf" => MergeStrategy.concat
+  case "META-INF/MANIFEST.MF" => MergeStrategy.discard
+  case "META-INF\\MANIFEST.MF" => MergeStrategy.discard
+  case _ => MergeStrategy.first
+}
```
lossyrob: Why put this here? I avoided using subproject `build.sbt` because it makes things hard to locate.

echeipesh: Spark subproject is the only one we ever build an assembly for, so it localized the settings (it was part of common settings before). Same story with the admin subproject and the `conflictWarning in ThisBuild := ConflictWarning.disable` setting.

lossyrob: I think these project-specific settings should just be added to the `Build.scala` subprojects specifically then. I explicitly felt pain about `build.sbt` files in some subprojects and not others containing pieces of data.

One idea to provide some separation would be to define the subproject settings in separate files under `project`, and import them into `Build.scala`. But really it's good to have all the build stuff in one place and not have danglers in the subproject folders.

build.sbt
```diff
@@ -0,0 +1,461 @@
+import Dependencies._
+
+lazy val commonSettings = Seq(
```
lossyrob: Ok I see now that we moved from a `Build.scala` to a `build.sbt`. Can you convince me this is a good idea?

lossyrob: Done (on Gitter). Now just to work out the subproject build.sbt's...

lossyrob:
+1

PR #1168, opened 2015-09-02
BinaryFileInputFormat, CellGrid supertype

lossyrob:
This PR has a couple changes that I made as part of work on OpenAerialMap:
- Modified GeotiffInputFormat to use a generic BinaryFileInputFormat. The useful functionality is generalized in BinaryFileInputFormat, and that can be used to create RDD's from other types (in my example, an `RDD[(Extent, MultiBandTile)]`
- Create a `CellGrid` trait that both `Tile` and `MultiBandTile` extent, which houses `cellType`, `cols`, `rows`, and related methods.
- Did a no-op replacement on the `Callback` type alias in the rasterizer, just because it's easier to read from an API standpoint. Keeping the `Callback` type alias for backwards compatibility, this might be deleted in a future PR.
- Increase the MaxPermSize for spark subproject, the tests were failing on PermGen space and this fixed.


echeipesh:
:+1:

PR #1171, opened 2015-09-08
Add support for additional projections in geotiffs

notthatbreezy:
This commit improves support for projections when reading and writing
GeoTiffs. Before this commit, it looks as though almost all projections
were read as custom user projections, ignoring and not setting the
`ProjectedCSTypeGeoKey` properly in the process. In the suggestions for
geocoding rasters in the GeoTiff spec, if this code is set (e.g. an EPSG
code) then other parameters are unecessary.

This commit adds support for reading/writing the `ProjectedCSTypeGeoKey`
value properly to enable reading/writing projections like EPSG:3857 (web
-mercator) properly. Tests have been added for this purpose.

In the process of writing this task, I found
http://download.osgeo.org/pub/geotiff/samples/README which describes
additional GeoTiffs GeoTrellis should aim to support reading/writing
with to be on par with GDAL.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/io/geotiff/tags/TiffTags.scala
+++ b/raster/src/main/scala/geotrellis/raster/io/geotiff/tags/TiffTags.scala
@@ -196,15 +196,15 @@ case class TiffTags(
     }
   }
 
-  def pcs: Int = try {
+  def projectedCoordinateSystem: Int = try {
     GeoTiffCSTags.pcs
   } catch {
     case e: Exception => 32767
   }
 
   lazy val crs: CRS = {
-    if (pcs != 32767) {
-      CRS.fromName(s"EPSG:${pcs}")
+    if (projectedCoordinateSystem != 32767) {
+      CRS.fromName(s"EPSG:${projectedCoordinateSystem}")
     } else {
       proj4String match {
         case Some(s) => CRS.fromString(s)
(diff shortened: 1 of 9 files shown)


raster/src/main/scala/geotrellis/raster/io/geotiff/writer/CoordinateSystemParser.scala
```diff
@@ -208,6 +228,17 @@ class Proj4StringParser(val proj4String: String) {
     (geoKeysInt, doublesLB.toList)
   }
 
+  private lazy val mercProps = {
```
notthatbreezy: Whoops, this needs to come out

raster/src/main/scala/geotrellis/raster/io/geotiff/tags/TiffTags.scala
```diff
@@ -181,16 +181,36 @@ case class TiffTags(
...
-    case None => LatLng
+  def proj4String: Option[String] = try {
+    GeoTiffCSTags.getProj4String
+  } catch {
+    case e: Exception => {
+      None
+    }
+  }
+
+  def pcs: Int = try {
```
lossyrob: What does `pcs` stand for?

lossyrob: Ah, I see. Projected Coordinate System. I think this name deserves not being an initialism.

PR #1178, opened 2015-09-16
Added GridBounds.distinct and GridBounds.minus

lossyrob:
Added some functionality to GridBounds that allows you to transform a sequence of potentially overlapping GridBounds into a set that covers the same area but the does not overlap. Useful if you want to count the number of merged tiles you will have in some tile operation without actually counting them (e.g. for determining partition size in an RDD).

Also created a main method you can `test:run` to force the replacement of the geotiff spec data (use when you merge in master and all the sudden have GeoTiff test errors)


echeipesh:
distrinct(:+1:, :+1:) should be (:+1:)

PR #1179, opened 2015-09-17
Fixed the resource file reading in geotrellis.proj4

lossyrob:
The `geotrellis-proj4` package was reading CRS info files as Files on local disk, not as resources. This fixes those couple of places where that mistake was being made. It also scoped down those resource files under the `geotrellis.proj4` namespace, and prevents exceptions resulting in resources not being found from being swallowed by the TiffTag code.


echeipesh:
+1

PR #1181, opened 2015-09-22
Adding applys to HadoopRasterCatalog

lossyrob:
I've just run into this compiler error (needing a `Path` instead of just taking a `String`) multiple times, this will save future headaches.


echeipesh:
+1

Issue #1149, opened 2015-07-28
ETL project can abstract transform step.


lossyrob:
Instead of
```scala
class GeoTiffHadoopIngest extends HadoopIngest {
  val format = "geotiff"
  val key = classTag[SpatialKey]

  def apply[K](lvl: StorageLevel, crs: CRS, layoutScheme: LayoutScheme, props: Map[String, String])(implicit sc: SparkContext) = {
    val source = sc.hadoopGeoTiffRDD(props("path"))
    val reprojected = source.reproject(crs).persist(lvl)
    val (layoutLevel, rasterMetaData) =
      RasterMetaData.fromRdd(reprojected, crs, layoutScheme, isUniform = false) { _.extent }
    val tiler = implicitly[Tiler[ProjectedExtent, SpatialKey]]
    layoutLevel -> tiler(reprojected, rasterMetaData).asInstanceOf[RasterRDD[K]]
  }
}
```
we could have something like ...
```scala
class GeoTiffHadoopIngest extends HadoopIngest with Transform[ProjectedExtent, SpatialKey] {
  def source: RDD[(ProjectedExtent, Tile)] =
    sc.hadoopPngRDD(props("path"))
}
```
If we abstracted out the Transform step between an InputKey and a target K (which really just requires a `Tiler[InputKey,K]`

PR #1182, opened 2015-09-23
Catalog Splitup with MultiBand support

echeipesh:
This pull request splits up catalog objects into readers and writers implementing `geotrellis.spark.io.FilteringLayerReader` and `geotrellis.spark.io.Writer` respectively.

## LayerReader/LayerWriter

- AccumuloCatalog becomes:
	- AccumuloLayerWriter
	- AccumuloLayerReader
	- AccumuloTileReader
- HadoopCatalog becomes:
	- HadoopLayerWriter
	- HadoopLayerReader
	- HadoopTileReader
- S3Catalog becomes:
	- S3LayerWriter
	- S3LayerReader
	- S3TileReader

These changes have following effects:
- Since all readers share an interface the query functionality is now lifted to that interface, `FilteringLayerReader`
- It is now possible to create a tile reader without access to `SparkContext`
- Common interface allows various methods of dependency injection for persistence layer. For example `PersistenceSpec` implements a test suite that is shared by all backends and replaces individual tests with better coverage.
- Notably TileReaders are ` Reader[LayerId, Reader[K, V]]` (they are reader readers) allowing them to be provided with layer readers/writers at points where `LayerId`s are not yet available.

## RDDReader/RDDWriter
All layer readers and writers depend on their respective RDDReaders and RDDWriters. These objects abstract over writing and reading `RDD[(K, V)]` to their respective backends without regard for metadata, which is handled by layer readers/writers.

These classes do not share a common interface as they leak both types and concerns from the underlying layers. For instance there are two strategies to write to Accumulo, one through `BatchWriter` and one through RFile import. However, their interface is still similar.

### RDDWriter
Writers require some way to map K to unique a unique identifier
	- S3RDDWriter accepts `K => String` mapping K to S3 key
	- AccumuloRDDWriter accepts `K => Text` mapping K to RowID
	- HadoopRDDWriter accepts `KeyIndex[K]` which is `K => Long` mapping K to an index in a Map file.

### RDDReader
Reader needs some way to express ability query multi-dimensional space. This is represented by two arguments:

-  `queryKeyBounds: Seq[KeyBounds[K]]` represents a series of hypercubes in said space.  These are produced by `RDDQuery` instances.
-  `decomposeBounds: KeyBounds[K] => Seq[(Long, Long)]` represents a function that maps each of those multi-dimensional regions to a sequence of one dimensional ranges. This directly maps to `KeyIndex[K].indexRanges`. 

Hopefully the very general format of these arguments support alternative indexing methods that are not yet implemented or will be implemented outside of GeoTrellis layer readers.

## MultiBand Support

Several steps are present to provide MultiBand support.
RDDReaders and RDDWriters treat `MultiBandTile` just as another instance of `V`. `geotrellis.spark.io.avro.codecs._` provides appropriate formats for them to just work.

RDDLayerReader requires to be generic on return type since single band rasters are wrapped in`RasterRDD[K]` and multi-band rasters are wrapped in`MultiBandRasterRDD[K]`. Consequently the signatures needs to look like this:

``` scala
val reader1 = new S3LayerReader[SpatialKey, Tile, RasterRDD[SpatialKey]](...)
```
Third argument is somewhat redundant so we provide overloads to `apply` that assumes a single type hole in return type for K.

```scala
val reader2 = S3LayerReader[SpatialKey, Tile, RasterRDD](...)
```
### Ingest
In order to support Ingest of multi-band data all the steps support working on arbitrary tiles. This support implemented through various type classes

- Collect RasterMetaData requires:
	- `V <: CellGrid`
- Reproject requires:
	- `ReprojectView[V]`
- Tiler requires:
	-  `MergeView[K]`, `CellGridPrototypeView[K]`
- Pyramid requires:
	- `MergeView[K]`, `CellGridPrototypeView[K]`

#### Ingest workflows
`MultiBandIngest` is provided to bring all of the above components together similar to `Ingest` object. However there exists a feeling is that this object only captures only of the possible workflows. To support more flexible ingests with preprocessing options extension methods for each stage are provided:

```scala
val source = sc.hadoopGeoTiffRDD("hdfs:/...")
val reprojected = source.reproject(LatLng)
val (zoom, rmd) = source.collectMetaData(LatLng, ZoomedLayoutScheme(LatLng, 512))
val tiled = source.tile[SpaceTimeKey](rmd)
val rdd = RasterRDD[SpatialKey](rdd, rmd)
val writer = S3LayerWriter[SpatialKey, Tile, RasterRDD]("my-bucket", "catalog")
writer.write(LayerId("myLayer", zoom), 
```
In this manner additional processing can be performed at each stage or stages may be omitted. For instance if you are targeting a specific raster layout you may construct your own `RasterMetaData` rather than collecting it from source images.

## Accumulo Format Changes
To keep record handling more consistent the handling of records in Accumulo has changed to include all the (key, tile) pairs that share the same `KeyIndex[K]` id in the Value as an Avro encoded array.

This has no impact on spatial layers since the nature of `SpatialKey(col, row)` means that two keys may never share an index under currently available index methods. However this changes the behavior of storing spatiotemporal data. Under the old format each pair that shares an index would be disambiguated by adding of column qualifier containing the date string. On query the format a server side iterator could be applied that filtered for only keys with column qualifier in the desired range.

Legacy behavior is still implemented by `SpaceTimeAccumuloRDDReader` and `SpaceTimeAccumuloRDDWriter` and can be used with the new layer readers if desired.

Furthermore the accumulo RowID is now encoded as `Array[Byte]` of length 8, derived from the key index. Old RowID format was to encode both the zoom level and the index in as a string.

## ETL Changes
ETL project has gained new arguments, `layoutScheme` can be provided as either `tms` or `floating` to use either `ZoomedLayoutScheme` or `FloatingLayoutScheme` respectively; it is no longer necessary to provide the layout scheme in code.

If `layoutScheme` is not provided then `layoutExtent`, `cellSize`, `cellType`, and `tileSize` will be used to define the desired `RasterMetaData`. This both skips the collect metadata phase and provides full control of the raster layout and resolution.

It is currently not possible to use ETL subproject to ingest multi-band data. This feature is forthcoming in near future.

fixes: #1149

## AttributeStore Changes
Instead of writing layer attribute (those attributes used for layer reading and writing) fields into their own a attributes, we centralize them in "metadata" attribute. The combined size of these is small enough that it makes sense to batch their reading and writing. Not to mention that they're almost always used together.

This "metadata" attribute is now has following top level members:

- `header`: contains keyClass and valueClass fields along with backend specific information needed to load the layer
- `metadata`: contains RasterMetaData or one of it's subclasses. 
- `keyIndex`: unchanged, contains keyIndex object
- `keyBounds`: unchanged, contains keyBounds of the layer
- `schema`: contains avro record schema with which the layer was written

Reading and writing these attributes is centrlized in `JsonAttributeStoreMethods` class and on read it will perform graceful failvoer to look for missing fields their own attribute entries. This should allow reading of layers attributes ingested with previous version of GeoTrellis.

AttributeStore is now accepts type parameter for the format it will use for encoding/decoding, which is easier to deal with than the type member.

`AttributeStore` now  extends `AttributeCaching` which provides `cacheRead` and `cacheWrite`. This makes caching available to all clients, rather than just the layer readers/writers. It also became clear that it is not possible to anticipate all fields that should be cached. Exposing these methods delegates the responsibility of the decision to the user.


spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloWriteStrategy.scala
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * BatchWriter is notified of the tablet migrations and will follow them around the cluster.
+ *
+ * @param config Configuration for the BatchWriters
+ */
+case class SocketWriteStrategy(
+  config: BatchWriterConfig = new BatchWriterConfig().setMaxMemory(128*1024*1024).setMaxWriteThreads(32) 
+) extends AccumuloWriteStrategy {
+  def write(kvPairs: RDD[(Key, Value)], instance: AccumuloInstance, table: String): Unit = {
+    val BC = KryoWrapper((instance, config))
```
lossyrob: BC is unclear, rename val

lossyrob: Verify what is un-serializable with closure serializer; comment on why were are KryoWrapping

spark/src/main/scala/geotrellis/spark/io/accumulo/spacetime/SpaceTimeAccumuloRDDWriter.scala
```diff
@@ -0,0 +1,47 @@
...
+import org.apache.accumulo.core.data.{Key, Value}
+import org.apache.hadoop.io.Text
+import org.apache.spark.rdd.RDD
+import org.joda.time.DateTimeZone
+
+/**
+ * This class implements writing SpaceTime keyed RDDs by using Accumulo column qualifier to store the dates of each tile.
+ * This allows the use of server side filters for refinement of spatio-temporal queries.
+ */
+class SpaceTimeAccumuloRDDWriter[TileType: AvroRecordCodec](
```
lossyrob: TileType -> V

spark/src/main/scala/geotrellis/spark/io/s3/S3LayerReader.scala
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Handles reading raster RDDs and their metadata from S3.
+ *
+ * @param attributeStore  AttributeStore that contains metadata for corresponding LayerId
+ * @param getCache        Optional cache function to be used when reading S3 objects.
+ * @tparam K              Type of RDD Key (ex: SpatialKey)
+ * @tparam TileType       Type of RDD Value (ex: Tile or MultiBandTile )
+ * @tparam Container      Type of RDD Container that composes RDD and it's metadata (ex: RasterRDD or MultiBandRasterRDD)
+ */
+class S3LayerReader[K: Boundable: AvroRecordCodec: JsonFormat: ClassTag, TileType: AvroRecordCodec: ClassTag, Container[_]](
```
lossyrob: Is it possible for the `Container` type to not specify type params (so the user would have to fully specify), and have object applys do some type translation so we get the generic machinery without sacrificing the cleaner API?

e.g.
```
      // This is the actual call to new
      val spaceTimeReader = new S3LayerReader[SpatialKey, Tile, RasterRDD[SpatialKey]](attributeStore){
        override val getS3Client = () => new MockS3Client
      }
  
     // This is an object apply that delegates to the above call
      val spaceTimeReader = S3LayerReader[SpatialKey, Tile, RasterRDD](attributeStore){
        override val getS3Client = () => new MockS3Client
      }
```

spark/src/main/scala/geotrellis/spark/RDDQuery.scala
```diff
@@ -0,0 +1,48 @@
+
+package geotrellis.spark
```
lossyrob: mv query stuff to `geotrellis.spark.io`

spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerWriter.scala
```diff
@@ -0,0 +1,63 @@
...
+  extends Writer[LayerId, Container[K] with RDD[(K, TileType)]] {
+
+  def write(id: LayerId, rdd: Container[K] with RDD[(K, TileType)]): Unit = {
+    val layerMetaData =
+      AccumuloLayerMetaData(
+        keyClass = classTag[K].toString(),
+        valueClass = classTag[TileType].toString(),
+        tileTable = table
+      )
+    val rasterMetaData = cons.getMetaData(rdd)
```
lossyrob: rm raster name

spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerWriter.scala
```diff
@@ -0,0 +1,63 @@
...
+class AccumuloLayerWriter[K: SpatialComponent: Boundable: JsonFormat: ClassTag, TileType: ClassTag, Container[_]](
+    val attributeStore: AttributeStore.Aux[JsonFormat],
+    rddWriter: BaseAccumuloRDDWriter[K, TileType],
+    keyIndexMethod: KeyIndexMethod[K],
+    table: String)
+  (implicit val cons: ContainerConstructor[K, TileType, Container])
+  extends Writer[LayerId, Container[K] with RDD[(K, TileType)]] {
+
+  def write(id: LayerId, rdd: Container[K] with RDD[(K, TileType)]): Unit = {
+    val layerMetaData =
```
lossyrob: LayerMetaData => StorageMetaData

spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerReader.scala
```diff
@@ -0,0 +1,64 @@
...
+      val rdd = rddReader.read(layerMetaData.tileTable, id.name, writerSchema, queryKeyBounds, decompose)
+      cons.makeContainer(rdd, keyBounds, metadata)
+    } catch { // TODO: Decide if this is actually helpful, this hides the real error
+      case e: AttributeNotFoundError => throw new LayerNotFoundError(id)
+    }
+  }
+}
+
+object AccumuloLayerReader {
+  def apply[K: Boundable: AvroRecordCodec: JsonFormat: ClassTag, V: AvroRecordCodec: ClassTag, C[_]](instance: AccumuloInstance)
```
lossyrob: Add object apply for `new AccumuloLayerReader`

lossyrob:
+1 after changes

PR #1184, opened 2015-10-07
Fix Reading GeoTiff NoData tags <= 4 bytes

echeipesh:
When a tiff tag value is less than or equal to four bytes the offset (int32) of the tag metadata contains the actual value rather then the byte offset of where the value is written.

For instance a common NoData value of "0" would be encoded as count=2 offset=48

In this case we read the wrong ByteBuffer looking for the value.


notthatbreezy:
+1

PR #1185, opened 2015-10-14
Fixed geotools test.

pomadchin:
Added generic unapply functions to `Feature` objects (vector project). Fixed geotools project test.


echeipesh:
:+1: Thank you for submitting this!

Test failure from ScalaCheck can not be produced locally, will require digging.

PR #1190, opened 2015-10-21
GeoTiff Byte NoData handling

echeipesh:
Fixes two use cases:
- ByteArrayTile.fromBytes, which used when we decompress GeoTiff into array tiles 
- NoDataByteGeoTiffSegment, which is used when reading GeoTiff segment by segment.

This also removes "& FF" on get from ByteGeoTiffArrayTile, leaving it as a signed byte container.

### DIFF
--- a/project/Environment.scala
+++ b/project/Environment.scala
@@ -22,7 +22,7 @@ object Environment {
 
   lazy val hadoopVersion  = either("SPARK_HADOOP_VERSION", "2.2.0")
   lazy val sparkVersion   = either("SPARK_VERSION", "1.4.1")
-  lazy val versionSuffix  = either("GEOTRELLIS_VERSION_SUFFIX", "-SNAPSHOT")
+  lazy val versionSuffix  = either("GEOTRELLIS_VERSION_SUFFIX", "-97834e6")
 
   lazy val javaGdalDir    = either("JAVA_GDAL_DIR", "/usr/local/lib")
 }
--- a/project/Environment.scala
+++ b/project/Environment.scala
@@ -22,7 +22,7 @@ object Environment {
 
   lazy val hadoopVersion  = either("SPARK_HADOOP_VERSION", "2.2.0")
   lazy val sparkVersion   = either("SPARK_VERSION", "1.4.1")
-  lazy val versionSuffix  = either("GEOTRELLIS_VERSION_SUFFIX", "-97834e6")
+  lazy val versionSuffix  = either("GEOTRELLIS_VERSION_SUFFIX", "-SNAPSHOT")
 
   lazy val javaGdalDir    = either("JAVA_GDAL_DIR", "/usr/local/lib")
 }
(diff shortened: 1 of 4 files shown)


project/Environment.scala
```diff
@@ -22,7 +22,7 @@ object Environment {
 
   lazy val hadoopVersion  = either("SPARK_HADOOP_VERSION", "2.2.0")
   lazy val sparkVersion   = either("SPARK_VERSION", "1.4.1")
-  lazy val versionSuffix  = either("GEOTRELLIS_VERSION_SUFFIX", "-SNAPSHOT")
+  lazy val versionSuffix  = either("GEOTRELLIS_VERSION_SUFFIX", "-97834e6")
```
lossyrob: This doesn't seem right.

raster/src/main/scala/geotrellis/raster/ByteArrayTile.scala
```diff
@@ -44,7 +44,8 @@ object ByteArrayTile {
         val v = bytes(i)
         if(v == replaceNoData)
           arr(i) = byteNODATA
-        arr(i) = bytes(i)
+        else
+          arr(i) = bytes(i)
```
lossyrob: Could be

```scala
arr(i) = v
```
That will save 1/1000th of a millisecond most likely ;)

Also, the bug this fixes is awful and hilarious. Thanks for catching that.

raster/src/main/scala/geotrellis/raster/io/geotiff/ByteGeoTiffSegment.scala
```diff
@@ -24,7 +33,7 @@ class ByteGeoTiffSegment(val bytes: Array[Byte]) extends GeoTiffSegment {
         val bs = new BitSet(size)
         cfor(0)(_ < size, _ + 1) { i => if ((get(i) & 1) == 0) { bs.set(i) } }
         bs.toByteArray()
-      case TypeByte | TypeUByte => 
+      case TypeByte | TypeUByte =>
```
lossyrob: Whitespace add

raster/src/main/scala/geotrellis/raster/io/geotiff/ByteGeoTiffSegment.scala
```diff
@@ -9,13 +9,22 @@ import java.util.BitSet
...
+}
+
 class ByteGeoTiffSegment(val bytes: Array[Byte]) extends GeoTiffSegment {
   val size: Int = bytes.size
 
   def getInt(i: Int): Int = get(i)
   def getDouble(i: Int): Double = i2d(get(i))
-
-  def get(i: Int): Int = bytes(i) & 0xFF
+  def get(i: Int): Int = bytes(i)
```
lossyrob: Wondering why this is different from https://github.com/geotrellis/geotrellis/blob/320072b10d64387ca1945bb42ad47f69ce28611b/raster/src/main/scala/geotrellis/raster/io/geotiff/ByteGeoTiffSegment.scala#L26
Can you think of a reason that the `get` should return an `Int`?

lossyrob:
:+1:

PR #1192, opened 2015-10-23
Update layer feature

pomadchin:
According to https://github.com/geotrellis/geotrellis/issues/1187

Function just for an Accumulo right now.

*TODO*: 
* Verify layer `keyClass` and `valueClass` match our instance type
* Verify `schema` matches existing layer
* ~~Spread it not only for Accumulo~~

The `Format` object is heavy and written with hands (bad copy pasted read & write functions), but there appears a little heavy call of the `Write` object `update` function (with rddReader as an argument).

By the way, the `update` function has a lot of code from the `read` function.

*UPD*: may be, to define the new exception types (layer existence, header and schema verification) for `read`, `write` and `update` functions is a good idea.

*UPD2*: refactor after pairing: was decided to make a `LayerUpdater`.


spark/src/main/scala/geotrellis/spark/ingest/Ingest.scala
```diff
@@ -74,13 +73,15 @@ object Ingest {
...
     def buildPyramid(zoom: Int, rdd: RasterRDD[K]): List[(Int, RasterRDD[K])] = {
-      if (zoom > 1) {
+      if (zoom >= 1) {
         rdd.persist(cacheLevel)
         sink(rdd, zoom)
         val pyramidLevel@(nextZoom, nextRdd) = Pyramid.up(rdd, layoutScheme, zoom)
         pyramidLevel :: buildPyramid(nextZoom, nextRdd)
-      } else
+      } else {
+        sink(rdd, zoom)
```
lossyrob: This will sink zoom 0, no?

pomadchin: yes, and we should not store it (same with zoom level 1, previously it was removed)?

lossyrob: Hmmm...yeah I don't know what the lowest zoom level users would want to store at. I guess 0 makes sense.

spark/src/main/scala/geotrellis/spark/io/ReaderWriter.scala
```diff
@@ -37,4 +38,17 @@ abstract class FilteringLayerReader[ID, K: Boundable, ReturnType] extends LayerR
...
\ No newline at end of file
+}
+
+abstract class LayerFormat[ID, K: Boundable, V, ReturnType] extends FilteringLayerReader[ID, K, ReturnType] with Writer[ID, ReturnType with RDD[(K, V)]] {
+  val layerReader: FilteringLayerReader[ID, K, ReturnType]
+  val layerWriter: Writer[ID, ReturnType with RDD[(K, V)]]
+
+  /**  Dirty cast, obviously MetaData types have to be equal */
+  def read(id: ID, rasterQuery: RDDQuery[K, MetaDataType], numPartitions: Int): ReturnType =
+    layerReader.read(id ,rasterQuery.asInstanceOf[RDDQuery[K, layerReader.MetaDataType]], numPartitions)
```
lossyrob: We have to figure out how to take out this caste. @echeipesh says it necessitates moving the MetaData type into a type param. He's going to write up an issue on it

lossyrob:
Have you tested the S3, Hadoop and Accumulo LayerFormats?

pomadchin:
Tested Accumulo, Hadoop is in process

lossyrob:
Nice! Thanks for the update

spark/src/main/scala/geotrellis/spark/io/s3/S3LayerHeader.scala
```diff
@@ -8,7 +8,16 @@ case class S3LayerHeader(
   valueClass: String,
   bucket: String,
   key: String
-)
+) {
+  def equals(that: S3LayerHeader) =
```
lossyrob: Why not override `Object.equals`?

pomadchin: Ouch, we don't even need overriding, case class implements eqality

pomadchin:
Hadoop tested

spark/src/main/scala/geotrellis/spark/io/ReaderWriter.scala
```diff
@@ -37,4 +39,14 @@ abstract class FilteringLayerReader[ID, K: Boundable, ReturnType] extends LayerR
 
   def query(layerId: ID, numPartitions: Int): BoundRDDQuery[K, MetaDataType, ReturnType] =
     new BoundRDDQuery(new RDDQuery, read(layerId, _, numPartitions))
+}
+
+abstract class UpdatingLayerWriter[ID, K: Boundable, V, ReturnType] extends Writer[ID, ReturnType] {
+  def update(id: ID, rdd: RDD[(K, V)]): Unit
+
+  def mergeUpdate(id: ID, reader: FilteringLayerReader[ID, K, RDD[(K, V)]], rdd: RDD[(K, V)])
```
lossyrob: Where is the `rdd` parameters used here?

spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerWriter.scala
```diff
@@ -41,17 +40,51 @@ class AccumuloLayerWriter[K: Boundable: JsonFormat: ClassTag, V: ClassTag, Conta
...
+      val keyBounds = boundable.getKeyBounds(rdd)
+
+      if (!boundable.includes(keyBounds.minKey, existingKeyBounds) || !boundable.includes(keyBounds.maxKey, existingKeyBounds))
+        throw new OutOfKeyBoundsError(id)
+
+      val getRowId = (key: K) => index2RowId(existingKeyIndex.toIndex(key))
+
+      rddWriter.write(rdd, table, columnFamily(id), getRowId, oneToOne = false)
+    } catch {
+      case e: HeaderMatchError[_] => throw e.initCause(e)
```
lossyrob: Why are we catching these errors here if we're just rethrowing?

spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerWriter.scala
```diff
@@ -41,17 +40,51 @@ class AccumuloLayerWriter[K: Boundable: JsonFormat: ClassTag, V: ClassTag, Conta
...
+      case e: OutOfKeyBoundsError => throw e.initCause(e)
+      case e: LayerNotExistsError => throw e.initCause(e)
+      case e: Exception => throw new LayerUpdateError(id).initCause(e)
```
lossyrob: Even here, I don't usually see the value of catching errors you'll throw, it generally obfuscates the stack trace for me.

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopLayerWriter.scala
```diff
@@ -46,11 +45,45 @@ class HadoopLayerWriter[K: SpatialComponent: Boundable: JsonFormat: ClassTag, V:
...
+        throw new OutOfKeyBoundsError(id)
+
+      rddWriter.write(rdd, layerPath, existingKeyIndex)
+    } catch {
+      case e: HeaderMatchError[_] => throw e.initCause(e)
+      case e: OutOfKeyBoundsError => throw e.initCause(e)
+      case e: LayerNotExistsError => throw e.initCause(e)
+      case e: Exception => throw new LayerUpdateError(id).initCause(e)
+    }
+  }
```
lossyrob: This seems to share a lot of code with the other implementation. Any way to abstract it out, so the subclasses only provide the necessary types/functionality?

pomadchin: The most obvious to make a `isUpdatable` function, but it is only a sugar (not sure it is necessary), and not reduces code; to make it more generic is harder, due to lots of different types.

pomadchin:
May be it is possible to make update function more generic: `ReturnType <: RDD[(K, V)]` will look how it goes with the client code.

echeipesh:
Can it use `Container with RDD[(K, V)]` like the `write` method or is there a reason that won't work?

pomadchin:
We should some how to get `KeyBounds` in the `mergeUpdate` function (so that's the reason to define subtyping), another point is to make a dirty typecast of the `ReturnType` (https://github.com/pomadchin/geotrellis/blob/da72896c73ed4258c2a6a24e55017324dfa8f2fe/spark/src/main/scala/geotrellis/spark/io/ReaderWriter.scala#L49); 
Yes, want the `update` function to be similar to the `writer` function: `def update(id: ID, rdd: ReturnType): Unit`, what do you think?

spark/src/test/scala/geotrellis/spark/io/PersistenceSpec.scala
```diff
@@ -47,6 +49,16 @@ abstract class PersistenceSpec[K: ClassTag, V: ClassTag] extends FunSpec with Ma
       actual should contain theSameElementsAs expected
     }
 
+    it("should update a layer") {
+      updater.update(layerId, sample)
+    }
+
+    it("should not update a layer") {
```
lossyrob: Can you make this test name more descriptive? What's the reason this update should fail?

lossyrob:
I'm wondering if there's some other cases we should be creating unit tests for. For instance, what happens when we try to update out of the bounds of the original RDD? Would be good to think through edge cases and hit them with unit tests.

Other then that, I think this is about ready. @echeipesh want to take a look and give your thoughts?

pomadchin:
There can be even two cases of out of bounds: completely (similar to an empty set update), and partly. Just need some sample low weight files for tests, or will update keys manually.

pomadchin:
So there can be some problems with the `update` function for `HDFS`, as @echeipesh pointed me:
* https://hadoop.apache.org/docs/current/api/org/apache/hadoop/io/MapFile.html
* https://www.quora.com/How-do-I-update-a-MapFile-in-Hadoop

pomadchin:
Actually, it is no possible to _update_ Hadoop MapFile, the only way to make a sort of an _update_ feature is to copy the entire tiles, merge em with the updated, and write to a new directory.
This functionality suits more to LayerManager (move, rename, delete, etc.).

echeipesh:
I am ok with removing the HadoopLayerUpdater. +1

PR #1194, opened 2015-10-26
Focal operation as flatMap with Neighborhood trimming

### LINKED ISSUE #1196: Focal Operation Optimization, Boundary Slivers
Focal operations on a tile require access to pixels of the neighboring tile when calculating values close to tile borders. We solve this problem by building a tile neighborhood for each tile that includes all of the surrounding tiles.

However, focal neighborhood rarely require access to more than 3 to 10 pixels past the tile boundary, meaning that 256x246 (depending on tile size) pixels of every neighboring tile go unused. It is likely that the processing cost of cutting neighboring tiles into tile slivers required by the neighborhood parameter would be more than made up by the savings in serialization and shipping costs in the RDD groupBy.

This would likely require enhancements to `CompositeTile` to handle stitching uneven tiles into a neighborhood tile.

echeipesh:
resolves #1196

This PR implements `FocalOperation` using `flatMap` and `groupByKey` instead of a series of `groupByKey`. 
Best effort is made to discard contributions to neighbors that are outside of bounds, but all focal operations will now succeed even if no tile bounds are provided.

New implementation exposes API for doing focal operations on `RDD[(K: SpatialComponent, Tile)]` in addition to `RasterRDD[K]`.

As an optimization, when contributing pixels to neighboring tiles only the pixels required by the `Neighborhood` extent will be cropped and serialized. This avoids shipping the entire neighboring tile. For instance for the corner neighbor tiles this results in shipping 5x5 pixels vs 256x256 pixels, given focal extent of 5 and tile size of 256.

In a micro-benchmark using local spark context focal operations time is reduced from ~2.5s to ~500ms.


spark/src/main/scala/geotrellis/spark/op/focal/FocalOperation.scala
```diff
@@ -10,160 +12,117 @@ import org.apache.spark.SparkContext._
...
-    }
-
-    val x = (xStart + to % 3) % 3 match {
-      case 2 => -1
-      case x => x
+  object Direction extends Enumeration {
+    type Direction = Value
+    val Center, Top, TopRight, Right, BottomRight, Bottom, BottomLeft, Left, TopLeft = Value
+
+    implicit class withMethods(direction: Value) {
```
lossyrob: Is there a specific reason to use `Enumeration` and implicits instead of `case objects` or something here?

spark/src/main/scala/geotrellis/spark/op/focal/FocalOperation.scala
```diff
@@ -10,160 +12,117 @@ import org.apache.spark.SparkContext._
...
+        addSlice(SpatialKey(col+1, row), Right, tile.crop(cols-mm, 0, cols, rows))
+        addSlice(SpatialKey(col, row-1), Top, tile.crop(0, 0,  cols, mm))
+        addSlice(SpatialKey(col, row+1), Bottom, tile.crop(0, rows-mm, cols, rows))
+
+        addSlice(SpatialKey(col-1, row-1), TopLeft, tile.crop(0, 0, mm, mm))
+        addSlice(SpatialKey(col+1, row-1), TopRight, tile.crop(cols-mm, 0, cols, mm))
+        addSlice(SpatialKey(col+1, row+1), BottomRight, tile.crop(cols-mm, rows-mm, cols, rows))
+        addSlice(SpatialKey(col-1, row+1), BottomLeft, tile.crop(0, rows-mm, mm, rows))
+
+        slivers.take(sliverCount)
```
lossyrob: Use `ListBuffer` here instead, so you don't have to keep track of `sliverCount` and do this `take`?

lossyrob:
Very cool. Looks good; there's a question and a nitpick, feel free to merge otherwise. :+1:

spark/src/main/scala/geotrellis/spark/op/focal/FocalOperation.scala
echeipesh: Really good question. I've read that enums will be serialized better and took it on faith. Your question prompted me to test it and I could not get Kryo to serialize the num sensibly. It would always produces something like: `??scala.Enumeration$Va￬???geotrellis.spark.op.focal.FocalOperation$Directionﾤ?????`

Perhaps the statement applies to java enums?

I switched to sealed trait and was able to register them and get it to serialize into 2 byte id through kryo. It didn't make anything faster locally but I bet it would have some non-zero impact on network traffic

PR #1198, opened 2015-11-02
Missing Avro codecs for unsigned tile types

pomadchin:


lossyrob:
Good call, thanks @pomadchin !

PR #1202, opened 2015-11-13
[1200] Fix support for modeltransformation

jdries:
Here's a fix for my geotiff reading issue, together with a test file. Perhaps the test could be improved, but at least it exposes the issue.


jdries:
There's a typo in the code, will fix first

PR #1207, opened 2015-11-17
Added ability to color double tiles with double color breaks.

lossyrob:
Previously we weren't allowed to color a tile of doubles with color breaks that were also doubles. For instance, if we had breaks that corresponding to the values of 0.0 - 1.0, and colors to match those breaks, we now are able to render a tile that has values 0.0 - 1.0 with those breaks.

ColorBreaks changes from being a `case class ColorBreaks(limits: Array[Int], colors: Array[Int])` to a trait, with `IntColorBreaks` and `DoubleColorBreaks` implementations. This is an API change.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/render/ColorBreaks.scala
+++ b/raster/src/main/scala/geotrellis/raster/render/ColorBreaks.scala
@@ -21,30 +21,39 @@ import java.util.Locale
 
 /**
   * ColorBreaks describes a way to render a raster into a colored image.
+  */
+trait ColorBreaks extends Serializable {
+  def colors: Array[Int]
+  def length: Int
+  def replaceColors(newColors: Array[Int]): ColorBreaks
+  def mapColors(f: Int => Int): ColorBreaks
+  def toColorMap(): ColorMap = toColorMap(ColorMapOptions.Default)
+  def toColorMap(options: ColorMapOptions): ColorMap
+}
+
+/**
+  * IntColorBreaks describes a way to render a raster into a colored image.
   *
-  * This class defines a set of value ranges and assigns a color to
+  * This class defines a set of Int value ranges and assigns a color to
   * each value range.
   *
   * @param limits  An array with the maximum value of each range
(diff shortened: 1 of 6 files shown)


raster/src/main/scala/geotrellis/raster/render/ColorBreaks.scala
```diff
@@ -21,30 +21,73 @@ import java.util.Locale
...
+    val colorsStr =
+      colors
+        .map("%08x" formatLocal(Locale.ENGLISH, _))
+        .mkString("Array(", ", ", ")")
+    s"IntColorBreaks($limitsStr, $colorsStr)"
   }
+}
+
+/**
+  * IntColorBreaks describes a way to render a raster into a colored image.
```
echeipesh: docstring

raster/src/main/scala/geotrellis/raster/render/ColorBreaks.scala
```diff
@@ -21,30 +21,73 @@ import java.util.Locale
...
+/**
+  * DoubleColorBreaks describes a way to render a raster into a colored image.
+  *
+  * This class defines a set of Double value ranges and assigns a color to
+  * each value range.
+  *
+  * @param limits  An array with the maximum value of each range
+  * @param colors  An array with the color assigned to each range
+  */
+class DoubleColorBreaks(val limits: Array[Double], val colors: Array[Int]) extends ColorBreaks {
```
moradology: Why not case classes on these things? I assume they are immutable and could thus benefit from free product/serializable

lossyrob: Good call on serializable. but not sure it's worth making it a case class (adds bytecode for equals, creates an object, not necessary if we're not going to use it)

lossyrob: Oh wait. ColorBreaks extends Serializable, so it should be fine

moradology:
+1

PR #1208, opened 2015-11-18
Zonal summary refactors

lossyrob:
This changes the way zonal summaries happen in Spark and under the covers in core. This is based off the insight that a `Raster` is just a `Feature[Extent, Tile]`, and we can do zonal summaries on the more general `RDD[Feature[G, D]]` or `RDD[(K, Feature[G, D])]`.


lossyrob:
@echeipesh ready for review

lossyrob:
Failing because #1220 needs to be merged.

echeipesh:
I like this a lot. Do you think the MultiPolygon would ever get large enough that it would be worth using Broadcast?
+1 otherwise

lossyrob:
Task bodies are serialized and broadcasted, so unless we were reusing the multipolygon between tasks, we don't need to broadcast.

PR #1220, opened 2015-11-24
Fixing Travis publishing; Elevation operation fix

### LINKED ISSUE #1197: Fix RowTransformCheck_LatLngToWebMercator scalacheck test nondeterministicly failing
This SalaCheck test fails travis randomly, and we aren't able to capture what arguments fail it, because it prints out array `.toString`'s:

```
[info] ! RowTransform.Stays within thresholds: Falsified after 18 passed tests.
[info] > ARG_0: TestCase(Extent(37.36787900133859,-78.97170323633226,74.55484139675815,-44.029243473735185),[D@3280030a,[D@31996cf3)
[info] > ARG_1: Threshold(4.945130587159675)
```

We need to fix that output, and then run it until it fails, to capture the failing parameters. If there's something wrong with the code, then it needs to be fixed, or if the test parameters are faulty, the test needs to be fixed.

lossyrob:
This fixes:
- The publish on travis was broken because something that compiled in scala 2.10 didn't in scala 2.11, and was only showing up when there was a publish step taking place.
- RowTransformCheck, which was failing randomly, was actually passing randomly. I rewrote the test to properly test RowTransform.
- `SurfacePointCalculation` was not consistently assigning neighbor values of NoData to the focal value.

Also, moved WKT and WKB to proper package names.

Fixes #1197


echeipesh:
Sorry if there is an obvious answer, but why are some of the east/base values not going through `getSafeVal`:
https://github.com/geotrellis/geotrellis/pull/1220/files#diff-73e453326b38c65458ab725608b793d6L199

lossyrob:
It's part of an optimization step that I had done with a lot of benchmarking, sweat and tears to try and get the replacement focal operation architecture to match the speeds of the old architecture. `getSafeVal` includes bounds checks that are not necessary for every cell, since some cells we know will be within valid cells. So, we save an if statement check in a tight fast loop. I could have switched every call to use getSaveVal, which would have done the unnecessary bounds check, but also did the `isData` check, which is necessary. Instead I moved the `isData` check to another place and out of `getSafeVal`. This saves on one conditional branch in the code for values in the middle of the area of interest, which may be splitting hairs, but I figured I'd go with the more logically optimized version.

echeipesh:
Ok, deep optimization magic works for me, +1

PR #1224, opened 2015-12-01
WIP: layer management functions (LayerDeleter)

pomadchin:


spark/src/main/scala/geotrellis/spark/io/ReaderWriter.scala
```diff
@@ -3,6 +3,7 @@ package geotrellis.spark.io
 import geotrellis.spark._
 import org.apache.spark.rdd.RDD
 
+/** TODO: split this files into separate files in its own package */
```
lossyrob: :+1:

spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerDeleter.scala
```diff
@@ -0,0 +1,51 @@
...
+    } catch {
+      case e: Exception => throw new LayerDeleteError(id).initCause(e)
+    }
+  }
+}
+
+object AccumuloLayerDeleter {
+  def apply[K: Boundable: JsonFormat: ClassTag](instance: AccumuloInstance) =
+    new AccumuloLayerDeleter[K](AccumuloAttributeStore(instance.connector), instance.connector)
+}
```
lossyrob: Add apply method for (attributeStore, connector)

spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerDeleter.scala
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.accumulo.core.security.Authorizations
+import spray.json.JsonFormat
+import org.apache.accumulo.core.data.{Range => AccumuloRange}
+import org.apache.hadoop.io.Text
+import geotrellis.spark.io.json._
+import spray.json.DefaultJsonProtocol._
+import scala.collection.JavaConversions._
+import scala.reflect.ClassTag
+
+/** attributeStore & connector: mb pass only connector */
```
lossyrob: Taken care of with object apply method

spark/src/main/scala/geotrellis/spark/io/AttributeStore.scala
```diff
@@ -9,6 +9,13 @@ trait AttributeStore[F[_]] extends AttributeCaching[F] {
   def readAll[T: Format](attributeName: String): Map[LayerId, T]
   def write[T: Format](layerId: LayerId, attributeName: String, value: T): Unit
   def layerExists(layerId: LayerId): Boolean
+  def delete(layerId: Option[LayerId], attributeName: Option[String]): Unit
+  def delete(layerId: LayerId, attributeName: String): Unit =
+    delete(Some(layerId), Some(attributeName))
+  def delete(layerId: LayerId): Unit =
+    delete(Some(layerId), None)
+  def delete(attributeName: String): Unit =
```
lossyrob: Don't think we should let the user blanket delete an attribute from all layers.

lossyrob: I think the API might be better as

```scala
def delete(layerId: LayerId): Unit // Deletes all attributes out of the store for LayerId
def delete(layerId: LayerId, attributeName: String): Unit // Deletes only attributeName attribute for LayerId
```

@echeipesh what do you think?

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopAttributeStore.scala
```diff
@@ -87,6 +94,16 @@ class HadoopAttributeStore(val hadoopConfiguration: Configuration, attributeDir:
...
+
+  def delete(layerId: Option[LayerId], attributeName: Option[String]): Unit = {
+    val path = optionAttributePath(layerId, attributeName)
+
+    if(fs.exists(path)) {
+      fs.delete(path, false)
+    }
+
+    clearCache()
+  }
```
lossyrob: Alternate:

```scala
  private def _delete(path: String): Unit = {
    if(fs.exists(path)) { fs.delete(path, false) }
    clearCache()
  }

  def delete(layerId: LayerId, attributeName: String): Unit =
    _delete(Path(s"${id.name}___${id.zoom}___${name}.json"))

  def delete(layerId: LayerId, attributeName: String): Unit =
    _delete(Path(s"${id.name}___${id.zoom}___*.json"))
```

spark/src/test/scala/geotrellis/spark/io/PersistenceSpec.scala
```diff
@@ -55,5 +68,21 @@ abstract class PersistenceSpec[K: ClassTag, V: ClassTag] extends FunSpec with Ma
       val expectedV: V = sample.filter(_._1 == key).values.first()
       readV should be equals expectedV
     }
+
+    // TODO: more specific exception type?
+    it("should delete a layer") {
+      deleter.delete(deleteLayerId)
+
+      // TODO: fix weird workaround, due to hadoop failure on count only
+      intercept[Exception] {
```
lossyrob: Should this catch a LayerNotFound error?

spark/src/test/scala/geotrellis/spark/io/PersistenceSpec.scala
```diff
@@ -55,5 +68,21 @@ abstract class PersistenceSpec[K: ClassTag, V: ClassTag] extends FunSpec with Ma
...
+        try {
+          reader.read(deleteLayerId)
```
lossyrob: Hadoop should def throw at this point if the layer isn't found, must be a bug with the HadoopLayerReader

spark/src/main/scala/geotrellis/spark/io/AttributeStore.scala
echeipesh: That makes sense. Does it throw if the layer/attribute does not exist or silent pass through?

pomadchin: Throws (path not found for Hadoop, and similar errors for Accumulo & S3), should we it be handled somehow, to throw more specific exception?

echeipesh: If it's going to throw when layer does not exist I would say it needs to chain the exception on `LayerNotExistsError`. As long as those specific exceptions exist they should be used when applicable. Also that probably should be called `LayerNotFoundError` to be consistent with `AttributeNotFoundError`.

pomadchin:
The question still remains, is it good to keep obfuscating exceptions via `try` blocks? Fixed it a little (tried to), @echeipesh and @lossyrob want to know your opinons

lossyrob:
I'm just this minute running into something where there's a generic `LayerReadError` that I really wish I could have a more specific exception for, but I don't. It's coming from https://github.com/geotrellis/geotrellis/blob/e6afeaf075036ac431974fbaf9035909ed1b9acd/spark/src/main/scala/geotrellis/spark/io/s3/S3LayerReader.scala#L56 . It says that the LayerMetaData is not found but it's clearly in the _attribute folder on S3, so that does not seem right at all. So this is immediately applicable pain that I'm feeling from the obfuscation.

Let's get rid of the obfuscation. We should catch errors that we know will not be helpful messages and make them helpful messages, but we should not catch all errors and take their helpfulness away.

pomadchin:
Should we continue catching smth like `AttributeNotFoundError` (https://github.com/geotrellis/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerReader.scala#L42) or completely to remove `try` blocks?

lossyrob:
There should be additional message information if we are going to wrap it. Like:

```scala
case e: AttributeNotFoundError => throw new LayerDeleteError(id, e.getMessage).initCause(e)
```

or something.

lossyrob:
Here's an example of an exception being caught and providing a more interesting message:
https://github.com/aws/aws-sdk-java/blob/54d3f55d7a03581291b9225b069769d9ff39f47f/aws-java-sdk-core/src/main/java/com/amazonaws/auth/InstanceProfileCredentialsProvider.java#L219

I think this idea of wrapping a whole lot of code in a try block is what the problem is. I don't think that's a good pattern. We should be wrapping only a small amount of code in trys and specifically saying what's wrong if what goes wrong needs to be in the context that it happened when trying to perform that bit of functionality (like the linked try block does).

spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerUpdater.scala
```diff
@@ -17,25 +17,29 @@ class AccumuloLayerUpdater[K: Boundable: JsonFormat: ClassTag, V: ClassTag, Cont
   extends LayerUpdater[LayerId, K, V, Container with RDD[(K, V)]] {
 
   def update(id: LayerId, rdd: Container with RDD[(K, V)]) = {
-    try {
-      if (!attributeStore.layerExists(id)) throw new LayerNotExistsError(id)
-      implicit val sc = rdd.sparkContext
-      implicit val mdFormat = cons.metaDataFormat
+    if (!attributeStore.layerExists(id)) throw new LayerNotFoundError(id)
+    if (rdd.isEmpty()) throw new LayerUpdateError(id, ": empty rdd update")
```
lossyrob: This actually computes a partition, is there a way to check this that doesn't perform an action?

pomadchin:
```scala
def isEmpty(): Boolean = withScope {
    partitions.length == 0 || take(1).length == 0
  }
```

For rdd it is possible to have one but empty partition. The other option we can set `try` block to wrap `boundable.getKeyBounds` function, and to catch unsupported operation exception: due to `reduce` used in its implementation. To provide a valid function application to the empty rdd we have to make an empty `KeyBounds` instance (to use `foldLeft` instead of `reduce`), @echeipesh also had some ideas about it.

lossyrob: I was thinking more along the lines of, catch the UnsupportedOperatinException around the line of code that would actually throw it.

spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloAttributeStore.scala
```diff
@@ -41,6 +41,19 @@ class AccumuloAttributeStore(connector: Connector, val attributeTable: String) e
     scanner.iterator.map(_.getValue)
   }
 
+  private def _delete(layerId: LayerId, attributeName: Option[String]): Unit = {
```
echeipesh: _ is not required here, the types of parameters is enough to disambiguate it.

spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerDeleter.scala
```diff
@@ -0,0 +1,52 @@
...
+        new AccumuloRange(new Text(long2Bytes(min)), new Text(long2Bytes(max)))
+      }
+
+    val ranges = decompose(keyBounds, keyIndex)
+    val numThreads = 1
+    val config = new BatchWriterConfig()
+    config.setMaxWriteThreads(numThreads)
+    val deleter = connector.createBatchDeleter(header.tileTable, new Authorizations(), numThreads, config)
+    deleter.fetchColumnFamily(columnFamily(id))
+    deleter.setRanges(ranges)
```
echeipesh: We should not need to decompose and set ranges here, we're deleting all the rows.

As a side note this is incompatible with `SpaceTimeAccumuloRDDWriter`, which will write multiple zoom levels to single column family. It should be noted in `SpaceTimeAccumuloRDDReader` and `SpaceTimeAccumuloRDDWriter` comment. But I don't think it's worth while to create abstractions to solve this issue. + @lossyrob

spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerDeleter.scala
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.accumulo.core.security.Authorizations
+import spray.json.JsonFormat
+import org.apache.accumulo.core.data.{Range => AccumuloRange}
+import org.apache.hadoop.io.Text
+import geotrellis.spark.io.json._
+import spray.json.DefaultJsonProtocol._
+import scala.collection.JavaConversions._
+import scala.reflect.ClassTag
+
+class AccumuloLayerDeleter[K: Boundable: JsonFormat: ClassTag]
```
echeipesh: `Boundable` type class is not actually used here.

pomadchin: :+1:  if we don't need to set keys ranges, we don't need the type param at all

spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerDeleter.scala
echeipesh: Updated: `SpaceTimeAccumuloRDDWriter` doesn't actually use zoom information anymore. Deleting all the rows for column family is always safe way to delete a layer.

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopAttributeStore.scala
```diff
@@ -27,10 +27,16 @@ class HadoopAttributeStore(val hadoopConfiguration: Configuration, attributeDir:
     new Path(attributeDir, fname)
   }
 
+  private def _delete(layerId: LayerId, path: Path): Unit = {
+    if(!layerExists(layerId)) throw new LayerNotFoundError(layerId)
+    HdfsUtils
+      .listFiles(new Path(attributeDir, path), hadoopConfiguration)
```
echeipesh: I think we can call `HdfsUtils.deletePath` here, it should support globing. And if it doesn't it would be good to know : )

pomadchin: It doesn't support globing ): tried it in late commits, then during not working test found that attributes are not deleting.

spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloAttributeStore.scala
pomadchin: Just wanted to keep code style: in the `S3AttributeStore` we need `_delete` function

```scala
private def _delete(layerId: LayerId, path: String): Unit = {
  if(!layerExists(layerId)) throw new LayerNotFoundError(layerId)
  s3Client.deleteObject(bucket, path)
}

def delete(layerId: LayerId): Unit =
  _delete(layerId, attributePath(layerId))

  def delete(layerId: LayerId, attributeName: String): Unit =
    _delete(layerId, attributePath(layerId, attributeName))
```

pomadchin: :+1:

spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerDeleter.scala
pomadchin: Just remembered: already tried to set only columnFamily, seems like it is not possible, or there is some workaround? 

https://github.com/apache/accumulo/blob/master/core/src/main/java/org/apache/accumulo/core/client/impl/TabletServerBatchReader.java#L107

spark/src/main/scala/geotrellis/spark/io/s3/S3AttributeStore.scala
```diff
@@ -33,6 +33,9 @@ class S3AttributeStore(bucket: String, rootPath: String) extends AttributeStore[
 
   def path(parts: String*) = parts.filter(_.nonEmpty).mkString("/")
 
+  def attributePath(id: LayerId): String =
+    path(rootPath, "_attributes", s"*__${id.name}__${id.zoom}.json")
```
echeipesh: I am pretty certain that `DeleteObjectRequest` to s3client does not glob. Which presents a really unfortunate problem here. The reason attribute name is listed first here is so we can use `listObjects` to implement `.readAll(attributeName)`. Now we need to do the opposite, get a list of all attributes per layer.

This would not be a problem if we stored attributes with each layer... which we switched from.

spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerDeleter.scala
echeipesh: I believe if you create a range from -inf to +inf will do the trick here:
https://github.com/geotrellis/geotrellis/pull/1224/files#diff-5c4305182b01c1f7c572a3e031799577R37

spark/src/main/scala/geotrellis/spark/io/s3/S3AttributeStore.scala
pomadchin: Yes, I see; so how to solve it, to change attribute storage and to make it depend on layerId? Or to search for some workaround?

spark/src/main/scala/geotrellis/spark/io/s3/S3LayerDeleter.scala
```diff
@@ -0,0 +1,54 @@
...
+    if (!attributeStore.layerExists(id)) throw new LayerNotFoundError(id)
+    val (header, _, keyBounds, keyIndex, _) = try {
+      attributeStore.readLayerAttributes[S3LayerHeader, Unit, KeyBounds[K], KeyIndex[K], Unit](id)
+    } catch {
+      case e: AttributeNotFoundError => throw new LayerDeleteError(id).initCause(e)
+    }
+
+    val bucket = header.bucket
+    val prefix = header.key
+    val numPartitions = 1
```
echeipesh: It's an interesting idea to make the executors do the work, but you need some parallelism here, `sc.defaultParallelism` should do

pomadchin: I wonder, can we make smth like 
```scala
s3Client
      .listObjectsIterator(bucket, prefix)
      .foreach { os => s3Client.deleteObject(bucket, os.getKey) }
```

but using `multiObjectDeleteRequest`?

echeipesh: I did not know about you can set multiple keys. Likely the best course of action will be to just do that from the driver.
If you have a big layer it would be really interesting to test if multi-threading makes it much faster like used here: https://github.com/geotrellis/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/io/s3/S3RDDWriter.scala

But I don't think the second part is critical for first implementation.

spark/src/main/scala/geotrellis/spark/io/s3/S3AttributeStore.scala
echeipesh: I would expect `readAll` to be used more often than delete. That means the only thing you can do is list all the keys in the attribute store, filter for this layer and delete those.

spark/src/main/scala/geotrellis/spark/io/s3/S3LayerDeleter.scala
```diff
@@ -0,0 +1,55 @@
...
+    if (!attributeStore.layerExists(id)) throw new LayerNotFoundError(id)
+    val (header, _, keyBounds, keyIndex, _) = try {
+      attributeStore.readLayerAttributes[S3LayerHeader, Unit, KeyBounds[K], KeyIndex[K], Unit](id)
+    } catch {
+      case e: AttributeNotFoundError => throw new LayerDeleteError(id).initCause(e)
+    }
+
+    val bucket = header.bucket
+    val prefix = header.key
+    val numPartitions = sc.defaultParallelism
```
pomadchin: I wonder, should we pass as the delete function argument?

echeipesh: You pointed out that it's possible to have multi-key delete request. Judging by listing request we can put around 1024 keys into single request. The main resource we're conserving here are number of S3 requests. I would expect that this would both out-perform, be cheaper, and simpler than distributing the tasks over executors.

pomadchin: :+1: i wonder, will it be better just to get list of keys for the `header.key`, then to iterate over them, group by 1024 keys per request?
This will help us to avoid index decomposing and to get rid of type param.

echeipesh: Yes, that's a more general solution, I'm down with that.

spark/src/main/scala/geotrellis/spark/io/s3/S3LayerDeleter.scala
```diff
@@ -0,0 +1,40 @@
...
+import geotrellis.spark.io._
+import spray.json.JsonFormat
+import spray.json.DefaultJsonProtocol._
+import scala.collection.JavaConversions._
+
+class S3LayerDeleter(attributeStore: AttributeStore[JsonFormat]) extends LayerDeleter[LayerId] {
+
+  def getS3Client: () => S3Client = () => S3Client.default
+
+  def deleteListing(s3: S3Client, bucket: String, listing: ObjectListing): Unit = {
```
echeipesh: This looks a method that should be on the S3Client. Could use `tailrec` tag to make certain can't blow the stack.

echeipesh:
+1

spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerDeleter.scala
```diff
@@ -0,0 +1,41 @@
...
+import spray.json.JsonFormat
+import org.apache.accumulo.core.data.{Range => AccumuloRange}
+import spray.json.DefaultJsonProtocol._
+import scala.collection.JavaConversions._
+
+class AccumuloLayerDeleter(attributeStore: AttributeStore[JsonFormat], connector: Connector) extends LayerDeleter[LayerId] {
+
+  def delete(id: LayerId): Unit = {
+    if (!attributeStore.layerExists(id)) throw new LayerNotFoundError(id)
+    val (header, _, _, _, _) = try {
```
lossyrob: Is there an exception that we can catch here that would be thrown if the layer doesn't exist? It's not ideal to make a listing call and then a fetch call, just because each listing costs $$ so if we can get away with not making it it would be the cheaper option

echeipesh: That's a good point, probably most applicable to S3 implementation. We should able to catch `AttributeNotFoundError` when attempting to read layer metadata. That should be generated by all the AttributeStores.

PR #1226, opened 2015-12-02
Handle S3Client IAM credential expiration

echeipesh:
S3 credentials are handled by InstanceProfileCredentialsProvider when they are expired.
If we pass the provider to AWS `AmazonS3Client` it will delegate to provider `.getCredentials` call on requests, triggering refresh logic.

Confusingly both AWS SDK and our code as `AmazonS3Client` but they follow naming convention in both packages.


lossyrob:
+1

PR #1227, opened 2015-12-02
Layer management functions (LayerCopier) 

pomadchin:
WARN: Depends on LayerDeleter pr (https://github.com/geotrellis/geotrellis/pull/1224)


pomadchin:
Jut wondering, do we need reindexer? Because this copier _can_ reindex data with just passed different `layerReader` and `layerWriter` passed?

lossyrob:
In order for Copier to behave like a Reindexer, it would need to copy, delete the old, then move the new into the place the old was. A reindex implementation of, say, Accumulo could do this more efficiently, no?

spark/src/main/scala/geotrellis/spark/io/SparkLayerCopier.scala
```diff
@@ -0,0 +1,36 @@
...
+import spray.json._
+import scala.reflect._
+
+class SparkLayerCopier[Header: JsonFormat, K: Boundable: JsonFormat: ClassTag, V: ClassTag, Container]
+  (attributeStore: AttributeStore[JsonFormat],
+   layerReader: FilteringLayerReader[LayerId, K, Container],
+   layerWriter: Writer[LayerId, Container with RDD[(K, V)]])
+  (implicit val cons: ContainerConstructor[K, V, Container], containerEv: Container => Container with RDD[(K, V)]) extends LayerCopier[LayerId] {
+
+  def copy(from: LayerId, to: LayerId): Unit = {
```
pomadchin: Is it possible to rewrite this function to: 

```scala
def copy(from: LayerId, to: LayerId): Unit = {
    if (!attributeStore.layerExists(from)) throw new LayerNotFoundError(from)
    if (attributeStore.layerExists(to)) throw new LayerExistsError(to)    
    try {
      layerWriter.write(to, layerReader.read(from))
    } catch { 
      case e: Exception => new LayerCopyError(from, to).initCause(e)
    }
    attributeStore.copy(from, to)
  }
```

For Accumulo, seems there no significant difference. But for other backends definitely lots of difference. 
We just need to decide, should we provide `inspark` copy, or not? If No, I will use `attributeStore.copy` function.
The pros of this `SparkLayerCopier` now that it is generic and can be used with any backend.

pomadchin:
WARN: has broken S3 test, depends on #1237

pomadchin:
After fail with test, decided to have a close look to all the code; found an issue (this idea somehow flew out of my head and not critical to Accumulo, so didn't catch it on my cluster with real data):
Header update: we have to update `path` for `Hadoop` and `(bucket, key)` for the `S3`.
As a consequence, same problem with the `LayerMover`.

pomadchin:
For Accumulo backend, probably possible to use: https://accumulo.apache.org/1.6/apidocs/org/apache/accumulo/core/iterators/user/TransformingIterator.html

pomadchin:
PR #1241 includes LayerCopier.

PR #1229, opened 2015-12-03
Slippy tile reader/writer + other changes

lossyrob:
This provides a way to write RasterRDD[SpatialKey] to Slippy map tiles to the local file system/hdfs/s3 using the Hadoop API. Also provides a reader that can read the RDD back in or just one tile.

Whey I say slippy map tiles, I mean tiles that are in the `/z/x/y` file system for storing tiles, the thing that is often mislabeled as a TMS. (See https://gist.github.com/tmcw/4954720 for insight, thanks @caseypt for pointing that article out)

Other changes contained in this PR:
- Allow saving to hadoop to be a map step. In another project I've found this as a useful way to pyramid up; this is the `setupWrite` methods that are added to saving to hadoop and the slippy tile writer.
- Pyramid has functionality that will pyramid up all the levels between some max and min zoom.
- Allow setting of the region when reading from S3InputFormat.
- Add SpaceTimeGeoTiffInputFormat for Hadoop.
- Refactored S3InputFormat and BinaryInputFormat to allow keys to be created off of path/s3key
- Removed `application.conf` from `geotrellis-spark`, as this can overwrite user settings.
- Fixed TileMerge to work with UByte an UShort types correctly

TODO: 
- [ ] Test S3 reader. Perhaps move things in the reader to mirror the writer and only use the Hadoop API?


spark/src/main/scala/geotrellis/spark/io/slippy/package.scala
```diff
@@ -0,0 +1,29 @@
...
+    def read(zoom: Int, key: SpatialKey, bytes: Array[Byte]): Tile =
+      SingleBandGeoTiff(bytes).tile
+  }
+
+  implicit object MultiBandTileTileByteReader extends TileByteReader[MultiBandTile] {
+    def read(zoom: Int, key: SpatialKey, bytes: Array[Byte]): MultiBandTile =
+      MultiBandGeoTiff(bytes).tile
+  }
+
+  implicit def tileTileByteWriter: TileByteWriter[Tile] =
```
moradology: Consider 'defaultTile' or the like - tileTile is kind of weird

project/Dependencies.scala
```diff
@@ -55,8 +55,9 @@ object Dependencies {
   val caliper       = ("com.google.code.caliper" % "caliper" % "1.0-SNAPSHOT"
     from "http://plastic-idolatry.com/jars/caliper-1.0-SNAPSHOT.jar")
 
-  val nscalaTime    = "com.github.nscala-time" %% "nscala-time" % "1.6.0"
+  val nscalaTime    = "com.github.nscala-time" %% "nscala-time" % "2.6.0"
 
+//  val awsSdkS3      = "com.amazonaws" % "aws-java-sdk-s3" % "1.10.11"
   val awsSdkS3      = "com.amazonaws" % "aws-java-sdk-s3" % "1.9.34"
```
echeipesh: Did you mean to upgrade this and changed your mind?

project/Environment.scala
```diff
@@ -21,7 +21,7 @@ object Environment {
     Properties.envOrElse(environmentVariable, default)
 
   lazy val hadoopVersion  = either("SPARK_HADOOP_VERSION", "2.2.0")
-  lazy val sparkVersion   = either("SPARK_VERSION", "1.4.1")
+  lazy val sparkVersion   = either("SPARK_VERSION", "1.5.0")
```
echeipesh: We should probably depend on 1.5.1 or 1.5.2. EMR only supports 1.5.0 as of right now but I'm there should not be API breaks with minor version.

raster/src/main/scala/geotrellis/raster/Raster.scala
```diff
@@ -33,6 +33,30 @@ case class Raster(tile: Tile, extent: Extent) extends Product2[Tile, Extent] {
 
   def asFeature(): Feature[Extent, Tile] = ExtentFeature(extent, tile)
 
+  def get(point: Point): Int =
+    get(point.x, point.y)
+
+  def get(col: Int, row: Int): Int =
+    tile.get(col, row)
+
+  def get(x: Double, y: Double): Int =
```
echeipesh: These two `get` methods should not have the same name. They do different things and with automatic upgrades from Int to Double there may be surprising results. I've had this happen with another `geotrellis.raster` class and it wasn't great. Maybe `getPointValue` and `getDoublePointValue`.

publish-local.sh
```diff
@@ -1,3 +1,16 @@
...
+#./sbt "project index" publish-local && \
+#./sbt "project engine" publish-local && \
+#./sbt "project testkit" publish-local && \
+#./sbt "project services" publish-local && \
+#./sbt "project jetty" publish-local && \
+./sbt "project spark" publish-local && \
+#./sbt "project spark-etl" publish-local && \
+./sbt "project gdal" publish-local #&& \
+#./sbt "project geotools" publish-local && \
+#./sbt "project slick" publish-local
```
lossyrob: TODO: Uncomment

spark/src/main/scala/geotrellis/spark/ingest/Pyramid.scala
```diff
@@ -67,4 +67,38 @@ object Pyramid extends Logging {
...
+        runLevel(nextRdd, nextZoom)
+      } else {
+        (f(thisRdd, thisZoom), thisZoom)
+      }
+
+    runLevel(rdd, startZoom)._1
+  }
+
+  def upLevels[K: SpatialComponent: ClassTag](rdd: MultiBandRasterRDD[K], layoutScheme: LayoutScheme, startZoom: Int)
+                                             (f: (MultiBandRasterRDD[K], Int) => MultiBandRasterRDD[K]): MultiBandRasterRDD[K] =
```
echeipesh: Minor, but took a little while to figure out how this pyramids, that `f` is expected to save in a `.map` step. Wish there was a good name for it, like `hook` or `visit` (for Visitor pattern, right?)

spark/src/main/scala/geotrellis/spark/io/s3/TemporalGeoTiffS3InputFormat.scala
```diff
@@ -6,13 +6,32 @@ import geotrellis.raster.io.geotiff._
...
+    job.getConfiguration.set(GEOTIFF_TIME_FORMAT, timeFormat)
+  }
+
+  def getTimeTag(job: JobContext) ={
+    job.getConfiguration.get(GEOTIFF_TIME_TAG, "TIFFTAG_DATETIME")
+  }
+
+  def getTimeFormatter(job: JobContext): DateTimeFormatter = {
+    val df = job.getConfiguration.get(GEOTIFF_TIME_FORMAT)
+    if(df == null) { DateTimeFormat.forPattern("YYYY:MM:dd HH:mm:ss") }
```
echeipesh: spacing

spark/src/main/scala/geotrellis/spark/io/slippy/SlippyTileWriter.scala
```diff
@@ -0,0 +1,32 @@
...
+import org.apache.spark.rdd._
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.mapreduce.Job
+import org.apache.hadoop.mapreduce.{InputSplit, TaskAttemptContext}
+import org.apache.hadoop.fs.Path
+
+import java.io.File
+import scala.collection.JavaConversions._
+
+trait TileByteWriter[T] extends Serializable {
```
echeipesh: This type class is dangling, unused. Also I'm not sure it's a good interface. Kind of implies that there is only one way to write a particular T instance to bytes, when it probably depends on the extension rather than provides it. Will this have life or a dead branch?

I'll also follow up on @moradology PR to see if it makes sense there anyway.

spark/src/main/scala/geotrellis/spark/io/slippy/FileSlippyTileReader.scala
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+
+  def read(zoom: Int)(implicit sc: SparkContext): RDD[(SpatialKey, T)] = {
+    val paths = {
+      listFiles(uri)
+        .flatMap { file =>
+          val path = file.getAbsolutePath
+          path match {
+            case TilePath(z, x, y) if z.toInt == zoom => Some((SpatialKey(x.toInt, y.toInt), path))
```
echeipesh: `TilePath` regex assumes that zoom is leading directory, which seems like the only alternative. So we can just append that to the root path and list only that zoom level. Is there a crazier structure for slippy tiles where that would not be true?

spark/src/main/scala/geotrellis/spark/io/s3/S3Client.scala
```diff
@@ -96,4 +119,26 @@ class AmazonS3Client(s3client: AWSAmazonS3Client) extends S3Client {
   def putObject(putObjectRequest: PutObjectRequest): PutObjectResult = {
     s3client.putObject(putObjectRequest)
   }
+
+  private def readInputStream(inStream: InputStream): Array[Byte] = {
```
echeipesh: Just just use `org.apache.commons.io.IOUtils.toByteArray`, same should be true for `S3FileInputFormat`

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopSparkContextMethods.scala
```diff
@@ -56,24 +70,13 @@ trait HadoopSparkContextMethods {
...
-      case p if tiffExtensions.exists(p.endsWith) => path
-      case p =>
-        val extensions = tiffExtensions.mkString("{", ",", "}")
-        new Path(s"$p/*$extensions")
-    }
-
-    val updatedConf =
-      sc.hadoopConfiguration.withInputDirectory(searchPath)
-
+  def hadoopMultiBandGeoTiffRDD(path: Path, tiffExtensions: Seq[String] = defaultTiffExtensions): RDD[(ProjectedExtent, MultiBandTile)] =
```
moradology: It may be sensible to remove spaces around the `=` on a default param - with word wrap, this can look pretty confusing.

project/Environment.scala
lossyrob: Both EMR and CDH depend on 1.5.0, this can be changed later if it needs to be...I'm actually compiling against 1.5.2 in the client app

spark/src/main/scala/geotrellis/spark/ingest/Pyramid.scala
lossyrob: Agree, I'm writing it down as a task, we should think of a better name.

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopSparkContextMethods.scala
lossyrob: This fits with the Scala formatting guidelines AFAIK

spark/src/main/scala/geotrellis/spark/io/s3/TemporalGeoTiffS3InputFormat.scala
lossyrob: ?

spark/src/main/scala/geotrellis/spark/io/slippy/FileSlippyTileReader.scala
lossyrob: No you're totally right

project/Dependencies.scala
lossyrob: If by "change my mind" you mean agonizingly run through versions of this dependency that would both recognize `eu-central-1` as a region and not throw signature or other crazy errors, then yes :)

PR #1230, opened 2015-12-03
SpaceRDD and SpatialPartitioner

echeipesh:
## SpatialPartitioner
This PR implements a partitioner for loosely based on quad/oct/tree regions which allows for efficient:
 - spatial filtering
 - spatial left join
 - spatial inner join

Since the key bounds are known at RDD construction time we can scale number of partitions up and down depending on the space volume, reducing typical "partition management" headache.

This partitioner works by grouping keys using "coarse" space filling curve.

Once SpaceRDD loses it's type it's partitioner essentially degrades to a version of HashPartitioner. As an upshot from this the RDD can still be used with `PairRDDFunctions`. This will still return correct answers, but require more shuffles.

Since joining two RDDs has no clear choice for metadata only the resulting KeyBounds are produced as result metadata. If other information is required it is up to the user to re-insert it in the result.

## KeyBounds
KeyBounds had to change to become an ADT. I can into several cases where empty KeyBounds were needed to describe either an RDD or it's partitioner. Similar needs rose when processing RDDs that may be empty. Where possible KeyBounds are used as nonEmpty class. Where not possible their type changed to Bounds, which could either be KeyBounds or EmptyBounds.

## Spark raster operations
Spark raster operations have been redefined to work on `RDD[(K, Tile)]` where no knowledge of metadata is required. Otherwise they work on `RDD[(K, Tile)] with Metadata[M]` where `M` may have some requirements, like convertible to `MapKeyTransform`.

When working with layers containing metadata, like those that are produced from GeoTrellis readers, it is necessary to use `.withContext { rdd => rdd}` method in order to preserve the metadata through operations.

## RDD Stitch Methods
Have been upgrade to work on `Tile` or `MultiBandTile`. There is a new `StitcherR` method that is usable in method extensions for all subclasses of `Tile`. The mechanics are described in: https://github.com/geotrellis/geotrellis/issues/1292

`StitcherR` should replace `Stitcher` type-class but I am leaving that work for another PR.


echeipesh:
Things I am unhappy about:

- KeyBounds needs to have an Empty state. Currently the only way to express this is with an Option
Since KeyBounds intersect possibility of empty bounds comes up all the time. Maybe a sealed trait hierarchy would serve us well here.

- Is it possible to base SpacePartitioner implementation on SFC type class rather than GridKey?
All we really need for this scheme is a "low resolution" SFC. Perhaps we can have a dedicated type class for this purpose. This would be more extensible than a match statement.

- Multi-reoslution RDDs
What needs to happen for this to be able to perform operations between RDDs of different resolution (i.e.: RDD that uses 2 bits per partition and 4 bits per partition)
Should this just shuffle is there more efficient way? Test coverage for this would be a great way to start.

- Benchmarks
They need to happen.

spark/src/main/scala/geotrellis/spark/ContextRDDMethods.scala
```diff
@@ -13,7 +13,7 @@ class ContextRDDMethods[K: ClassTag, V: ClassTag, M](val rdd: RDD[(K, V)] with M
   def mapKeys[R: ClassTag](f: K => R): RDD[(R, V)] with Metadata[M] =
     rdd.withContext { rdd => rdd.map { case (key, tile) => f(key) -> tile } }
 
-  def mapValues(f: V => V): RDD[(K, V)] with Metadata[M] =
+  def mapValues[W](f: V => W): RDD[(K, W)] with Metadata[M] =
```
lossyrob: Curious on how something like this plays out for, say, changing the cell type of the tile. So we would have a cell type in the raster metadata would then be invalid, and it's carried through by default. I'm wondering if we shouldn't override mapValues? Is that going against stuff we decided on in the many long talks we've had about this?

echeipesh: Good catch, this should have better documented. Here is the use-case:

Given instance of  `RDD[(K, V)] with Metadata[M]`
call to `.mapValues` will always resolve to `ContextRDDMethods.mapValues` as it is the most specific implicit class. Scala will not go a-looking elsewhere for the function passed in changes the value type.

Which means that if a user is intending to use `.mapValues` in the way that spark defines it, with mapping `V => W`, it will fail with something like: 
```scala
[error]  found   : (Int, None.type)
[error]  required: Int
[error]     val expected = pr1.mapValues(v => (v, None))
```

I think we talked about enforcing `.mapValues(f: V => V)` type signature and we can still do it if we change the name. However, it’s not clear how much this is buying us. Let me argue against that goal.

Keeping the type of values is not guarantee that metadata is valid. For instance `.convert` function will change the content of the tile but not the type. So we have to do some book keeping and adjust the metadata with it. https://github.com/echeipesh/geotrellis/blob/feature/partitioner/spark/src/main/scala/geotrellis/spark/BaseRasterRDDMethods.scala#L7-L8

Some of the problem stems from the fact that even now part of the metadata, in case of `RasterMetaData`, refers to the keys (`extent`, `layout`, `crs`) and part refers to values (`cellType`). In future we may have `bandCount` which also refers to values only.

One way to deal with this is to maybe to provide two traits to be extended by metadata, one for keys and one for values and provide API that changes one or the other depending on the function. If our objective remains to support user extensible metadata we basically have no control of what that might be and would have to rely on lenses most likely.

Another way to look at this is as user concern. In this case we provide a way to mutate metadata with the keys and/or values and rely on the user to be a good programmer. Our library code can again rely on lenses but this time lenses for the whole metadata type. For instance to change layout during pyramiding (https://github.com/echeipesh/geotrellis/commit/4d252901427d6337175accb42535d9814f92422d).

I’m starting to be in favor of the second approach as it more accurately reflects the reality, metadata is liable to go invalid without care. Plus splitting the metadata type into key and value metadata doesn’t seem to buy much as both approaches would fall-back on lenses.

raster/src/main/scala/geotrellis/raster/stitch/Stitcher.scala
```diff
@@ -33,3 +33,37 @@ object Stitcher {
   }
 
 }
+
+
+trait StitcherR[T <: CellGrid, R <: CellGrid] extends Serializable {
```
lossyrob: We shouldn't have both Stitcher and StitcherR

echeipesh: Do you think that can happen in upcoming PR or it should be taken care of here?

lossyrob: Perhaps an upcoming PR, if we want to think through and talk out that issue
a bit before committing to that path/design pattern.

lossyrob: Do we need to add StitcherR now, or can we just type the test code correctly to avoid having the duplicate classes and keep with Stitcher?

echeipesh: You're totally correct. I saw this problem and was a little afraid of how prevalent it was with our use of type classes and not having a solution for it. In retrospect we can just change the test and roll this in as it's own PR.

spark/src/main/scala/geotrellis/spark/SpaceTimeKey.scala
```diff
@@ -63,8 +63,8 @@ object SpaceTimeKey {
 
     def getKeyBounds(rdd: RDD[(SpaceTimeKey, X)] forSome {type X}): KeyBounds[SpaceTimeKey] = {
       rdd
-        .map{ case (k, tile) => KeyBounds(k, k) }
-        .reduce { combine }
+        .map { case (k, tile) => KeyBounds(k, k) }
+        .reduce { _ combine _ }
```
pomadchin: Should not be there a `fold(EmptyBounds)(_ combine _)` ?

spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerUpdater.scala
```diff
@@ -32,7 +32,7 @@ class AccumuloLayerUpdater[K: Boundable: JsonFormat: ClassTag, V: ClassTag, M: J
       case e: UnsupportedOperationException => throw new LayerUpdateError(id, ": empty rdd update").initCause(e)
```
pomadchin: Probably we can avoid catching this exception: it is workaround for the `getKeyBounds` function applied to an empty rdd.

raster/src/main/scala/geotrellis/raster/MethodExtensions.scala
```diff
@@ -1,5 +1,6 @@
 package geotrellis.raster
 
 trait MethodExtensions[+T] extends Serializable {
+  type Self = T
```
lossyrob: What's this for?

echeipesh: Such that methods extending this can make use of type:

```scala
  def localXor(other: Self): Self = localXor(other, None)
```

spark/src/main/scala/geotrellis/spark/KeyBounds.scala
```diff
@@ -3,26 +3,93 @@ package geotrellis.spark
...
+      case KeyBounds(otherMin, otherMax) =>
+        val newMin = b.minBound(minKey, otherMin)
+        val newMax = b.maxBound(maxKey, otherMax)
+        KeyBounds(newMin, newMax)
+
+      case EmptyBounds =>
+        this
+    }
+
+  def combine(other: Bounds[K])(implicit b: Boundable[K]): KeyBounds[K] =
```
lossyrob: What's the overload used for? I'm actually a bit surprised there's no ambiguous compiler errors when trying to call this, because it seems to fit the above overload type sig as well.

spark/src/main/scala/geotrellis/spark/io/RDDFilter.scala
```diff
@@ -77,7 +80,10 @@ object Intersects {
         val queryBounds = KeyBounds(
           kb.minKey updateSpatialComponent SpatialKey(bounds.colMin, bounds.rowMin),
           kb.maxKey updateSpatialComponent SpatialKey(bounds.colMax, bounds.rowMax))
-        implicitly[Boundable[K]].intersect(queryBounds, kb).toSeq
+        (queryBounds intersect kb) match {
+          case kb: KeyBounds[K] => List(kb)
+          case EmptyBounds => Nil
+        }
```
lossyrob: this is happening a lot. Maybe ad a `toList` to Bounds, that will do exactly this?

spark/src/main/scala/geotrellis/spark/op/Implicits.scala
```diff
@@ -0,0 +1,34 @@
...
+  implicit class withCombineMethods[K: ClassTag, V: ClassTag](val self: RDD[(K, V)])
+    extends CombineMethods[K, V]
+
+  implicit class withCombineTraversableMethods[K: ClassTag, V: ClassTag](rs: Traversable[RDD[(K, V)]]) {
+    def combineValues[R: ClassTag](f: Traversable[V] => R): RDD[(K, R)] =
+      rs.head.combineValues(rs.tail)(f)
+  }
+
+  implicit class withMapValuesTupleMethods[K: ClassTag, V: ClassTag](val self: RDD[(K, (V, V))]) extends MethodExtensions[RDD[(K, (V, V))]] {
+    // TODO these needs names
```
lossyrob: Is this TODO done..are we sticking with combineValues?

spark/src/main/scala/geotrellis/spark/op/Implicits.scala
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  implicit class withMapValuesOptionMethods[K: ClassTag, V: ClassTag](val self: RDD[(K, (V, Option[V]))]) extends MethodExtensions[RDD[(K, (V, Option[V]))]] {
+    def updateValues(f: (V, V) => V): RDD[(K, V)] =
+      self.mapValues { case (v1, ov2) => 
+        ov2 match {
+          case Some(v2) => f(v1, v2)
+          case None => v1
+        }
+      //ov2.fold(v1)(f(v1, _)) }
```
lossyrob: delete

spark/src/main/scala/geotrellis/spark/KeyBounds.scala
echeipesh: I'm a little surprised this works as well, I'm guessing not requiring a type param is enough to disambiguate it.

This allows combine on something that is `KeyBounds` to produce `KeyBounds` instead of `Bounds`, as in generic case above. This is something you mentioned would be nice to have in the API and I agree. It is strange though.

spark/src/main/scala/geotrellis/spark/io/RDDFilter.scala
echeipesh: I was on the fence about this so I left it out. This happens a lot in context of `RDDFilter` shape. So may be a companion helper function here. Do you think its generic enough to be on `Bounds`?

spark/src/main/scala/geotrellis/spark/op/Implicits.scala
echeipesh: Yep, orphaned TODO. I think these names make sense. Thoughts?

spark/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerUpdater.scala
echeipesh: This is in reference to getKeyBounds comment above. I wasn't going to change this because I expect that function to be removed soon. But I think you're right. This site should match and the `getKeyBounds` should return Bounds

spark/src/main/scala/geotrellis/spark/op/README.md
```diff
@@ -0,0 +1,158 @@
+# GeoTrellis Spark Raster Operations
+
+In `geotrellis.spark` we represent raster layers as a distributed collection of non-intersecting raster tiles indexed by keys from a given layout. For instance a spatial raster is represented as `RDD[(SpatialKey, Tile)]` where `SpatialKey(col: Int, row: Int)` from a `TileLayout`. In this setup we represent operations between raster layers as a join.
```
moradology: I think this should read: "In `geotrellis.spark` we represent a distributed, non-intersecting collection of raster layers as tiles indexed by keys from a given layout." The mode of implementation seems to me orthogonal to the means of representation

spark/src/main/scala/geotrellis/spark/op/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Metadata
+
+If we read a previously tiled and saved GeoTrellis raster RDD from layer reader see that in fact we also get some metadata mixed in with a `Metadata[RasterMetaData]` trait. This metadata describes the `TileLayout` used by the layer, what extent it covers and what `CRS` of its projection, and what the `CellType` of each tile is. This metadata allows you to verify that you are working with compatible layers.
```
moradology: A previously tiled and saved GeoTrellis Raster RDD read in through an instance of `geotrellis.spark.io.LayerReader` will be mixed in with the `Metadata[RasterMetaData]` trait. This metadata describes the `TileLayout` used by the layer, the extent it covers, the `CRS` of its projection, and what the `CellType` of each tile is. This metadata allows you to verify that you are working with compatible layers.

PR #1233, opened 2015-12-08
WIP - Add imageio jpg rendering

moradology:
This PR should allow the rendering of JPG data from a tile and/or colorbreaks.


raster/src/main/scala/geotrellis/raster/render/shared/Renderer.scala
```diff
@@ -14,8 +14,10 @@
  * limitations under the License.
  */
 
-package geotrellis.raster.render.png
+package geotrellis.raster.render.shared
```
lossyrob: I'd suggest, instead of having a "shared" package, that this just go in `geotrellis.raster.render`

lossyrob:
Looking good so far

lossyrob:
need to update `geotrellis-benchmark`

moradology:
Closed because superseded in PR #1239

PR #1232, opened 2015-12-07
Change Return Type of RDDFilter

jamesmcclain:
RDDFilter now returns a Seq[KeyBound] instead of an Option[KeyBound].

Tests for the `spark` project all passed locally except for `geotrellis.spark.op.local.spatial.LocalSpatialSpec`.

PR #1234, opened 2015-12-08
Use Rasterizer to Find Keys Covering Polygon

### LINKED ISSUE #1228: Query LayerReader by Polygon and MultiPolygon
We should be able to do something like this:

```scala
val polygon: Polygon = ???
val query = new RDDQuery[SpatialKey, RasterMetaData].where(Intersects(polygon))
```
and so

```scala
val layerReader: S3LayerReader[SpatialKey, Tile, RasterRDD] = ???
val layerId: LayerId = ???
layerReader.query(layerId)
  .where(Intersects(polygon))
  .toRDD
```

jamesmcclain:
A method to provide intersection queries against polygons has been added.  It does so by rasterizing the polygon into an extent fashioned to mirror the set of keys that meet the polygon.

Still Requires:
   - [x] More Testing
   - [x] Unit Tests

Fixes #1228

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/io/RDDFilter.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/RDDFilter.scala
@@ -1,12 +1,14 @@
 package geotrellis.spark.io
 
 import com.github.nscala_time.time.Imports._
-import geotrellis.raster.GridBounds
+import geotrellis.raster.{GridBounds, RasterExtent}
 import geotrellis.spark._
 import geotrellis.spark.tiling.MapKeyTransform
-import geotrellis.vector._
+import geotrellis.vector.{Extent, Point, MultiPolygon}
 
 import scala.annotation.implicitNotFound
+
+import scala.collection.mutable
 
 @implicitNotFound("Unable to filter ${K} by ${F} given ${M}, Please provide RDDFilter[${K}, ${F}, ${T}, ${M}]")
 trait RDDFilter[K, F, T, M] {
@@ -53,6 +55,9 @@ object RDDFilter {
 
 
 object Intersects {
+  import geotrellis.raster.rasterize.{Rasterizer, Callback}
+  import scala.collection.mutable.ListBuffer
+
   def apply[T](value: T) = RDDFilter.Value[Intersects.type, T](value)
 
   /** Define Intersects filter for KeyBounds */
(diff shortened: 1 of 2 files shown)


lossyrob:
Note: Depends on #1232

lossyrob:
@jamesmcclain looking good. How does this handle partial intersection with tiles, where the polygon does not contain the tile center?

lossyrob:
Also, please reference issues in the PR that will close them when they get merge. I've updated the PR description to include the appropriate "Fixes" line that will auto-close the issue.

jamesmcclain:
I have not confirmed it yet, but it appears that the (Multi)Polygon rasterizer code already has provision for this: you can pass a Boolean that appears to widen the envelope of pixels over all pixels that edges pass through, not just ones whose centers are passed through.

I will seek to confirm this in the morning.

lossyrob:
`includeExterior` means to include the exterior line of the polygon, not of the raster cell. Perhaps that is named poorly. That just means, if the center point of the cell lies on the polygon border, and that is true, it's included, if it's false, it's not. So it still wouldn't work with cells that don't have their center cells intersected at all

jamesmcclain:
This code still occasionally misses keys with very small intersections with the query polygon.  I believe that this is due to the fact that the polygon and the boxes around keys are not guaranteed to be axis-aligned (the two large rectangles are the query multi-polygon, and the small square is a key which intersects the multi-polygon but is not reported by the code):

screenshot from 2015-12-11 15 04 54

Modulo that, this should be fairly solid (a few issues with the rasterizer have been addressed, but I am sure that more remain).

jamesmcclain:
@lossyrob Ready for review.  The tests passed locally, but seem to have failed here due to what appears to be a transient error (missing package).

jamesmcclain:
screenshot from 2015-12-16 18-33-47

This is the result that I get when this PR is rebased on top of my other PR #1244 .  Superficially it looks worse than the result pictured above, but it is actually much better (because it is returning this result without any special tricks).

Numerous missing keys can still be seen, but I have been able to confirm that the idealized pixel boxes (that the rasterizer considers) for the missing keys do not actually intersect the polygon.  Looking closely at the screenshot, it looks as though the keys are not perfect squares.  Are they really not squares in real-life, or are the polygons that I am generating from the respective key extents just warped?

lossyrob:
@jamesmcclain awesome. Can you show me the code that you're using to generate the image? It looks like the key space you are considering is not in web mercator, and there's some reprojecting going on to put it into web mercator, in which case the Keys, which are rectangles (Extents) in the CRS of the source layer, might be warped into non-rectangular polygons.

lossyrob:
Also, this PR depends on #1228, but it doesn't look like it has the commits of #1228...confused about this.

jamesmcclain:
I corrected another small bug in the rasterizer: I was failing to include the right-most endpoint of some intervals.  Here are the current results:

screenshot from 2015-12-17 09 38 51

With a closeup of each missing key:

screenshot from 2015-12-17 09 24 56

screenshot from 2015-12-17 09 25 04

screenshot from 2015-12-17 09 25 18

screenshot from 2015-12-17 09 25 31

screenshot from 2015-12-17 09 28 24

I think that at this point we are clearly dealing with numerical issues here:  In fact, it can be seen that there is actually one key that does not appear to intersect with the polygon at all.

jamesmcclain:
This is the code that I have been using to generate the pictures: https://github.com/jamesmcclain/gtAssembly/tree/rasterizer

jamesmcclain:
> Also, this PR depends on #1228, but it doesn't look like it has the commits of #1228...confused about this.

Did you mean #1244 ?  I have rebased on top of it.

lossyrob:
:+1: :100:

PR #1236, opened 2015-12-09
Allow extents in GeometryCollections; Style GeoJson

lossyrob:
A recent change made `Extent` extend `Geometry`, and this PR fixes a resulting bug that was caused by not correctly modifying the `GeometryCollectionBuilder` logic to handle that change.

Also, added a `Style` object that makes it simple to style GeoJson geometries so that they can have colors and more in web viewers like geojson.io and gists, like this: https://gist.github.com/lossyrob/df261af1a9a02e088686 (Thanks to @nocrosstalk for that idea)

### DIFF
--- a/vector/src/main/scala/geotrellis/vector/io/json/Style.scala
+++ b/vector/src/main/scala/geotrellis/vector/io/json/Style.scala
@@ -1,0 +1,86 @@
+package geotrellis.vector.io.json
+
+import spray.json._
+import scala.collection.mutable
+
+case class Style(
+  strokeColor: Option[String],
+  strokeWidth: Option[Int],
+  strokeOpacity: Option[Double],
+  fillColor: Option[String],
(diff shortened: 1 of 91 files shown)


vector/src/main/scala/geotrellis/vector/io/json/Style.scala
```diff
@@ -0,0 +1,86 @@
+package geotrellis.vector.io.json
+
+import spray.json._
+import scala.collection.mutable
+
+case class Style(
+  strokeColor: Option[String],
+  strokeWidth: Option[Int],
```
echeipesh: This can be Unit, which is "1", "1.5", "5%"
http://wiki.openstreetmap.org/wiki/Geojson_CSS
http://www.w3.org/TR/SVG/painting.html#StrokeWidthProperty

echeipesh:
+1 after strokeWidth change

PR #1239, opened 2015-12-10
Feature/render jpg docs

moradology:
This changeset includes modifications to the directory structure of raster.render, the inclusion of jpeg rendering, documentation around the use of raster.render, and benchmarks for jpeg rendering.

Note that it supersedes the prior PR 'feature/render-jpg'


engine/src/main/scala/geotrellis/engine/render/RenderRasterSourceSharedMethods.scala
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+package geotrellis.engine.render
+
+import geotrellis.raster._
+import geotrellis.raster.render._
+import geotrellis.engine._
+
+trait RenderRasterSourceSharedMethods extends RasterSourceMethods {
```
lossyrob: The `shared` naming scheme is a bit confusing, and especially if something moves around later, might be even more so. How about `ColorRasterSourceMethods`?

raster/src/main/scala/geotrellis/raster/render/JpgRenderMethods.scala
```diff
@@ -0,0 +1,97 @@
...
+    * Use this operation when you have a raster of data that you want to visualize
+    * with an image.
+    *
+    * To render a data raster into an image, the operation needs to know which
+    * values should be painted with which colors.  To that end, you'll need to
+    * generate a ColorBreaks object which represents the value ranges and the
+    * assigned color.  One way to create these color breaks is to use the
+    * [[geotrellis.raster.stats.op.stat.GetClassBreaks]] operation to generate
+    * quantile class breaks.
+    */
```
lossyrob: +1 on the doc string

raster/src/main/scala/geotrellis/raster/render/MultiBandJpgRenderMethods.scala
```diff
@@ -0,0 +1,54 @@
...
+          if(z < gMin) { gMin = z }
+        } else if(band == 2) {
+          if(z > bMax) { bMax = z }
+          if(z < bMin) { bMin = z }
+        }
+      }
+    }
+
+    val rgb =
+      tile.convert(TypeInt).combine(0, 1, 2) { (r, g, b) =>
```
lossyrob: Where is this scaling logic coming from? Is that from somewhere else in the codebase? I feel like the scaling to 0 - 255 should be another operation, and renderJpg should just assume it's already pre-processed into the correct RGB values.

lossyrob:
@moradology waiting on the Spark component to this in order to review

lossyrob:
+1

PR #1240, opened 2015-12-10
Add zonal stats and implement count for Statistics methods

moradology:
1. Implement a zonalStatistics method
2. Remove no-op loop inside the statistics generating code
3. Add a data count on the Statistics object

### DIFF
--- a/raster-test/src/test/scala/geotrellis/raster/op/stats/StatsMethodsSpec.scala
+++ b/raster-test/src/test/scala/geotrellis/raster/op/stats/StatsMethodsSpec.scala
@@ -39,7 +39,7 @@ class StatsMethodsSpec extends FunSpec
       val stats = RasterSource.fromPath("raster-test/data/quad8.arg").get.statistics
 
       val dev = math.sqrt((2 * (0.5 * 0.5) + 2 * (1.5 * 1.5)) / 4)
-      val expected = Statistics(2.5, 3, 1, dev, 1, 4)
+      val expected = Statistics(400, 2.5, 3, 1, dev, 1, 4)
 
       stats should be (expected)
     }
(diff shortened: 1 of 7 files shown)


raster-test/src/test/scala/geotrellis/raster/op/stats/StatsMethodsSpec.scala
```diff
@@ -39,7 +39,7 @@ class StatsMethodsSpec extends FunSpec
       val stats = RasterSource.fromPath("raster-test/data/quad8.arg").get.statistics
 
       val dev = math.sqrt((2 * (0.5 * 0.5) + 2 * (1.5 * 1.5)) / 4)
-      val expected = Statistics(2.5, 3, 1, dev, 1, 4)
+      val expected = Statistics(400, 2.5, 3, 1, dev, 1, 4)
```
lossyrob: What happened here? Was a test passing when it shouldn't have been?

raster-test/src/test/scala/geotrellis/raster/op/zonal/ZonalStatisticsSpec.scala
```diff
@@ -0,0 +1,84 @@
...
+import geotrellis.raster.op.stats._
+
+import geotrellis.raster._
+
+import org.scalatest._
+
+import geotrellis.testkit._
+
+import scala.collection.mutable
+
```
lossyrob: Shouldn't space out all the imports like this

raster/src/main/scala/geotrellis/raster/histogram/Histogram.scala
```diff
@@ -136,12 +134,12 @@ abstract trait Histogram extends Serializable {
       var needMedian = true
       val limit = getTotalCount() / 2
 
-      var i = 0
       val len = values.length
 
-      while (i < len) {
+      cfor(0)(_ < len, _ + 1) { i =>
```
lossyrob: Sometimes a while loop is faster than a cfor. Don't need to change these. We'll keep these in, but then benchmark the old vs new histogram and maybe change back, just wanted to give a heads up

moradology: Good to know. I had looked into it and figured out that cfor was basically just being compiled into a while loop but figured that the consistency may be worth it. Will bench this and post something on this PR.

PR #1243, opened 2015-12-14
Ephemeral Spark Test Files

echeipesh:
Auto-generates spark test rasters on access, avoids writing an HDFS test catalog during tests.


echeipesh:
@lossyrob ready for review, fails are related to the Mask spec.

PR #1246, opened 2015-12-15
Split ResampleMethods into aggregate and point-based resample

moradology:
Implement an AggregateResample class which is responsible for generating bins over which aggregations can be done


raster/src/main/scala/geotrellis/raster/render/README.md
```diff
@@ -1,6 +1,6 @@
 #geotrellis.raster.render
 
-## Rendering Common Image Formats
+##Rendering Common Image Formats
```
lossyrob: Is there a reason for no space vs space?

raster/src/main/scala/geotrellis/raster/resample/README.md
```diff
@@ -0,0 +1,56 @@
+#geotrellis.raster.resample
+
+##Resampling
+Often, when working with raster data, it is useful to change the number
+of rows and/or columns. For everything there is a price, however, and
+changing the resolution of a tile is no exception: there will (almost)
+always be a loss of information (or representativeness) when conducting
+an operation which changes the number of cells on a tile.  
+
```
lossyrob: Edits:

Often, when working with raster data, it is useful to change the resolution, crop the data, transform the data to a different projection, or to do all of that at once. This all relies on our ability
to _resample_, which is the act of changing the spatial resolution and layout of the raster cells, and interpolating the values of the modified cells from the original cells.
For everything there is a price, however, and
changing the resolution of a tile is no exception: there will (almost)
always be a loss of information (or representativeness) when conducting
an operation which resamples raster data.

raster/src/main/scala/geotrellis/raster/resample/README.md
```diff
@@ -0,0 +1,56 @@
...
+####Remember What Your Data Represents
+Along with the formal characteristics of these methods, it is important
+to keep in mind the specific character of the data that you're working
+with. After all, it doesn't make sense to use a method like
+`Bilinear` resampling if you're dealing primarily with categorical data.
+In this instance, your best bet is to choose an aggregate method (and
+keep in mind that the values generated don't necessarily mean the same
+thing as the data being operated on) or a forgiving (though
+unsophisticated) method like `NearestNeighbor`.
```
lossyrob: Sweet docs. Can you add it/link to it to the main index on the repo README?

raster/src/main/scala/geotrellis/raster/render/README.md
moradology: I think it will make a slight difference (an extra space in the header)

echeipesh:
+1

Issue #1249, opened 2015-12-16
Have SpatialIndex be able to handle Extents, not just points.


lossyrob:
Currently the implementation of `SpatialIndex` (https://github.com/geotrellis/geotrellis/blob/6d1c011aa2637ae1cceb7c2223d123bab7e475fe/vector/src/main/scala/geotrellis/vector/SpatialIndex.scala), which wraps JTS's `STRtree`, only allows for elements to be associated with points in the index, when it should support both point and extents. The interface needs to change/be refactored to allow for a SpatialIndex with a call like

```scala
def insert(v: T, extent: Extent) 
```

lossyrob:
Closed by #1375

PR #1252, opened 2015-12-17
WIP: Allow adding entries to SpatialIndex by Extent

dwins:
Addresses #1249. Please don't merge yet, I'm happy with this but have not completed the CLA.


lossyrob:
@dwins can you tie your email address to your github account? The commit isn't picking up that it's you that made the change; ideally you'd be committing with the same email address that you'd have registered with an Eclipse account that has the CLA of their's signed, and that is tied to your github account.

lossyrob:
There's a couple of `apply` methods in the `SpatialIndex` object that make newing up a spatial index easier on the client if what they have is Iterable of point-like things. This API functionality should be replicated by the Extent capabilities of this feature.

lossyrob:
We currently keep track of a set of `points` inside the instance; this was for some `geotrellis-transit` functionality, and I'm not sure it should be in there. We should probably remove it; if we don't, we should rename it to reflect that they aren't just points. My vote is for removal though.

lossyrob:
A unit test proving this functionality should be made.

dwins:
Associated my email with my github account.  I'll look into the changes you suggested.

Issue #1248, opened 2015-12-16
Implement CGAlgorithms in geotrellis.vector


lossyrob:
JTS has functions in it's `CGAlgorithms`:
  - [ ] isPointInRing (RING)
  - [ ] isOnLine (LINE)
  - [ ] isCCW (RING)
  - [ ] distancePointLine (LINE)
  - [ ] distancePointLinePerpendicular (LINE)
  - [ ] distanceLineLine (LINE)

These should be implemented on `Point` and `Line` in geotrellis.vector, where appropriate. Part of this issue is tracking down exactly when these methods should be called on which types.

dwins:
Looking into this a bit, I see that:
* isPointInRing should be called with a Point and a closed LineString.
* isOnLine should be called with a Point and any LineString (seems this is conceptually equivalent to linestring.contains(point) but I didn't look into the relative performance yet.)
* isCCW should be called with a valid ring (closed LineString that does not cross itself)
* distancePointLine should be called with a point and two more points defining a line segment (NOT a line string)
* distancePointLinePerpendicular should be called with a point and two more points defining an infinite line (still NOT a linestring)
* distanceLineLine should be called with exactly 4 points, two pairs defining line segments.

In order to express these constraints fully in the Scala type system I think it would appropriate to introduce a couple of new Geometry types - Segment for a line segment defined by only its start and end, and Ring for a LineString that is closed and does not self-intersect - both would produce a JTS LineString when asked for the jtsGeometry.  We could add asSegment: Option[Segment] and asRing: Option[Ring] to LineString to allow developers to convert.

lossyrob:
Closed by #1255

PR #1255, opened 2015-12-18
Expose JTS CGAlgorithms methods in Geometry

dwins:
See discussion on #1248

### DIFF
--- a/vector/src/main/scala/geotrellis/vector/Point.scala
+++ b/vector/src/main/scala/geotrellis/vector/Point.scala
@@ -17,6 +17,7 @@
 package geotrellis.vector
 
 import com.vividsolutions.jts.{geom => jts}
+import com.vividsolutions.jts.algorithm.CGAlgorithms
 import GeomFactory._
 
 object Point {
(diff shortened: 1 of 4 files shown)


lossyrob:
Same thing as the last PR, need the commit to be made with the email of the eclipse CLA, tied to the github account.

lossyrob:
The issue with adding Geometry types is that it effects anywhere in the library we are matching over geometry types to do a thing (e.g. in GeoJson creation, any Segment or Ring would fail).

Another case against adding the geometry types is simplicity: I found it a burden to have to have two separate types for a LineString and a LinearRing. This seemed like some surprising type restrictions when say creating a polygon and requiring a linear ring, given that a lot of JTS deals with the supertype Geometry and isn't big on typing things specifically. So I created `Line`, which simply represents a line in all it's iterations; you can check if it's closed if you require it. It also makes more sense in my mind map of how Geometries work, Point < Line < Polygon, MultiPoint < MultiLine < MultiPolygon, GeometryCollection. Adding separate Line types, where there aren't separate types to the other geometry types, seems like a bit of a deviation, and I'm not sure for what gain.

This is all to say, I'm hesitant to add new Geometry types, especially specific Line types, for the above reasons. I'm willing to be convinced though, we'd have to talk it out.

vector/src/main/scala/geotrellis/vector/Point.scala
```diff
@@ -321,4 +322,16 @@ case class Point(jtsGeom: jts.Point) extends Geometry
    */
   def within(g: Geometry): Boolean =
     jtsGeom.within(g.jtsGeom)
+
+  def isInRing(l: Line): Boolean =
+    CGAlgorithms.isPointInRing(jtsGeom.getCoordinate, l.jtsGeom.getCoordinates)
```
lossyrob: I'd be curious what a benchmark between this and another, more general version (Polygon.contains). If there's no significant performance improvement, then I'm not sure we need to implement this. If there is a significant performance improvement, we'll have to think about when the user would want to use this function (and the others), and try to guide that usage through the API, the type system and/or documentation.

You can benchmark this using the [geotrellis-benchmark](https://github.com/geotrellis/geotrellis-benchmark) project.

dwins:
Thanks for pointing me at geotrellis-benchmark.  I did write a benchmark comparing CGAlgorithms.isPointInRing with Polygon.contains(Point) and it seems that CGAlgorithms is pretty consistently about 3 times faster, except the general contains method performs a preliminary bbox check that CGAlgorithms does not.

dwins:
I added another benchmark for testing with a [PreparedGeometry](http://tsusiatsoftware.net/jts/javadoc/com/vividsolutions/jts/geom/prep/PreparedGeometry.html) and actually it seems to be another factor of 3 faster than isPointInRing.  I guess we would be better off integrating PreparedGeometry into geotrellis-vector than the lower-level CGAlgorithms so I'll think about what that might look like.

dwins:
Added a wrapper for the JTS PreparedGeometry classes. With this I think we're ready to merge?

lossyrob:
+1 after travis says OK

PR #1260, opened 2015-12-22
Upgrade to JTS version 1.14.0

dwins:
No compile problems or test failures after upgrading.  I looked over the release notes and did a quick review of geotrellis-vector and didn't see anything that needed to be updated.  I'm going to look into how the new LineDissolver in JTS compares with the LineDissolve operation in geotrellis, if the JTS version seems clearly superior it might make sense to include that change in this PR.


lossyrob:
@dwins actually the LineDissolver we wrote was a port of the JTS LineDissolver as it existed in the repository, since we needed it and it wasn't released yet. I think if you don't find any notable improvements in the GeoTrellis implementation, switching to the now-released JTS implementation is a good choice.

PR #1262, opened 2015-12-23
Fixes and changes to Reprojection

lossyrob:
This PR includes some changes to raster reprojection, mainly

- The extent for the reprojected raster uses a densified version of the extent, reducing the code from the GDAL ported method.
- There was a "bug" in how we were determining the pixel size (I say "bug" in quotes because it's just an estimate). This more closely mirrors what GDAL is doing, which is verified by some tests around reprojecting raster extents from tiles of a landsat image.

Other changes include:
- Added a `fromEpsgCode` to CRS
- Added a `MapKeyTransform` that takes a layout level
- Fixed a GeoTiff writing bug that was happening when we were writing GeoTiffs without any GeoDoubleKey tags.

Also, this PR adds an `.sbtignore` that aims to reduce the output of the build (especially in travis). This will potentially have the effect of hiding information that one would expect from the build, so if this ends up biting us I'll take it back out.

### DIFF
--- a/.sbtignore
+++ b/.sbtignore
@@ -1,0 +1,28 @@
+# sbt ignore regexps
+Starting sbt with output filtering enabled
+^\[info\][ ]+(Resolving|Loading|Done|Attempting|Formatting|Updating)[ ]
+^\[info\] Main Scala API documentation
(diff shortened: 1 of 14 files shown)


echeipesh:
+1

PR #1268, opened 2016-01-03
Create buffered tiles, reproject indexed tiled sets with accuracy

lossyrob:
### Changes in this PR

- `BufferTiles` operation: this does what `FocalOperation` used to do, just abstracted
  so other operations can use it. It takes each of the tiles of a RDD and gives
  them a buffer of a specific size, with a grid bounds that tells you where the original
  tile is inside of the larger tile (represented by the type `BufferedTile`). You can
  either give it a constant buffer size (which will be modified for border tiles, e.g.
  if you have a constant buffer size of 2 pixels, a tile A will only have a top buffer
  if there's some other contributing tile B that is directly above it that can provide
  the pixels of the buffer, i.e. the spatial component SpatialKey(col, row) of the key of B is
  SpatialKey(col, row + 1) relative to the spatial component of the key of A). Another way
  to specify a buffer size is to give it a function `K => BufferSizes`, or to pass in
  an RDD that already has all the buffer sizes per key filled out, `RDD[(K, BufferSizes)]`.
- The way we reproject RDD[(K, V)] of index tiles of type V is different now. The problem
  was that, if you just reproject the tile, you're going to get some border resampling
  going on that could use the pixels that are represented by the border tiles of the target
  tile, but aren't there. So enter in the BufferTiles operation. We can now repreoject
  while taking into account the surrounding pixels of a tile; that buffer is determined
  either by a constant buffer size from the user, or figuring out what sort of
  buffer each tile needs...the way that is done is by reprojecting the extent,
  of the tile, taking that envelope, doing the inverse reprojection of that, getting
  it's envelope, and using that to determine the buffer size needed so that no information is
  missing in the reprojection resampling.
- Implemented `crop` operation for MultiBandTile and rasters, moved that functionality on
  `Tile` into the `crop` subpackage of `geotrellis.raster`.
- `Raster`/`MultiBandRaster` are now just `Raster[T <: CellGrid]`
- Implemented new implicit method addition structure in a number of places (see note below)
- Option structure (see other note below)
- Created a `GridExtent` type, which is basically a RasterExtent without the integer
  `cols` and `rows`. They both implement `GridDefinition`, which represents defining
  a grid in map space by defining the extent and cell size. `LayoutDefinition` is also
  a `GridDefinition`. This lets me do things like reproject a `GridDefinition` that
  may have too many cols and rows to be represented by integers (e.g. sub-2 cm
  raster layers with a world-wide web mercator tile layout).
- Fixing some bugs in RasterExtent methods that I'm suprised made it this far.
- Respect the order of the sequence when creating a JsonFeatureCollection
- Move `geotrellis.spark` TestSparkContext into TestEnvironment, create
  and shutdown a spark context per suite (slower but safer).

The `TileRDDReproject` really shows off the new `Raster[(K, V)] with Metadata[M]` change: it's abstracted over RDD's that contain Tiles and MultiBandTiles, all just by specifying a generic type `V` and the things that it needs. It needs a lot, as you can tell by the type signature for apply:

```scala
def apply[
  K: SpatialComponent: ClassTag,
  V <: CellGrid: ClassTag: Stitcher: (? => TileReprojectMethods[V]): (? => CropMethods[V]): (? => MergeMethods[V]): (? => CellGridPrototype[V])
](
  bufferedTiles: RDD[(K, BufferedTile[V])],
  metadata: RasterMetaData,
  destCrs: CRS,
  layoutScheme: LayoutScheme,
  options: Options
): (Int, RDD[(K, V)] with Metadata[RasterMetaData])
 ```

This means that, if anybody anywhere has a type that has those capabilities, you could use this reprojection method to reproject an RDD of instances of that type.

#### Note on new implicit method addition structure

- implicit classes that add methods to a type `T` should extent `MethodExtensions[T]`.
- implicit classes that add methods should be namd with the prefix `with`, e.g. `implicit class withTileMethods` or `implicit class withTileRDDReprojectMethods`

When adding methods to an core type in non-spark package
-Methods should be made available as a trait or class in a relavent subpackage with a `Methods` name, e.g. `SingleBandRasterReprojectMethods.scala`.
- Create Methods trait in package extending MethodExtensions[T]
- implement that trait on the base implicit MethodExtensions[T] class in root package

When adding methods to an core type in spark package
- Create an object that houses the base implementation of the functionality.
- Create Methods trait in package extending MethodExtensions[T] that calls to the object method
- implement that trait on the base implicit MethodExtensions[T] class in root package
_This avoids serialization issues of calling methods with direct implementations on implicit classes_

When adding methods, but you need implicit parameters on the implicit class
- e.g. `implicit class withSomethingMethods[T: (? => CropMethods[T])](val self: T)`
- create a `Methods` class instead of a trait in the relevant subpackage.
- create Implicits trait/object in package if it's not already there
  - declare an `object Implicits extends Implicits` in the subpackage
  - extent the trait in the package object of the root package
- Add the implicit class with the `with` prefix into the `Implicits` trait of the subpackage.

#### Note on Options

The best way to mix required and optional parameters when you overloads is not clear, and [I tried a thing](https://gist.github.com/lossyrob/63f0d40506ff8b4c3b22), and it didn't work out. This is because the funky method I thought up doesn't handle implicit parameter lists. So, unless somebody comes up with something better, we're stuck with using options classes everywhere.

When you need to overload and have a single optional parameter with a default argument, then just lose the default argument and create the overloads. If you have two optional parameters, then it's time to make an `Options` object. For example:

```scala
object SomeOperation {
  case class Options(
    theCommonOne: Foo = Foo("something"),
    theUncommonOne: Int = 45
  )

  object Options {
    def DEFAULT = Options()

    // Make life a little easier by adding implicit conversions from
    // common options that would be used on their own to Options with
    // that option set but the rest as default
    implicit def commonToOptions(theCommonOne: Foo): Options =
      Options(theCommonOne = theCommmonOne)
  }

  def doAThing(x: Int): Int
    doAThing(x, Options.DEFAULT)

  def doAThing(x: Int, options: Options): Int = ???

  // etc
}
```


raster/src/main/scala/geotrellis/raster/crop/CropMethods.scala
```diff
@@ -0,0 +1,35 @@
+package geotrellis.raster.crop
+
+import geotrellis.vector._
+import geotrellis.raster._
+
+trait CropMethods[T] extends MethodExtensions[T] {
```
echeipesh: The best thing about `MethodExtensions[T]` is that it gives something to search for and tags the trait with a noun.

raster/src/main/scala/geotrellis/raster/reproject/TileReprojectMethods.scala
```diff
@@ -0,0 +1,31 @@
+package geotrellis.raster.reproject
+
+import geotrellis.raster._
+import geotrellis.proj4._
+import geotrellis.raster.resample.ResampleMethod
+import geotrellis.vector.Extent
+
+trait TileReprojectMethods[T] extends MethodExtensions[T] {
```
echeipesh: If `T <: CellGrid` the reproject methods could be returning `Raster[T]`, which is nicer API.

raster/src/main/scala/geotrellis/raster/resample/RasterResampleMethods.scala
```diff
@@ -0,0 +1,23 @@
...
+  def resample(target: RasterExtent, method: ResampleMethod): T
+
+  def resample(target: RasterExtent): T =
+    resample(target, ResampleMethod.DEFAULT)
+
+  def resample(target: Extent): T =
+    resample(target, ResampleMethod.DEFAULT)
+
+  def resample(target: Extent, method: ResampleMethod): T =
+    resample(RasterExtent(self.extent, self.cols, self.rows).createAlignedRasterExtent(target), method)
```
echeipesh: This signature feels like it's a crop rather than resample.

spark/src/main/scala/geotrellis/spark/buffer/BufferSizes.scala
```diff
@@ -0,0 +1,6 @@
+package geotrellis.spark.buffer
+
+case class BufferSizes(left: Int, right: Int, bottom: Int, top: Int) {
+  def totalWidth = left + right
```
echeipesh: These aren't used and kind of sound like they might be totalWidth and totalHeight of the whole tile, which they are not of course.

spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
```diff
@@ -0,0 +1,191 @@
...
+
+import org.apache.spark.rdd._
+import org.apache.spark.storage.StorageLevel
+
+import scala.reflect.ClassTag
+
+object TileRDDReproject {
+  import geotrellis.raster.reproject.Reproject.Options
+
+  /** Reproject a set of buffered
```
echeipesh: Very generic : )

echeipesh:
:+1:

PR #1271, opened 2016-01-06
Add raw (no NoData) types

moradology:
This PR will include new types which lack the concept of missing data (e.g. all 8 bits in a byte are meaningful). This should make GT compatible with TIFFs which lack the GDAL-internal `nodata` tag. The most significant and furthest reaching change to the library is in the raster/CellType.scala. CellType is now a type alias for the conjunction of some DataType specification and some NoDataHandling specification.

In practice, this allows for more readable means of ensuring type safety for common operations. Consider the case in which we know that we must have `ByteCells` but in which we want to be polymorphic over the `NoDataHandling` policy. With the prior architecture, the best option would be to either make some (safe) assumptions about the what is possible or to throw exceptions:
```scala
SomeByteOnlyFunc(cellType: CellType) = cellType match {
  case ByteRawCellType => ???
  case ByteUserDefinedNoDataCellType(nd) => ???
  case _ => ??? // Handling our only other possible (given some assumptions) case
}
// or
SomeByteOnlyFunc(cellType: CellType) = cellType match {
  case ByteRawCellType => ???
  case ByteUserDefinedNoDataCellType(nd) => ???
  case ByteConstantNoDataCellType => ???
  case _ => throw BadCellTypeException
}
```

As of this PR, we can now specify things like:
```scala
SomeByteOnlyFunc(cellType: ByteCells with NoDataHandling) = cellType match {
  case ByteRawCellType => ???
  case ByteUserDefinedNoDataCellType(nd) => ???
  case ByteConstantNoDataCellType => ???
}
// conversely
SomeRawOnlyFunc(cellType: DataType with NoNoData) = ??? // you get the idea
```
And we completely exhaust the match.


raster/src/main/scala/geotrellis/raster/CellType.scala
```diff
@@ -46,18 +46,55 @@ sealed abstract class CellType(val bits: Int, val name: String, val isFloatingPo
...
-case object TypeUByte extends CellType(8, "uint8", false)
-case object TypeShort extends CellType(16, "int16", false)
-case object TypeUShort extends CellType(16, "uint16", false)
-case object TypeInt extends CellType(32, "int32", false)
-case object TypeFloat extends CellType(32, "float32", true)
-case object TypeDouble extends CellType(64, "float64", true)
+case class TypeDynamicByte(noDataValue: Double)   extends DynamicCellType(8, "int8dynamic", false)
+case class TypeDynamicUByte(noDataValue: Double)  extends DynamicCellType(8, "uint8dynamic", false)
+case class TypeDynamicShort(noDataValue: Double)  extends DynamicCellType(16, "int16dynamic", false)
+case class TypeDynamicUShort(noDataValue: Double) extends DynamicCellType(16, "uint16dynamic", false)
```
moradology: @lossyrob The hierarchy here could use another set of eyes

raster/src/main/scala/geotrellis/raster/VectorToRaster.scala
```diff
@@ -41,14 +41,14 @@ object VectorToRaster {
...
-    *                              If you use a Raster with a Double CellType (TypeFloat, TypeDouble)
+    *                              If you use a Raster with a Double CellType (FloatConstantNoDataCellType, DoubleConstantNoDataCellType)
     *                              the data values will be rounded to integers.
     */
   def kernelDensity[D](points: Seq[PointFeature[D]],
                        transform: D => Int, 
                        kernel: Kernel, 
                        rasterExtent: RasterExtent): Tile = {
-    val stamper = KernelStamper(TypeInt, rasterExtent.cols, rasterExtent.rows, kernel)
+    val stamper = KernelStamper(IntConstantNoDataCellType, rasterExtent.cols, rasterExtent.rows, kernel)
```
echeipesh: Does this mean I can't use this with Int tiles that are Raw or User Defined?

raster/src/main/scala/geotrellis/raster/io/geotiff/BitGeoTiffSegment.scala
```diff
@@ -13,8 +13,8 @@ class BitGeoTiffSegment(val bytes: Array[Byte], cols: Int, rows: Int) extends Ge
     bytesWidth * 8
   }
 
-  def getInt(i: Int): Int = b2i(get(i))
-  def getDouble(i: Int): Double = b2d(get(i))
+  def getInt(i: Int): Int = get(i).toInt
+  def getDouble(i: Int): Double = get(i).toDouble
```
echeipesh: Why not use the macro here? Shouldn't it always be better to inline this promotion?

raster/src/main/scala/geotrellis/raster/VectorToRaster.scala
moradology: This code doesn't deal with conversion of tiles to other tiles - rather, this code generates a kernel density using vector points and a raster extent. The alternative forms of celltype are really, primarily focused on geotiff interop - there's not an obvious argument to be made for extending this to Raw/UserDefinedNoData and not extending it to bytes, shorts, floats and doubles

moradology:
@lossyrob This beast is - I believe - ready for a full once-over.

raster/src/main/scala/geotrellis/raster/DoubleArrayTile.scala
```diff
@@ -39,27 +40,100 @@ final case class DoubleArrayTile(array: Array[Double], cols: Int, rows: Int)
...
+  def updateDouble(i: Int, z: Double) { arr(i) = d2udd(z) }
+}
+
 object DoubleArrayTile {
-  def ofDim(cols: Int, rows: Int): DoubleArrayTile = 
-    new DoubleArrayTile(Array.ofDim[Double](cols * rows), cols, rows)
+  def apply(arr: Array[Double], cols: Int, rows: Int): DoubleArrayTile =
+    apply(arr, cols, rows, DoubleConstantNoDataCellType)
+
+  def apply(arr: Array[Double], cols: Int, rows: Int, cellType: CellType with DoubleCells): DoubleArrayTile =
```
echeipesh: `DoubleCells with  NoDataHandling`

raster/src/main/scala/geotrellis/raster/FloatArrayTile.scala
```diff
@@ -42,32 +42,109 @@ final case class FloatArrayTile(array: Array[Float], cols: Int, rows: Int)
...
+      case FloatCellType =>
+        new FloatRawArrayTile(Array.ofDim[Float](cols * rows).fill(v), cols, rows)
+      case FloatConstantNoDataCellType =>
+        new FloatConstantNoDataArrayTile(Array.ofDim[Float](cols * rows).fill(v), cols, rows)
+      case udct @ FloatUserDefinedNoDataCellType(_) =>
+        new FloatUserDefinedNoDataArrayTile(Array.ofDim[Float](cols * rows).fill(v), cols, rows, udct)
+    }
 
-  def fromBytes(bytes: Array[Byte], cols: Int, rows: Int): FloatArrayTile = {
+  def constructFloatArray(bytes: Array[Byte]): Array[Float] = {
```
echeipesh: this was private in double case

raster/src/main/scala/geotrellis/raster/io/arg/ArgReader.scala
```diff
@@ -113,72 +101,73 @@ object ArgReader {
...
 
-  final def read(path: String, typ: CellType, cols: Int, rows: Int): Tile = {
-    ArrayTile.fromBytes(Filesystem.slurp(path), typ, cols, rows)
+  final def read(path: String, cellType: CellType, cols: Int, rows: Int): Tile = {
+    ArrayTile.fromBytes(Filesystem.slurp(path), cellType, cols, rows)
   }
 
-  final def read(path: String, typ: CellType, rasterExtent: RasterExtent, targetExtent: RasterExtent): Tile = {
-    val size = typ.numBytes(rasterExtent.size)
+  final def read(path: String, cellType: CellType, rasterExtent: RasterExtent, targetExtent: RasterExtent): Tile = {
```
echeipesh: Currently we only support writing arg with `ConstantNoData` type, so we should constrain on that trait here to avoid erroneous reading.

raster/src/main/scala/geotrellis/raster/io/geotiff/ByteConstantNoDataGeoTiffSegment.scala
```diff
@@ -0,0 +1,102 @@
...
+
+  protected def convertToConstantNoData(cellType: DataType with ConstantNoData): Array[Byte] =
+    cellType match {
+      case ByteConstantNoDataCellType =>
+        val arr = Array.ofDim[Byte](size)
+        cfor(0)(_ < size, _ + 1) { i => arr(i) = get(i) }
+        arr
+      case UByteConstantNoDataCellType =>
+        val arr = Array.ofDim[Byte](size)
+        cfor(0)(_ < size, _ + 1) { i => arr(i) = get(i) }
```
echeipesh: missing conversion

raster/src/main/scala/geotrellis/raster/io/geotiff/ByteConstantNoDataGeoTiffSegment.scala
```diff
@@ -0,0 +1,102 @@
...
+  def getDouble(i: Int): Double = b2i(get(i))
+
+  protected def intToByteOut(v: Int): Byte = i2b(v)
+  protected def doubleToByteOut(v: Double): Byte = d2b(v)
+
+  protected def convertToConstantNoData(cellType: DataType with ConstantNoData): Array[Byte] =
+    cellType match {
+      case ByteConstantNoDataCellType =>
+        val arr = Array.ofDim[Byte](size)
+        cfor(0)(_ < size, _ + 1) { i => arr(i) = get(i) }
```
echeipesh: do we need defensive copy here?

raster/src/main/scala/geotrellis/raster/ArrayTile.scala
```diff
@@ -277,3 +221,67 @@ trait ArrayTile extends Tile with Serializable {
...
+      case ct: ByteCells => ByteArrayTile.fromBytes(bytes, cols, rows, ct)
+      case ct: UByteCells => UByteArrayTile.fromBytes(bytes, cols, rows, ct)
+      case ct: ShortCells => ShortArrayTile.fromBytes(bytes, cols, rows, ct)
+      case ct: UShortCells => UShortArrayTile.fromBytes(bytes, cols, rows, ct)
+      case ct: IntCells => IntArrayTile.fromBytes(bytes, cols, rows, ct)
+      case ct: FloatCells => FloatArrayTile.fromBytes(bytes, cols, rows, ct)
+      case ct: DoubleCells => DoubleArrayTile.fromBytes(bytes, cols, rows, ct)
+    }
+
+  def fromBytes(bytes: Array[Byte], t: CellType, cols: Int, rows: Int, replaceNoData: Double): MutableArrayTile =
```
echeipesh: @lossyrob:  + for comment

This isn't actually used anywhere. It seems like since we support user defined nodata values explicitly now there is no use case for this. I started thinking that the name is confusing for this overload but do we even need to provide this now?

moradology: I second this thought, given the added support. If there's not another use case in mind, I think we can be rid of it

raster/src/main/scala/geotrellis/raster/io/geotiff/ByteConstantNoDataGeoTiffSegment.scala
```diff
@@ -0,0 +1,17 @@
...
+import geotrellis.raster.io.geotiff.utils._
+import geotrellis.raster.io.geotiff.compression._
+
+import java.util.BitSet
+
+import spire.syntax.cfor._
+
+class ByteConstantNoDataCellTypeGeoTiffSegment(bytes: Array[Byte]) extends ByteGeoTiffSegment(bytes) {
+  def getInt(i: Int): Int = b2i(get(i))
+  def getDouble(i: Int): Double = b2i(get(i))
```
echeipesh: `b2d`

raster/src/main/scala/geotrellis/raster/io/geotiff/ByteGeoTiffSegment.scala
```diff
@@ -4,57 +4,24 @@ import geotrellis.raster._
...
-        arr.toArrayByte()
-      case TypeDouble =>
-        val arr = Array.ofDim[Double](size)
-        cfor(0)(_ < size, _ + 1) { i => arr(i) = getDouble(i) }
-        arr.toArrayByte()
-    }
+  def getInt(i: Int): Int
+  def getDouble(i: Int): Double
+
+  protected def intToByteOut(v: Int): Byte
```
echeipesh: Is this a remnant of a refactor? This is the only segment that has a method for this rather than just inlining the macros.

raster/src/main/scala/geotrellis/raster/io/geotiff/Float64UserDefinedNoDataGeoTiffSegment.scala
```diff
@@ -0,0 +1,20 @@
...
+
+import java.nio.ByteBuffer
+import spire.syntax.cfor._
+
+import java.util.BitSet
+
+class Float64UserDefinedNoDataGeoTiffSegment(bytes: Array[Byte], val userDefinedDoubleNoDataValue: Double)
+    extends Float64GeoTiffSegment(bytes)
+       with UserDefinedDoubleNoDataConversions {
+  def getInt(i: Int): Int = d2i(get(i))
```
echeipesh: seems this should be `udd2i`

spark/src/main/scala/geotrellis/spark/ingest/CellGridPrototype.scala
```diff
@@ -0,0 +1,8 @@
+package geotrellis.spark.ingest
+
+import geotrellis.raster.CellType
+
+trait CellGridPrototype[T] {
+  def prototype(cols: Int, rows: Int): T
+  def prototype(cellType: CellType, cols: Int, rows: Int): T
+}
```
echeipesh: What is going on here? Looks like something from raster but we have those `TilePrototypeMethods`.

spark/src/main/scala/geotrellis/spark/io/avro/codecs/TileCodecs.scala
```diff
@@ -16,12 +17,14 @@ trait TileCodecs {
       .fields()
       .name("cols").`type`().intType().noDefault()
       .name("rows").`type`().intType().noDefault()
+      .name("celltype").`type`().stringType().noDefault()
```
echeipesh: This will need a default for us to be able to read existing records, which are all constant no data types

spark/src/main/scala/geotrellis/spark/io/avro/codecs/TileCodecs.scala
```diff
@@ -32,7 +35,9 @@ trait TileCodecs {
         .asScala // notice that Avro does not have native support for Short primitive
         .map(_.toShort)
         .toArray
-      new ShortArrayTile(array, rec[Int]("cols"), rec[Int]("rows"))
+      val cellType = CellType.fromString(rec[Utf8]("celltype").toString)
```
echeipesh: Since this is stored per record I'm going to pinch bits here. The first part of cell type, the data type is irrelevant because you already know that. We have to specialize avro codecs per data type so we allocate arrays of min size. Looks like what is needed here is some kind of enum for the type of no data, `RAW`, `CONSTANT`, `USER_DEFINED` (with default being `CONSTANT`) and then an optional NoData field. This would also make it easier to read for a hypothetical client that used our avro schema.

raster/src/main/scala/geotrellis/raster/ArrayTile.scala
lossyrob: :+1:

spark/src/main/scala/geotrellis/spark/io/avro/codecs/TileCodecs.scala
echeipesh: Actually perhaps even more economical is to include optional NoData value and then inspect it's value to determine if constant or user defined tile is appropriate.

raster/src/main/scala/geotrellis/raster/io/geotiff/ByteConstantNoDataGeoTiffSegment.scala
lossyrob: Is this just converting to the same type? If so we should just pass it through and not copy.

spark/src/main/scala/geotrellis/spark/io/avro/codecs/TileCodecs.scala
moradology: A couple issues:
1. the enum doesn't seem like it would be capable of supporting the user defined case (we'd need to send the user defined nodata in addition to the enum)
2. the optional nodata approach breaks backward compatibility since a default would prevent us from encoding raw tiles

raster/src/main/scala/geotrellis/raster/io/arg/ArgReader.scala
moradology: This is perhaps tough to do, given the constraints that supporting a BitType force upon us

echeipesh:
+1

echeipesh:
Since you're going to have to look at this again to figure out the error, could you fix the `CutTiles.scala` problem where it should use the cellType in the prototype. That is made more problematic by this PR.

PR #1273, opened 2016-01-07
Pending IO refactor: KeyIndex attributeStore Format

pomadchin:
Now we store `KeyIndex[K]` as a serialized Java object, we should try to store it more generic (`json` / `avro`).


pomadchin:
It's WIP: 
current problems: 
 * Reindexer
 * Failed tests

spark/src/main/scala/geotrellis/spark/io/json/package.scala
```diff
@@ -1,21 +1,228 @@
...
+          if (id != KeyIndexIds.zSpatialKeyIndex) throw new DeserializationException("12: Wrong KeyIndex type.")
+          new ZSpatialKeyIndex()
+        }
+        case _ =>
+          throw new DeserializationException("13: Wrong KeyIndex type.")
+      }
+    }
+  }
+
+  implicit def keyIndexFormat[K: ClassTag] = new RootJsonFormat[index.KeyIndex[K]] {
```
pomadchin: May be it is possible to make it similar to our `AvroUnionCodec[T]`

pomadchin:
@lossyrob NOT: fixed a little a new `file` backend.

pomadchin:
@lossyrob @echeipesh and others, waiting for your comments

lossyrob:
@pomadchin what did you fix in the file backend?

pomadchin:
@lossyrob think we dont need type params in a LayerDeleter: https://github.com/pomadchin/geotrellis/blob/feature/keyindex-json/spark/src/main/scala/geotrellis/spark/io/file/FileLayerDeleter.scala

lossyrob:
I don't like how this adds the type parameter everywhere. Which is a direct result of the KeyIndexMethod not being in the signatures of the method calls instead of in the class creations. we should use this as an opportunity to fix that.

For example, writers should not take a key index method, which breaks the base `Writer` triat. But that's fine. We can have it implement another trait, and if someone wants a (LayerId, RDD) => Unit call, we can provide a `writer` method that takes a KeyIndex or a KeyIndexMethod that gives them that sort of writer. This should be made in collaboration with @jamesmcclain, since a writer should be able to take either a KeyIndexMethod or a KeyIndex (which would allow key indexes of other layers to be passed in, for instance when updating or copying a layer).

Also, "custom" is not a very good name...we should figure out a better API for that.

pomadchin:
TODO:
* fix tests
* clean up api

pomadchin:
Guys, what about a name for default overload? 
* `defaultWrite`
 * `write`, and for defining special `KeyIndex` `_write`

Some other variants are possible.

pomadchin:
`BoundedKeyIndex` issue: https://github.com/geotrellis/geotrellis/issues/1291

spark/src/main/scala/geotrellis/spark/io/FilteringLayerReader.scala
```diff
@@ -1,17 +1,26 @@
...
+import geotrellis.spark.io.json._
+import spray.json.JsonFormat
 
-abstract class FilteringLayerReader[ID, K: Boundable, M, ReturnType] extends LayerReader[ID, ReturnType] {
+abstract class FilteringLayerReader[ID, K: Boundable, M, ReturnType] extends LayerReader[ID, K, ReturnType] {
 
-  def read(id: ID, rasterQuery: RDDQuery[K, M], numPartitions: Int): ReturnType
+  def read[I <: KeyIndex[K]: JsonFormat](id: ID, rasterQuery: RDDQuery[K, M], numPartitions: Int, format: JsonFormat[I]): ReturnType
+
+  def read[I <: KeyIndex[K]: JsonFormat](id: ID, numPartitions: Int, format: JsonFormat[I]): ReturnType =
```
echeipesh: Aren't you asking for `JsonFormat[I]` implicitly here? Why ask for it as an explicit parameter ?

pomadchin: Just a way to resolve overloads @lossyrob suggested this idea (may be I loosed smth). But any comments are appreciated: there is another way of solving this issue to give special names for functions with same types.

echeipesh: If you're getting it explicitly, you shouldn't need to ask for it implicitly.
Further, if you're basically asking for `JsonFormat[KeyIndex[K]]`, do you still need `I`? It doesn't look like you care what the subtype is, you just want to use it through the base interface.

pomadchin: We should know how to deserialise the `KeyIndex`, and we need to know its type. But we can use generic `KeyIndex[K]` for in-built types (look at the json format for the `KeyIndex[K]` type).

echeipesh: I'm not crazy about the idea of providing the format per read instruction, its one of those cases where you couple code structure to data structure. 

Perhaps we should consider making the `JsonFormat[KeyIndex[K]]` extensible. That class already has to partially read and match some some tag to figure out what parser to use. Why not expose that to the user and allow them to bring that in through higher precedence import. Something like:

```scala
class ChainedJsonFormat[T](
  base: JsonFormat[T], 
  fread: PartialFunction[JsValue,T], 
  fwrite PartialFunction[T,JsValue]
) extends JsonFormat[T] {
  def read(json: JsValue) = fread.applyOrElse(base.read(json))
  def write(thing: T) = fwrite.applyOrElse(base.write(thing))
}

object package spark
   implicit class withExtentFormat[T](fmt: JsonFormat[T]) {
     def extendFormat(fread: PartialFunction[JsValue,T], fwrite PartialFunction[T,JsValue]) =
		new ChainedJsonFormat(fmt, fread, fwrite)  
   }
}

object SomeUserModule {
  implicit val myProjectKeyFormat: JsonFormat[BoundKeyIndex[K]] = geotrellis.spark.json.baseKeyIndexFormat.extendFormat {
    json => {//parse me some json for y KeyIndex},
    thing => {// write it out using my rules}
  }

}
```

lossyrob: Here's an idea. We have the the `JsonFormat[KeyIndex[K]]` as a constructor parameter. It's never really used because the object applys will pass in the default one (through implicit resolution):

```
object SomethingLayerReader {
  def apply[K,V,M](path: String)(implicit keyIndexFormat: JsonFormat[KeyIndex[K]]) =
    new SomethingLayerReader(path, keyIndexFormat)
}
```

but we have something like a `withKeyIndexFormat` that allows the user to supply a custom format type (if they don't use the implicit resolution to define a newer one that is higher on the implicit scope). So it would look like

```scala
SomethingLayerReader(path).withKeyIndexFormat(customKeyIndexFormatter)
```

That way they could pass in a new formatter that the layer reader could use. If build up some API like Eugene is suggesting to make it extendable, then the user could create a JsonFormat[KeyIndex[K]] by adding new "type" property resolutions to it. Something like

```scala

implicit val customKeyIndexFormat: JsonFormat[KeyIndex[K]] =
  KeyIndexFormat.addType[K, CustomKeyIndexFormat[K]]("custom")
```

where that addType would look for the implicit type-specific resolution and add it to a Keyindex format with the new type tag:

```scala
object KeyIndexFormat {
  def addType[K, I <: KeyIndex[K]: JsonFormat](typeName: String)(implicit generalKeyIndexFormat: JsonFormat[KeyIndex[K]]): JsonFormat[KeyIndex[K]] =
    generalKeyIndexFormat.extent(typeName, implicitly[JsonFormat[I]])
}
```

lossyrob:
There should be a pause in development on this. I think I have a good solution but I need to know where the work to get KeyBounds in to the RasterMetaData is in order to move forward with it. We can talk about it on Monday.

spark/src/main/scala/geotrellis/spark/io/FilteringLayerReader.scala
echeipesh: I very much like the idea of `SomethingLayerReader(path).withKeyIndexFormat(customKeyIndexFormatter)`, fluent APIs are hot right now.

I would however keep the machinery of extending the KeyIndexFormat as simple as possible. Having `KeyIndexFormat.addType` do an implicit resolution doesn't seems necessary. At this point the user is extending a known thing with a known thing at a specific point before passing in a concrete object to the reader.

lossyrob:
Superseded by #1346

PR #1275, opened 2016-01-08
Check the cast in Geometry#apply for earlier failures

dwins:
Geometry#apply includes an unsafe cast for convenience when converting general JTS Geometry instances into geotrellis.vector.Geometry.  However because the cast is done with type parameter instead of a concrete class, no check is actually performed in the apply method, and in the event of a programmer error the runtime exception may happen far from the call.  With this PR, Geometry.apply would perform the check before the converted object is returned, throwing an exception on failure.

### DIFF
--- a/vector/src/main/scala/geotrellis/vector/affine/AffineTransformation.scala
+++ b/vector/src/main/scala/geotrellis/vector/affine/AffineTransformation.scala
@@ -22,7 +22,7 @@ trait AffineTransformation {
   def transform(geom: MultiLine): MultiLine = MultiLine(transform(geom.jtsGeom))
   def transform(geom: MultiPolygon): MultiPolygon = MultiPolygon(transform(geom.jtsGeom))
   def transform(geom: GeometryCollection): GeometryCollection = GeometryCollection(transform(geom.jtsGeom))
-  def transform(geom: Geometry): Geometry = Geometry(transform(geom.jtsGeom))
+  def transform(geom: Geometry): Geometry = Geometry[Geometry](transform(geom.jtsGeom))
 
   private def transform[D <: jts.Geometry](g: D): D = trans.transform(g).asInstanceOf[D]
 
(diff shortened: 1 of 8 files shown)


lossyrob:
@dwins  Looks like there's some travis errors; one recommendation is to run `./buildall.sh` locally before finishing off the PR to make sure travis won't throw errors.

vector/src/main/scala/geotrellis/vector/affine/AffineTransformation.scala
```diff
@@ -22,7 +22,7 @@ trait AffineTransformation {
   def transform(geom: MultiLine): MultiLine = MultiLine(transform(geom.jtsGeom))
   def transform(geom: MultiPolygon): MultiPolygon = MultiPolygon(transform(geom.jtsGeom))
   def transform(geom: GeometryCollection): GeometryCollection = GeometryCollection(transform(geom.jtsGeom))
-  def transform(geom: Geometry): Geometry = Geometry(transform(geom.jtsGeom))
+  def transform(geom: Geometry): Geometry = Geometry[Geometry](transform(geom.jtsGeom))
```
lossyrob: This call looks a bit awkward...maybe a reason to not use apply and have an explicit method call?

```scala
def transform(geom: Geometry): Geometry =
  Geometry.wrap[Geometry](transform(geom.jtsGeom))
```

also we have implicit conversions for JTS geometries to wrapped types, except for the base class for some reason. e.g. https://github.com/dwins/geotrellis/blob/fix-geometry-apply/vector/src/main/scala/geotrellis/vector/Point.scala#L29

If we did have this conversion for Geometry, this could be written as 

```scala
def transform(geom: Geometry): Geometry =
  transform(geom.jtsGeom)
```

although, I think being explicit is better in this case. Even if it's not used here it would be good to add that implicit conversion to match the rest of the geometries.

I'm not really in love with any of these, but leaning towards `Geometry.wrap`. Thoughts?

dwins:
How about not casting at all in Geometry.apply? It can just wrap jts.Geometry to a geotrellis.vector.Geometry and let calling code worry about downcasting if needed.  This gives some symmetry with the apply methods for specific Geometry types.  I've updated the branch showing what this would look like and it doesn't seem too disruptive.

lossyrob:
This is perhaps a better solution. I'd like to try and minimize the number of times a client would have to use `asInstanceOf` in their code, since it's considered Bad practice. This might not be bad because the user would only have to do this if they are doing JTS stuff directly with the JTS geometries, and so someone will have to cast at some point.

Interested if others have thoughts.

echeipesh:
+1 Exactly what we were talking about last and it still looks pretty.

PR #1279, opened 2016-01-11
Added filesystem backend support.

lossyrob:
This PR adds a `geotrellis.spark.io.file` backend, which works off of a file system. This should be much faster than using the hadoop backend for the local case; especially for reading single tiles off disk. It carries over some caching from S3, so there's an opportunity there to load a small-to-medium dataset off of disk into memory, which would serve out tiles as fast as 0.9 catalogs with layers cached in memory.


spark/src/main/scala/geotrellis/spark/io/file/FileLayerMover.scala
```diff
@@ -0,0 +1,72 @@
...
+
+        // Read the metadata file out.
+        val (header, metadata, keyBounds, keyIndex, writerSchema) = try {
+          sourceAttributeStore.readLayerAttributes[FileLayerHeader, M, KeyBounds[K], KeyIndex[K], Schema](from)
+        } catch {
+          case e: AttributeNotFoundError => throw new LayerReadError(from).initCause(e)
+        }
+
+        // Delete the metadata file in the source
+        sourceMetadataFile.delete()
```
moradology: This should probably happen only once we've written the metadata out successfully, it seems to me

spark/src/main/scala/geotrellis/spark/io/index/IndexRanges.scala
```diff
@@ -0,0 +1,43 @@
...
+      (range._1, range._1 + take - 1) -> (range._1 + take, range._2)
+    }
+
+    val arr = Array.fill(count)(Nil: List[(Long, Long)])
+    var sum = 0l
+    var i = 0
+    while (stack.nonEmpty) {
+      if (len(stack.head) + sum <= binWidth){
+        val take = stack.head
+        arr(i) = take :: arr(i)
```
moradology: This seems better, cosmetically:
```scala
val head = stack.head
if (len(head) + sum <= binWidth){
  arr(i) = head :: arr(i)
  sum += len(head)
}
```

moradology:
That test looks like it was some kind of anomaly. +1 after considering the couple of things i had questions about

PR #1283, opened 2016-01-11
Remove 'Extent' from the Geometry type hierarchy

dwins:
This does have an impact throughout the codebase since some parts of the code convert tiles to features and back.  In this branch I've simply used PolygonFeature for this instead of ExtentFeature - I could also look into extending Feature to support "geometries" that are not part of the Geometry type hierarchy to avoid converting back-and-forth.


echeipesh:
+1

PR #1285, opened 2016-01-12
Improved Tiler API, refactor SpaceTimeKey

lossyrob:
This PR improves the Tiler API, matching it more closely with the way we're adding implicit methods in other places in the library, and getting rid of a lot of complexity around using the Tiler. The functionality is now accessed through a "cutTiles" method.

This also renames the `mosiac` packages to `merge` in both `geotrellis.raster` and `geotrellis.spark`, and fits them to the new `MethodExtensions` design pattern.

Also:
- Changes SpaceTimeKey to internally store the instant of the time as a Long (the long returned by DateTime.getMillis).
- Added a ZCurveIndexMethod implementation for SpaceTimeKey indexing that uses division by millisecond resolution instead of working with the date times (/cc @pomadchin)
- Allows the user to not try and match the layer extent on tile RDD reproject. On large enough layers, this produces artifacts.


spark/src/main/scala/geotrellis/spark/reproject/Reproject.scala
```diff
@@ -0,0 +1,23 @@
+package geotrellis.spark.reproject
+
+import geotrellis.raster.reproject.Reproject.{Options => RasterReprojectOptions}
```
lossyrob: Not psyched about this. I don't the name clashing that is going on with `geotrellis.vector.reproject.Reproject`, `geotrellis.raster.reproject.Reproject`, and now this Reproject. Need to think of a better pattern here.

spark/src/main/scala/geotrellis/spark/ingest/Ingest.scala
```diff
@@ -61,15 +61,14 @@ object Ingest {
...
-    (implicit tiler: Tiler[T, K, Tile]): Unit =
+    (sink: (RasterRDD[K], Int) => Unit): Unit =
   {
     sourceTiles.persist()
     val reprojectedTiles =
       sourceTiles.reproject(destCRS, resampleMethod).cache()
     val (zoom, rasterMetaData) =
       RasterMetaData.fromRdd(reprojectedTiles, destCRS, layoutScheme)(_.projectedExtent.extent)
-    val tiledRdd = tiler(reprojectedTiles, rasterMetaData, resampleMethod).cache()
+    val tiledRdd = reprojectedTiles.cutTiles(rasterMetaData, resampleMethod).cache()
```
echeipesh: Isn't this missing something that will `.reduceByKey` on the cut tiles?
It looks like this could a call to `tileToLayout`.

lossyrob: Good catch, that should be a `tileToLayout` call

echeipesh:
+1 after tileToLayout change. Really like the tiler change, makes it much easier to use both on dev and client side.

PR #1286, opened 2016-01-14
Update Spark and Accumulo

pomadchin:
Update up to Spark 1.5.2 and to Accumulo 1.7.0


pomadchin:
Want to test it on a real Accumulo before merge.

pomadchin:
Tested [here](https://github.com/pomadchin/accumulo-spark/tree/single-node-512m-accumulo-1.7) on a chatta demo project building it in a [java-8-container](https://hub.docker.com/r/daunnc/java-8-oracle/), will publish the complete guide how I use docker environment.

lossyrob:
Doesn't 1.7 Accumulo have what we needed BatchInputFormat for? @echeipesh could we get rid of this?

echeipesh:
Yep, we should be able to use this: https://accumulo.apache.org/1.7/apidocs/org/apache/accumulo/core/client/mapreduce/InputFormatBase.html#setBatchScan(org.apache.hadoop.mapreduce.Job,%20boolean)

We haven't changed our methodology since then.

lossyrob:
Marking this as WIP, since we are going to test using `setBatchScan` instead of `BatchInputFormat`

pomadchin:
Tested on Accumulo 1.6.4 and Accumulo 1.7.0, 2 node cluster configuration. :+1:

spark/build.sbt
```diff
@@ -7,7 +7,8 @@ libraryDependencies ++= Seq(
     exclude("org.apache.hadoop", "hadoop-client"),
   "org.apache.spark" %% "spark-core" % Version.spark % "provided",
   "org.apache.hadoop" % "hadoop-client" % Version.hadoop % "provided",
-  "de.javakaffee" % "kryo-serializers" % "0.27",
+  "de.javakaffee" % "kryo-serializers" % "0.27" exclude("com.esotericsoftware.kryo", "kryo"),
```
pomadchin: Can downgrade to `javakaffee 0.24` only after we decide what to do with `org.joda.time.Interval` (is it really necessary to use it)

PR #1290, opened 2016-01-20
MultiBand subset Method

jamesmcclain:
`subset(...)` methods for the the `ArrayMultiBandTile` and `MultiBandGeoTiff` types have been created.  These methods allow for the efficient creation of new instances from a subset of the bands of some parent instance.


raster-test/src/test/scala/geotrellis/raster/io/geotiff/GeoTiffMultiBandTileSpec.scala
```diff
@@ -12,11 +12,11 @@ import geotrellis.testkit._
 
 import org.scalatest._
 
-class GeoTiffMultiBandTileSpec extends FunSpec
+class XGeoTiffMultiBandTileSpec extends FunSpec
```
lossyrob: X?

raster/src/main/scala/geotrellis/raster/ArrayMultiBandTileSubsetMethod.scala
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package geotrellis.raster
+
+
+trait ArrayMultiBandTileSubsetMethod[T <: ArrayMultiBandTile] extends MethodExtensions[T] {
+
+  def subset(m: Map[Int, Int]): ArrayMultiBandTile = {
```
lossyrob: Why should this be a map? I was thinking it would be a sequence of ints, with overloads to support passing values in directly:

```scala
def subset(bands: Seq[Int]): ArrayMultiBandTile = ???
def subset(bands: Int*)(implicit d: DummyImplicit): ArrayMultiBandTile = ???
```

You might need the dummy implicit to help resolve the overloads (scalac will tell you if it can't because signatures have the same type after type erasure).

raster-test/src/test/scala/geotrellis/raster/io/geotiff/GeoTiffMultiBandTileSpec.scala
jamesmcclain: I just wanted to give this one a unique prefix compared to the other tests which match the pattern `*GeoTiffMultiBandTileSpec` so that I could run this one without the others.  I don't know if there is an easier way to do this.  This change does not need to be permanent (can be removed before merging).

raster/src/main/scala/geotrellis/raster/ArrayMultiBandTileSubsetMethod.scala
jamesmcclain: I guess that I wanted to treat this more as a permutation with subset being a special case.  I can add these overloads, as well.

raster-test/src/test/scala/geotrellis/raster/io/geotiff/GeoTiffMultiBandTileSpec.scala
lossyrob: Yeah, we should have a clearer name for this, or they should be the same test.

raster/src/main/scala/geotrellis/raster/io/geotiff/MultiBandGeoTiffSubsetMethod.scala
```diff
@@ -0,0 +1,47 @@
...
+
+package geotrellis.raster.io.geotiff
+
+import geotrellis.raster.{MethodExtensions, Tile, ArrayMultiBandTile}
+import geotrellis.raster.io.geotiff._
+
+
+trait MultiBandGeoTiffSubsetMethod[T <: MultiBandGeoTiff] extends MethodExtensions[T] {
+
+  def permutation(m: Map[Int, Int]): ArrayMultiBandTile = {
```
lossyrob: Why do we need a map argument at all? I don't understand where we'd want to call permutation.

Also these methods are core enough to MultiBandTile to have directly in the MultiBandTile trait.

jamesmcclain: Okay, will update in the morning.

raster/src/main/scala/geotrellis/raster/io/geotiff/MultiBandGeoTiff.scala
```diff
@@ -20,31 +20,48 @@ class MultiBandGeoTiff(
       case gtt: GeoTiffMultiBandTile => gtt
       case _ => GeoTiffMultiBandTile(tile)
     }
+
+  def subset(mapping: Seq[Int]): ArrayMultiBandTile = {
```
lossyrob: `mapping` => `bands`

jamesmcclain:
The `X` has been removed from the beginning of the `GeoTiffMultiBandTileSpec` test.

raster/src/main/scala/geotrellis/raster/ArrayMultiBandTile.scala
```diff
@@ -269,4 +269,20 @@ class ArrayMultiBandTile(bands: Array[Tile]) extends MultiBandTile {
     }
     result
   }
+
+  def subset(bands: Seq[Int]): ArrayMultiBandTile = {
+    val newBands = Array.ofDim[Tile](bands.size)
+    var i = 0
+
+    assert(bands.size <= this.bandCount)
```
echeipesh: this should use `require`, that produces `IllegalArgumentException` and gives a chance for an error message.

raster/src/main/scala/geotrellis/raster/io/geotiff/MultiBandGeoTiff.scala
```diff
@@ -20,31 +20,48 @@ class MultiBandGeoTiff(
       case gtt: GeoTiffMultiBandTile => gtt
       case _ => GeoTiffMultiBandTile(tile)
     }
+
+  def subset(bands: Seq[Int]): ArrayMultiBandTile = {
+    val mbTile = this.tile
+    val newBands = Array.ofDim[Tile](bands.size)
+    var i = 0
+
+    assert(bands.size <= mbTile.bandCount)
```
echeipesh: ditto on `require`

jamesmcclain: Okay, will-do

raster-test/src/test/scala/geotrellis/raster/io/geotiff/GeoTiffMultiBandTileSpec.scala
```diff
@@ -110,6 +110,50 @@ class GeoTiffMultiBandTileSpec extends FunSpec
...
 
+  describe("MutliBand subset methods") {
+
+    it("subset should be inexpensive") {
+      val tile0 = MultiBandGeoTiff(geoTiffPath("3bands/int32/3bands-striped-pixel.tif"))
+      val tile1 = tile0.subset(List(1, 2, 0))
+
+      tile0.band(0) should be (tile1.band(2))
+      tile0.band(1) should be (tile1.band(0))
+      tile0.band(2) should be (tile1.band(1))
```
moradology: I think you might need to use 'theSameInstanceAs' for this test: http://www.scalatest.org/user_guide/using_matchers

lossyrob: Seems like it's just doing the object equals anyway, but it would be better to explicitly state in the test that it's an object reference equality, +1

jamesmcclain: Okay, will-do.

PR #1297, opened 2016-01-26
Partitioners set on any operations that have that ability

pomadchin:
Depends on #1230


vector/src/main/scala/geotrellis/vector/Projected.scala
```diff
@@ -3,8 +3,11 @@ package geotrellis.vector
...
 object Projected {
-  implicit def fromTupleA[T](tup: (T, CRS)):ProjectedExtent = Projected(tup._1, tup._2)
-  implicit def fromTupleB[T](tup: (CRS, T)):ProjectedExtent = Projected(tup._2, tup._1)
+  implicit def fromTupleA[T](tup: (T, CRS)): Projected[T] = Projected(tup._1, tup._2)
+  implicit def fromTupleB[T](tup: (CRS, T)): Projected[T] = Projected(tup._2, tup._1)
 }
 
-case class Projected[T](obj: T, crs: CRS) extends Product2[T, CRS]
\ No newline at end of file
+case class Projected[T](obj: T, crs: CRS) extends Product2[T, CRS] {
```
pomadchin: @echeipesh there is an important fix, think we can rebase other pr's on top of this / make this small fix as a separate pr.

spark/src/main/scala/geotrellis/spark/op/local/LessTileRDDMethods.scala
```diff
@@ -73,13 +74,13 @@ trait LessTileRDDMethods[K] extends TileRDDMethods[K] {
...
+  def localLess(other: RDD[(K, Tile)], partitioner: Option[Partitioner] = None): RDD[(K, Tile)] =
+    self.combineValues(other, partitioner)(Less.apply)
 
   /**
     * Returns a Tile with data of TypeBit, where cell values equal 1 if
     * the corresponding cell valued of the rasters are less than the next
     * raster, else 0.
     */
-  def <(other: RDD[(K, Tile)]) = localLess(other)
+  def <(other: RDD[(K, Tile)], partitioner: Option[Partitioner] = None): RDD[(K, Tile)] = localLess(other, partitioner)
```
echeipesh: Not needed here. Infix operations can't take partitioner parameter.

echeipesh:
+1 after comment

spark/src/main/scala/geotrellis/spark/op/local/LessTileRDDMethods.scala
pomadchin: Ok, missed it ):

PR #1299, opened 2016-01-26
Pending IO Refactor: Preallocated Keybounds

jamesmcclain:
The following changes have been made:

   * A new `Option[KeyBounds[K]]` parameter has been added to the write method of the layer writer superclass (`Writer.scala`).  The motivation here is to allow the user to write into a key range other than that of the rdd that is being written (so that key space can be preallocated).
   * The x- and y-dimensions had been transposed in the Hilbert Index creation code.  This was probably not detected earlier because the two dimensions are typically about the same size?
   * The calculation of the location on the Hilbert curve corresponding to a particular key has been adjusted.
   * A corner-case in the calculation of the number of bits required to represent numbers has been adjusted.
   * ~~Some of the Hilbert index tests seem to have been incorrect: they have been changed to be consistent with [this page](http://mathworld.wolfram.com/HilbertCurve.html), where it says:~~  Embedding points into a smaller or larger space seems to cause rotations and reflections of the space with respect to curve-order.  Some changes were made to the Hilbert curve implementation which caused the tests to need to be updated, as well.  The reference for the "new" expected results is [this page](http://mathworld.wolfram.com/HilbertCurve.html), where it says:

   > Traversing the polyhedron vertices of an n-dimensional hypercube in Gray code order produces a generator for the n-dimensional Hilbert curve.

See also: https://github.com/pomadchin/geotrellis/issues/2

Still needs:
   - [ ] #1273 to be merged
   - [x] A `README.md` file in the `index/hilbert` directory documenting behavior, current limitations, and factoids about the underlying `uzaygezen` library.
   - [x] Change the `resolution` method to take the two endpoints rather than their difference
   - [x] Determine definitively whether it is possible to index out of bounds


pomadchin:
Depends on #1273

lossyrob:
Superseded by #1346

PR #1303, opened 2016-01-29
Hilbert Curve Fixes

jamesmcclain:
Contains the following changes:

   * The x- and y-dimensions had been transposed in the Hilbert Index creation code.  This was probably not detected earlier because the two dimensions are typically about the same size?
   * The calculation of the location on the Hilbert curve corresponding to a particular key has been adjusted.
   * A corner-case in the calculation of the number of bits required to represent numbers has been adjusted.
   * The resolution function has been changed to take the minimum and maximum of the range as parameters, rather than the length as a single parameter.  This makes usage mistakes less likely.

https://github.com/geotrellis/geotrellis/pull/1303 ⊂ https://github.com/geotrellis/geotrellis/pull/1299
https://github.com/geotrellis/geotrellis/pull/1303 ⊂ https://github.com/pomadchin/geotrellis/pull/2


echeipesh:
+1

jamesmcclain:
Many thanks

PR #1306, opened 2016-02-04
Port grid-shifting related functionality from proj.4

### LINKED ISSUE #1101: Issue with projecting to EPSG:27700
From @mfsny on Gitter

I have my data in the British EPSG:27700 and my leaflet is using WSG84 for passing bbox. I tried following first: 

```scala
val src = CRS.fromName("EPSG:3857") 
val dst = CRS.fromName("EPSG:27700") 
val oldx = Extent.fromString("-39135.75848201024,6692214.700423751,0,6731350.458905762") 
val newx = oldx.reproject(src,dst) 
```
but all I get is `geotrellis.vector.ExtentRangeError: y: -1.3747887674479786E7 to -1.374961923978705E7`.

So I checked the proj4 string in /nad/epsg and it looks like the value (`<27700> +proj=tmerc +lat_0=49 +lon_0=-2 +k=0.9996012717 +x_0=400000 +y_0=-100000 +datum=OSGB36 +units=m +no_defs <>`) is not correct. I found some other defs online and this one worked the best: `+proj=tmerc +lat_0=49 +lon_0=-2 +k=0.9996012717 +x_0=400000 +y_0=-100000 +ellps=airy +towgs84=446.448,-125.157,542.06,0.15,0.247,0.842,-20.489 +units=m +no_defs`. But it has still some southwards.
I get 514302.4577355564,188783.730894182,537991.4618278733,213686.72817825677 when I reproject using geotrellis. http://cs2cs.mygeodata.eu/ says it should be : 514771.247816,167974.826075,538560.963803,192915.540317 .

dwins:
As discussed elsewhere, the lack of grid shift functionality in proj4j is causing erroneous output for some projection transformations.  This is the branch where I'm working on porting proj.4's grid shift system to proj4j.

Fixes #1101


lossyrob:
@dr-jts in case your interested.

dr-jts:
Great!

dwins:
@lossyrob want to take a look at this? One big question I had for you was about the script I wrote to generate `proj4-epsg.csv`, the MetaCRS test suite using the original proj.4 as an oracle.  I just put a `Main.scala` in the test directory which I've been running with `sbt "project proj4" test:run` but maybe you'd prefer to manage it differently.

Anyway, the current approach to testing is using the `testMethod` field of the MetaCRS file to indicate the expected status of each test - `passing` for tests that are within tolerance, `failing` for tests that successfully generate results outside of the tolerance, and `error` for tests that throw unhandled exceptions instead of generating results.  This was generated using the script I mentioned earlier.

The test script reads in the entire epsg database and attempts to convert (1E, 1S) from LatLng to each projection in the epsg database using the `cs2cs` command line utility from the proj.4 project.  If cs2cs generates non-NaN results, then we attempt to repeat the conversion using proj4j.  A tolerance is set for each test by inspecting the CRS information - basically angular coordinate systems set it to 1e-6 and linear ones set it to 1e-2.

While investigating I've been manually editing the generated MetaCRS to change `failing` or `error` statuses to `passing` in order to get a test failure for one CRS at a time.

proj4/src/test/scala/Main.scala
```diff
@@ -0,0 +1,123 @@
...
+
+  def main(args: Array[String]): Unit = {
+    val registeredCodes = 
+      loan(scala.io.Source.fromInputStream(getClass().getResourceAsStream("/geotrellis/proj4/nad/epsg"))) { source =>
+        source.getLines
+          .filter { _ startsWith "<" }
+          .map { s => s.tail.take(s.indexOf('>') - 1) }
+          .to[Set]
+      }
+    val knownCodes =
```
lossyrob: Should this file be moved into `test/resources`?

proj4/src/main/java/org/osgeo/proj4j/datum/Axes.java
```diff
@@ -0,0 +1,112 @@
+package org.osgeo.proj4j.datum;
+
+import org.osgeo.proj4j.ProjCoordinate;
+
+public final class Axes {
```
lossyrob: This, and other classes, must be marked as Serializable (spark tests will fail with serialization errors)

proj4/src/test/scala/Main.scala
dwins: No, this is used at runtime for resolving CRSs from the EPSG authority.  The output does go to the test resource directory.

proj4/src/main/java/org/osgeo/proj4j/datum/Axes.java
dwins: Ok. This explains why all the spark tests are failing. Easy fix.

proj4/src/test/scala/Main.scala
```diff
@@ -0,0 +1,123 @@
...
+  def main(args: Array[String]): Unit = {
+    val registeredCodes = 
+      loan(scala.io.Source.fromInputStream(getClass().getResourceAsStream("/geotrellis/proj4/nad/epsg"))) { source =>
+        source.getLines
+          .filter { _ startsWith "<" }
+          .map { s => s.tail.take(s.indexOf('>') - 1) }
+          .to[Set]
+      }
+    val knownCodes = 
+      loan(scala.io.Source.fromFile("/Users/dwins/Projects/radiantblue/proj.4/nad/epsg")) { source =>
```
lossyrob: Sorry, the above comment was about this line. `/Users/dwins/` probably shouldn't be there, so where should this file go?

lossyrob:
There's a hanging unused directory in `proj4/resources`, `nad`, we should remove that

proj4/src/main/java/org/osgeo/proj4j/Grid.java
```diff
@@ -0,0 +1,322 @@
...
+        // search path for grid definition files, but for now we only check the
+        // working directory and the classpath (in that order.)
+        File file = new File(gridName);
+        if (file.exists()) return new DataInputStream(new FileInputStream(file));
+        InputStream resource = Grid.class.getResourceAsStream("/geotrellis/proj4/nad/" + gridName);
+        if (resource != null) return new DataInputStream(resource);
+
+        return null;
+    }
+}
```
lossyrob: Move to the `datum` folder (already in the package)

lossyrob:
Notes for docs:

_Ellipsoid_ models the 3D planet via a polar and equatorial radii
_Datum_ defines the mapping between the LatLng euclidean plane (WGS84) and the polar coordinates of an Ellipsoid.

The transformation of the Datum can be represented by 3 or 7 (the latter being a Helmert  parameters, or a Grid.
- 3 parameter transformation, which is a translation in 3D space.
- 7 parameter transformation, which is a [Helmert transformation](https://en.wikipedia.org/wiki/Helmert_transformation)
- A grid shift

A _Grid Shift_ defines the transformation based on a grid defined in a grid file found in the resources. Grids are meant to model the irregularities of the surface of the earth.

A _Grid_ is a set of conversion tables. A _conversion table_ defines some subset of the surface that it is valid over, and gives a set of transformed points between the Ellipsoid to WGS84, so that interpolation can be used to convert arbitrary points between the Ellipsoid and WGS84 over that area.

A _Projection_ defines a family of formulas for mapping between polar coordinates and some 2D plane. 
A projection may not be valid for the entire space of polar coordinates.

A _Coordinate Reference System_ combines a projection and a datum to translate points between some 2D plane and WGS84.

proj4/src/main/java/org/osgeo/proj4j/BasicCoordinateTransform.java
```diff
@@ -156,11 +160,20 @@ private void datumTransform(ProjCoordinate pt)
...
-        if (srcCRS.getDatum().isEqual(tgtCRS.getDatum()))
+        if (srcCRS.getDatum().isEqual(tgtCRS.getDatum())
+                || srcCRS.getDatum().getTransformType() == Datum.TYPE_UNKNOWN
+                || tgtCRS.getDatum().getTransformType() == Datum.TYPE_UNKNOWN)
             return;
     
-        // TODO: grid shift if required
+        if (srcCRS.getDatum().getTransformType() == Datum.TYPE_GRIDSHIFT) {
+            srcCRS.getDatum().shift(pt);
+            srcGeoConv.overrideWithWGS84Params();
```
lossyrob: Override WGS84 params in constructor of basic coordinate transform, instead of doing it per point transformed.

proj4/src/main/java/org/osgeo/proj4j/NTV1.java
```diff
@@ -0,0 +1,28 @@
...
+    private static boolean containsAt(byte[] needle, byte[] haystack, int offset) {
+        if (needle == null || haystack == null) return false;
+
+        int maxoffset = Math.min(needle.length - 1, haystack.length - offset - 1);
+        for (int i = 0; i < maxoffset; i++) {
+            if (needle[i] != haystack[offset + i]) return false;
+        }
+        return true;
+    }
+}
```
lossyrob: This is incomplete work, would be good to back out of this PR and store in a branch for post-0.10 work.

proj4/src/main/java/org/osgeo/proj4j/PolarCoordinate.java
```diff
@@ -0,0 +1,20 @@
+package org.osgeo.proj4j.util;
+
+import java.io.Serializable;
+
+public final class PolarCoordinate implements Serializable {
```
lossyrob: This could use Complex

proj4/src/main/java/org/osgeo/proj4j/datum/Axes.java
```diff
@@ -0,0 +1,113 @@
+package org.osgeo.proj4j.datum;
+
+import org.osgeo.proj4j.ProjCoordinate;
+import java.io.Serializable;
+
+public final class Axes implements Serializable {
```
lossyrob: rename to AxisOrder

proj4/src/main/java/org/osgeo/proj4j/datum/Datum.java
```diff
@@ -193,14 +193,9 @@ public boolean isEqual(Datum datum)
...
       }
       return true;
-    } 
-    /* 
-     //TODO: complete
-    else if( this.datum_type == Proj4js.common.PJD_GRIDSHIFT ) {
-      return strcmp( pj_param(this.params,"snadgrids").s,
-                     pj_param(dest.params,"snadgrids").s ) == 0;
+    } else if (getTransformType() == TYPE_GRIDSHIFT) {
+        return grids.equals(datum.grids);
```
lossyrob: Should make sure that Grid and ConversionTable have structural comparison methods.

proj4/src/main/java/org/osgeo/proj4j/parser/Proj4Keyword.java
```diff
@@ -49,6 +51,7 @@
   public static final String nadgrids = "nadgrids";
   public static final String no_defs = "no_defs";
   public static final String wktext = "wktext";
+  public static final String no_uoff = "no_uoff";
```
lossyrob: Add TODO to implement

proj4/src/main/java/org/osgeo/proj4j/parser/Proj4Keyword.java
```diff
@@ -87,12 +90,15 @@ public static synchronized Set supportedParameters()
...
       
       supportedParams.add(title);       // no-op
       supportedParams.add(no_defs);     // no-op
       supportedParams.add(wktext);      // no-op
-      supportedParams.add(nadgrids);    // no-op for now
-
+      supportedParams.add(pm);          // no-op
+      supportedParams.add(axis);        // no-op
+      supportedParams.add(gamma);       // no-op
+      supportedParams.add(no_uoff);     // no-op
```
lossyrob: Remove "no-op" for supported features.

proj4/src/main/java/org/osgeo/proj4j/PolarCoordinate.java
dwins: After further review we've decided to leave this separate, as Complex and PolarCoordinate have slightly different semantics.

dwins:
@lossyrob I've made the corrections we identified during the review earlier and also added some info and examples to ``docs/proj4`` .  I can spend some time improving the Javadoc comments as well but I don't think that we need to defer the merge for that.

PR #1307, opened 2016-02-06
Upgrade OpenCSV to 3.4

adamkozuch:


adamkozuch:
Not sure if we need dependency in https://github.com/geotrellis/geotrellis/blob/master/raster/build.sbt. I don't see any use of it in that subdirectory.

lossyrob:
Thanks @adamkozuch!

One thing we would need for you to do is sign an eclipse CLA. We're moving to [LocationTech](https://www.locationtech.org/), which is under the eclipse foundation, and are starting to require that every contributor electronically sign a CLA.

If you visit http://www.eclipse.org/legal/CLA.php, you'll see the link on the right to log in and fill out the form. You'll have to create an eclipse account with the same email as you do git commits with. Let me know if you need any help with this.

adamkozuch:
Done

lossyrob:
Great, +1

PR #1308, opened 2016-02-08
Register Classes with Kryo

### LINKED ISSUE #1145: Kryo registration for all our types
When using the library and having `kryo.setRegistrationRequired(true)` for a project that I am using a `MultiPolygonFeature` in, this is what I had to register:

```scala
    kryo.register(classOf[MultiPolygon])
    kryo.register(classOf[Point])
    kryo.register(classOf[com.vividsolutions.jts.geom.Point])
    kryo.register(classOf[com.vividsolutions.jts.geom.MultiPolygon])
    kryo.register(classOf[com.vividsolutions.jts.geom.GeometryFactory])
    kryo.register(classOf[com.vividsolutions.jts.geom.impl.CoordinateArraySequenceFactory])
    kryo.register(classOf[com.vividsolutions.jts.geom.PrecisionModel])
    kryo.register(classOf[com.vividsolutions.jts.geom.PrecisionModel.Type])
    kryo.register(classOf[com.vividsolutions.jts.geom.Polygon])
    kryo.register(classOf[Array[com.vividsolutions.jts.geom.Polygon]])
    kryo.register(classOf[com.vividsolutions.jts.geom.LinearRing])
    kryo.register(classOf[Array[com.vividsolutions.jts.geom.LinearRing]])
    kryo.register(classOf[com.vividsolutions.jts.geom.Envelope])
    kryo.register(classOf[com.vividsolutions.jts.geom.impl.CoordinateArraySequence])
    kryo.register(classOf[com.vividsolutions.jts.geom.Coordinate])
    kryo.register(classOf[Array[com.vividsolutions.jts.geom.Coordinate]])
    kryo.register(classOf[Feature[_, _]])
```

That's a whole lot. We should provide some functionality to register all the types we'd need to. Not registering classes causes a huge waste of space.

A note on how I encountered this: I was using an `rdd.persist(MEMORY_ONLY)` and was getting 4 partitions that wouldn't fit into memory. I changed to `rdd.persist(MEMORY_ONLY_SER)`, and then __8__ partitions wouldn't fit. Curious, as the serialized versions should take up less space. I changed to Java Serialization and...it all fit into memory. No bueño. So I set the registration required, and then went through the process that produced the above list. Once I got the registration working (which also required doing `kryo.register(classOf[Array[scala.Tuple2[_, _]]])`, it ran and fit all in memory, but wasn't any faster than using the java serialization.

jamesmcclain:
A great number of the Geotrellis objects have been registered with Kryo.  If the environment variable `GEOTRELLIS_KRYO_REGREQ` is set when tests are run, a runtime error will be raised if an attempt is made to serialize a class with Kryo without registering it first.  The set of registered objects is currently sufficient to allow all unit tests to pass.

Closes #1145


spark/src/main/scala/geotrellis/spark/io/hadoop/KryoRegistrator.scala
```diff
@@ -70,6 +75,207 @@ class KryoRegistrator extends SparkKryoRegistrator {
...
+      kryo.register(Schema.createUnion(schemas).getClass)             // Union
+
+      kryo.register(java.lang.Class.forName("org.apache.avro.Schema$LockableArrayList"))
+      kryo.register(java.lang.Class.forName("org.apache.avro.Schema$Name"))
+    }
+
+    val reqRegistration = Properties.envOrElse("GEOTRELLIS_KRYO_REGREQ", null)
+
+    if (reqRegistration != null) {
+      List(
```
lossyrob: This should be in the test project, right?

jamesmcclain: Ideally.  The issue was that it wasn't clear how to get access to the `kryo` variable from there.  I can investigate this further.

spark/src/main/scala/geotrellis/spark/io/hadoop/KryoRegistrator.scala
```diff
@@ -70,6 +74,165 @@ class KryoRegistrator extends SparkKryoRegistrator {
...
+    List(
+      "geotrellis.proj4.LatLng$",
+      "geotrellis.spark.EmptyBounds$",
+      "scala.collection.immutable.Nil$",
+      "scala.math.Ordering$Double$",
+      "scala.math.Ordering$Float$",
+      "scala.math.Ordering$Int$",
+      "scala.math.Ordering$Long$",
+      "scala.None$"
+    ).foreach({ c => kryo.register(java.lang.Class.forName(c)) })
```
lossyrob: Let's try to find a way to get rid of the strings.

echeipesh: `geotrellis.proj4.LatLng.getClass`

jamesmcclain: Ah, okay.

echeipesh:
Backlog task: "Move KryoRegistrator out of hadoop, put in geotrellis.spark.io.kryo"

jamesmcclain:
All comments prior to this one have been addressed, with the exception of two class names in strings here: https://github.com/jamesmcclain/geotrellis/blob/a74fecac8456cf00922b6c72184b0f2ec00dc2f8/spark/src/main/scala/geotrellis/spark/io/hadoop/KryoRegistrator.scala#L232-L233 .

pomadchin:
It is possible to use `getDeclaredClasses` method (but not sure it is easier / simplier):

```scala
val dc = classOf[org.apache.avro.Schema].getDeclaredClasses()
val lal: Class[_] = dc(1) //> class org.apache.avro.Schema$LockableArrayList
val lal: Class[_] = dc(2) //> class org.apache.avro.Schema$Names
```

jamesmcclain:
> It is possible to use getDeclaredClasses method (but not sure it is easier / simplier):

Thanks for the information, I will do that!

echeipesh:
+1

PR #1298, opened 2016-01-26
KeyComponents api renames

pomadchin:

PR #1309, opened 2016-02-09
Spark-ETL Multiband ingest

pomadchin:
Depends on #1298


echeipesh:
README needs to be updated:
- with code samples
- removal of accumulo catalog input


Also I was thinking it would be good to have a `geotrellis.spark-etl.SinglebandEtl` and `geotrellis.spark-etl.MultibandEtl` App objects that go through load,reproject,tile,save stages without modifications. This would allow uses to build `spark-etl` assembly and use it for vanilla jobs right from checking out geotrellis repo. It would also provide some code documentation, compiled use case and a clear sample that can be copied/pasted into their own, more complicated jobs.

pomadchin:
@echeipesh suggested to try an api version with type params moved to ETL functions declarations

pomadchin:
Current API usage example:

```scala
object Ingest extends App {
  val etl = Etl[ProjectedExtent, SpatialKey, Tile](args)
  implicit val sc = SparkUtils.createSparkContext("GeoTrellis ETL", new SparkConf(true))
  
  val (zoom, tiledRdd) = etl.tile(etl.reproject(etl.load()
  val result = tiledRdd.localAdd(1)))
  etl.save(LayerId(etl.conf.layerName(), zoom), tiledRdd, ZCurveKeyIndexMethod)
  sc.stop()
}
```

With type params moved to functions:

```scala
object Ingest extends App {
  val etl = Etl(args)
  implicit val sc = SparkUtils.createSparkContext("GeoTrellis ETL", new SparkConf(true))

  val (zoom, tiledRdd) = etl.tile[ProjectedExtent, SpatialKey, Tile](etl.reproject(etl.load[ProjectedExtent, Tile]()))
  val result = tiledRdd.localAdd(1)
  etl.save(LayerId(etl.conf.layerName(), zoom), tiledRdd, ZCurveKeyIndexMethod)
  sc.stop()
}
```

echeipesh:
How about this:

```scala
object Ingest extends App {
  val etl = Etl(args)
  implicit val sc = SparkUtils.createSparkContext("GeoTrellis ETL", new SparkConf(true))

  val tiles = etl.load[ProjectedExtent, Tile]
  val reprojected = etl.reproject(tiles)
  val (zoom, metadata) = etl.collectMetadata(reprojected)
  val tiled = tiles.cutToLayout(metadata)
  etl.save(LayerId(etl.conf.layerName(), zoom), tiled, ZCurveKeyIndexMethod)

  sc.stop()
}
```

Here we get rid of `etl.tile` and replace it with `etl.collectMetadata` which is what is required to call `cutToLayout`. That used the native geotrellis tile function that is always valid. Also this allows the `save` to have different type parameters than `load`. This is useful when user maps from `SpatialKey` to `SpaceTimeKey` or from `MultiBandTile` to `Tile`.

`collectMetadata` will sometimes trigger collect or just construct the metadata from command line params, but that seems like an ok ambiguity.

spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala
```diff
@@ -43,35 +71,59 @@ class Etl[K: TypeTag: SpatialComponent](args: Seq[String], modules: Seq[TypedMod
...
+    }
   }
 
-  def save(id: LayerId, rdd: RasterRDD[K], method: KeyIndexMethod[K]): Unit = {
-    val attributes = outputPlugin.attributes(conf.outputProps)
-    def savePyramid(zoom: Int, rdd: RasterRDD[K]): Unit = {
-      val currentId = id.copy( zoom = zoom)
+  def save[
+    K: SpatialComponent: TypeTag,
+    V <: CellGrid: TypeTag: ? => TileMergeMethods[V]: ? => TilePrototypeMethods[V]
```
echeipesh: `save` does't require `? => TileMergeMethods[V]` and other similar types.

pomadchin: But `Pyramid.up` requires: https://github.com/geotrellis/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/ingest/Pyramid.scala#L18

echeipesh: Ah, good point.

echeipesh:
+1

PR #1312, opened 2016-02-10
Feature/render api refactor

moradology:
This PR is still being heavily edited; will rebase for coherent commits when the shape of things is more solid.


lossyrob:
Eugene code reviewed and +1'd so I'm merging.

PR #1317, opened 2016-02-12
Remove GDAL Dependency from geotrellis-spark

jamesmcclain:
The dependency on `geotrellis-gdal` has been removed from `geotrellis-spark`.  In order to preserve all functionality, dependencies on `spark` and `etl` have been added to `gdl`.

### GDAL with Java bindings on Ubuntu ###
One can get the GDAL functionality working on Ubuntu 14.04 (and presumably others) by building GDAL and the Java bindings from source.  Download [this tarball](http://download.osgeo.org/gdal/2.0.2/gdal-2.0.2.tar.xz), untar it, and type:
```shell
./configure --prefix=$HOME/local/gdal-2.0.2 --with-java && make
cd swig/java
make
```

(Note that `make install` was never typed above because it does not seem to be strictly necessary.)

Then building Geotrellis and run unit tests by typing:
```shell
LD_LIBRARY_PATH=$HOME/local/src/gdal-2.0.2/swig/java:$LD_LIBRARY_PATH ./buildall.sh
```

### Still Need To ###
   - [x] Move commented-out NetCdf-dependent tests from `IngestSpec` to somewhere in the `gdal` project.


jamesmcclain:
Changing this to a "WIP" PR because I forgot to move some NetCdf-dependant tests from the `spark` project to the `gdal` project.

PR #1318, opened 2016-02-13
Multiband combine functions macro generation

pomadchin:
## A Short description

In order to avoid code copy-pasting of the code that is in fact differs only by *arity* (arity in each case can mean a bit different thing: number of trait `val's`, arity of constructor type for a `class`, any `function` arity).

To solve that problem (when it is really necessary to have lots of similar code: [example](https://github.com/geotrellis/geotrellis/blob/master/macros/src/main/scala/geotrellis/macros/MultiBandTileMacros.scala)) and not to copy-paste this code manually (increasing arity) we can use `sbt` code generation. This approach is used in [shapeless](https://github.com/milessabin/shapeless) and [cats](https://github.com/typelevel/cats) to generate functions and classes of different arity: https://github.com/milessabin/shapeless/blob/master/project/Boilerplate.scala

*NOTE 1: Current combine functions arity is limited by 10 by design ([exmple link](https://github.com/geotrellis/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultiBandTile.scala#L809-L810)). We are limited by `Function` arity as for `Combiner` of arity 22 we need functions with arity 43 (`combinerArity * 2 + 1`).*

```scala
// cause of 10 arity limit 
def setDouble(targetIndex: Int, 
                        s1: GeoTiffSegment, i1: Int, 
                        s2: GeoTiffSegment, i2: Int, 
                        s3: GeoTiffSegment, i3: Int, 
                        s4: GeoTiffSegment, i4: Int)(combiner: DoubleTileCombiner4): Unit
```

*NOTE 2: To avoid this limit I suggest to refactor these functions:*

```scala
// new curried variant which removes 10 arity limit
def setDouble(s1: GeoTiffSegment,
                       s2: GeoTiffSegmen,
                       s3: GeoTiffSegment,
                       s4: GeoTiffSegment)
                       (i1: Int, i2: Int, i3: Int, i4: Int)
                       (targetIndex: Int)(combiner: DoubleTileCombiner4): Unit 
```

*NOTE 3: Branched the 22 arity solution into a separate branch, though agree with @lossyrob that there is no much sense in it, basically because of beautiful https://github.com/geotrellis/geotrellis/pull/1326*

- [x] move combine functions to boilerplate generator
- [x] increase arity (up to 10)
- [x] code clenaup
- [x] test coverage
- [x] ~~increase arity up to 22~~
- [x] ~~benchamrking~~


pomadchin:
https://github.com/pomadchin/geotrellis/commit/455591209c71c987d34cf01976ea1376753d4e5a

project/Boilerplate.scala
```diff
@@ -0,0 +1,410 @@
...
+
+      block"""
+         |package geotrellis.raster
+         |import geotrellis.macros._
+         |import spire.syntax.cfor._              
+         |trait MacroMultibandCombiners { self: MultiBandTile =>      
+        -  def combineIntTileCombiner(combiner: IntTileCombiner${arity}): Tile = {
+        -    ${bandVals}
+        -    val result = ArrayTile.empty(cellType, cols, rows)
+        -    val arr = Array.ofDim[Int](bandCount)
```
pomadchin: I am curious about this line, think we don't need it, `arr` has no usages in the function body.

pomadchin:
With arity 22 (a bit refactored function: curried and moved some params a bit): https://github.com/pomadchin/geotrellis/commit/c6f5b3381fca612d84f6a3a2486dd1f3b9f18d1e

lossyrob:
The body of the PR says that you can only have 10, but the Specs have 22. A bit confused.

pomadchin:
@lossyrob if you mean the last pr, it is from another branch with refactored problem function

pomadchin:
Will make benchmarking to see what's with curried functions speed. Anyway we can revert to https://github.com/pomadchin/geotrellis/commit/cef16b9b7897092a0448e7a82b1a44f0cc82e281 anytime

pomadchin:
Decided to revert it now, at to merge this without refactor. Will do benchmarking a bit later, but as already said #1326 solves it better. /cc @lossyrob

PR #1320, opened 2016-02-15
Create geotrellis-util Subproject

jamesmcclain:
A new subproject called `geotrellis-util` has been created.  It contains the files formerly-known as:
   * `raster/MethodExtensions.scala`
   * `raster/io/Filesystem.scala`

Another possible candidate is:
   * ~~`raster/PixelSampleType.scala` (doubtful)~~

The methodology for identify candidates was to run the command `grep -L 'import geotrellis' --include='*.scala' -r .` to identify files that do not make use of any imported objects/classes/methods, then to manually look through that list.  That pattern is exhibited by both `MethodExtensions.scala` and `Filesystem.scala` and seems like a reasonable criterion for identifying other candidates because it identifies files that are possibly project-agnostic.

### Still Need To ###
   - [x] Get yea or nay on `raster/PixelSampleType.scala`
   - [x] Possibly identify other candidates

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -117,7 +117,7 @@ lazy val admin = Project("admin", file("admin")).
   settings(commonSettings: _*)
 
 lazy val spark = Project("spark", file("spark")).
-  dependsOn(raster).
+  dependsOn(util, raster).
   settings(commonSettings: _*)
 
 lazy val sparkTestkit: Project = Project("spark-testkit", file("spark-testkit")).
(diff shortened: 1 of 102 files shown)


lossyrob:
PixelSampleType is raster specific, so I think it should stay put.

jamesmcclain:
Some deleted files got recreated when I rebased.  Will correct on the morn.

PR #1321, opened 2016-02-16
Make Spurious Failure Less Likely

jamesmcclain:
This *should* make a spurious failure in the `StreamingHistogramSpec` tests very unlikely.

### DIFF
--- a/raster-test/src/test/scala/geotrellis/raster/histogram/StreamingHistogramSpec.scala
+++ b/raster-test/src/test/scala/geotrellis/raster/histogram/StreamingHistogramSpec.scala
@@ -84,7 +84,7 @@ class StreamingHistogramSpec extends FunSpec with Matchers {
 
       Iterator.continually(list1)
         .flatten.take(list1.length * 10000)
-        .foreach({ i => h.countItem(i + r.nextDouble / 1000.0) })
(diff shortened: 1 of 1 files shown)


lossyrob:
...what are these magic numbers? can you give some info on why there are spurious failures, and what this does to resolve them? 

I'm very in favor of deterministic tests that don't have a chance to spuriousness

jamesmcclain:
> ...what are these magic numbers? can you give some info on why there are spurious failures, and 
> what this does to resolve them?
>
> I'm very in favor of deterministic tests that don't have a chance to spuriousness

Okay, I will add the following information as comments tomorrow.  For both of these tests, the random numbers are just small perturbations design to ensure a stream of well-structured unique stream of numbers are added to the Histogram.

For the median, the (x + 3) / 60000 is shifting the random number up (so that it is roughly between 0 and 6 instead of -3 and 3) then scaling it down by a factor of 60000 so that it is between 0 and 1e-4.  This is small enough to keep it form interfering with the median, while still making sure that the numbers are different.

For the mean case, the random number is being scaled down by a factor of 10000.  There are 10000 numbers being chosen, so the mean of difference of the perturbed list of numbers versus the unperturbed list of numbers should be zero and the standard deviation 1e-6.  The test passes as long as the difference is less than 1e-4, or 100 standard deviations.  The test should never fail unless the histogram code is really broken.

PR #1314, opened 2016-02-11
Remove engine dependency from raster-test and testkit projects

pomadchin:
Based on a #1266

PR #1324, opened 2016-02-17
Publishment scripts fix

pomadchin:
Due to #1314 we got rid of a `testkit`, this pr fixes local and travis publishment.


echeipesh:
+1, thanks for a quick fix

PR #1328, opened 2016-02-18
Etl and Ingest object buffered tiles reprojection

pomadchin:


spark-etl/src/main/scala/geotrellis/spark/etl/EtlConf.scala
```diff
@@ -48,6 +48,8 @@ class EtlConf(args: Seq[String]) extends ScallopConf(args){
   val tileSize     = opt[Int]("tileSize",
                       descr = "Tile size in pixels (default: 256)",
                       default = Some(256))
+  val bufferSize   = opt[Int]("bufferSize",
```
echeipesh: Is there a reason for user to be able to control this?

pomadchin: To be honest, not sure, think just preventing of its calculation depending on tiles metadata. Probably for `Etl` we can remove it, but left for the `Ingest` object.

raster/src/main/scala/geotrellis/raster/Tile.scala
```diff
@@ -292,4 +292,10 @@ trait Tile extends CellGrid with IterableTile with MappableTile[Tile] {
     }
     s
   }
+
+  def isNoDataTile = {
+    var empty = true
+    foreach { z => if (isData(z)) empty = false }
```
echeipesh: There should be a way to break out of this early if we're going to provide this method.

pomadchin: Have you an idea how to call it? I like the `isEmpty`, but not sure that it is a correct function name and can be confusing.

echeipesh: +@lossyrob I think the name is appropriate. I remember wanting a function like this but stopped just short of committing it.

echeipesh:
+1

raster/src/main/scala/geotrellis/raster/Tile.scala
lossyrob: The much faster way to do this is to have a return inside of a while loop, or some other way to break immediately upon finding a non-nodata value

lossyrob:
I believe this was still in progress as we were figuring out where the difference in NODATA tiles was coming from with modifying buffered tiles. @pomadchin if there's updates with that, those can be in a separate PR.

raster/src/main/scala/geotrellis/raster/Tile.scala
pomadchin: @lossyrob ok will make a probably small pr, yes throwing an exception can be significant :/

pomadchin:
@lossyrob:+1: mb it is even better to perform it as a separate pr (if there would be a pr at all)

PR #1329, opened 2016-02-18
Put s3 and accumulo Backends Into Their Own Projects

jamesmcclain:
### Outstanding ###
   - [ ] Depends on https://github.com/geotrellis/geotrellis/pull/1320


echeipesh:
As far as slippy/s3 question. What I hope can happen is that only the s3 relevant code for slippy reader gets moved into the s3 package. The base and file slippy classes don't appear to have any s3 dependency.

If we keep all the package structure the same including the `geotrellis-s3` package should just augment packages with s3 enabled implementations.

backends/s3/src/main/scala/geotrellis/s3/S3Client.scala
```diff
@@ -14,10 +14,10 @@ import com.amazonaws.ClientConfiguration
 import org.apache.commons.io.IOUtils
 import scala.collection.JavaConversions._
 
-trait S3Client extends LazyLogging {
+trait S3Client extends LazyLogging with Serializable {
```
echeipesh: This seems like a bug. S3Client wraps an object that has references to sockets which are no transferable between machines, even if serializable. This is the reason that s3 reader/writer code will ask for `() => S3Client` instead of `S3Client` directly. The former can be serialized and used to construct a new client on the workers and the latter will cause errors during distributed execution.

Whatever code prompted this change probably has a bug that will reveal itself on a cluster but not on a local test.

jamesmcclain:
> As far as slippy/s3 question. What I hope can happen is that only the s3 relevant code for slippy reader gets moved into the s3 package. The base and file slippy classes don't appear to have any s3 dependency.

> If we keep all the package structure the same including the geotrellis-s3 package should just augment packages with s3 enabled implementations.

Okay, I interpenetrates that to mean that the `geotrellis-slippy` project that I created should go, and it should be split between `geotrellis-s3` and `geotrellis-spark`.

spark-test/src/test/resources/test-catalog/attributes/all-ones___8___metadata.json
```diff
@@ -0,0 +1 @@
+[{"name":"all-ones","zoom":8},{"keyBounds":{"minKey":{"col":1,"row":1},"maxKey":{"col":6,"row":7}},"schema":null,"metadata":{"cellType":"float32","extent":{"xmin":-135.00000125,"ymin":-89.99999,"xmax":134.99999125,"ymax":67.49999249999999},"layoutDefinition":{"extent":{"xmin":-180.0,"ymin":-89.99999,"xmax":179.99999,"ymax":89.99999},"tileLayout":{"layoutCols":8,"layoutRows":8,"tileCols":3,"tileRows":4}},"crs":"+proj=longlat +datum=WGS84 +no_defs "},"header":{"keyClass":"geotrellis.spark.SpatialKey","valueClass":"geotrellis.raster.Tile","path":"file:/home/likewise-open/AZVA-INT/jmcclain/local/src/geotrellis/spark/src/test/resources/test-catalog/all-ones/8"},"keyIndex":{"obj":"rO0ABXNyADpnZW90cmVsbGlzLnNwYXJrLmlvLmluZGV4LnJvd21ham9yLlJvd01ham9yU3BhdGlhbEtleUluZGV4DTH0g5hxQLACAANJAApsYXlvdXRDb2xzSQAGbWluQ29sSQAGbWluUm93eHAAAAAGAAAAAQAAAAE="}}]
```
echeipesh: some gitignore got messed up. We shouldn't be committing test catalog. In fact we shouldn't even need to be generating this file. I think `catalog-test` will be written from the actual hadoop catalog spec where as this file used to be the source of layers for the tests. Layers for the test are now generated and kept in memory.

backends/s3/src/main/scala/geotrellis/s3/S3Client.scala
jamesmcclain: Ah, okay.  That was to enable to `saveToS3` method to work in the context of a unit test.  I wanted to be able to switch between a mock S3 and a real one.

jamesmcclain: The code in question is here: https://github.com/jamesmcclain/geotrellis/blob/feature/jwm/backends/spark-test/src/test/scala/geotrellis/spark/render/RenderedImageSpec.scala#L23

spark-test/src/test/resources/test-catalog/attributes/all-ones___8___metadata.json
jamesmcclain: Hmm ... it looks like those were actually already committed, but I removed them.

backends/s3/src/main/scala/geotrellis/s3/S3Client.scala
echeipesh: I see. I would suggesting moving the actual logic to companion object and making parameter of type `() => S3Client`. The companion object can be tested. My feeling has been that `() => S3Client` is sufficiently weird signature that I want to minimize how often it shows up in the API. So I would even say that the extension method need not expose it.

jamesmcclain: I do not understand the point about moving the logic to a companion object.  Can you give me some idea of what that might look like?

jamesmcclain:

backends/s3/src/main/scala/geotrellis/s3/package.scala
```diff
@@ -1,11 +1,13 @@
...
 import org.apache.spark.rdd.RDD
 
+
 package object s3 {
   private[s3]
   def makePath(chunks: String*) =
     chunks.filter(_.nonEmpty).mkString("/")
 
-  implicit class S3RDDExtensions[K,V](rdd: RDD[(K,V)]) extends SaveToS3Methods[K, V](rdd)
+  implicit class S3RDDExtensions[K](rdd: RenderedImages[K]) extends SaveToS3Methods[K](rdd)
```
echeipesh: we're settling on `withSaveToS3Methods` type of naming pattern for these implicit classes.

jamesmcclain: Ah yes, that slipped my attention, thank you.

build.sbt
```diff
@@ -120,12 +120,24 @@ lazy val spark = Project("spark", file("spark")).
   dependsOn(util, raster).
   settings(commonSettings: _*)
 
+lazy val s3 = Project("s3", file("backends/s3")).
+  dependsOn(spark).
```
echeipesh: I wonder if dependency should be % "provided", like the test kits. Not sure on all the implications. I suppose most directly it would allow us to use different version of geotrellis-s3 library with geotrellis-spark without class path conflicts.

jamesmcclain: I don't understand all of the implications either, but I can certainly make that change.  Will do the same for the accumulo backend.

jamesmcclain:

cleanall.sh
```diff
@@ -43,4 +47,3 @@ rm -r tasks/target
 rm -r util/target
 rm -r raster-testkit/target
 rm -r vector-testkit/target
-rm -r spark-testkit/target
```
echeipesh: looks like erroneous change

spark/build.sbt
```diff
@@ -18,10 +15,10 @@ libraryDependencies ++= Seq(
   scalatest % "test")
```
echeipesh: Does spark project still require `awsSdkS3` dep? One of the motivations was to remove it to minimize version conflicts if our s3 module is not used.

s3/src/main/scala/geotrellis/s3/SaveToS3Methods.scala
```diff
@@ -0,0 +1,99 @@
...
+        requests map write
+      }
+
+      results.run.run
+      pool.shutdown()
+    }
+  }
+}
+
+class SaveToS3Methods[K](images: RenderedImages[K]) {
```
echeipesh: I think this is unnecessarily specific. We should be able to save anything thats `Array[Byte]`, not just images. Which I think makes `RenderedImages` unused.

spark/src/main/scala/geotrellis/spark/io/hadoop/SaveToHadoopMethods.scala
```diff
@@ -1,34 +1,74 @@
...
+      pathTemplate
+        .replace("{x}", key.col.toString)
+        .replace("{y}", key.row.toString)
+        .replace("{z}", id.zoom.toString)
+        .replace("{name}", id.name)
+    }
+  }
+}
+
+class SaveToHadoopMethods[K](images: RenderedImages[K]) {
```
echeipesh: Same comment here re: `RenderedImages`

echeipesh:
This is a little pedantic but: the s3 and accumulo sub-project folder structures do not match their package path, they're missing spark folder.

+1 after comments and merge with upstream.

cleanall.sh
jamesmcclain: Indeed.

spark/build.sbt
jamesmcclain: Ah, most probably lnot.

s3/src/main/scala/geotrellis/s3/SaveToS3Methods.scala
jamesmcclain: Okay, I will remove this case class.

jamesmcclain:
> This is a little pedantic but: the s3 and accumulo sub-project folder structures do not match their package path, they're missing spark folder.

Did you want an additional `spark/io` for each of those, or just `spark`?

jamesmcclain:

PR #1332, opened 2016-02-19
Renamed geotools subproject to shapefile

lossyrob:
Also removed unnecessary dependencies in that subproject.


echeipesh:
publish-m2, publish-local, publish-snapshot still have references to geotools

lossyrob:
Whoops I didn't push that commit

echeipesh:
+1, the rename looks good to me

PR #1333, opened 2016-02-22
Different fixes and review changes

pomadchin:
- [x] Fix publish scripts (+util subproject)
- [x] Fix Ingest object persist strategy re-assign in a pyramid function call
- [x] Fix isNoDataTile function


echeipesh:
+1, fixes look good

PR #1335, opened 2016-02-22
Add scaladoc generation on building of master branch

moradology:
This PR adds automated building of scaladocs. Two steps remain:

1. setting up CI_GH_TOKEN on travis for the azaveadev github account
2. cleaning up the output to exclude akka and some other third party libs that get pulled in

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,7 @@ _SUCCESS
 *.swo
 *.sublime-*
 .vagrant
+
+lib
+index.html
+index.js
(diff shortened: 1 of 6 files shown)


moradology:
Number 2 is proving very difficult to track down answers on and it looks like the [last time we had unified scaladocs being generated](https://github.com/geotrellis/scaladocs/commit/1578065518d818d66cede028694e0ec29d132cb3), akka was pulled in as well. If we're good with docs being generated as before, the only thing necessary now is setting up CI_GH_TOKEN as an environment variable on Travis. @lossryob you might need to be the one to do this.

lossyrob:
Superseded by #1357

PR #1338, opened 2016-02-23
WIP: Input src in etl and ingest objects

pomadchin:
Currently input crs forces to be equal dest crs. 

- [x] fix inputs crs in Etl and in Ingest objects
- [x] test it on tiles in different projections
  - [x] webmerc (chatta demo)
  - [x] sinus (modis, MOD13Q1: ingested `2015 001 18 03`)
  - [x] utm (lc8: ingested `LC82010352015188LGN00_B2.TIF`)
  - [ ] latlng

PR #1340, opened 2016-02-24
RasterMetaData contains KeyBounds

echeipesh:
Add collection of KeyBounds to collection of other MetaData.
This change implies that RasterMetaData must be parametric on K, because it contains KeyBounds[K].


echeipesh:
@pomadchin 
This clashes with https://github.com/geotrellis/geotrellis/pull/1338 somewhat.
Now that `RasterMetaData` is parametric on `K` it is even more integrated with the tiling process. It makes more sense to not break it out. Could you take a look over the ETL portion and make some comments on this approach since you've been thinking about it the most lately.

spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala
```diff
@@ -24,27 +24,25 @@ object Etl {
...
     implicit def classTagK = ClassTag(typeTag[K].mirror.runtimeClass(typeTag[K].tpe)).asInstanceOf[ClassTag[K]]
     implicit def classTagV = ClassTag(typeTag[V].mirror.runtimeClass(typeTag[V].tpe)).asInstanceOf[ClassTag[V]]
 
     val etl = Etl(args)
     val sourceTiles = etl.load[I, V]
-    val (_, metadata) = etl.collectMetadata(sourceTiles)
-    val tiled = sourceTiles.cutTiles[K](metadata, NearestNeighbor)
-    val (zoom, reprojected) = etl.reproject(ContextRDD(tiled, metadata))
-    etl.save[K, V](LayerId(etl.conf.layerName(), zoom), reprojected, keyIndexMethod)
+    val (tiledZoom, tiled) = etl.tile(sourceTiles)
```
pomadchin: Probably that's what ew really wanted, awesome that it derives type `K` here. Think that we really don't need to get *only* metadata with etl object, but probably that could be useful. 
Anyway your approach looks nicer think :+1:

spark/src/main/scala/geotrellis/spark/ingest/Ingest.scala
```diff
@@ -66,7 +66,7 @@ object Ingest {
     (sink: (RasterRDD[K], Int) => Unit): Unit =
   {
     val (_, rasterMetaData) =
-      RasterMetaData.fromRdd(sourceTiles, destCRS, layoutScheme)(_.projectedExtent.extent)
+      RasterMetaData.fromRdd(sourceTiles, destCRS, layoutScheme)
```
pomadchin: think we can not to merge #1338 as it breaks compilation anyway until further investigation, but there we have to use: `RasterMetaData.fromRdd(sourceTiles, layoutScheme)` and this breaks `LatLng` `Ingest` tests .

spark/src/test/scala/geotrellis/spark/io/LayerUpdateSpaceTimeTileTests.scala
```diff
@@ -7,17 +7,19 @@ import geotrellis.vector._
...
 
-  def dummyRasterMetaData: RasterMetaData =
+  def dummyRasterMetaData =
     RasterMetaData(
       IntConstantNoDataCellType,
       LayoutDefinition(RasterExtent(Extent(0,0,1,1), 1, 1), 1),
       Extent(0,0,1,1),
-      LatLng
+      LatLng,
+      KeyBounds(SpaceTimeKey(0,0, new DateTime()), SpaceTimeKey(1,1, new DateTime()))
```
pomadchin: Seem like keyBounds are not empty, that causes problems with test `should not update a layer (keys out of bounds)`

echeipesh: thanks, just fixed that

spark-etl/README.md
```diff
@@ -20,10 +20,9 @@ object GeoTrellisETL extends App {
   
   val etl = Etl(args)
   val tiles = etl.load[ProjectedExtent, Tile]
-  val reprojected = etl.reproject(tiles)
-  val (zoom, metadata) = etl.collectMetadata(reprojected)
-  val tiled = ContextRDD(tiles.cutTiles[SpatialKey](metadata, NearestNeighbor), metadata)
-  etl.save(LayerId(etl.conf.layerName(), zoom), tiled, ZCurveKeyIndexMethod)
+  val (zoom, tiled) = etl.tile(tiles)
+  val (reprojectZoom, reprojected) = etl.reproject(tiled)
```
echeipesh: These should be merged into one step

spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala
```diff
@@ -72,35 +75,75 @@ case class Etl(args: Seq[String], @transient modules: Seq[TypedModule] = Etl.def
...
+    I: ProjectedExtentComponent: ? => TilerKeyMethods[I, K],
+    V <: CellGrid: Stitcher: ClassTag: (? => TileMergeMethods[V]): (? => TilePrototypeMethods[V]):
+      (? => TileReprojectMethods[V]): (? => CropMethods[V]),
+    K: SpatialComponent: Boundable: ClassTag
+  ](
+    rdd: RDD[(I, V)], method: ResampleMethod = NearestNeighbor
+  )(implicit sc: SparkContext): (Int, RDD[(K, V)] with Metadata[RasterMetaData[K]]) = {
+    val targetCellType = conf.cellType.get
+
+    val (_, rmd: RasterMetaData[K]) = {
```
lossyrob: Should add an option so that we can reproject the source tiles via ProjectedComponentMethods reproject vs the buffered tile reprojection method. The latter requires that all the incoming CRS's are the same; we could handle the case where a buffered reproject is requested but not all the CRS's are the same inside the metadata collect, and report that error.

spark/src/main/scala/geotrellis/spark/ingest/package.scala
```diff
@@ -15,21 +15,25 @@ package object ingest {
...
-      key &|-> _projectedExtent.lens set (pe)
+      _projectedExtent.lens.set(pe)(key)
   }
 
   implicit object ProjectedExtentComponent extends IdentityComponent[ProjectedExtent]
 
-  implicit class withCollectMetadataMethods[K: ProjectedExtentComponent, V <: CellGrid](rdd: RDD[(K, V)]) extends Serializable {
-    def collectMetaData(crs: CRS, layoutScheme: LayoutScheme): (Int, RasterMetaData) = {
-      RasterMetaData.fromRdd(rdd, crs, layoutScheme)(_.projectedExtent.extent)
+  implicit class withCollectMetadataMethods[
```
lossyrob: Should we make a `metadata` package that would have these methods? I'm a bit uncomfortable with still putting stuff in `ingest`, which we have talked a bunch about getting rid of.

lossyrob:
TODO:
- Separate out the ETL code changes into it's own PR, make the option to reproject against source tiles
- Tag that version, talk on the mailing list about how this version is the last that won't break metadata badly
- When we release RC1, create a tool to convert metadata from that version to the RC1 version

lossyrob:
Changes merged into #1346

PR #1342, opened 2016-02-25
Support for temporal hadoop geotiffs

jasonbeverage:
Added new TemporalGeoTiffHadoopInput to support reading temporal geotiffs.


pomadchin:
:+1: we need to wait travis

PR #1345, opened 2016-02-29
Fix: remove headerUpdate function from LayerManagement stuff

pomadchin:


lossyrob:
I'm doing this as part of the IO refactor PR.

PR #1346, opened 2016-03-01
IO Refactor

### LINKED ISSUE #1291: Layer KeyBounds storage
Currently `KeyBounds` stored for a certain layer, PR https://github.com/geotrellis/geotrellis/pull/1273 allows to store `KeyIndex` as a json. For `HilbertCurve` and for `RowMajor` indexes we have to store constructor arguments (in particular, `KeyBounds` is a part of it for these indexes). For layers indexed with this curves `KeyBounds` twice (for layer, in a layer `Metadata`, and for `KeyIndex`). 
May be we can use `KeyIndex.keyBounds` to store all _preallocated (James PR) / possible_ `KeyBounds` defined for a layer; and in `Metadata` to store _current (already ingested)_ `KeyBounds` for a layer? The only problem is a `ZCurve`, for it we should not store `KeyBounds` (that's the reason I am not sure in this approach).

```scala
trait KeyIndex[K] extends Serializable {
  val id: String = this.getClass.getName
  val keyBounds: KeyBounds[K] // key space for index
  def toIndex(key: K): Long 
  def indexRanges(keyRange: (K, K)): Seq[(Long, Long)]
}
```

### LINKED ISSUE #1322: Accumulo layer updater won't handle avro schema updates correctly
We save the avro schema with the layer metadata. Say the following happens:

- Write a layer with a specific schema for the tiles.
- Change the schema. We can still read the layer because the old schema is stored with the layer, and avro encoding will allow us to use the codec with the old schema to read tiles out.
- Update the layer with new tiles. These tiles will use the new schema.
- When we go to read out the layer now, there are some tiles that were encoded with the old schema (which is read in with the layer metadata), and some that are written with the new schema. The current code will try to read in the ones written with the new schema as if they had the old schema, and this is sure to fail.

A potential fix is to have the updater check to see if the schema has changed, and if so, to require the entire layer be rewritten. Otherwise we'll have to do something like save the schema per tile.

lossyrob:
This PR gets the `K,V,M` type parameters out of Reader and Writers.

- Adds LayerUpdater for File and Hadoop backends
- LayerManger is a trait now.
- Hadoop uses Avro Codecs exclusively, gets rid of HadoopWritable
- Includes changes from #1340, so RasterMetaData has KeyBounds
- KeyIndex has KeyBounds
- Removed the separate "keybounds" attribute we were storing in layer metadata
- Allows writing to a key index whose keybounds might not match the layer (allows for updating of layers outside of the initial layer keybounds, you need to set the keybounds of the key index on the initial write to the full space)
- Clean up of layer management functions
- Serialize KeyIndex via JSON (and some weird java reflection stuff, but it's configurable if you want to add custom key indexes or custom keys)
- Added the `Compenent[T, C]` architecture, which generalizes the KeyComponent design we had before. This is used for metadata, where most functions require KeyBounds in metadata and so declare a context bound `M: Component[?, Bounds[K]]`

Fixes #1322 
Fixes #1291


lossyrob:
@echeipesh ready for review.

accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloKeyEncoder.scala
```diff
@@ -0,0 +1,46 @@
...
+object AccumuloKeyEncoder {
+  final def long2Bytes(x: Long): Array[Byte] =
+    Array[Byte](x>>56 toByte, x>>48 toByte, x>>40 toByte, x>>32 toByte, x>>24 toByte, x>>16 toByte, x>>8 toByte, x toByte)
+
+  final def index2RowId(index: Long): Text = new Text(long2Bytes(index))
+
+  def encode[K](id: LayerId, key: K, index: Long): Key =
+    new Key(index2RowId(index), columnFamily(id))
+
+  def setupQuery[K](id: LayerId, queryKeyBounds: Seq[KeyBounds[K]], keyIndex: KeyIndex[K])(getJob: () => Job): Seq[Job] = {
```
echeipesh: orphan method

accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerCopier.scala
```diff
@@ -1,55 +1,120 @@
...
-      layerWriter    = layerWriter
-    ) {
-      def headerUpdate(id: LayerId, header: AccumuloLayerHeader): AccumuloLayerHeader = {
-        val writerHeader = writerAttributeStore.readLayerAttribute[AccumuloLayerHeader](id, Fields.header)
-        header.copy(tileTable = writerHeader.tileTable)
-      }
+class AccumuloLayerCopier(
+  attributeStore: AttributeStore[JsonFormat],
+  layerReader: AccumuloLayerReader,
+  getLayerWriter: LayerId => AccumuloLayerWriter
```
echeipesh: What is the thinking behind asking for a reader as function of LayerId? It isn't used internally. Is it user API?

accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerCopier.scala
```diff
@@ -1,55 +1,120 @@
...
+    apply(
+      AccumuloAttributeStore(instance.connector),
+      layerReader,
+      _ => layerWriter
     )
 
-  def apply[K: Boundable: AvroRecordCodec: JsonFormat: ClassTag, V: AvroRecordCodec: ClassTag, M: JsonFormat](
+  def apply(
+    instance: AccumuloInstance,
+    table: String,
```
echeipesh: While it's the only logical choice, since we must know the table which we copy from I think it would help the API readability to rename all these `table` arguments to `targetTable`.

accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerReindexer.scala
```diff
@@ -1,41 +1,91 @@
...
 import scala.reflect.ClassTag
 
 object AccumuloLayerReindexer {
-  def apply[K: Boundable: AvroRecordCodec: JsonFormat: ClassTag, V: AvroRecordCodec: ClassTag, M: JsonFormat](
+  def apply(
+    instance: AccumuloInstance,
+    attributeStore: AttributeStore[JsonFormat],
+    options: AccumuloLayerWriter.Options
+  )(implicit sc: SparkContext): AccumuloLayerReindexer =
+    new AccumuloLayerReindexer(instance, AccumuloAttributeStore(instance), options)
```
echeipesh: Disregarding `attributeStore` argument here and creating a new one form the instance here.

accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloRDDWriter.scala
```diff
@@ -15,57 +15,41 @@ import org.apache.accumulo.core.data.{Key, Value}
...
-      oneToOne: Boolean = false): Unit = {
+object AccumuloRDDWriter {
+
+  def write[K: AvroRecordCodec, V: AvroRecordCodec](
+    raster: RDD[(K, V)],
+    instance: AccumuloInstance,
+    encodeKey: K => Key,
+    writeStrategy: AccumuloWriteStrategy,
+    table: String,
+    oneToOne: Boolean = false
```
echeipesh: Unused option, writer now defaults to doing a precaution groupBy

s3/src/main/scala/geotrellis/spark/io/s3/S3RDDWriter.scala
```diff
@@ -1,32 +1,36 @@
...
-class S3RDDWriter [K: AvroRecordCodec: ClassTag, V: AvroRecordCodec: ClassTag]() {
 
-  def getS3Client: () => S3Client = () => S3Client.default
-  val codec  = KeyValueRecordCodec[K, V]
-  val schema = codec.schema
+trait S3RDDWriter {
+
+  def getS3Client: () => S3Client
+
+  def write[K: AvroRecordCodec: ClassTag, V: AvroRecordCodec: ClassTag](rdd: RDD[(K, V)], bucket: String, keyPath: K => String, oneToOne: Boolean): Unit = {
```
echeipesh: If we're going to remove `oneToOne` from accumulo RDD writer, should we do the same here?

spark/src/main/scala/geotrellis/spark/KeyBounds.scala
```diff
@@ -23,10 +25,40 @@ sealed trait Bounds[+A] extends Product with Serializable {
 
   def getOrElse[B >: A](default: => KeyBounds[B]): KeyBounds[B] =
     if (isEmpty) default else this.get
+
+  @inline final def map[B >: A](f: (A, A) => (B, B)): Bounds[B] =
```
echeipesh: should map argument be `f: (A, A) => (B, B)` or `f: KeyBounds[A] => KeyBounds[B]` here?

spark/src/main/scala/geotrellis/spark/ingest/package.scala
```diff
@@ -10,26 +10,15 @@ import monocle.syntax._
...
-    def updateProjectedExtent(pe: ProjectedExtent): T =
-      key &|-> _projectedExtent.lens set (pe)
-  }
-
-  implicit object ProjectedExtentComponent extends IdentityComponent[ProjectedExtent]
-
-  implicit class withCollectMetadataMethods[K: ProjectedExtentComponent, V <: CellGrid](rdd: RDD[(K, V)]) extends Serializable {
-    def collectMetaData(crs: CRS, layoutScheme: LayoutScheme): (Int, RasterMetaData) = {
-      RasterMetaData.fromRdd(rdd, crs, layoutScheme)(_.projectedExtent.extent)
+  implicit class withCollectMetadataMethods[K1, V <: CellGrid](rdd: RDD[(K1, V)]) extends Serializable {
```
echeipesh: I think we can put this in `geotrellis.spark` package so `ingest` import is not required.

spark/src/main/scala/geotrellis/spark/io/LayerUpdater.scala
```diff
@@ -2,18 +2,57 @@ package geotrellis.spark.io
...
+    V: AvroRecordCodec: ClassTag,
+    M: JsonFormat: Component[?, Bounds[K]]
+  ](id: ID, rdd: RDD[(K, V)] with Metadata[M], keyBounds: KeyBounds[K]): Unit
 
-  def mergeUpdate(id: ID, reader: FilteringLayerReader[ID, K, M, Container], rdd: Container)
-                 (merge: (Container, Container) => Container) = {
-    val bounds = implicitly[Boundable[K]].collectBounds(rdd)
-    val existing = reader.query(id).where(Intersects(bounds)).toRDD
-    update(id, merge(existing, rdd))
+  protected def schemaHasChanged[K: AvroRecordCodec, V: AvroRecordCodec](writerSchema: Schema): Boolean = {
```
echeipesh: This is used in all the updaters in exactly this way, why not move the whole thing to the avro package as function and get rid of the `fingerprintMatches` as an implicit method

spark/src/main/scala/geotrellis/spark/io/json/KeyIndexFormats.scala
```diff
@@ -0,0 +1,268 @@
...
+  val indexClassTag = implicitly[ClassTag[T]]
+}
+
+trait KeyIndexRegistrator {
+  def register(keyIndexRegistry: KeyIndexRegistry)
+}
+
+class KeyIndexJsonFormat[K](entries: Seq[KeyIndexFormatEntry[K, _]]) extends RootJsonFormat[KeyIndex[K]] {
+  def write(obj: KeyIndex[K]): JsValue =
+    entries.find(_.indexClassTag.runtimeClass.isAssignableFrom(obj.getClass)) match {
```
echeipesh: Maybe too paranoid, but there is a possible failure case where more than one type is assignable from the the value we got. This should either be called out in the comment as a feature or we should do a `.filter` and throw on multiple matches.

spark/src/main/scala/geotrellis/spark/io/json/KeyIndexFormats.scala
```diff
@@ -0,0 +1,268 @@
...
+    entryRegistry register KeyIndexFormatEntry[SpatialKey, ZSpatialKeyIndex](ZSpatialKeyIndexFormat.TYPE_NAME)
+    entryRegistry register KeyIndexFormatEntry[SpatialKey, RowMajorSpatialKeyIndex](RowMajorSpatialKeyIndexFormat.TYPE_NAME)
+
+    entryRegistry register KeyIndexFormatEntry[SpaceTimeKey, HilbertSpaceTimeKeyIndex](HilbertSpaceTimeKeyIndexFormat.TYPE_NAME)
+    entryRegistry register KeyIndexFormatEntry[SpaceTimeKey, ZSpaceTimeKeyIndex](ZSpaceTimeKeyIndexFormat.TYPE_NAME)
+
+    // User defined here
+    val conf = ConfigFactory.load()
+    if(conf.hasPath(REG_SETTING_NAME)) {
+      val userRegistratorClassName = conf.getString("geotrellis.spark.io.index.registrator")
```
echeipesh: should use `REG_SETTING_NAME` as well

spark/src/main/scala/geotrellis/spark/io/LayerUpdater.scala
lossyrob: `fingerprintMatches` was based on [this documentation](https://avro.apache.org/docs/1.7.6/spec.html#Schema+Fingerprints), and I think is more generally useful than this case (although it is currently only used in this case). `schemaHasChanged` deals explicitly with the way writers pull together the K and V codecs to get the schema, so I think this separation is warranted.

pomadchin:
```
git mv spark-testkit/src/main/scala/geotrellis/spark/testkit/TileRDDBuilders.scala spark-testkit/src/main/scala/geotrellis/spark/testkit/TileLayerRDDBuilders.scala
git mv spark-testkit/src/main/scala/geotrellis/spark/testkit/TileRDDMatchers.scala spark-testkit/src/main/scala/geotrellis/spark/testkit/TileLayerRDDMatchers.scala
git mv spark/src/main/scala/geotrellis/spark/TileRDDMethods.scala spark/src/main/scala/geotrellis/spark/TileLayerRDDMethods.scala
git mv spark/src/main/scala/geotrellis/spark/TileRDDSeqMethods.scala spark/src/main/scala/geotrellis/spark/TileLayerRDDSeqMethods.scala
git mv spark/src/main/scala/geotrellis/spark/crop/TileRDDCropMethods.scala spark/src/main/scala/geotrellis/spark/crop/TileLayerRDDCropMethods.scala
git mv spark/src/main/scala/geotrellis/spark/io/filter/TileRDDFilterMethods.scala spark/src/main/scala/geotrellis/spark/io/filter/TileLayerRDDFilterMethods.scala
git mv spark/src/main/scala/geotrellis/spark/mask/TileRDDMaskMethods.scala spark/src/main/scala/geotrellis/spark/mask/TileLayerRDDMaskMethods.scala
git mv spark/src/test/scala/geotrellis/spark/TileRDDFilterMethodsSpec.scala spark/src/test/scala/geotrellis/spark/TileLayerRDDFilterMethodsSpec.scala
git mv spark/src/test/scala/geotrellis/spark/crop/TileRDDCropMethodsSpec.scala spark/src/test/scala/geotrellis/spark/crop/TileLayerRDDCropMethodsSpec.scala
```

PR #1347, opened 2016-03-01
Raster RDD Cropping

jamesmcclain:
These changes add a `crop` method to the `RasterRDD` type that allows the same to be cropped by a given extent.


lossyrob:
Crop methods/code can be in `geotrellis.spark.crop`

spark/src/main/scala/geotrellis/spark/filter/RasterRDDCropMethods.scala
```diff
@@ -0,0 +1,45 @@
...
+    val md = self.metadata
+    val mt = md.mapTransform
+    val rdd = self
+      .filter({ case (key, tile) =>
+        val srcExtent = mt(SpatialKey(key.col, key.row))
+        extent.interiorIntersects(srcExtent) })
+      .map({ case (key, tile) =>
+        val srcExtent = mt(SpatialKey(key.col, key.row))
+        val newTile = tile.crop(srcExtent, extent, options)
+        (key, newTile) })
```
lossyrob: In spark, if we do something with a Key Value RDD, we can lose the partitioner of we don't explicitly state to keep it (in the case where we are potential changing keys), or if we use functions that don't change keys.

In this case, we need the key value, so I would write this as a mapPartitions with `preservePartitioning = true`

```scala
val rdd = self
  .mapPartitions({ partition =>
    partition.flatMap { case (key, tile) =>
      val srcExtent = mt(SpatialKey(key.col, key.row))
      if(extent.interiorIntersects(srcExtent)) {
        if(extent.contains(srcExtent)) {
          Some((key, tile))
        } else {
          Some((key, tile.crop(srcExtent, extent, options)))
        }
      } else {
        None
      }
    }
  }), preservesPartitioning = true)
```

spark/src/main/scala/geotrellis/spark/filter/RasterRDDCropMethods.scala
```diff
@@ -0,0 +1,45 @@
...
+import geotrellis.raster._
+import geotrellis.raster.crop.Crop.Options
+import geotrellis.spark._
+import geotrellis.util.MethodExtensions
+import geotrellis.vector.Extent
+
+import org.apache.spark.rdd._
+
+
+abstract class RasterRDDCropMethods[K <% SpatialKey] extends MethodExtensions[RasterRDD[K]] {
```
lossyrob: `RasterRDDCropMethods[K: SpatialComponent]`

spark/src/main/scala/geotrellis/spark/filter/RasterRDDCropMethods.scala
```diff
@@ -0,0 +1,45 @@
...
+  def crop(extent: Extent, options: Options): RasterRDD[K] = {
+    val md = self.metadata
+    val mt = md.mapTransform
+    val rdd = self
+      .filter({ case (key, tile) =>
+        val srcExtent = mt(SpatialKey(key.col, key.row))
```
lossyrob: `val srcExtent = mt(key)`

spark/src/main/scala/geotrellis/spark/package.scala
```diff
@@ -127,6 +128,12 @@ package object spark
       }, preservesPartitioning = true)
   }
 
+  implicit class withRasterRDDSpatialKeyCropMethods(val self: RasterRDD[SpatialKey])
+      extends RasterRDDCropMethods[SpatialKey]
+
+  implicit class withRasterRDDSpaceTimeKeyCropMethods(val self: RasterRDD[SpaceTimeKey])
```
lossyrob: these implicits should be defined in an `Implicits` trait in the package (`crop`). An object named Implicits in that package extends that (companion object), and then we extent the package with the Implicits trait

spark/src/test/scala/geotrellis/spark/filter/RasterRDDCropMethodsSpec.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  describe("RasterRDD Crop Methods") {
+    val path = "raster-test/data/aspect.tif"
+    val gt = SingleBandGeoTiff(path)
+    val originalRaster = gt.raster.resample(500, 500)
+    val (_, rdd) = createRasterRDD(originalRaster, 5, 5, gt.crs)
+    val md = rdd.metadata
+    val overall = md.extent
+    val half = Extent(633750, 218375, 641250, 225125)
+    val small = Extent(639000.0, 217700.0, 642000.0, 220399.0)
```
lossyrob: Should do computation to get new extents, in the unlikely event that the extent of the underlying raster would ever change.

spark/src/main/scala/geotrellis/spark/filter/Implicits.scala
```diff
@@ -0,0 +1,10 @@
+package geotrellis.spark.filter
+
+import geotrellis.spark.SpaceTimeKey
+
+
+object Implicits extends Implicits
+
+trait Implicits {
+  implicit def SpaceTimeToSpatial(st: SpaceTimeKey) = st.spatialKey
```
lossyrob: No need to have this (can use `SpatialComponent`).

This is where the implicit classes would live, then root package extends this trait

jamesmcclain:

spark/src/main/scala/geotrellis/spark/crop/RasterRDDCropMethods.scala
```diff
@@ -0,0 +1,48 @@
...
+abstract class RasterRDDCropMethods[K: SpatialComponent] extends MethodExtensions[RasterRDD[K]] {
+  def crop(extent: Extent, options: Options): RasterRDD[K] = {
+    val md = self.metadata
+    val mt = md.mapTransform
+    val rdd = self
+      .mapPartitions({ partition =>
+        partition.flatMap({ case (key, tile) =>
+          val srcExtent = mt(key)
+          if (extent.interiorIntersects(srcExtent)) {
+            val newTile = tile.crop(srcExtent, extent, options)
```
lossyrob: This does an unnecessary crop if the srcExtent is contained fully within the extent

jamesmcclain: Okay, corrected.

PR #1348, opened 2016-03-01
Dynamic partition sizing with S3InputFormat

echeipesh:
Resolves: https://github.com/geotrellis/geotrellis/issues/1313

In this PR S3InputFormat loses the `.setMaxKeys` configuration options in favor of `.setPartitionCount` or `.setPartitionBytes` options.

## PartitionCount
In this mode of operation we will divide all keys evenly to create the desired number of partitions

## PartitionBytes
In this mode of operations we attempt to make partitions of at most specified number of bytes. At least one key will be placed in a partition, so it is possible for a partition to exceed the requested size under some conditions.

This is the default mode of operations and will attempt to create 256MB partitions.

## ETL
ETL module has been updated to expose both of these options. However PartitionBytes will be used as a default best effort method if nothing is specified.


echeipesh:
@jamesmcclain could I ask for a review please?

jamesmcclain:
> @jamesmcclain could I ask for a review please?

Sure, will-do.

s3/src/main/scala/geotrellis/spark/io/s3/S3InputFormat.scala
```diff
@@ -48,51 +51,81 @@ abstract class S3InputFormat[K, V] extends InputFormat[K,V] with LazyLogging {
...
 
-      splits = split :: splits
-      request.setMarker(listing.getNextMarker)
-    } while (listing.isTruncated)
+    var splits: Vector[S3InputSplit] = Vector(makeNewSplit)
+
+    if (null == partitionCountConf) {
+      // By default attempt to make partitions the same size
+      val maxSplitBytes = if (null == partitionSizeConf) S3InputFormat.DEFAULT_PARTITION_BYTES else partitionSizeConf.toLong
+      logger.info(s"Building partitions of at most $maxSplitBytes bytes")
```
jamesmcclain: You mentioned it in the PR comment, but perhaps this string should be changed to make clear that the split size is not guaranteed to be \leq maxSplitBytes.

jamesmcclain:
Overall, the code looks both functional and inoffensive.  Besides the picayune comment about the string, the only overall comment that I would have is about the efficiency of the packing strategy.  I don't know how much we care about this, but the strategy being used for the byte-based partitioning seems to the "first fit" strategy.  There are approximation algorithms available with [slightly better gurantees](https://en.wikipedia.org/wiki/Bin_packing_problem).

lossyrob:
@jamesmcclain would be good to write a low priority issue for the more efficient packing algorithm.

lossyrob:
Supersede by #1355

PR #1349, opened 2016-03-01
Raster RDD Filter Methods

jamesmcclain:
These changes add a `filterByKeyBounds` method to the `RasterRDD` type that allows the same to be filtered by one or more `KeyBounds`.

~~The `.filterByKeyBounds(...)` method returns a `BoundedRDDQuery` instead of an `RDDQuery` because the former has a `toRDD` method and the latter does not.~~

### Still Need To ###
   - [x] ~~Confirm that this covers the requirement ".filter method which returns an RDDQuery, whose .toRDD method will produce the filtered RDD"~~

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/KeyBounds.scala
+++ b/spark/src/main/scala/geotrellis/spark/KeyBounds.scala
@@ -19,9 +19,9 @@ sealed trait Bounds[+A] extends Product with Serializable {
   def intersects[B >: A](other: KeyBounds[B])(implicit b: Boundable[B]): Boolean =
     intersect(other).nonEmpty
 
-  def get: KeyBounds[A] 
+  def get: KeyBounds[A]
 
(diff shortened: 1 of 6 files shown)


lossyrob:
Desired API:

```scala

val rdd: RDD[(K, V)] with Metadata[M] = ???

rdd
  .filter
  .where(Between(time1, time2))
  .where(Interesects(polygon))
  .toRDD
```

lossyrob:
In this PR, remove `asInstanceOf` in polygon filtering:

https://github.com/jamesmcclain/geotrellis/blob/feature/jwm/spatial-filtering-2/spark/src/main/scala/geotrellis/spark/io/RDDFilter.scala#L121

lossyrob:
For this PR, we'll just do a `filterByKeyBounds` that returns an RDD (not a query). This should have overloads for `KeyBounds[K]` and `Seq[KeyBounds[K]]`. Later we'll figure out a good way to implement the `filter` with the query mechanism.

jamesmcclain:

PR #1351, opened 2016-03-01
Streamline ETL Interface and extends README

echeipesh:
# ETL Interface
With this PR the ETL interface changes back to more familiar format of `load`, `tile`, and `save`:
```scala
object GeoTrellisETL extends App {
  implicit val sc = SparkUtils.createSparkContext("GeoTrellis ETL", new SparkConf(true))
    
  /* parse command line arguments */
  val etl = Etl(args)
  /* load source tiles using input module specified */
  val sourceTiles = etl.load[ProjectedExtent, Tile]
  /* perform the reprojection and mosaicing step to fit tiles to LayoutScheme specified */
  val (zoom, tiled) = etl.tile(sourceTiles)
  /* save and optionally pyramid the mosaiced layer */
  etl.save(LayerId(etl.conf.layerName(), zoom), tiled, ZCurveKeyIndexMethod())
    
  sc.stop()
}
```

# Reproject Method
ETL now allows for two methods of projection: `buffered` and `pertile`, with `buffered` being the default. `pertile` method allows sourcing heterogenous CRS datasets with some tradeoff in accuracy.

# ETL README
README has been expanded to cover common questions and provide more context for the project.


echeipesh:
@moradology I would appreciate some comments on the README update since you're dealing with the documentation now.

lossyrob:
@jasonbeverage this PR will make the change we were talking about, where an option is added to ETL to do the per source tile (for heterogeneous CRSs), or buffered in the case where there is only one CRS.

spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala
```diff
@@ -72,35 +80,92 @@ case class Etl(args: Seq[String], @transient modules: Seq[TypedModule] = Etl.def
...
+      scheme match {
+        case Left(layoutScheme) =>
+          RasterMetaData.fromRdd(source, layoutScheme)
+        case Right(layoutDefinition) =>
+          RasterMetaData.fromRdd(source, layoutDefinition)
+      }
+    }
+    val adjustedMetadata = targetCellType.fold(rmd){ ct => rmd.copy(cellType = ct) }
+    val tiled: RDD[(K, V)] with Metadata[RasterMetaData] =
+      ContextRDD(source.tileToLayout[K](adjustedMetadata, method), adjustedMetadata)
```
lossyrob: This isn't right. If we are doing a buffered reproject, then we always tile to a floating layout scheme using the source tile CRS. *Then* we reproject and tile to the desired layout. If we reproject against the source tiles, this will work, but for the buffered tile case, the first tileToLayout is wrong.

echeipesh: Good catch. The way I understand it is that current approach would be produce results but they would be suboptimal?

lossyrob:
+1

PR #1352, opened 2016-03-01
BufferTiles Fix

jamesmcclain:
These changes (attempt to) fix a bug in the `BufferTile` functionality.  That bug would cause exceptions when trying to process certain non-rectangular RDDs.


jasonbeverage:
Just to follow up, I tried my failed test case with this change and it appears to work correctly now.  Thanks!

jamesmcclain:
You are most welcome, sir.  Glad to help.

PR #1356, opened 2016-03-02
application.conf and reference.conf

jamesmcclain:
Changes:
   * Various configuration files are removed in this pull request.  Those files which are required to compile and pass tests have been retained untouched.
   * A stray file from the deleted `vector-benchmarks` project has been removed.
   * A typo in the recent `BufferTiles` fixes, https://github.com/geotrellis/geotrellis/pull/1352, is corrected


spark/src/main/resources/reference.conf
```diff
@@ -1 +0,0 @@
-geotrellis.accumulo.catalog = "metadata"
```
lossyrob: We actually need this. This is a default that will be pulled into the user's `application.conf`, even if they don't supply one or unless they overwrite it.

raster/src/main/resources/reference.conf
```diff
@@ -1,11 +0,0 @@
...
-include "akka-remote-reference"
-
-# Set the geotrellis catalog here.
-# Required for reading any data from GeoTrellis by name.
-#geotrellis.catalog = "src/test/resources/catalog.json"
-
-akka {
-  log-dead-letters = 0
-  log-dead-letters-during-shutdown = off
-}
```
lossyrob: This `reference.conf` should actually be moved to the `engine` project's not-yet-existing `reference.conf`

jamesmcclain: Okay, will-do.

spark/src/main/resources/reference.conf
jamesmcclain: It shall be restored.

lossyrob: sorry...this actually doesn't belong in `spark`, but belongs in `accumulo` now

jamesmcclain: Okay, it shall go to accumulo.

PR #1360, opened 2016-03-03
[Fix] Small fixes

pomadchin:
- [x] Allowed empty extents (added test coverage)
- [x] Catching cropped tiles to combine
- [x] Remove merge strategy from subproject build files that don't need it

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/ArrayTile.scala
+++ b/raster/src/main/scala/geotrellis/raster/ArrayTile.scala
@@ -138,7 +138,9 @@ trait ArrayTile extends Tile with Serializable {
       case ct: ConstantTile =>
         ct.combine(this)(f)
       case ct: CompositeTile =>
-        ct.combine(this)((z1, z2)=>f(z2, z1))
+        ct.combine(this)((z1, z2) => f(z2, z1))
+      case ct: CroppedTile =>
+        ct.combine(this)((z1, z2) => f(z2, z1))
     }
   }
 
--- a/vector-test/src/test/scala/spec/geotrellis/vector/ExtentSpec.scala
+++ b/vector-test/src/test/scala/spec/geotrellis/vector/ExtentSpec.scala
@@ -16,8 +16,6 @@
 
 package geotrellis.vector
 
-import java.awt.Rectangle
-
 import geotrellis.vector.io.json._
 import org.scalatest.FunSpec
 import org.scalatest.Matchers
--- a/vector-test/src/test/scala/spec/geotrellis/vector/ExtentSpec.scala
+++ b/vector-test/src/test/scala/spec/geotrellis/vector/ExtentSpec.scala
@@ -16,7 +16,8 @@
 
 package geotrellis.vector
 
-import geotrellis.vector.io._
+import java.awt.Rectangle
+
 import geotrellis.vector.io.json._
 import org.scalatest.FunSpec
 import org.scalatest.Matchers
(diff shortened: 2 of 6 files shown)


pomadchin:
@lossyrob Quick question about merge strategy: now we have it only for `spark-etl` and for `spark` sub projects; should I left these lines only for a `spark-etl` or we need `spark` assembly too?

vector/src/main/scala/geotrellis/vector/Extent.scala
```diff
@@ -43,9 +43,9 @@ object ProjectedExtent {
  */
 case class Extent(xmin: Double, ymin: Double, xmax: Double, ymax: Double) {
 
-  // Validation: Do not accept extents min values greater than max values.
-  if (xmin > xmax) { throw ExtentRangeError(s"Invalid Extent: xmin must be less than xmax (xmin=$xmin, xmax=$xmax)")  }
-  if (ymin > ymax) { throw ExtentRangeError(s"Invalid Extent: ymin must be less than ymax (ymin=$ymin, ymax=$ymax)") }
+  // Validation: Do not accept extents min values greater than or equal to max values (allows empty extents).
+  if (xmin >= xmax) { throw ExtentRangeError(s"Invalid Extent: xmin must be less than or equal to xmax (xmin=$xmin, xmax=$xmax)")  }
```
pomadchin: @lossyrob Want to note, that `xmin >= xmax` *not* allowing empty extents; the previous condition `xmin > xmax` accepted: other words all extents with `xmin <= xmax` were accepted; should I rollback changes?

vector-test/src/test/scala/spec/geotrellis/vector/ExtentSpec.scala
```diff
@@ -95,6 +99,11 @@ class ExtentSpec extends FunSpec with Matchers {
       assert(e.contains(10.0, 10.0) === false)
     }
 
+    it("empty extent should contain empty extent") {
```
pomadchin: Should empty extent contain an empty extent? How the contain function should behave on extent bounds?

vector/src/main/scala/geotrellis/vector/Extent.scala
lossyrob: Whoops, yes we were already accepting empty extents then, so roll that back.

vector-test/src/test/scala/spec/geotrellis/vector/ExtentSpec.scala
lossyrob: No I don't think an empty extent should contain anything, including itself

lossyrob:
Merge strategy only applies to ETL

raster/src/main/scala/geotrellis/raster/ArrayTile.scala
```diff
@@ -138,7 +138,9 @@ trait ArrayTile extends Tile with Serializable {
       case ct: ConstantTile =>
         ct.combine(this)(f)
       case ct: CompositeTile =>
-        ct.combine(this)((z1, z2)=>f(z2, z1))
+        ct.combine(this)((z1, z2) => f(z2, z1))
+      case ct: CroppedTile =>
```
lossyrob: Can you add a small test that will cover this code?

vector-test/src/test/scala/spec/geotrellis/vector/ExtentSpec.scala
pomadchin: And what about non empty extents? Should any extent contain itself?

raster/src/main/scala/geotrellis/raster/ArrayTile.scala
pomadchin: :+1: thanks for a quick review

vector-test/src/test/scala/spec/geotrellis/vector/ExtentSpec.scala
lossyrob: yes non empty extents contain themselves

PR #1362, opened 2016-03-03
Upgrade All Dependencies

jamesmcclain:
The following dependencies have not been upgraded:
   * com.typesafe.config due to incompatible byte code
   * com.typesafe.slick due to use of an extension that does not support the newer version
   * hadoop due to the fact that it is marked "provided"
   * spark due to the fact that it is marked "provided"
   * com.google.guava.guava due to the fact that it isn't clear which version to upgrade to

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -42,7 +42,7 @@ object Dependencies {
   val sprayHttpx    = "io.spray"        %% "spray-httpx"   % Version.spray
   val sprayJson     = "io.spray"        %% "spray-json"    % Version.sprayJson
 
-  val apacheMath    = "org.apache.commons" % "commons-math3" % "3.5"
+  val apacheMath    = "org.apache.commons" % "commons-math3" % "3.6"
 
   val jettyWebapp   = "org.eclipse.jetty" % "jetty-webapp" % "8.1.0.RC4"
   val jerseyBundle  = "com.sun.jersey"    % "jersey-bundle" % "1.11"
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -48,7 +48,7 @@ object Dependencies {
   val jerseyBundle  = "com.sun.jersey"    % "jersey-bundle" % "1.11"
   val slf4jApi      = "org.slf4j"         % "slf4j-api" % "1.6.0"
   val asm           = "asm"               % "asm"       % "3.3.1"
-  
+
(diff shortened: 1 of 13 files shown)


project/Dependencies.scala
```diff
@@ -42,27 +42,27 @@ object Dependencies {
...
+  val postgresql    = "postgresql"         % "postgresql"  % "9.2-1002.jdbc4"
 
   val caliper       = ("com.google.code.caliper" % "caliper" % "1.0-SNAPSHOT"
     from "http://plastic-idolatry.com/jars/caliper-1.0-SNAPSHOT.jar")
 
-  val nscalaTime    = "com.github.nscala-time" %% "nscala-time" % "2.6.0"
+  val nscalaTime    = "com.github.nscala-time" %% "nscala-time" % "2.10.0"
 
-  val awsSdkS3      = "com.amazonaws" % "aws-java-sdk-s3" % "1.9.34"
+  val awsSdkS3      = "com.amazonaws" % "aws-java-sdk-s3" % "1.10.57"
```
lossyrob: We should revert this dependency back. This has always been a tricky one with hadoop, and this version was meaningful, it was the one that let things work against `geotrellis-ec2-cluster`. Can you throw a comment in saying "Using this version of aws-java-sdk-s3 for compatibility reasons" in there?

jamesmcclain: Certainly.

spark/build.sbt
```diff
@@ -4,8 +4,8 @@ name := "geotrellis-spark"
 libraryDependencies ++= Seq(
   "org.apache.spark" %% "spark-core" % Version.spark % "provided",
   "org.apache.hadoop" % "hadoop-client" % Version.hadoop % "provided",
-  "de.javakaffee" % "kryo-serializers" % "0.27" exclude("com.esotericsoftware.kryo", "kryo"),
-  "com.esotericsoftware.kryo" % "kryo" % "2.21",
+  "de.javakaffee" % "kryo-serializers" % "0.37" exclude("com.esotericsoftware.kryo", "kryo"),
+  "com.esotericsoftware.kryo" % "kryo" % "2.24.0",
```
pomadchin: Spark uses kryo `2.21`, that was the reason to exclude kryo version.

jamesmcclain: Thank you for catching this.

PR #1367, opened 2016-03-04
Rasterizer API Updates

jamesmcclain:
This pull request ensures the existence of `foreach` and/or `rasterize` methods for the `Raster`, `RasterExtent`, `Feature`, and `Geometry` types.

   - [x] Add tests for new code, where needed
   - [x] `foreachCell` -> `foreach`
   - [x] Make `rasterize` methods return objects of the type `Raster` rather than `Tile`


raster/src/main/scala/geotrellis/raster/rasterize/RasterRasterizerMethods.scala
```diff
@@ -0,0 +1,56 @@
...
+    val cols = tile.cols
+    val rows = tile.rows
+
+    cfor(0)(_ < cols, _ + 1) { col =>
+      cfor(0)(_ < rows, _ + 1) { row =>
+        fn(col, row, tile.getDouble(col, row))
+      }
+    }
+  }
+}
```
lossyrob: We don't need these methods. There is already `foreach` and `foreachDouble` methods on tile. For the `(Int, Int, Int) => Int` and `(Int, Int, Double) => Double` cases, they are macro-ized to avoid boxing. So we can delete this trait.

raster/src/main/scala/geotrellis/raster/rasterize/SingleBandRasterRasterizeMethods.scala
```diff
@@ -56,12 +56,12 @@ trait SingleBandRasterRasterizeMethods[T <: Tile, S <: Raster[T]] extends Method
...
 
-    Rasterizer.foreachCellByGeometry(geom, re, options)(new Callback {
+    geom.foreach(re, options)(new Callback {
       def apply(col : Int, row : Int): Unit = {
         val z = tile.get(col, row)
         mutableTile.setDouble(col, row, fn(z))
       }
     })
-    mutableTile
+    Raster(mutableTile, extent)
```
lossyrob: We don't need rasterize methods on `Raster`, this is something people will call on the geometry or feature, so remove this trait.

raster/src/main/scala/geotrellis/raster/rasterize/TileRasterizeMethods.scala
```diff
@@ -23,17 +23,17 @@ import geotrellis.util.MethodExtensions
...
-    geom : Geometry,
-    extent : Extent,
+  def rasterizeDouble(
+    geom: Geometry,
+    extent: Extent,
     options: Options = Options.DEFAULT
-  )(fn : Double => Double) : Tile =
-    Raster(self, extent).foreachCellDouble(geom, options)(fn)
+  )(fn : Double => Double): Raster[MutableArrayTile] =
+    Raster(self, extent).rasterizeDouble(geom, options)(fn)
```
lossyrob: We don't need rasterize methods on Raster, this is something people will call on the geometry or feature, so remove this trait.

raster/src/main/scala/geotrellis/raster/rasterize/RasterRasterizerMethods.scala
jamesmcclain: Okay, it will go.

raster/src/main/scala/geotrellis/raster/rasterize/SingleBandRasterRasterizeMethods.scala
jamesmcclain: It shall be removed.

raster/src/main/scala/geotrellis/raster/rasterize/TileRasterizeMethods.scala
jamesmcclain: > We don't need rasterize methods on Raster

Did you mean "Tile" rather than "Raster"?

jamesmcclain:

PR #1369, opened 2016-03-06
Implicits reorganization

pomadchin:
It's enough just to import `geotrellis.spark.io._` to provide all implicits necessary for readers and writers
- [x] test it in integrations tests project


lossyrob:
:+1:  w00t

PR #1370, opened 2016-03-07
Remove references to gdal module

### LINKED ISSUE #1135: Dependency issues: GDAL and ScalaZ stream
We depend on the following in geotrellis-spark 0.10.0-M1.

This causes errors when trying to use the library without some repos in the `resolvers`. We should require our users to have unfound dependencies and have to debug to figure out how to include these transitive dependencies.

```scala
  "org.scalaz.stream" %% "scalaz-stream" % "0.7a"
  "org.gdal"         % "gdal"       % "1.10.1"
```

The solution to the GDAL one is, don't have geotrellis-spark depend on the gdal project (it shouldn't anyway). geotrellis-gdal clearly needs it, but there should be documentation about needing to add the opengeo repository.

@echeipesh, what about scalaz-stream? Is this library not on maven central? In order to use it, will we require users to have to put in the "Scalaz Bintray Repo" at "https://dl.bintray.com/scalaz/releases" resolver? That would not be ideal. Also, what does the 'a' in the version string mean? Is this a stable release?

Moving forward, we should be wary of adding any dependency that is not a released version and is not on maven central.

moradology:
This should be merged in once [`geotrellis.gdal`](https://github.com/geotrellis/geotrellis-gdal) is pinned against a specific version of geotrellis (rather than SNAPSHOT)

Fixes #1135


lossyrob:
Note: `geotrellis-gdal` functionality has moved to https://github.com/geotrellis/geotrellis-gdal

PR #1373, opened 2016-03-07
Histogram API Improvements

jamesmcclain:
Numerous changes have been made to the Histogram API.  Most notably from an external perspective, the "get" and "generate" verbs have been removed from various methods, unused histogram implementations have gone, the `toJSON` method has gone, and the `merge` method has been added.

Also, "double" versions (to accompany "int" versions) of various `geotrellis.raster` and `geotrellis.spark` methods that make use of histograms have been created.  Notably, the `StrictColorClassifier.fromQuantileBreaks` function can now produce color breaks using the double histogram implementation.  Tests have also been created for those.

Various functions/methods in the png rendering code, jpeg rendering code, and `geotrellis.engine` project which use histograms have not yet been given "double" versions.


raster/src/main/scala/geotrellis/raster/histogram/ArrayHistogram.scala
```diff
@@ -79,10 +75,10 @@ class ArrayHistogram(val counts: Array[Int], var total: Int)
     counts(i) += count
   }
 
-  def getItemCount(i: Int) = counts(i)
+  def itemCount(i: Int) = counts(i)
 
   // REFACTOR: use Option
```
lossyrob: We should remove these REFACTOR lines, unless what they are trying to do make sense, then we should do them and remove them.

raster/src/main/scala/geotrellis/raster/histogram/ArrayHistogram.scala
```diff
@@ -101,4 +97,11 @@ class ArrayHistogram(val counts: Array[Int], var total: Int)
...
     return Int.MinValue
   }
+
+  def merge(histogram: Histogram[Int]): Histogram[Int] = {
+    val total = ArrayHistogram(histogram.bucketCount())
+
+    total.update(this); total.update(histogram)
+    total
+  }
 }
```
lossyrob: I see you kept ArrayHistogram around...it was on the chopping block, what was the justification for keeping it in?

raster/src/main/scala/geotrellis/raster/histogram/StreamingHistogram.scala
```diff
@@ -396,25 +437,29 @@ class StreamingHistogram(
...
   }
 
-  def getPercentile(q: Double): Double =
-    getPercentileBreaks(List(q)).head
+  def percentile(q: Double): Double =
+    percentileBreaks(List(q)).head
 
-  def getQuantileBreaks(num: Int): Array[Double] =
-    getPercentileBreaks(List.range(0,num).map(_ / num.toDouble)).toArray
+  def quantileBreaks(num: Int): Array[Double] =
```
lossyrob: Can we put some scaladocs that talk about the guarantee that we talked about this morning, that there will be no item in the value set this histogram represents that is outside of the bounds of the quantile breaks returned?

raster/src/main/scala/geotrellis/raster/render/color/ColorClassifier.scala
```diff
@@ -429,9 +429,11 @@ object StrictColorClassifier {
     colorClassifier
   }
 
-  def fromQuantileBreaks(histogram: Histogram[Int], colors: Array[RGBA]) = {
-    val breaks = histogram.getQuantileBreaks(colors.length)
-    apply(breaks zip colors)
+  def fromQuantileBreaks[X <: AnyVal : ClassTag](histogram: Histogram[X], colors: Array[RGBA]) = {
+    histogram.quantileBreaks(colors.length) match {
+      case breaks: Array[Int] => apply(breaks zip colors)
+      case breaks: Array[Double] => apply(breaks zip colors)
```
lossyrob: Need to handle the default case here.

raster/src/main/scala/geotrellis/raster/render/color/ColorClassifier.scala
```diff
@@ -429,9 +429,11 @@ object StrictColorClassifier {
     colorClassifier
   }
 
-  def fromQuantileBreaks(histogram: Histogram[Int], colors: Array[RGBA]) = {
-    val breaks = histogram.getQuantileBreaks(colors.length)
-    apply(breaks zip colors)
+  def fromQuantileBreaks[X <: AnyVal : ClassTag](histogram: Histogram[X], colors: Array[RGBA]) = {
```
lossyrob: Not into this generic function that just checks two types.

I think if there were simply an overload that took `Array[Double], Array[RGBA]` (or traversables in general), and did the zip for you, that would be sufficient.

spark/src/main/scala/geotrellis/spark/mapalgebra/zonal/ZonalTileRDDMethods.scala
```diff
@@ -14,12 +14,12 @@ import spire.syntax.cfor._
 
 trait ZonalTileRDDMethods[K] extends TileRDDMethods[K] {
 
-  private def mergeMaps(a: Map[Int, Histogram[Int]], b: Map[Int, Histogram[Int]]) = {
+  private def mergeMaps[X <: AnyVal](a: Map[Int, Histogram[X]], b: Map[Int, Histogram[X]]) = {
```
lossyrob: Why X? `T` seems more appropriate here.

raster/src/main/scala/geotrellis/raster/histogram/ArrayHistogram.scala
jamesmcclain: Evidently I overlooked it at the beginning of the list.

raster/src/main/scala/geotrellis/raster/histogram/ArrayHistogram.scala
jamesmcclain: Okay, will-do.

raster/src/main/scala/geotrellis/raster/histogram/StreamingHistogram.scala
jamesmcclain: Will-do.

spark/src/main/scala/geotrellis/spark/mapalgebra/zonal/ZonalTileRDDMethods.scala
jamesmcclain: It shall be changed.

jamesmcclain:
I believe that all comments prior to this one have been addressed.  The refactor suggested in the comments has been made, as well as one logically implied by it.  The respective commits can be removed if those changes are not desired.

PR #1377, opened 2016-03-10
Supply MultiBandTile bands (reorder) Method

### LINKED ISSUE #1112: MultiBandTile method to subset/reorder the bands.
For instance, if I have an 8 band tile, I might want to do something like this:

```scala
val eightBandTile: MultiBandTile = ???
val reorderedThreeBandTile: MultiBandTile = eightBandTile.bands(2,1,3)
```

jamesmcclain:
This pull request adds a `bands` (reorder) method to `MultiBandTiles`, where possible.

In https://github.com/geotrellis/geotrellis/commit/b78a1ef7b3d4ab792de19b5e4ea3f6a0bb6ecb82, a trait is added to gurantee the existence of the method where it is possible to do so.  That commit can be removed if it is not desired.

Closes #1112

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/io/geotiff/MultiBandGeoTiff.scala
+++ b/raster/src/main/scala/geotrellis/raster/io/geotiff/MultiBandGeoTiff.scala
@@ -11,7 +11,7 @@ class MultiBandGeoTiff(
   val crs: CRS,
   val tags: Tags,
   options: GeoTiffOptions
-) extends GeoTiff[MultiBandTile] {
+) extends GeoTiff[MultiBandTile] with SimpleMultiBandTile {
   val cellType = tile.cellType
 
   def mapTile(f: MultiBandTile => MultiBandTile): MultiBandGeoTiff =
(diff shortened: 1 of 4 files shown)


raster/src/main/scala/geotrellis/raster/io/geotiff/MultiBandGeoTiff.scala
```diff
@@ -37,8 +37,14 @@ class MultiBandGeoTiff(
...
 
-  def subset(bands: Int*)(implicit d: DummyImplicit): ArrayMultiBandTile =
-    subset(bands)
+  def bands(bandSequence: Int*)(implicit d: DummyImplicit): ArrayMultiBandTile =
+    bands(bandSequence)
+
+  def subset(bandSequence: Seq[Int]): ArrayMultiBandTile =
+    bands(bandSequence.sorted)
+
+  def subset(bandSequence: Int*)(implicit d: DummyImplicit): ArrayMultiBandTile =
```
lossyrob: you know what, I think `bands` is enough, I don't think we need a `subset` method.

jamesmcclain: `subset` shall be removed.

jamesmcclain:

PR #1378, opened 2016-03-10
Provide toSpatial Method

### LINKED ISSUE #1057: Add a method to convert a subset of RasterRDD[SpaceTimeKey] -> RasterRDD[SpatialKey]
It is potentially useful to be able to change dimensions on an RDD. Since currently we only support 3D (spatiotemporal) and 2D (spatial) keys, this would mean taking an RDD[(SpaceTimeKey, V)] and subsetting it to a RDD[(SpatialKey, V)]. This is basically a filter on a specific time slice, and then a mapping of keys.

```scala
rdd
  .filter { case (k, _) => k.instant == someInstant }
  .map { case (k, v) => (k.getComponent[SpatialKey], v) }
```

jamesmcclain:
These changes equip every `RasterRDD` keyed by `SpaceTimeKey` with a `toSpatial` (formerly `filterByInstant`) method that filters out all entries not occurring at the given instant, and also projects the keys to the `SpatialKey` type.

Closes #1057

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/TileLayout.scala
+++ b/raster/src/main/scala/geotrellis/raster/TileLayout.scala
@@ -1,12 +1,12 @@
 /*
  * Copyright (c) 2014 Azavea.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


lossyrob:
I feel weird about calling this Filter, because it actually changes key types.

I'm not sure a good name. It is sort of dimension reduction in a general case, where you have some mapping of K1 -> K2 and fix one dimension of K1. Is there a good work for taking a plane (or N-1 hyperplane) in 3D space (or N dimensional space)?

jamesmcclain:
> Is there a good work for taking a plane (or N-1 hyperplane) in 3D space (or N dimensional space)?

I think that the word normally used is "projection".

jamesmcclain:
`.toSpatial` sounds good.

jamesmcclain:

PR #1379, opened 2016-03-10
ETL Accumulo write strategy configuration option

echeipesh:
#### Accumulo Output

Accumulo output module has two write strategies: 
  - `hdfs` strategy uses Accumulo bulk import
  - `socket` strategy uses Accumulo `BatchWriter`
    
When using `hdfs` strategy `ingestPath` argument will be used as the temporary directory where records will be written
for use by Accumulo bulk import. This directory should ideally be an HDFS path.

Fixes: https://github.com/geotrellis/geotrellis/issues/1339
Depends On: https://github.com/geotrellis/geotrellis/pull/1346


lossyrob:
Looks like this is agains io refactor, can you make the pr against that branch.

lossyrob:
Merged into IO refactor, closing.

PR #1381, opened 2016-03-11
Documentation for geotrellis.util

jamesmcclain:
### Still Needs ###
   - [x] README.md
   - [x] Scaladoc for `MethodExtensions.scala`

### DIFF
--- a/docs/util/util-intro.md
+++ b/docs/util/util-intro.md
@@ -1,0 +1,107 @@
+# geotrellis.util
+
+## Introduction ##
+
+The `geotrellis.util` package contains utility functionality that does not properly belong to any single project.
+At present, there are only two source files housed in this project:
+[`util/src/main/scala/geotrellis/util/Filesystem.scala`](https://github.com/geotrellis/geotrellis/blob/master/util/src/main/scala/geotrellis/util/Filesystem.scala) and
+[`util/src/main/scala/geotrellis/util/MethodExtensions.scala`](https://github.com/geotrellis/geotrellis/blob/master/util/src/main/scala/geotrellis/util/MethodExtensions.scala).
+The motivation-for and usage-of those two pieces of code is provided below.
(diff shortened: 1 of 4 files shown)


docs/util/util-intro.md
```diff
@@ -0,0 +1,107 @@
...
+(compare the example below with the `package.scala` file linked-to above):
+```scala
+package object raster
+    extends buffer.Implicits
+    with display.Implicits // Add Our Extension Methods
+    with Serializable
+{
+...
+}
+```
```
lossyrob: There are couple of caveats that I would want to mention:

- If your method extensions need [ContextBounds](http://docs.scala-lang.org/tutorials/FAQ/context-and-view-bounds.html) or other implicit parameters, you can't define them as a trait. In this case, define them as an `abstract class` and implement the `withFooMethods` in the `Implicits` as you mentioned, extending the abstract class. Example: https://github.com/geotrellis/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/stitch/StitchRDDMethods.scala#L38

- If you don't require context bounds or other implicit parameters, but you're defining the extension methods on a core type of the library, then look to see if there is already an implicit class that gives method extensions to core types in the package object, and don't define a new implicit class in an Implicits object. Example: https://github.com/geotrellis/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/package.scala#L70

- If you are going to be working with RDD's, and potentially serializing implicit parameters of the implicit class that will be extending the trait, then you should define an object that contains the functionality, and the methods trait or abstract class should call out to that object. Example: https://github.com/geotrellis/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/filter/TileLayerRDDFilterMethods.scala#L35. This avoids serialization issues.

jamesmcclain:

PR #1386, opened 2016-03-14
0.10 cleanup

### LINKED ISSUE #1062: Optimize foreach, foreachDouble, map, mapDouble in CompositeTile
Currently `CompositeTile` has implementations of these methods that just use `get`. This is not an optimal way to do things. It should really happen per tile like you see in the `toArray` and `toArrayDouble` implementations. `map` and `mapDouble` should also keep the return type to `CompositeTile` as well, instead of creating one big ArrayTile.

Once the versions of `map( (Int, Int, Int) => Int )` etc are implemented, these need to be optimized as well.

lossyrob:
These are final 0.10 cleanup tasks, including:
- Moving code around to the appropriate packages
- Making code fit the design patterns of the library
- Fixing RDD filtering/cropping by having them modify metadata

Fixes #1062


build.sbt
```diff
@@ -54,7 +55,22 @@ lazy val commonSettings = Seq(
 )
 
 lazy val root = Project("geotrellis", file(".")).
-  dependsOn(raster, vector, proj4, spark, sparkEtl, s3, accumulo).
+  dependsOn(
```
echeipesh: Recently realized that `aggregate` is more appropriate here: http://stackoverflow.com/questions/26477388/is-the-use-of-aggregate-following-by-dependson-redundant-with-the-same-modul

That would also mean that including % "test" shouldn't be required.

PR #1387, opened 2016-03-14
Vector documentation

moradology:
This PR includes scaladoc and markdown documentation fixes.


docs/vector/vector-intro.md
```diff
@@ -45,28 +46,28 @@ val nonIntersection = myPoint & Point(200, 300)
 
 At this point, the values `selfIntersection` and `nonIntersection`
 are `GeometryResult` containers. These containers are what many JTS
-operations on `Geometry` objects will wrap their results in. Source
-for the To idiomatically destructure these wrappers, we do the following:
+operations on `Geometry` objects will wrap their results in.
+To idiomatically destructure these wrappers, we can use the
+`as[G <: Geometry]` function which either returns `Some(G)` or `None`.
```
dwins: `as` is convenient but it gives up some typesafety - `(r: PointGeometryIntersectionResult).as[Line]` will compile fine but has zero possibility of returning a `Some` .  I think we should at least mention the benefit of the `match` based approach (exhaustiveness checking) as this is the entire reason the Result types exist.

docs/vector/vector-intro.md
```diff
@@ -45,28 +46,28 @@ val nonIntersection = myPoint & Point(200, 300)
...
+val pointIntersection = (Point(1.0, 2.0) & Point(1.0, 2.0)).as[Point]
+val pointNonIntersection = (Point(1.0, 2.0) & Point(12.0, 4.0)).as[Point]
+
+assert(pointIntersection == Some(Point(1.0, 2.0)))  // Either some point
+assert(pointNonIntersection == None)                // Or nothing at all
 ```
 
 Beyond the methods which come with any `Geometry` object there are
 implicits in many geotrellis modules which will extend Geometry
 capabilities. For instance, after importing `geotrellis.vector._`,
```
dwins: This is the wrong import, isn't it?

docs/vector/vector-intro.md
```diff
@@ -83,122 +85,87 @@ myML.unionGeometries
...
 
-##GeoTrellis Extents
+## GeoTrellis Extents
 
 There's one more piece to the `geotrellis.vector` puzzle: `Extent`.
 A `geotrellis.vector.Extent` is nothing more than a rectangular
-polygon which is projected (look
-[here](../../../../../proj4/src/main/scalageotrellis/proj4) for more
-on projection). This is useful mainly as a tool for defining the
+polygon on projection). This is useful mainly as a tool for defining the
```
dwins: Something wonky here - this closing parenthesis does not correspond to any opening parenthesis.

vector/src/main/scala/geotrellis/vector/Geometry.scala
```diff
@@ -24,19 +24,29 @@ import GeomFactory._
 
 import scala.reflect.{ ClassTag, classTag }
 
+/** A trait inhereted by classes which wrap a jts.Geometry */
```
dwins: "inherited"

docs/vector/vector-intro.md
lossyrob: I changed this to be similar to the rest of the library, where we are now placing implicits in the root package, so this is actually correct.

docs/vector/vector-intro.md
moradology: Added below. good call

PR #1390, opened 2016-03-15
Fix avro field constructor compatibility

pomadchin:
Hadoop 2.6.x and 2.7.x depend on `avro 1.7.4`. We have newer `1.8.0` version and we use new, constructor function https://github.com/geotrellis/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/io/kryo/KryoRegistrator.scala#L212 which is available only in `avro 1.8.0`. That may cause a problem during ingest which looks like that: 

```
Exception in thread "main" java.lang.NoSuchMethodError: org.apache.avro.Schema$Field.<init>(Ljava/lang/String;Lorg/apache/avro/Schema;Ljava/lang/String;Ljava/lang/Object;)V
	at geotrellis.spark.io.kryo.KryoRegistrator.registerClasses(KryoRegistrator.scala:212)
```

Due to tested compatibility we can use the old constructor.


jamesmcclain:
Good find.

PR #1393, opened 2016-03-15
Fixed issue with Accumulo ingest ordering

pomadchin:


accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloRDDWriter.scala
```diff
@@ -33,8 +33,8 @@ object AccumuloRDDWriter {
         // Map and sort first, so that partitioner is carried over to groupByKey
         raster
           .map { case (key, value) => (encodeKey(key), (key, value)) }
-          .sortByKey()
           .groupByKey()
+          .sortByKey()
```
pomadchin: Using `.sortByKey().groupByKey()` the order is not guaranteed actually, only which keys end up in each partition; reducers may fetch tasks in an arbitrary order, depending on which ones are available first, that's why to guarantee ordering we should sort after group.

lossyrob: This actually ruins the optimization that the write strategy `requiresSort` was going for, so this sort should go back into the write strategy (since what I was stupidly going for was to carry the partitioner over from the sort to the group so I wouldn't require a shuffle). I'm going to make that change now.

lossyrob:
Superseded by #1394

PR #1407, opened 2016-03-18
Fix/color

moradology:
This PR adds further testing of rendering capabilities and renames the method `toRGB` to the more accurate `toARGB`


jamesmcclain:
Let me ask a philosophical question:  Is it the PR doing all of those things, or you doing those things in the PR?

PR #1409, opened 2016-03-18
Fix: focal floating point errors

echeipesh:
A collection of fixes for focal operations, especially to do with floating point cell values

- focal Sum, Min, Max preserve the input tile cell type. Other focal operations change the semantics of the cell
- added `ArrayTileResult` that preserves the input cell type but allows overwriting
- focal Sum fixed to no longer assume starting sum is 0
- focal Max fixed to no longer assume start min is Double.MinValue
- StandardDeviation starts with NoData value
- added tests for allowing Tiles with user defined NoData values


echeipesh:
@lossyrob I had to change the test for the focal SumSpec because the test assumed you could remove cells without adding any to see a result. Current test enforces the invariant that if you're removing a cell there must have been a addition for that cell beforehand. Consequently focal Sum will not allow to remove cells when the accumulator is NODATA. Worth while to double check me on this logic.

lossyrob:
Superseded by #1412

PR #1413, opened 2016-03-21
Add ability to handle fallback colors on indexed pngs

moradology:
I noticed that fallback colors were being handled improperly in indexed PNGs. Logic was added to handle those cases and tests constructed to ensure that both nodata and fallback cases are handled appropriately.


raster/src/main/scala/geotrellis/raster/render/png/PngColorEncoding.scala
```diff
@@ -78,6 +86,10 @@ object PngColorEncoding {
         grey &&= c.isGrey
         i += 1
       }
+      opaque &&= fallbackColor.isOpaque
+      grey &&= fallbackColor.isGrey
+      opaque &&= noDataColor.isOpaque
+      grey &&= noDataColor.isGrey
```
lossyrob: nice catch

lossyrob:
:+1:

PR #1417, opened 2016-03-22
Fix saveToS3 path template parsing 

echeipesh:
Fixes incorrect parsing of URI template when saving to s3, adds tests to cover the failures.

Also changes `SaveToS3Methods` object to `SaveToS3` to better describe what it does. Extension method `.saveToS3()`  remains unchanged.

### DIFF
--- a/s3/src/main/scala/geotrellis/spark/io/s3/package.scala
+++ b/s3/src/main/scala/geotrellis/spark/io/s3/package.scala
@@ -7,14 +7,5 @@ package object s3 {
   def makePath(chunks: String*) =
     chunks.filter(_.nonEmpty).mkString("/")
 
-  implicit class withSaveToS3Methods[K](rdd: RDD[(K, Array[Byte])]) {
-    /**
-      * Saves each RDD value to an S3 key.
-      *
-      * @param keyToPath A function from K (a key) to an S3 URI
-      */
-    def saveToS3(keyToPath: K => String): Unit = {
-      SaveToS3(keyToPath, rdd, { () => S3Client.default })
-    }
-  }
+  implicit class withSaveToS3Methods[K](rdd: RDD[(K, Array[Byte])]) extends SaveToS3Methods(rdd)
 }
--- a/s3/src/main/scala/geotrellis/spark/io/s3/package.scala
+++ b/s3/src/main/scala/geotrellis/spark/io/s3/package.scala
@@ -7,5 +7,14 @@ package object s3 {
   def makePath(chunks: String*) =
     chunks.filter(_.nonEmpty).mkString("/")
 
-  implicit class withSaveToS3Methods[K](rdd: RDD[(K, Array[Byte])]) extends SaveToS3Methods[K](rdd)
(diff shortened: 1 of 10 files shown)


echeipesh:
@longcmu This should cover the problem you ran into.

s3/src/main/scala/geotrellis/spark/io/s3/package.scala
```diff
@@ -7,5 +7,14 @@ package object s3 {
   def makePath(chunks: String*) =
     chunks.filter(_.nonEmpty).mkString("/")
 
-  implicit class withSaveToS3Methods[K](rdd: RDD[(K, Array[Byte])]) extends SaveToS3Methods[K](rdd)
+  implicit class withSaveToS3Methods[K](rdd: RDD[(K, Array[Byte])]) {
```
lossyrob: Why move this into the package? The pattern has been to have the object, abstract class, and the Implicits with the `with` prefix. Which lets you reference a `? => SaveToS3Methods` if you need it.

echeipesh: Good catch, wasn't thinking about that. Not sure the `Implicits` is appropriate here though. This is a separate project this has to be imported by name and nothing can extend it to bring it in.

lossyrob: Yeah, I think how you have it, where we just add the `with` implicit class to the package object of the specific backend, makes the most sense.

echeipesh:
Looking at related classes and found these issues:
[ ] There is no `setupSaveToS3` as there is to Hadoop, this has been useful it should be mirrored.
[ ] `SaveToHadoopMethods` needs separated from `SaveToHadoop`
[ ] `keyToPath` should be curried so it can given in function body form
[ ] `SaveToHadoop` should not need `scheme` as an argument, we should be able to manage this on per partition/key basis.

This is mostly code juggling that makes the API more nicer and consistent.

longcmu:
It works.
The only thing need to be changed is in SpatialRenderOutput.scala:
[error] geotrellis-master/spark-etl/src/main/scala/geotrellis/spark/etl/hadoop/SpatialRenderOutput.scala:62: not found: value SaveToS3Methods
[error]       val keyToPath = SaveToS3Methods.spatialKeyToPath(id, props("path"))
[error]                       ^
[error] one error found

echeipesh:
Backing out from the idea that `keyToUri` parameter needs to be curried, for a single argument function it does not give marked improvements to compensate for non-standard call.

On other hand `getBytes` in SaveToHadoop benefits significantly from being in second parameter list because scala has trouble inferring it's arguments otherwise and requires messy annotations.

Also `saveToS3` does not currently have `setup` methods because it uses scalaz streams to push the iterator records in parallel. It's not immediately obvious to me how to change that without also going back to redoing a lot of benchmarking so I am punting on this task for me.

lossyrob:
@echeipesh is this ready for merge? I see unchecked checkboxes in a previous comment.

echeipesh:
It's ready, bailed on those tasks as described in last comment. Not sure if I can strike out the check box

PR #1418, opened 2016-03-23
Add Conus Albers CRS

jamesmcclain:
This change adds the Conus Albers CRS to the roster of available ones.

### DIFF
--- a/proj4/src/main/scala/geotrellis/proj4/ConusAlbers.scala
+++ b/proj4/src/main/scala/geotrellis/proj4/ConusAlbers.scala
@@ -1,0 +1,7 @@
+package geotrellis.proj4
+
+object ConusAlbers extends CRS {
+  lazy val proj4jCrs = factory.createFromName("EPSG:5070")
+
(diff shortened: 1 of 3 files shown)


proj4/src/main/scala/geotrellis/proj4/ConusAlbers.scala
```diff
@@ -0,0 +1,7 @@
+package geotrellis.proj4
+
+object ConusAlbers extends CRS {
+  lazy val proj4jCrs = factory.createFromName("EPSG:5070")
+
+  def epsgCode: Option[Int] = CRS.getEPSGCode(toProj4String + " <>")
```
lossyrob: This is a weird way to call this...I know this comes from the other defined CRS's, but shouldn't they all be something like:

```scala
def epsgCode: Option[Int] = Some(5070)
```

Can you change this and the others to do that?

jamesmcclain: Sure, will-do

PR #1421, opened 2016-03-24
Improve proj4 performance

dwins:
Add NTV1 file parsing and a grid file in this format. Also, regenerate the test cases, since this has not been done since I added the tolerance heuristic based on the units used.


proj4/src/main/java/org/osgeo/proj4j/datum/Datum.java
```diff
@@ -71,20 +72,22 @@
   public static final Datum NZGD49 = new Datum("nzgd49", 59.47,-5.04,187.44,0.47,-0.1,1.024,-4.5993, Ellipsoid.INTERNATIONAL, "New Zealand Geodetic Datum 1949");
   public static final Datum OSEB36 = new Datum("OSGB36", 446.448,-125.157,542.060,0.1502,0.2470,0.8421,-20.4894, Ellipsoid.AIRY, "Airy 1830");
 
+  static {
+      Datum temp = null;
+      try {
+          temp = new Datum("NAD27", Grid.fromNadGrids("@conus,@alaska,@ntv2_0.gsb,@ntv1_can.dat"), Ellipsoid.CLARKE_1866,"North_American_Datum_1927");
+      } catch (IOException e) {
+          // TODO: Logging
```
lossyrob: Should we really be swallowing here and setting NAD27 to NULL? When would this throw?

dwins: It should be pretty rare to see exceptions here - the throwing expression is the call to `Grid.fromNadGrids` which can throw if there are IO errors while reading resources from the class path.  So the things that would prompt `IOException` here are along the lines of hard disks failing during the loading of an app using proj4j.  I think an invalid file could also prompt `IOException` but since we only load from the classpath I think these should be rarely encountered as well.

Some other options would be:

* making NAD27 an identity transform in the case of loading failures (which is roughly what happens with an empty grid list anyway)
* a lazy loading scheme so that if exceptions are thrown, it wouldn't be during classloading.  Since this is in a static block the exception would simply prevent the class from being loaded and not get propagated to where a user can see it.  However, I am hesitant to make this change at this point in the release process.

dwins:
I set it to default to a `Datum` with an empty grid list if an `IOException` is thrown, rather than null.  I think this should be pretty safe.  Please let me know if you'd prefer a different approach.

lossyrob:
:+1:

PR #1423, opened 2016-03-24
Add addSplits Function into AccumuloUtils

jamesmcclain:
A function called `addSplits` has been added to the `AccumuloUtils` object.  This function splits a given Accumulo table into a given number of tablets to improve ingest performance.

   - [ ] This is a necessary but not sufficient chance for optimal performance
   - [ ] Some additional notes will be added to this pull request comment


lossyrob:
Merging this in; if there are additional notes, the could perhaps be written into an Accumulo Optimization issue.

jamesmcclain:
> Merging this in; if there are additional notes, the could perhaps be written into an Accumulo Optimization issue.

Okay, will-do.

PR #1429, opened 2016-03-28
Add TileFeature Type

jamesmcclain:
In this pull request the `TileFeature[T, D]` type is added.

### Points to Examine ###

   - ~~The `TileFeature[T <: CellGrid,D: ? => BundleMethods[D]]` type takes two type parameters, but the `merge` methods (in both `raster` and `spark` assume that `T <: Tile`.  There are no assumptions on `D` other than it being a member of the [bundle](https://en.wikipedia.org/wiki/Fiber_bundle) type class.~~
   - ~~There is a question of range and precision of the data D.  Right now they are treated as integer entities.  Double versions are possible as well, but since one of the supported types of `D` is `Seq[Tile]`, there is a question of how (best) to handle the situation when the data are mixed integers and floats.~~
   - ~~I found that I had to add [these](https://github.com/jamesmcclain/geotrellis/blob/feature/ordered-tile/raster/src/main/scala/geotrellis/raster/bundle/Implicits.scala#L8-L17) secondary conversions from subclasses of `BundleMethods[D]` to `BundleMethods[D]` in order to get the constraints on `TileFeatureMergeMethods` to be satisfied.  The issue seems to be that the Scala compiler does not look more than one layer deep when trying to satisfy these constraints.  If there is a better way to handle this, I am open to it.~~

### Still Need To ###

   - ~~Provide Bundle Methods for `Seq[Tile]`~~
   - [x] Provide Appropriate I/O support for `TileFeature`s
   -  ~~Provide Merge Methods for RDDs of `TileFeature`s~~
   - ~~Move bundle methods to their own directory~~
   - [x] Write ScalaDocs
   - [x] Integrate Tests into `PersistenceSpec`


spark/src/test/scala/geotrellis/spark/io/LayerUpdateSpaceTimeTileFeatureTests.scala
```diff
@@ -0,0 +1,113 @@
...
+import geotrellis.spark.testkit._
+import geotrellis.spark.tiling._
+import geotrellis.spark.util._
+import geotrellis.util._
+import geotrellis.vector._
+
+import org.apache.spark.rdd._
+import org.joda.time.DateTime
+
+trait LayerUpdateSpaceTimeTileFeatureTests
```
lossyrob: FunSpec, should be named `Spec` instead of `Tests`. `Tests` represents `FunSuite` tests.

spark/src/test/scala/geotrellis/spark/io/file/FileTileFeatureSpaceTimeSpec.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import com.github.nscala_time.time.Imports._
+
+
+class FileTileFeatureSpaceTimeSpec
+    extends PersistenceSpec[SpaceTimeKey, TileFeature[Tile, Tile], TileLayerMetadata[SpaceTimeKey]]
+    with SpaceTimeKeyIndexMethods
+    with TestEnvironment
+    with TestTileFeatureFiles
+//    with CoordinateSpaceTimeTests
```
lossyrob: Why commented out?

spark/src/test/scala/geotrellis/spark/io/hadoop/HadoopTileFeatureSpatialSpec.scala
```diff
@@ -0,0 +1,45 @@
...
+  describe("HDFS layer names") {
+    it("should handle layer names with spaces") {
+      val layer = AllOnesTestFile
+      val layerId = LayerId("Some layer", 10)
+
+      writer.write[SpatialKey, TileFeature[Tile, Tile], TileLayerMetadata[SpatialKey]](layerId, layer, ZCurveKeyIndexMethod)
+      val backin = reader.read[SpatialKey, Tile, TileLayerMetadata[SpatialKey]](layerId)
+    }
+
+    it("should fail gracefully with colon in name") {
```
lossyrob: Was this a fix/test that your shoehorning in? Or is this coming from another commit? Bit confused about this one.

lossyrob: Oh I see, this was copied from the non-feature tile reader. I don't think we have to test this twice.

lossyrob:
+1 after comments addressed

spark/src/test/scala/geotrellis/spark/io/LayerUpdateSpaceTimeTileFeatureTests.scala
jamesmcclain: I named this one after the file `.../spark/io/AllOnesTestTileTests.scala` from which it was copied (same is true for `.../spark/io/CoordinateSpaceTimeTileFeatureTests.scala` and `.../spark/io/LayerUpdateSpaceTimeTileFeatureTests.scala`).  I can change the names of one or both of these sets of tests, if desired.

spark/src/test/scala/geotrellis/spark/io/file/FileTileFeatureSpaceTimeSpec.scala
jamesmcclain: This has been un-commented.  I went ahead and did the same in `FileSpaceTimeSpec`, as well.

jamesmcclain:
All comments have been addressed with the exception of the `LayerUpdateSpaceTimeTileFeatureTests.scala` naming issue which is discussed above.

jamesmcclain:
> All comments have been addressed with the exception of the LayerUpdateSpaceTimeTileFeatureTests.scala naming issue which is discussed above.

Just to amplify, I am curious to know whether I should just change the names of the files that I added, or also change the name sof the files that they are based on.

lossyrob:
The latter choice, change all the incorrect ones. Thanks.

jamesmcclain:
Okay, will-do.

jamesmcclain:
All comments prior to this one have been addressed.

PR #1430, opened 2016-03-29
Define target cells for focal operations

lokifacio:
Focal operations may be applied only to data/no data cells or both of them.


echeipesh:
This is the way I'm understanding the use case for this functionality, please correct me if I'm wrong:

When applying focal operations on rasters that have NoData regions (ex: continents with NoData for water) the resulting focal raster is going to be "blurred" in so far that a NoData cell in the source raster may have a value in the target raster if it has some values in the focal neighborhood. So the main concern here is to prevent this blurring effect and maintain the NoData regions?

Is there a clear motivation to output a focal operation only for the NoData cells or is that the logical completeness?

lokifacio:
I find it useful for some cases where you have some cells with missing data and you want to complete them with surrounding data but without blurring already available data.

I also found this functionality in other raster packages, i.e. R, so I tried to mimic that behavior.

lossyrob:
@lokifacio can you point to the R package that has this feature, and any other places you've seen it? TBH I'm cautious about adding a parameter to all focal operations for something I haven't encountered before, could be a case of putting something that is not generally useful that pervades through a lot of the API, but that's just me being cautious and doing do diligence. 

You could solve these types of problems by doing a focal operation and then combining it with the values of the original tile, so really this would be a feature for performance gain in these situations. For an example of how you'd solve this without the feature:

```scala
// Fill in NoData holes with mean value
val tile: Tile = ???
tile.focalMean(Square(1)).combine(tile) { (focalValue, originalValue) =>
  if(isNoData(originalValue)) { focalValue } else { originalValue }
}
```

Given that this feature is then a performance optimization for specific use cases when calling focal operations, and that the feature generally effects the focal operations API (By adding to the parameter list of each focal operation call), I'd want to be confident that the use case justifies the code change. Seeing how this option is implemented in other software would be a good way to justify it's inclusion.

@rcheetham you might have some good input here; is there a wide use for an option in focal operations that determines which focal values to operate over? e.g. Run the focal operation, but pass through any value that is no NoData (to fill missing values), or only consider focal cells that are not NoData, otherwise pass the NoData through.

lokifacio:
@lossyrob this is the focal doc of R raster package http://www.inside-r.org/packages/cran/raster/docs/focal

They only allow to target All/NoData but I think it could be also useful to target Data (which can be useful to prevent adding cells which weren't on the original tile) or even cells with specific criteria.

I was aware about the alternative procedure to apply focal and then combine with original raster (though I didn't know about geotrellis combine which is nice btw), but I thought it would be nice to have such option for both performance and also to ease this task to other users.

dwins:
I've been looking over this PR since it is showing some changes that seem unrelated to the proposed functionality (some of the diffs leading up to 0.10 are somehow showing up on @lokifacio 's branch.)  I made a new branch that I hope reflects the intended changes without the extraneous diffs: https://github.com/dwins/geotrellis/tree/feature-focal-calculation-targets 

It's my intent that this will make it a little easier to discuss the change - if it would be easier I can also issue a new PR from my branch.

lossyrob:
Superseded by #1506. We need to move the unit tests from this PR into that branch before merging #1506 or a PR that supersedes it.

PR #1433, opened 2016-04-01
rdd.collectMetadata[K] covers all cases

echeipesh:
This patch includes a fix to API, `TileLayerMetadata.fomRdd[K1, V, K](layoutDefinition)` now returns `TileLayerMetadata[K]` instead of `(Int, TileLayerMetadata[K])`. This is both consistent with the other invocation and more correct since layoutDefinition bypasses the zoom level generation process.

Also `rdd.collectMetadata[K](layoutScheme)` and `rdd.collectMetadata[K](layoutDefinition)` are added to cover those cases when `crs` must be collected.


lossyrob:
@echeipesh what happened to your git email? Those are ghost commits. can you amend?

PR #1437, opened 2016-04-05
Remove constructor type parameters from tile readers

echeipesh:
This change makes tile readers consistent with layer readers in that final types need to be provided to the read method rather than in the constructor. 

This also means the tile readers may no longer extent `Reader[LayerId, Reader[K, V]]` but instead get to extend `TilerReader[LayerId]` trait. Again this is consistent with layer readers.

### DIFF
--- a/accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloTileReader.scala
+++ b/accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloTileReader.scala
@@ -23,7 +23,7 @@ class AccumuloTileReader(
 
   val rowId = (index: Long) => new Text(AccumuloKeyEncoder.long2Bytes(index))
 
-  def read[K: AvroRecordCodec: JsonFormat: ClassTag, V: AvroRecordCodec](layerId: LayerId): Reader[K, V] = new Reader[K, V] {
+  def reader[K: AvroRecordCodec: JsonFormat: ClassTag, V: AvroRecordCodec](layerId: LayerId): Reader[K, V] = new Reader[K, V] {
     val header = attributeStore.readHeader[AccumuloLayerHeader](layerId)
     val keyIndex = attributeStore.readKeyIndex[K](layerId)
     val writerSchema = attributeStore.readSchema(layerId)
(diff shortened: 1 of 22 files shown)


lossyrob:
@echeipesh To be consistent with readers, we should have a `reader` method that actually returns the readers for a given parameter set: https://github.com/geotrellis/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/io/LayerReader.scala#L29

echeipesh:
hmm, so that means there is no such method as `TileReader.read` only `TileReader.reader`, since currently `read` already acts as a `reader` would.

lossyrob:
@echeipesh wouldn't read then just be the shortcut to give a LayerId and K to get the tile? That would be in line with the layer readers right

echeipesh:
I thought about that, but the overhead of setting up the reader for single tile is much higher vs reading RDD. For Hadoop it would be down right horrible. Do you think it is worth providing such a dangerous method, from performance perspective, for completeness?

Having said that I did not bench the overhead.

lossyrob:
Yeah, I guess that makes sense.

This API is super confusing. Can you document it, and provide an example for tile reading in a markdown doc?

lossyrob:
@echeipesh ping on this, would like to merge this into master tomorrow for RC4

echeipesh:
update forthcoming, writing a couple of words on the readers/writers to put the tile readers into context.

echeipesh:
I want to draw your attention to this line: https://github.com/geotrellis/geotrellis/pull/1437/commits/21975f14084f6ee9dce1147729cf9ec9d45cb576#diff-95bb7594c81afabd7e07efe2903bf3ccR28

AFAIK this is the rationale we used for this feature and it makes sense. But having written that the word that jumps out at me is "unhelpful".

echeipesh:
Another issue I've just realized is the name `TileReader`, in two ways:
1. We've been careful to not call anything `Tile`. So really it's a `KeyValueReader`
2. As you've pointed out this is a reader that gives a reader, so maybe a provider?

As such the following rename begs to be made: `KeyValueReaderProvider`

PR #1438, opened 2016-04-05
0.10 cleanup

lossyrob:
See commits for description.


raster/src/main/scala/geotrellis/raster/histogram/StreamingHistogram.scala
```diff
@@ -254,23 +265,28 @@ class StreamingHistogram(
...
+      val lo = _buckets.lowerEntry(item)
+      val hi = _buckets.higherEntry(item)
+
       val raw = {
-        if (lo == null && hi == null) 0
-        else if (lo == null) {
+        if (lo == null && hi == null) {
+          // If we somehow landed right on a bucket value,
+          // the interpolated value is the count of that bucket.
+          if(_buckets.containsKey(item)) _buckets.get(item)
```
lossyrob: @jamesmcclain In the case we hit the bucket, we return the bucket value.

raster/src/main/scala/geotrellis/raster/histogram/StreamingHistogram.scala
```diff
@@ -462,64 +478,81 @@ class StreamingHistogram(
...
+          else if (q == 1.0) { result += maxValue().getOrElse(Double.PositiveInfinity) }
+          else {
+            if(q < curr._2._2) {
+              result += getValue(q)
+            } else {
+              while(data.hasNext && curr._2._2 <= q) { curr = data.next }
+              result += getValue(q)
+            }
+          }
+        }
```
lossyrob: The algorithm has changed to use the current val until the quantile passes the current value's quantile.

PR #1444, opened 2016-04-12
Add Support for the GridCoverage2D Type

jamesmcclain:
In this pull request, support is added for converting the GeoTools `GridCoverage2D` to a Geotrellis `Raster` and vice-versa.

   - [x] `GridCoverage2D` to `Raster`
   - [x] Add proper support for NODATA
   - [x] Add proper support for CRSs
   - ~~Confirm Kryo serializability of `GridCoverage2DMultibandTile` and `GridCoverage2DTile`~~ **`GridCoverage2D` is not Kryo Serializable** (See experimental diff below)
   - [x] `Raster` to `GridCoverage2D`
   - [x] Add Tests for `GridCoverage2D` to `Raster`
   - [x] Add Tests for `Raster` to `GridCoverage2D`
   - [x] Add ScalaDocs

### Points of Interest ###
   - When converting form signed byte Geotrellis `Raster`s to GeoTools `GridCoverage2D`s, the NODATA value must be non-negative.  This is as a result of what seems to be a bug in Java's `SamleDimension` implementation.
   - Due to limitations on banked rasters and interleaved rasters, conversion form a float- or double-valued Geotrellis `MultibandTile` is not supported. I suggest converting the individual bands to separate `GridCoverage2D` objects.

### To Do (2) ###
   - [x] Take a pass over the ScalaDocs to make sure that linkages are legal

### To Do (1) ###
   - [x] NODATA Concerns
   - [x] Thread Safety
   - [x] ColorModel, SampleDimension correctness

### More ###

The following diff can be applied to allow experimentation with `GridCoverage2D` vis-a-vis Spark
```diff
diff --git a/build.sbt b/build.sbt
index 05cfc79..dce90e1 100644
--- a/build.sbt
+++ b/build.sbt
@@ -131,7 +131,7 @@ lazy val slick = Project("slick", file("slick")).
   settings(commonSettings: _*)
 
 lazy val spark = Project("spark", file("spark")).
-  dependsOn(util, raster).
+  dependsOn(util, raster, geotools).
   settings(commonSettings: _*)
 
 lazy val sparkTestkit: Project = Project("spark-testkit", file("spark-testkit")).
diff --git a/spark/src/test/scala/geotrellis/spark/buffer/BufferTilesSpec.scala b/spark/src/test/scala/geotrellis/spark/buffer/BufferTilesSpec.scala
index 426dc80..d72b4bc 100644
--- a/spark/src/test/scala/geotrellis/spark/buffer/BufferTilesSpec.scala
+++ b/spark/src/test/scala/geotrellis/spark/buffer/BufferTilesSpec.scala
@@ -19,10 +19,51 @@ package geotrellis.spark.buffer
 import geotrellis.spark._
 import geotrellis.spark.io._
 import geotrellis.raster.io.geotiff.SinglebandGeoTiff
+import geotrellis.geotools._
 
+import org.geotools.coverage.grid._
+import org.geotools.coverage.grid.io._
+import org.geotools.gce.geotiff._
 import org.scalatest.FunSpec
 
 
+class ExperimentSpec extends FunSpec with TestEnvironment {
+  def getImage(path: String): GridCoverage2D = {
+    val policy = AbstractGridFormat.OVERVIEW_POLICY.createValue
+    val gridSize = AbstractGridFormat.SUGGESTED_TILE_SIZE.createValue
+    val useJaiRead = AbstractGridFormat.USE_JAI_IMAGEREAD.createValue
+    val file = new java.io.File(path)
+    val reader = new GeoTiffReader(file)
+
+    policy.setValue(OverviewPolicy.IGNORE)
+    gridSize.setValue("1024,1024")
+    useJaiRead.setValue(true)
+
+    val image = reader.read(List(policy, gridSize, useJaiRead).toArray)
+    val renderedImage = image.getRenderedImage
+
+    require(renderedImage.getHeight <= 1024)
+    require(renderedImage.getWidth <= 1024)
+
+    image
+  }
+
+  describe("experiment") {
+    it("should") {
+      val list = List(
+        "./raster-test/data/geotiff-test-files/nex-pr-tile.tif",
+        "./raster-test/data/geotiff-test-files/3bands/uint32/3bands-tiled-band.tif")
+        .map({ path => GridCoverage2DToRaster(getImage(path)) })
+        .map({ raster => raster.tile.map({ (_, z) => z }) })
+        // .map({ path => getImage(path) })
+      val rdd = sc.parallelize(list)
+      val bands = rdd.map(_.bandCount).collect()
+      // val bands = rdd.map(_.getRenderedImage.getSampleModel.getNumBands).collect
+      bands should be (List(1,3))
+    }
+  }
+}
+
 class BufferTilesSpec extends FunSpec with TestEnvironment {
 
   describe("The BufferTiles functionality") {
```


echeipesh:
I would like to talk through the NoData handling. The way this reads is that each band in `GridCoverage2D` may have its own NoData value and in fact its own cell type. Is this common from what you've seen so far?

So then we have a problem if we try to map such a tile to a raster, which NoData value to pick? The way `MultibandTile` is specified our tiles have a single cell type and single nodata value. 

We've had no end of trouble with incorrect interpretation of NoData values, so I don't think we should fudge it and pick the first one as `GridCoverage2DToRaster.noData` seems to do. 
If more correct return type for `GridCoverage2DToRaster.apply(gridCoverage: GridCoverage2D)` would be is `Array[Raster[Tile]]` I think that's fine. Converting  `GridCoverage2D` -> `Raster[MultibandTile]` with mismatched cellTypes and NoData fields should be an error.

What are your thoughts on this?

geotools/src/main/scala/geotrellis/geotools/GridCoverage2DToRaster.scala
```diff
@@ -0,0 +1,113 @@
...
+        case DataBuffer.TYPE_SHORT => ShortCellType
+        case DataBuffer.TYPE_INT => IntCellType
+        case DataBuffer.TYPE_FLOAT => FloatCellType
+        case DataBuffer.TYPE_DOUBLE => DoubleCellType
+        case _ => throw new Exception("Unknown CellType")
+      }
+    }
+    else {
+      val noData = noDataValue.get
+      typeEnum match {
```
echeipesh: What if the `noDataValue` matches our convention? Can we ever recover tile with `DoubleConstantNoDataCellType` for instance?

geotools/src/main/scala/geotrellis/geotools/GridCoverage2DMultibandTile.scala
```diff
@@ -0,0 +1,336 @@
...
+  * @param  gridCoverage2D  The GeoTools GridCoverage2D object
+  */
+class GridCoverage2DMultibandTile(gridCoverage: GridCoverage2D)
+    extends MultibandTile with MacroMultibandCombiners {
+
+  val renderedImage = gridCoverage.getRenderedImage
+  val buffer = renderedImage.getData.getDataBuffer
+  val sampleModel = renderedImage.getSampleModel
+
+  private val _array = Array.ofDim[Int](sampleModel.getNumBands)
```
echeipesh: Just a comment: This pretty much makes this not thread safe. We don't really write much multi-threaded code. Spark mostly sweeps things around so we have one-data per on-thread thingy. Not sure if `scala.util.DynamicVariable` would be justified here or even if `GridCoverage2D` is thread safe in the first place since they may fall back on the same pattern.

jamesmcclain:
> What are your thoughts on this?

Sorry for not responding earlier.  I am now working on a task that will need this code, so I will have a chance to do some more work with this.

geotools/src/main/scala/geotrellis/geotools/GridCoverage2DMultibandTile.scala
lossyrob: Best way to handle this is to do a `clone` in each function, which is faster than creating a new array. So keep these around, but `getFiber` should clone the array and return the cloned version.

jamesmcclain: Okay, that makes sense.

jamesmcclain:
> I would like to talk through the NoData handling. The way this reads is that each band in GridCoverage2D may have its own NoData value and in fact its own cell type. Is this common from what you've seen so far?

I don't know how common different permutations are, but I agree that the closest Geotrellis analogue is an array of tiles.

jamesmcclain:

geotools/src/main/scala/geotrellis/geotools/GridCoverage2DToRaster.scala
```diff
@@ -0,0 +1,108 @@
...
+import org.osgeo.proj4j._
+
+import scala.collection.JavaConverters._
+import scala.math.{min, max}
+
+
+/**
+  * The [[GridCoverage2DRaster]] object, contains a constructor, as
+  * well as other functions for converting a [[GridCoverage2D]] to a
+  * Geotrellis [[Raster]].
```
lossyrob: I've noticed you do a lot of linking between things in Scaladocs. This is good, but there is one limitation that you should be aware of: you can't do this sort of linking between classes in different projects. For instance, look at the docs for `Rasterizer.foreachCellByMultiPoint`. In code there is `[[MultiPoint]]`, but if you look at the scaladocs we don't get the linkage. `[[RasterExtent]]` is linked however.

It's my assumption that that is how the functionality works (only types inside the same project link). It would be worth it to look into how that scaladoc functionality works since you use it a lot...also because it dumps a lot of compiler warnings when the linkage can't be found.

lossyrob:
*runs tests*

```
[info] Total number of tests run: 274
```

geotools/src/main/scala/geotrellis/geotools/GridCoverage2DToRaster.scala
```diff
@@ -0,0 +1,108 @@
...
+object GridCoverage2DToRaster {
+
+  /**
+    * This constructor takes a [[GridCoverage2D]] and returns a
+    * sequence of Geotrellis [[Raster]]s.
+    *
+    * @param  gridCoverage  The  GeoTools GridCoverage2D object
+    * @return               A sequence of rasters of singleband Geotrellis [[Tile]]s
+    */
+  def apply(gridCoverage: GridCoverage2D): Seq[Raster[Tile]] = {
```
lossyrob: Is there a reason why this should be `Seq[Raster[Tile]]` and not a `Raster[MultibandTile]`? Is it because the cell types can be different?

geotools/src/main/scala/geotrellis/geotools/GridCoverage2DToRaster.scala
jamesmcclain: Okay, I will review the ScalaDocs docs and go back through the comments.

geotools/src/main/scala/geotrellis/geotools/GridCoverage2DToRaster.scala
jamesmcclain: Yeah, that is the reason.

jamesmcclain:
> [info] Total number of tests run: 274

Is 274 a special number?

lossyrob:
> Is 274 a special number?

It's just a good (large) number

lossyrob:
The GridCoverage2DTile does not handle NoData correctly. For instance, getDouble does not check to see if the pixel value corresponds to the "NoData" value saved off with the GridCoverage2D, and translate that to NODATA as the rest of the tiles do. Also, there was a technique I figured would be more performant, which also solves the NoData issue, by using the arrays of DataBuffers directly; spiking it shows it to be significantly faster:

```scala
class GridCoverage2DConvertersBenchmark extends Benchmarks with ConsoleReport {
  import GridCoverage2DConvertersBenchmark._
  val paths =
    List(
      BenchmarkUtil.dataPath("geotiff/SBN_inc_percap.tif"),
      BenchmarkUtil.dataPath("geotiff/elevation-comp.tif"),
      BenchmarkUtil.dataPath("geotiff/r-nir-wm-clipped.tif")
    )

  for(path <- paths) {
    benchmark(s"Converting a GeoTools GridCoverage2D to a GeoTrellis type: ${new java.io.File(path).getName}") {
      run("Foreach over GeoToolsTile") {
        new Benchmark {
          val p = path
          var gridCoverage: GridCoverage2D = _

          override def setUp(): Unit = {
            gridCoverage = getImage(p)
          }

          def run(): Unit = {
            // The previous way
            val Raster(tile, extent) = GridCoverage2DToRaster(gridCoverage).head
            var s = 0
            tile.foreach { z => if(isData(z)) s += z }
          }
        }
      }

      run("Foreach over GeoToolsTile Refactor") {
        new Benchmark {
          val p = path
          var gridCoverage: GridCoverage2D = _

          override def setUp(): Unit = {
            gridCoverage = getImage(p)
          }

          def run(): Unit = {
            // The refactored way
            val Raster(tile, extent) = gridCoverage.toRaster(0)
            var s = 0
            tile.foreach { z => if(isData(z)) s += z }
          }
        }
      }
    }
  }
}
```

```console
Running benchmarks for Converting a GeoTools GridCoverage2D to a GeoTrellis type: SBN_inc_percap.tif...
  1 of 2: Foreach over GeoToolsTile     54130140.55 ns; σ=57058178.07221 ns @ 10 trials     
  2 of 2: Foreach over GeoToolsTile     4550679.99 ns; σ=4796837.89136 ns @ 10 trials               

  Name                                    ms  linear runtime
  Foreach over GeoToolsTile            54.13  ===========================
  Foreach over GeoToolsTile Refa        4.55  ==
Running benchmarks for Converting a GeoTools GridCoverage2D to a GeoTrellis type: elevation-comp.tif...
  1 of 2: Foreach over GeoToolsTile     36791265.45 ns; σ=38781398.93927 ns @ 10 trials       
  2 of 2: Foreach over GeoToolsTile     3368608.23 ns; σ=3550824.85014 ns @ 10 trials               

  Name                                    ms  linear runtime
  Foreach over GeoToolsTile            36.79  ============================
  Foreach over GeoToolsTile Refa        3.37  ==
Running benchmarks for Converting a GeoTools GridCoverage2D to a GeoTrellis type: r-nir-wm-clipped.tif...
  1 of 2: Foreach over GeoToolsTile     2954836000.00 ns; σ=3114670624.08710 ns @ 10 trials
  2 of 2: Foreach over GeoToolsTile     974585600.00 ns; σ=1027303423.60060 ns @ 10 trials

  Name                                     s  linear runtime
  Foreach over GeoToolsTile             2.95  =========================
  Foreach over GeoToolsTile Refa        0.97  ========
```

These are big changes to this PR, so it would be best to freeze work on this until my changes take place. Let me know if there's any significant difference to the branch I pulled yesterday vs this branch (I see a bunch of commits after my comment, but can't tell if those are actually newly pushed commits or old commits that I have).

lossyrob:
It seems like NoData is handled in GeoTools differently then the GridSampleModel method: from what I have read, it is recommended to use the `CoverageUtilities.getNoDataProperty(image)` to get the NoData value(s) from a GridCoverage2D (and likewise set that property when creating a GridCoverage2D). Is there are reason to not use this method?

http://docs.geotools.org/stable/userguide/tutorial/raster/jaiext.html
https://github.com/geotools/geotools/blob/master/modules/library/coverage/src/main/java/org/geotools/resources/coverage/CoverageUtilities.java#L222

lossyrob:
Superseded by #1502

PR #1448, opened 2016-04-15
Few cleanup items before release.

lossyrob:
- Change the default classBoundryType for color maps to `LessThanOrEqualTo`. The behavior of `LessThan` was incorrect for a while (until RC3 I believe) and exhibited the behavior that LessThanOrEqualTo should have exhibited. This was fixed, but the default needs to change to the correct but previously incorrectly labeled behavior. (thanks @RickMohr for this catch)
- ESRI written EPSG:3857 don't record the code in the usual place; there's a hack in GDAL to account for this, and now we also do it. (Again thanks @RickMohr)
- In the combineValues call, follwed the rule to always use `SparkContext.union` to union a set of RDD's with order greater than 2; do not reduce over them. Also a small performance optimization in the case that only one of the set has a partitioner.
- Since we created `histogramExactInt` to deal with extracting integer value exact count histograms from RDDs, we should have also had a classBreaks method that uses this. Now we do.


echeipesh:
+1

PR #1449, opened 2016-04-18
JsonFormat for Histogram[Double] instead of StreamingHistogram

lossyrob:
This follows how Histogram[Int] behaved.

This way we can serialize the results of operations like `rdd.histogram` directly, without having to do strange casting.


raster/src/main/scala/geotrellis/raster/io/json/HistogramJsonFormats.scala
```diff
@@ -0,0 +1,100 @@
...
+import geotrellis.raster._
+import geotrellis.raster.histogram._
+
+import spray.json._
+
+
+trait HistogramJsonFormats {
+  implicit object HistogramIntFormat extends RootJsonFormat[Histogram[Int]] {
+    def write(h: Histogram[Int]): JsValue = {
+      var pairs: List[JsArray] = Nil
```
echeipesh: Converting from FastMapHistogram this could have a lot elements as I'm reading it, so this should be a Vector/ArrayBuilder.

echeipesh:
+1

PR #1451, opened 2016-04-19
Optimize reading from compressed Bit geotiffs

shiraeeshi:
This pull request is about [this issue](https://github.com/geotrellis/geotrellis/issues/1425)


lossyrob:
Thanks @shiraeeshi, have you tested that this makes the reading faster?

shiraeeshi:
Yes, I've tested reading, on my machine the program reads the compressed file faster than uncompressed, which is weird.

shiraeeshi:
And also I ran the tests from raster-test, all tests passed.

lossyrob:
@shiraeeshi great, thanks. I just tested too and it looks good.

Can you follow the CLA instructions here https://github.com/geotrellis/geotrellis/blob/master/contributing.md ? Otherwise +1 this change

shiraeeshi:
I'll do it today, one little question about eclipse CLA: I've signed it in electronic form, but I should print it anyway, right? It end with an expression 
> You represent that the information provided below is accurate.
I don't know what that mentioned information is. I'm going to double the information from geotrellis CLA, is it correct?
BTW, how to fill the "Point of contact" form field?

lossyrob:
Eclipse CLA only needs to be electronic.
Don't worry about the Point of Contact field, or just put yourself.
The information is in reference to your ownership of the code changes.
Doubling the information is correct; unfortunately we're unclear on whether the Eclipse CLA covers all of our bases, but need it for the move to LocationTech, so we're asking people to fill out both.

Thanks for this change, this was a great catch.

shiraeeshi:
> Eclipse CLA only needs to be electronic.

I should print just one CLA? Then CLA instructions should be modified because it says "print CLAs" in plural form, which is misleading.

lossyrob:
Good point, I'll change the docs. You do not need to print the eclipse CLA

shiraeeshi:
Now it's clear to me, thank you.

shiraeeshi:
I've sent the scanned document by email.

lossyrob:
👍

PR #1452, opened 2016-04-19
Cassandra support

pomadchin:


cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraRDDReader.scala
```diff
@@ -0,0 +1,66 @@
...
+        queryKeyBounds.flatMap(decomposeBounds)
+
+    val bins = IndexRanges.bin(ranges, numPartitions.getOrElse(sc.defaultParallelism))
+
+    val rdd =
+      sc.parallelize(bins, bins.size)
+        .mapPartitions { partition: Iterator[Seq[(Long, Long)]] =>
+          val session = instance.session
+
+          val statement = session.prepare(s"SELECT value FROM ${instance.keySpace}.${table} WHERE key = ?")
```
pomadchin: probably that would be faster to use range queries there, but not sure (the only way to figure it out -- to try)

lossyrob:
Hey @allixender, would you be able to review this? We're also having issues getting the unit tests to run on Travis, any thoughts? Thanks!

allixender:
Hi @lossyrob , @pomadchin ,
sure thing. Now that 0.10.0 is stable it'd be great to look into it again. I might need some time to get up to speed with the setup and the code again. I have seen you've added a lot of documentation on the website, too. Love it.
Umm, where should I start, which branch should I look at? 
Cheers,
Alex

pomadchin:
@allixender First of all you can look [diff](https://github.com/geotrellis/geotrellis/pull/1452/files) of this pr. Second point: we have separate sub projects for [s3](https://github.com/geotrellis/geotrellis/tree/master/s3/src/main/scala/geotrellis/spark/io/s3), [accumulo](https://github.com/geotrellis/geotrellis/tree/master/accumulo/src/main/scala/geotrellis/spark/io/accumulo), and [cassandra](https://github.com/pomadchin/geotrellis/tree/feature/cassandra-nmr/cassandra/src/main/scala/geotrellis/spark/io/cassandra) backends, [hadoop](https://github.com/geotrellis/geotrellis/tree/master/spark/src/main/scala/geotrellis/spark/io/hadoop) and [file](https://github.com/geotrellis/geotrellis/tree/master/spark/src/main/scala/geotrellis/spark/io/file) are included into [spark](https://github.com/geotrellis/geotrellis/tree/master/spark/src/main/scala/geotrellis/spark/io/) sub project. 

What different from your pr: we don't use datastax connector, but we use [datastax java driver](https://github.com/datastax/java-driver). Main differences in [rdd reader](https://github.com/pomadchin/geotrellis/blob/feature/cassandra-nmr/cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraRDDReader.scala) and [rdd writer](https://github.com/pomadchin/geotrellis/blob/feature/cassandra-nmr/cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraRDDWriter.scala)

So possible questions are: 
* is it really fast (we haven't checked it on a huge amount of data right now (think will do it this week / next week)), and what do you think all in all about that approach reading / writing, can it be faster?
* how to test it properly: I tried embedded cassandra and to start cassandra as a service for travis ci. First variant is rally slow, the second is much faster (on a local machine), but ci is out of mem.

By the way, cassandra space time tests are much slower than others, but it's understandable: other backends has mock clients (accumulo / s3), but for cassandra we need a real / embedded cassandra, probably that's one of the reasons why cassandra tests are all in all a bit slower.

P.S. I tried to load LC8 tiles, and moved [chatta demo](https://github.com/pomadchin/geotrellis-chatta-demo/tree/feature/cassandra-support) to use any backend (including cassandra) and there were no significant differences tile read speed / tile ingest (but i hadn't checked exact timings).

cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraLayerHeader.scala
```diff
@@ -0,0 +1,36 @@
+package geotrellis.spark.io.cassandra
+
+import geotrellis.spark.io.LayerHeader
+
+import spray.json._
+
+case class CassandraLayerHeader(
+  keyClass: String,
+  valueClass: String,
+  tileTable: String
```
echeipesh: I think we need to store keyspace here as well. The idea behind the header is that it has enough information to locate the records by itself. AFAIK tables are not unique among keyspaces. Is that right?

cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraRDDReader.scala
```diff
@@ -0,0 +1,84 @@
...
+          instance.withSession { session =>
+            val statement = session.prepare(query)
+
+            val tileSeq: Iterator[Seq[(K, V)]] =
+              for {
+                rangeList <- partition // Unpack the one element of this partition, the rangeList.
+                range <- rangeList
+                index <- range._1 to range._2
+              } yield {
+                // mb to use iterator there?
```
echeipesh: Stale comment? Otherwise what does it mean here?

project/Version.scala
```diff
@@ -30,6 +30,7 @@ object Version {
   val jackson     = "1.6.1"
   val monocle     = "1.2.1"
   val accumulo    = "1.7.1"
+  val cassandra   = "3.4" //"1.2.19"
```
echeipesh: whats the 1.2.19 comment ?

.travis.yml
```diff
@@ -24,6 +24,18 @@ cache:
...
+   - docker
+
+before_install:
+  - cat /etc/hosts # optionally check the content *before*
+  - sudo hostname "$(hostname | cut -c1-63)"
+  - sed -e "s/^\\(127\\.0\\.0\\.1.*\\)/\\1 $(hostname | cut -c1-63)/" /etc/hosts | sudo tee /etc/hosts
+  - cat /etc/hosts # optionally check the content *after*
+
+before_script:
+  - docker run -d --restart=always --net=host -m 1G --memory-swap -1 --env="MAX_HEAP_SIZE=500M" --env="HEAP_NEWSIZE=100M" --env="CASSANDRA_LISTEN_ADDRESS=127.0.0.1" --name dc-cassandra cassandra:latest
```
echeipesh: This should be a script that is referred to in error message for the tests, it can be handled in the same matter as slick tests. Otherwise you have to know to run this command to run the local tests.

cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraLayerHeader.scala
pomadchin: Yep, that's true, that's a good idea to store namespace as a part of header information.

allixender:
Hi @pomadchin ,

apologies for laying low. 

One problem is also that cassandra "sees" all the available memory in the JVM and makes assertion about its own performance settings. However, I have collected hints to down-size cassandra requirements and found that this has been helpful in testing.

https://gist.github.com/allixender/6a9a504ef017482f9760a78b9f9dd70d

Maybe try this as the cassandra bootstrap config in the testing?

Cheers,
Alex

pomadchin:
@allixender will check it, but however the best solution seems to start a real cassandra server: 

```bash
docker run -d --restart=always --net=host -m 1G --memory-swap -1 --env="MAX_HEAP_SIZE=500M" --env="HEAP_NEWSIZE=100M" --env="CASSANDRA_LISTEN_ADDRESS=127.0.0.1" cassandra:latest
```

pomadchin:
Includes fix open jdk 7: https://github.com/travis-ci/travis-ci/issues/5227
This fix yet not works in PR's: https://github.com/travis-ci/travis-ci/issues/5669

pomadchin:
Rollback to an old version and added links to travis issues.

lossyrob:
@pomadchin when I try to run the tests I get these types of errors:

```console
DeferredAbortedSuite:
[info] Exception encountered when attempting to run a suite with class name: org.scalatest.DeferredAbortedSuite *** ABORTED ***
[info]   com.datastax.driver.core.exceptions.NoHostAvailableException: All host(s) tried for query failed (tried: /127.0.0.1:9042 (com.datastax.driver.core.exceptions.TransportException: [/127.0.0.1] Cannot connect))
[info]   at com.datastax.driver.core.ControlConnection.reconnectInternal(ControlConnection.java:231)
[info]   at com.datastax.driver.core.ControlConnection.connect(ControlConnection.java:77)
[info]   at com.datastax.driver.core.Cluster$Manager.init(Cluster.java:1414)
[info]   at com.datastax.driver.core.Cluster.init(Cluster.java:162)
[info]   at com.datastax.driver.core.Cluster.connectAsync(Cluster.java:333)
[info]   at com.datastax.driver.core.Cluster.connectAsync(Cluster.java:308)
[info]   at com.datastax.driver.core.Cluster.connect(Cluster.java:250)
[info]   at geotrellis.spark.io.cassandra.CassandraInstance$class.getSession(CassandraInstance.scala:18)
[info]   at geotrellis.spark.io.cassandra.BaseCassandraInstance.getSession(CassandraInstance.scala:47)
[info]   at geotrellis.spark.io.cassandra.CassandraInstance$class.withSessionDo(CassandraInstance.scala:34)
```

What's with that?

pomadchin:
@lossyrob you just has no running cassandra instance on you local machine https://github.com/pomadchin/geotrellis/blob/dcf5f45ef985939a0a8135926eff48473fd1493e/cassandra/src/test/scala/geotrellis/spark/io/cassandra/CassandraSpaceTimeSpec.scala#L18-L24

Use https://github.com/pomadchin/geotrellis/blob/dcf5f45ef985939a0a8135926eff48473fd1493e/scripts/cassandraTestDB.sh to start cassandra in docker

lossyrob:
@pomadchin can we have the unit tests fail in a more specific way when there is no environment? That informs users how to make them pass?

This is my worry: There is a new user that wants to work with GeoTrellis. As I often do, I say, step 1 is to fork the repo, clone your fork, add upstream, and then run the unit tests.

So they `buildall.sh`, and it all the sudden fails with a sudo-cryptic error message. They freak out and flee to the wilderness.

Is there a way we can capture the scenario of not-having-started-up-cassandra-docker-yet and explain that to users that are simply running all unit tests?

pomadchin:
@lossyrob think yes, we can add special exception type, and to check cassandra availability before all tests run.

lossyrob:
@pomadchin also we need the script to be inside the geotrellis project. It should be a one liner to run the cassandra instance for the unit tests; it should also say in a README.md in the cassandra subproject about why and how to run the cassandra docker container and the unit tests.

pomadchin:
@lossyrob what do you mean by "inside geotrellis project"? [this](https://github.com/pomadchin/geotrellis/blob/feature/cassandra-nmr/scripts/cassandraTestDB.sh) one liner is not enough? Good information about README.md edition.

lossyrob:
@pomadchin nevermind, I thought you had linked to a gist. Disregard the script request.

lossyrob:
What's with the 

```console
00:14:50 DCAwareRoundRobinPolicy: Using data-center name 'datacenter1' for DCAwareRoundRobinPolicy (if this is incorrect, please provide the correct datacenter name with DCAwareRoundRobinPolicy constructor)
```
a bunch of times in the last build failure logs?

pomadchin:
It is related to loadBalancing policy, will add configuration settings, forgot about it. However these annoying messages like would be present, nothing to do with it, it's `logger.info` level: 

```
DCAwareRoundRobinPolicy: Using provided data-center name 'datacenter1' for DCAwareRoundRobinPolicy
```

PR #1454, opened 2016-04-20
Fix/idw interpolation

lokifacio:
- Add double idw interpolation
- Round mean values for int tiles intead of truncate them
- Fix div by 0 which invalidated interpolation at sampling positions


lokifacio:
First commit uses first sampling value at interpolation position. I'm not sure which approach is the proper one (I guess it may vary depending on use case).
I think averaging samples at interpolation position it makes more sense than using its actual value for the general case.

raster/src/main/scala/geotrellis/raster/VectorToRaster.scala
```diff
@@ -129,13 +135,16 @@ object VectorToRaster {
...
+                while (i < length) {
                   val point = pts(i)
                   val dX = (destX - point.geom.x)
                   val dY = (destY - point.geom.y)
                   val d = dX * dX + dY * dY
+                  i += 1
+
                   if (d < rr) {
-                    val w = 1 / d
+                    val w = if (d == 0) w0 else 1 / d
```
echeipesh: This fixes the NoData problem for when we have a point exactly for the co-ordinates we're trying to interpolate but not sure this is totally correct. Probably we should short-circuit the interpolation and return the actual value of the point. As d tends to 0 the weight tends to infinity, so it should dominate any other values. Also currently I think its possible for a point that is slightly off the focus co-ordinates to have more weight than the actual point.

echeipesh:
I think as a project we've become more sensitive to the memory usage of tiles and consequently to always having an ability to specify the narrows cell type. In that light the expanded signature to be:

```scala
 def idwInterpolate(points: Seq[PointFeature[Int]], re: RasterExtent, ct: CellType = IntConstantNoDataCellType): Tile 
```
@lossyrob: Do you have thoughts on this?

Then the implementation can switch on `ct.isFloatingPoint` to the unrolled code, which should be moved to `geotrellis.raster.interpolation.InverseDistanceWeighting` object.

echeipesh:
Also the double code looks correct but there should really be a test that runs through it. Even if it just maps the Int points to Double and then applies the Math.round to make sure they produce the same/close enough results.

raster/src/main/scala/geotrellis/raster/VectorToRaster.scala
lokifacio: That's what I actually did on the first commit, but then I thought than
there are cases where you may have several point values for the same cell
and in that case we would be ignoring other points if there is one which
has coordinates equal to cell center...

In my opinion, which solution is better for you depends on what you are
trying to do with the interpolation.

For that reason I posted the two solutions in separate commits so we can
comment them.

lokifacio: My initial thought was just using the value and short-circuit as IDW definition states. This works fines and I think it is the most efficient way when you have some raster where you want to interpolate missing cells (you'll end up with a single point for each data cell).

On the other hand, there is a case were you have a set of points and you want to create an interpolation raster from them. In that case, each cell has to represent all the points that lie on its region. Thus, if we have N points which lie on the region and none has 0 distance from the center of the cell, we would assign its weighted avg to that cell. However, if only one of those points would have a 0 distance we would ignore the other points. This could be seen like contradiction, don't you think so? I'm no expert on this filed so my opinion can be biased by current use case.

About the `Infinity` problem, I don't know. But we have the same problem in case the point is not a that exact location, haven't we?.

raster/src/main/scala/geotrellis/raster/VectorToRaster.scala
```diff
@@ -84,14 +81,52 @@ object VectorToRaster {
...
     * https://en.wikipedia.org/wiki/Inverse_distance_weighting for
     * more details.
+    *
+    * @param   points  A collection of known-points
+    * @param   re      The study area
+    * @param   radius  Interpolation radius in cells
+    * @return          The data interpolated across the study area
     */
-  def idwInterpolate(points: Seq[PointFeature[Int]], re: RasterExtent, radius: Int): Tile =
+  def idwInterpolate(points: Seq[PointFeature[Int]], re: RasterExtent, radius: Circle): Tile = {
```
echeipesh: This is a convenient overload, but I don't think the math is correct. `Circle(1)` defines a 3x3 cell `Neighborhood`, which means the radius should be 1.5 cell width, to account for half a cell from the center point.

echeipesh:
Ok, it makes sense as an option to this interpolation. We need an argument `minRadius: Double = 0` which defines a minimum radius around the cell center where all points will receive equal weight, as if they were `minRadius` away. In case of it there being a point 0 distance away from the cell center and minRadius == 0 we would use that point. This gives a nice knob to turn, a likely value might be half a cell width.

So I think this makes the target API something like:
```scala
object InverseDistanceWeighting {
    def apply[D](
      points: Seq[PointFeature[D]], 
      re: RasterExtent, 
      radius: Option[Double], 
      minRadius: Double = 0, 
      cellType: CellType = IntConstantNoDataCellType
    )(implicit ev: D => Double): Tile = ???
}
```

Since we're going to be doing Double math on the point features, we might as well encode that in the type. This gives us automatic support for all primitives without overloads as well.

`minRadius` is a strange name, needs good documentation to explain exactly what it does.

raster/src/main/scala/geotrellis/raster/interpolation/InverseDistanceWeighting.scala
```diff
@@ -0,0 +1,139 @@
...
+    * @param   points    A collection of known-points
+    * @param   rasterExtent The study area
+    * @param   radius    Interpolation radius in cell units
+    * @param   equalWeightRadius If any points lie at that distance from a cell, the interpolation result for it would
+    *                            be the mean of those points
+    * @param   cellType  Interpolation radius in coordinate unit
+    * @param   f         Function to be applied before storing interpolation values
+    * @return            The data interpolated across the study area
+    */
+  def closerThan[D](
```
lokifacio: another option could be:
```scala
def withCellRadius[D](points: Seq[PointFeature[D]],
                    rasterExtent: RasterExtent,
                    radius: Circle,
                    equalWeightRadius: Double = 0,
                    cellType: CellType = IntConstantNoDataCellType,
                    f: Double => Double = x => x)
                   (res: Double = rasterExtent.cellheight)
                   (implicit ev: D => Double): Tile = ???
```

raster/src/main/scala/geotrellis/raster/interpolation/InverseDistanceWeighting.scala
```diff
@@ -0,0 +1,139 @@
...
+    * https://en.wikipedia.org/wiki/Inverse_distance_weighting for
+    * more details.
+    *
+    * @param   points    A collection of known-points
+    * @param   rasterExtent The study area
+    * @param   radius    Interpolation radius in coordinate unit
+    * @param   equalWeightRadius If any points lie at that distance from a cell, the interpolation result for it would
+    *                            be the mean of those points
+    * @param   cellType  Interpolation radius in coordinate unit
+    * @param   f         Function to be applied before storing interpolation values
```
echeipesh: To be clear this exists because when setting a double to an int based raster the default behavior is to truncate. I can see why that would not always be desirable. The name should reflect something about when it's used `onSet` should do it.

raster/src/main/scala/geotrellis/raster/interpolation/InverseDistanceWeighting.scala
echeipesh: I like the idea behind this overload but I don't think it belongs in the API. Here are some strikes against it:
 - The units of radius and equalWeightRadius are not the same
 - `Circle` is the only subtype of `Neighborhood` that is admirable here
 - `Circle` isn't actually being used as a cell mask, but is used for it's incidental property of `extent` and even then we have to use max of width or height
 - It's difficult to come up with a specific name for it.

I can imagine there could be a version of IDW where we actually used a cell `Neighborhood`, as in the actual cells mask where you could use instances other than Circle. But it would be a different implementation and it's not clear to me what use case it would really support.

At the moment I would suggest dropping this overload and getting the rest of the changes in, which already include a number of improvements and fixes.

echeipesh:
This is a +1 from me pending the renames in the last set of comments and the CLA.

raster/src/main/scala/geotrellis/raster/interpolation/InverseDistanceWeighting.scala
lokifacio: It seems reasonable to me. I'll drop it.

PR #1457, opened 2016-04-21
Store FastMapHistogram counts as longs

jpolchlo:
FastMapHistogram was storing items and counts in a single array, with element 2\*i holding the item and 2\*i+1 storing the count. Counts could roll over for values above Int.MaxValue. This change splits the items and counts into two arrays, where the latter now stores Long values.


echeipesh:
+1 now with correct author names

PR #1460, opened 2016-04-22
Fixes to user defined float/double CellType parsing

echeipesh:
Handle negative, +/-Infinity, NaN.
Specifically `float64udNaN` will parse as `float64` and `float32udNaN` will parse as `float32`

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/CellType.scala
+++ b/raster/src/main/scala/geotrellis/raster/CellType.scala
@@ -316,15 +316,21 @@ object CellType {
       }
       IntUserDefinedNoDataCellType(ndVal.toInt)
     case ct if ct.startsWith("float32ud") =>
-      val ndVal = new Regex("\\d*.?\\d+$").findFirstIn(ct).getOrElse {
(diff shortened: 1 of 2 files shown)


raster/src/main/scala/geotrellis/raster/CellType.scala
```diff
@@ -316,15 +316,21 @@ object CellType {
       }
       IntUserDefinedNoDataCellType(ndVal.toInt)
     case ct if ct.startsWith("float32ud") =>
-      val ndVal = new Regex("\\d*.?\\d+$").findFirstIn(ct).getOrElse {
-        throw new IllegalArgumentException(s"Cell type $name is not supported")
+      try {
+        val ndVal = ct.stripPrefix("float32ud").toDouble.toFloat
```
echeipesh: This is to keep in line with this approach:
https://github.com/geotrellis/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/io/geotiff/writer/TiffTagFieldValue.scala#L55

PR #1461, opened 2016-04-22
Pass resampling method argument to merge in CutTiles.

lossyrob:
We weren't passing the resample method to the merge call in CutTiles, which was an oversight. This fixes it. Reprojecting still forces NearestNeighbor for the `tileToLayout` call when using a FloatingLayoutScheme, since the cells will line up and the bilinear computation shouldn't happen because of potential floating point errors or NoData overriding behavior.


lossyrob:
Thanks to @owcm for finding this one.

echeipesh:
+1

PR #1464, opened 2016-04-29
Documentation/raster quickstart

jpolchlo:
This branch creates a new subtree of docs that intends to give a useful introduction for new geotrellis users.  This first pass describes a simple example of building a kernel density estimate in both a single tile and a subdivided tile.  A future update will go over the same function within a spark context.


docs/quickstart/README.md
```diff
@@ -0,0 +1,155 @@
...
+      val p = ptf.geom
+      Extent(p.x - krnwdth * ld.cellwidth / 2,
+             p.y - krnwdth * ld.cellheight / 2,
+             p.x + krnwdth * ld.cellwidth / 2,
+             p.y + krnwdth * ld.cellheight / 2)
+    }
+    val ptfToExtent = { p: PointFeature[Double] => pointFeatureToExtent(9, ld, p ) }
+```
+
+The reason for thinking about points as extents is that any point near enough to a tile boundary will need to have its kernel applied to each of the overlapped tiles.  To discover the tiles that a given point's kernel extents overlap, LayoutDefinition provides the `mapTransform` member.  Among the methods of mapTransform is the ability to determine the index of the tile which overlaps a given extent.  Note that the tiles in a given layout are indexed by `SpatialKey`s, which are effectively `(Int,Int)` pairs giving the (column,row) of each tile as follows:
```
echeipesh: This line has inconsistent use use of `var` markings.

echeipesh:
This is good stuff. I like the use of `Map` to approximate `RDD` in code examples. That could probably use an explicit callout in the text. Also some code changes made in with the docs.

Going to wait until https://github.com/geotrellis/geotrellis/pull/1469 is merged to make sure the code comments there are reflected in these docs.

docs/quickstart/README.md
```diff
@@ -0,0 +1,298 @@
+Geotrellis Quickstart Guide
+===========================
+
+Geotrellis is a large and evolving package, which makes it difficult to maintain current documentation.  To make up for this deficiency, this quickstart guide will provide a set of documents describing case studies that accomplish some simple goals.  These examples are meant to highlight the fundamental Geotrellis classes as well as show some basic Scala techniques for those new to the language.
```
lossyrob: Convention is to use line breaks after each sentence, so that diffs/comments can be made per sentence. I see you use double spaces for sentence separation, so this should quick change by just replacing `  ` with a new line.

docs/quickstart/README.md
```diff
@@ -0,0 +1,298 @@
...
+
+Getting Started
+---------------
+
+In the following, we'll be providing some Scala code samples that make use of the GeoTrellis library.  This requires having a running Scala instance with GeoTrellis set as a library dependency.  It is recommended that the user download and install SBT, the Scala Build Tool.  Once that is installed, it is recommended that you use `git` to clone the repository at `github.com/geotrellis/geotrellis-sbt-template repository`.  Once available, from the root directory of the cloned repo, MacOSX and Linux users may launch the sbt script contained therein.  This will start an SBT session.  Once SBT is started, it is advisable to issue the `console` command; this will start the Scala interpreter.  At this point, you should be able to issue the command `import geotrellis.vector._` without raising an error.
```
lossyrob: Confused about `This requires having a running Scala instance with GeoTrellis set as a library dependency`. Does "Scala instance" mean Scala installation? This is actually not necessary.

You go on to recommend installing sbt, but then talk about the sbt script included. If you use the `./sbt` script, you don't have to install SBT.

`Once SBT is started, it is advisable ...` not sure if you're giving advice or giving steps to follow here. One way to make it clearer would be to list out the steps to verify your GeoTrellis clone is working, and bullet point the steps, ending with the import.

docs/quickstart/README.md
```diff
@@ -0,0 +1,298 @@
...
+
+Basic Geotrellis Types
+----------------------
+
+Geotrellis itself is a library for working with raster data over maps.  Raster data is essentially a regular, rectangular grid containing scalar values (i.e., numerical data) in each grid cell.
+
+The base type for these grid structures is `CellGrid` which offers basic dimension information.  The `Tile` trait gives the general interface for working with data over a grid.  Actual usage of Tiles requires an understanding of the NoData handling options defined in `CellType` and described in more detail [elsewhere](../raster/celltype.md) in the documentation.  Generally speaking, though, for Tiles containing integer data types (including bits and bytes), the `IntArrayTile` companion object is used to construct Tiles, while floating point Tiles are created using the `DoubleArrayTile` companion object.
```
lossyrob: `IntArrayTile companion object is used to construct Tiles` should this read, `is used to construct integer value Tiles`?

docs/quickstart/README.md
```diff
@@ -0,0 +1,298 @@
...
+
+A Tile needs additional context if it is to represent data on a map; specifically, it needs an `Extent` which defines the map coordinates of the corners of the Tile.  `Extent(xmin,ymin,xmax,ymax)` represents the bounding box of the Tile *expressed in some coordinate system*.  To apply the Tile to a map, it'll be necessary to define which coordinate system was used, though that will happen at a later stage.  A Tile and Extent come together to form a `Raster.`
```
lossyrob: `To apply the Tile to a map` is a bit confusing, perhaps "In order for the Tile to represent data on a portion of some map`?

docs/quickstart/README.md
```diff
@@ -0,0 +1,298 @@
...
+     val ptextent = ptfToExtent(ptf)
+      val gridBounds = ld.mapTransform(ptextent)
+      for ((c, r) <- gridBounds.coords;
+           if r < tl.totalRows;
+           if c < tl.totalCols) yield (SpatialKey(c,r), ptf)
+    }
+
+    val keyfeatures = pts.flatMap(ptfToSpatialKey).groupBy(_._1).map {
+      case (sk,v) => (sk,v.unzip._2)
+    }: Map[SpatialKey,List[PointFeature[Double]]]
```
lossyrob: Place the `: Map[SpatialKey,List[PointFeature[Double]]]` after the val, i.e.:

```scala
val keyfeatures: Map[SpatialKey,List[PointFeature[Double]]] = 
  pts
    .flatMap(ptfToSpatialKey)
    .groupBy(_._1)
    .map { case (sk,v) => (sk,v.unzip._2) }
```

lossyrob: Also, formatting: spaces after comma:

`case (sk, v) => (sk, v.unzip._2)`

docs/quickstart/README.md
```diff
@@ -0,0 +1,298 @@
...
+
+Our concern falls on how to write code to exploit the structures provided by Spark, specifically *Resilient Distributed Datasets*, or RDDs.  An RDD is a distributed collection, with all of the usual features of a collection---i.e., mapping, folding, reducing.  In order to create an RDD, one will call the `parallelize()` method on a `SparkContext` object, which can be generated by the following set of statements.
```
lossyrob: RDD's actually don't have a fold, which is a key difference (fold requires sequential processing). Instead they have `aggregate`. There actually is a "fold" method, but that's just a reduce with a seed value. I would just maybe change the wording to not set up the user expectation that RDDs have all the features of collections...`with many of the useful features of collections -- e.g. map, reduce -- as well as distributed versions of other types of sequential collection methods -- e.g. aggregate, which is a distributed version of something like foldLeft`.

docs/quickstart/README.md
```diff
@@ -0,0 +1,298 @@
...
+  val conf = new SparkConf().setMaster("local").setAppName("Kernel Density")
+  val sc = new SparkContext(conf)
+```
+
+In our case, we have a collection of PointFeatures that we wish to parallelize, so we issue the command
+```scala
+  import org.apache.spark.rdd.RDD
+  val pointRdd = sc.parallelize(pts, 10)
+```
+Here, the 10 indicates that we want to distribute the data in 10 pieces to the available workers.  Some of those chunks may end up on different physical machines if they are available.  If we exclude this value, the default parallelism will be used, which is typically the number of processors, though in this local example, it defaults to one.
```
lossyrob: pieces -> partitions. Might was well use the verbiage outright. Would be worth it to quickly explain, `a partition is a subset of the data in the RDD that will be processed by one of the workers, where partitions can be processed in parallel over all workers`.

docs/quickstart/README.md
```diff
@@ -0,0 +1,298 @@
...
...much the same way: points will be assigned an extent corresponding to the extent of the associated kernel, those points will be assigned SpatialKeys based on which subtiles their kernels overlap, and each kernel will be applied to the tile corresponding to its assigned SpatialKey.  Earlier, this process was effected by a flatMap followed by a groupBy and then a map.  This very same procedure could be used here, save for the fact that groupBy, when applied to an RDD, triggers an expensive, slow, network-intensive shuffling operation which collects items with the same key on a single node in the cluster.  Instead, a fold-like operation will be used: `aggregateByKey`, which has a signature of <center style="padding-bottom: 1em">`RDD[(K, U)] => T => ((U, T) => T, (T, T) => T) => RDD[(K, T)]`.</center>  That is, we begin with an RDD of key/value pairs, provide a zero value for an accumulator, and two functions: (1) a function for inserting values from the original RDD into the accumulator and (2) a function to merge accumulators.  In our case, `U = PointFeature[Double]` and `T = Tile`; this implies that the insertion function is a kernel stamper and the merging function is a tile adder.
```
lossyrob: I hesitate to use the word "accumulator", since it has a specific meaning in Spark (http://spark.apache.org/docs/latest/programming-guide.html#accumulators). It's usually what it's called in the functional context, but in the ScalaDocs for the aggregate functions they are called "combiners". I actually like calling it an "accumulator", but in this case not to confuse new users, I'd say go with the Spark language and call them combiners:

`provide a zero value for a combiner (the type of instance that will combine and aggregate the values of the RDD), and ...`

docs/quickstart/README.md
```diff
@@ -0,0 +1,298 @@
...
+      .flatMap(ptfToSpatialKey)
+      .mapPartitions({ partition =>
+        partition.map { case (spatialKey, pointFeature) =>
+          (spatialKey, (spatialKey, pointFeature))
+        }
+      }, preservesPartitioning = true)
+      .aggregateByKey(ArrayTile.empty(DoubleCellType, ld.tileCols, ld.tileRows))(stampPointFeature, sumTiles)
+      .mapValues{ tile: MutableArrayTile => tile.asInstanceOf[Tile] }
+```
+The `mapPartitions` operation simply applies a transformation to an RDD without triggering any kind of shuffling operation.  Here, it is necessary to make the SpatialKey available to `stampPointFeature` so that it can properly determine the pixel location in the corresponding tile.
```
lossyrob: It would be good to call out specifically why `asInstanceOf` is used here, since it is considered Bad and we should point it out as Bad but Necessary.

docs/quickstart/README.md
```diff
@@ -0,0 +1,298 @@
...
+                                   ld,
+                                   ld.extent,
+                                   LatLng,
+                                   KeyBounds(SpatialKey(0,0),
+                                             SpatialKey(ld.layoutCols-1,
+                                                        ld.layoutRows-1)))
+```
+To combine the RDD and the metadata, we write `val resultRdd = ContextRDD(tileRdd, metadata)`.
+
+This resulting RDD is essentially the object of interest, though it is possible to write `resultRDD.stitch` to produce a single merged tile.  In the general case, however, the RDD may cover an area so large and in sufficient resolution that the result of stitching would be too large for working memory.  In that case, it would be necessary to write subtiles off to disk and stitch them using some external program like GDAL, or a custom application would need to be written to make use of the tiles that were produced.
```
lossyrob: `In that case, ...`

Instead of mentioning building giant rasters w/ GDAL, we should mention that in that case, we generally save off layers to one of our supported distributed backends (HDFS, S3, Accumulo, etc) and work with the tiles in a distributed way, or pull tiles out individually (for example, when serving on a web map).

docs/quickstart/README.md
```diff
@@ -0,0 +1,298 @@
...
+  assembly
+  project vector
+  assembly
+  project raster
+  assembly
+  project spark
+  set test in assembly := {}
+  set assemblyMergeStrategy in assembly := { case "META-INF/MANIFEST.MF" => MergeStrategy.discard ; case _ => MergeStrategy.first }
+  assembly
+```
```
lossyrob: No nee to run all those assemblies. Just the one will package all that stuff into the one jar:

```
project spark
 set test in assembly := {}
 set assemblyMergeStrategy in assembly := { case "META-INF/MANIFEST.MF" => MergeStrategy.discard ; case _ => MergeStrategy.first }
 assembly
```
and then
```bash
spark-shell --jars spark/target/scala-2.10/geotrellis-spark-assembly-[version].jar
```

The spark assembly will package up all the dependencies, including all the geotrellis subprojects that you listed.

I would recommend putting instead to package up spark-etl, and then you don't have to manually set sbt-assembly properties like merging:

```scala
project spark-etl
assembly
```

```bash
spark-shell --jars spark-etl/target/scala-2.10/geotrellis-spark-etl-assembly-[version].jar
```

raster/src/main/scala/geotrellis/raster/mapalgebra/focal/Kernel.scala
```diff
@@ -89,15 +89,14 @@ object Kernel {
...
-        output.set(c+w, r+w, if (r * r + c * c <= rad2) 1 else 0)
+    var r = 0
+    var c = 0
+    while(r < size) {
+      while(c < size) {
+        output.set(c, r, if (r * r + c * c < rad2) 1 else 0)
         c += 1
       }
-      c = -w
+      c = 0
```
lossyrob: What's going on with this change? Isn't this reverting something you did in an earlier PR?

PR #1466, opened 2016-05-03
Handle Special Characters

jamesmcclain:
`geotrellis/proj4/src/main/java/org/osgeo/proj4j/util/PolarCoordinate.java` line 18 contains both a lambda and a phi (the actual greek characters, not the words), which results in the following compilation error on some systems: "unmappable character for encoding ASCII".

This change prevents that error.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -91,7 +91,8 @@ lazy val vectorTest = Project("vector-test", file("vector-test")).
   dependsOn(vector, vectorTestkit)
 
 lazy val proj4 = Project("proj4", file("proj4")).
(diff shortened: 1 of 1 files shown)


pomadchin:
cool catch; this thing also could be solved by setting default locale UTF-8 on your machine:

```bash
locale-gen en_US 
update-locale LANG=en_US.UTF-8 LC_CTYPE=en_US.UTF-8
```

jamesmcclain:
Yeah, I figured that it was a problem with my locale, but I figured that this would make it easier on people.   :smile:

PR #1468, opened 2016-05-04
Fix nodata values in codecs

shiraeeshi:
There is an inconsistency between encode and decode methods of ArrayTileCodecs for int and byte types: when encoding, it writes zero, but when decoding it expects min value for a type.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/ArrayMultibandTile.scala
+++ b/raster/src/main/scala/geotrellis/raster/ArrayMultibandTile.scala
@@ -435,6 +435,8 @@ class ArrayMultibandTile(_bands: Array[Tile]) extends MultibandTile with MacroMu
     } else other match {
       case that : ArrayMultibandTile =>
         var result = true
+        result = (bandCount == that.bandCount)
+
         cfor(0)(result && _ < bandCount, _ + 1) { i =>
           if (band(i) != that.band(i)) result = false
         }
--- a/raster/src/main/scala/geotrellis/raster/ArrayMultibandTile.scala
+++ b/raster/src/main/scala/geotrellis/raster/ArrayMultibandTile.scala
@@ -429,7 +429,7 @@ class ArrayMultibandTile(_bands: Array[Tile]) extends MultibandTile with MacroMu
     new ArrayMultibandTile(newBands)
   }
 
-  override def equals(other: Any): Boolean = {
+  override def equals(other: Any): Boolean = other match {
     case that : ArrayMultibandTile =>
       var result = (bandCount == that.bandCount)
 
--- a/raster/src/main/scala/geotrellis/raster/ArrayMultibandTile.scala
+++ b/raster/src/main/scala/geotrellis/raster/ArrayMultibandTile.scala
@@ -430,7 +430,7 @@ class ArrayMultibandTile(_bands: Array[Tile]) extends MultibandTile with MacroMu
   }
 
   override def equals(other: Any): Boolean = other match {
-    case that : ArrayMultibandTile =>
+    case that: ArrayMultibandTile =>
       var result = (bandCount == that.bandCount)
 
       cfor(0)(result && _ < bandCount, _ + 1) { i =>
--- a/raster/src/main/scala/geotrellis/raster/ArrayMultibandTile.scala
+++ b/raster/src/main/scala/geotrellis/raster/ArrayMultibandTile.scala
@@ -71,7 +71,7 @@ object ArrayMultibandTile {
 /**
   * The [[ArrayMultibandTile]] type.
   */
-class ArrayMultibandTile(private val _bands: Array[Tile]) extends MultibandTile with MacroMultibandCombiners {
+class ArrayMultibandTile(_bands: Array[Tile]) extends MultibandTile with MacroMultibandCombiners {
   val bandCount = _bands.size
 
   assert(bandCount > 0, "Band count must be greater than 0")
--- a/raster/src/main/scala/geotrellis/raster/ArrayMultibandTile.scala
+++ b/raster/src/main/scala/geotrellis/raster/ArrayMultibandTile.scala
@@ -71,7 +71,7 @@ object ArrayMultibandTile {
 /**
   * The [[ArrayMultibandTile]] type.
   */
-class ArrayMultibandTile(_bands: Array[Tile]) extends MultibandTile with MacroMultibandCombiners {
+class ArrayMultibandTile(private val _bands: Array[Tile]) extends MultibandTile with MacroMultibandCombiners {
   val bandCount = _bands.size
 
   assert(bandCount > 0, "Band count must be greater than 0")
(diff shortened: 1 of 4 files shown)


shiraeeshi:
I just noticed the default values defined in schemas
```
.name("noDataValue").`type`().unionOf().intType().and().nullType().endUnion().intDefault(Int.MinValue)
```
Does it change zero to default value implicily? Probably I need to test it.

shiraeeshi:
How to test it:

1. Update all the changed files except TileCodecs.scala and run TileCodecsSpec from spark's tests. The test should fail.
2. After updating TileCodecs.scala the test should succeed.

spark/src/test/scala/geotrellis/spark/io/avro/AvroTools.scala
```diff
@@ -1,14 +1,172 @@
...
     val fromBytes = AvroEncoder.fromBinary[T](bytes)
-    fromBytes should be equals thing
+    fromBytes shouldBe thing
     val json = AvroEncoder.toJson(thing)
     val fromJson = AvroEncoder.fromJson[T](json)
-    fromJson should be equals thing
+    fromJson shouldBe thing
+  }
+
+  def roundTripWithNoDataCheck[T <% NoDataValueChecker[T]](thing: T)(implicit codec: AvroRecordCodec[T]): Unit = {
```
lossyrob: Viewable is actually deprecated (or being deprecated) in Scala, so we avoid using it (`<%`)

A pattern I would use here is:

```scala
import geotrellis.util.MethodExtensions

trait AvroTools {
  def roundTripWithNoDataCheck[T: AvroRecordCodec: (? => AvroNoDataCheckMethods[T])](thing: T): Unit = [
    // Now we can call those methods on thing
  }
}

object AvroTools {
  trait NoDataValueChecker[T] extends MethodExtensions[T] { ... }
  implicit class ShortNoDataValueChecker(val self: ShortArrayTile) extends NoDavaValueChecker[ShortArrayTile] {
    override def doCheck(nodata: Option[Any]): Unit = { ... }
  }
}
```

This would fit the method extension structure of the rest of the library.

raster/src/main/scala/geotrellis/raster/ArrayMultibandTile.scala
```diff
@@ -428,4 +428,25 @@ class ArrayMultibandTile(_bands: Array[Tile]) extends MultibandTile with MacroMu
...
+    } else other match {
+      case that : ArrayMultibandTile =>
+        case class NotEqualException() extends Exception()
+        var result = true
+        try {
+          for (b <- 0 until bandCount) {
+            if (band(b) != that.band(b)) throw NotEqualException()
+          }
+        } catch {
+          case e : NotEqualException =>
```
lossyrob: We don't use exceptions for control flow. That's considered a "Bad Thing" in Scala.

You could fold:

```scala
val result = 
  (0 until bandCount)
    .foldLeft(true) { (acc, b) => acc && (band(b) == that.band(b)) }
```

That is doing a foreach over integer values, and while the size should be small, the actually fastest way to do it is not as pretty (because of the var):

```scala
var result = true
cfor(0)(_ < bandCount, _ + 1) { b =>
  result = result && (band(b) == that.band(b))
}
```

shiraeeshi: It was a terrible idea to use an exception there, the intention was to short-circuit when the result is known right away.

shiraeeshi: I decided to use a while loop, the drawback in the new code is that the use of "return" keyword makes the code too imperative, but I think it's not an issue for such a small method, the imperativity will not spread out of its little sandbox :).

lossyrob: Note that `false && {println("hey"); true}` will not print and evaluate to false, right side is not executed if left is false

raster/src/main/scala/geotrellis/raster/ArrayMultibandTile.scala
```diff
@@ -428,4 +428,20 @@ class ArrayMultibandTile(_bands: Array[Tile]) extends MultibandTile with MacroMu
...
+    if (other == null) {
+      false
+    } else other match {
+      case that : ArrayMultibandTile =>
+        var i = 0
+        while (i < bandCount) {
+          if (band(i) != that.band(i)) return false
+          i += 1
+        }
+        true
```
lossyrob: Would prefer yhe `cfor` method (cfor is macro which unrolls to structure which is basically as fast as while loop). Loop condition can be modified to break if result becomes false.

raster/src/main/scala/geotrellis/raster/ArrayMultibandTile.scala
shiraeeshi: My concern was about iterating that expression over and over again until we reach bandCount, but it's ok, editing looping condition did the trick, thank you for pointing that out.

raster/src/main/scala/geotrellis/raster/ArrayMultibandTile.scala
```diff
@@ -428,4 +428,21 @@ class ArrayMultibandTile(_bands: Array[Tile]) extends MultibandTile with MacroMu
 
     new ArrayMultibandTile(newBands)
   }
+
+  override def equals(other: Any): Boolean = {
+    if (other == null) {
```
moradology: Seems like there are two cases: case `that: ArrayMultibandTile` and case `else` - perhaps remove `if (other == null)`?

moradology:
A really solid PR. +1 after simplifying `equals`

raster/src/main/scala/geotrellis/raster/ArrayMultibandTile.scala
shiraeeshi: It's been a long story with this method :) I bet it can't be made simpler now.

raster/src/main/scala/geotrellis/raster/ArrayMultibandTile.scala
```diff
@@ -428,4 +428,16 @@ class ArrayMultibandTile(_bands: Array[Tile]) extends MultibandTile with MacroMu
 
     new ArrayMultibandTile(newBands)
   }
+
+  override def equals(other: Any): Boolean = other match {
+    case that : ArrayMultibandTile =>
+      var result = (bandCount == that.bandCount)
```
lossyrob: There was a null check here before, was that so that `that.bandCount` would not throw? That seems reasonable.

Also for types we don't put a right space, so `case that: ArrayMultibandTile`. Nitpicking :)

I propose

```scala
override def equals(other: Any): Boolean =
  other match {
    case that: ArrayMultibandTile if that != null =>
      var result = (bandCount == that.bandCount)
      
      cfor(0)(result && _ < bandCount, _ + 1) { i =>
        if (band(i) != that.band(i)) result = false
      }

      result
    case _ =>
      false
  }
```

shiraeeshi: I've checked the null case in the console and if I'm not mistaken, the null check seems unnecessary because null goes to "case _" part.

PR #1474, opened 2016-05-06
k-Nearest Neighbors

jpolchlo:
This PR adds k-nearest neighbors methods.  These come in two forms: an addition to SpatialIndex that allows for querying the k-nearest elements of the index to some target geometry, implemented as an R-tree; and the addition of a naïve, O(n) k-nearest neighbor algorithm over an RDD of geometry (or, rather, any type which can be coerced into an Extent).  The former version should be preferred for repeated queries, the latter for one-shot use.


spark/src/main/scala/geotrellis/spark/knn/KNearestRDD.scala
```diff
@@ -0,0 +1,27 @@
...
+import geotrellis.spark._
+import geotrellis.vector._
+import com.vividsolutions.jts.geom.Envelope
+import com.vividsolutions.jts.geom.Coordinate
+
+import org.apache.spark.rdd.RDD
+
+object KNearestRDD {
+  def kNearest[T](rdd: RDD[T], x: Double, y: Double, k: Int)(f: T => Extent): List[T] =
+    kNearest(rdd, new Envelope(new Coordinate(x, y)), k)(f)
```
echeipesh: We should not use naked `jts` types, that way madness lies. Results of most operations is a `Geometry`, which then requires you to test for types through reflection. This is one of the main reasons we ended up wrapping it, so we can use scala type system to reason about the result types: https://github.com/geotrellis/geotrellis/blob/master/vector/src/main/scala/geotrellis/vector/Results.scala.

lossyrob: It actually looks like you're just creating an Extent, but using the implicit jts Envelope => Extent conversion to do so.

Should just use `Extent(x, y, x, y)`

vector/src/main/scala/geotrellis/vector/SpatialIndex.scala
```diff
@@ -88,6 +91,62 @@ class SpatialIndex[T](val measure: Measure = Measure.Euclidean) extends Serializ
...
+        if (elem.d < best.head.d) {
+          best.enqueue(elem)
+          best.dequeue()
+        }
+      }
+    }
+    def ibToPQitem (ib: ItemBoundable): PQitem[T] = {
+      PQitem(DistanceOp.distance(gf.toGeometry(env), gf.toGeometry(ib.getBounds.asInstanceOf[Envelope])), ib.getItem.asInstanceOf[T])
+    }
+    def anToPQitem (nd: AbstractNode): PQitem[AbstractNode] = {
```
lossyrob: These names are really hard to follow. I'd prefer more verbose names that explained a bit more what they are doing than things like `ib` and `an`

vector/src/main/scala/geotrellis/vector/SpatialIndex.scala
```diff
@@ -88,6 +91,62 @@ class SpatialIndex[T](val measure: Measure = Measure.Euclidean) extends Serializ
...
+    val gf = new GeometryFactory()
+    val env = ex.jtsEnvelope
+    val pq = (new mutable.PriorityQueue(): mutable.PriorityQueue[PQitem[AbstractNode]]).reverse
+    val best = (new mutable.PriorityQueue(): mutable.PriorityQueue[PQitem[T]])
+
+    def addToBest (elem: PQitem[T]) = {
+      if (best.size < k)
+        best.enqueue(elem)
+      else {
+        if (elem.d < best.head.d) {
```
lossyrob: Doesn't this only enqueue a new point into a queue with k points only if the new element's distance is shorter than the closest point? What if we have k = 5 and the new element is further then the current 5 best (closest) points, but better (closer) than the other 5 that are in the queue?

spark/src/main/scala/geotrellis/spark/knn/KNearestRDD.scala
```diff
@@ -0,0 +1,27 @@
...
+  def kNearest[T](rdd: RDD[T], p: (Double, Double), k: Int)(f: T => Extent): List[T] =
+    kNearest(rdd, new Envelope(new Coordinate(p._1, p._2)), k)(f)
+
+  /**
+   * Determines the k-nearest neighbors of an RDD of objects which can be coerced into Extents.
+   */
+  def kNearest[T](rdd: RDD[T], ex: Extent, k: Int)(f: T => Extent): List[T] = {
+    val candidates = rdd.glom.map { 
+        arr => SpatialIndex.fromExtents(arr)(f)
+      }.map (_.kNearest (ex, k)).reduce( _ ++ _ )
```
lossyrob: It's unclear to me whether creating a SpatialIndex for each of the elements, and then using that spatial index to avoid doing distance calculations against some elements, will be less costly than just doing the distance calculation for each partition and reducing over them. What was the reason for going the spatial index route?

vector/src/main/scala/geotrellis/vector/SpatialIndex.scala
lossyrob: Or is the `head` of the queue the furthest-away best item?

spark-testkit/src/main/scala/geotrellis/spark/testkit/TestKryoSerializer.scala
```diff
@@ -0,0 +1,682 @@
...
+import org.apache.spark.util.collection.CompactBuffer
+
+/**
+ * A Spark serializer that uses the [[https://code.google.com/p/kryo/ Kryo serialization library]].
+ *
+ * Note that this serializer is not guaranteed to be wire-compatible across different versions of
+ * Spark. It is intended to be used to serialize/de-serialize data within a single
+ * Spark application.
+ */
+class TestKryoSerializer(conf: SparkConf)
```
lossyrob: Is this used anymore? We wouldn't want to keep it around if not.

spark/src/main/scala/geotrellis/spark/knn/KNearestMethods.scala
```diff
@@ -0,0 +1,14 @@
...
+
+import geotrellis.util.MethodExtensions
+import geotrellis.vector.{Extent, Point}
+
+import org.apache.spark.rdd.RDD
+
+trait KNearestMethods[T] extends MethodExtensions[RDD[T]] {
+  def kNearest(x: Double, y: Double, k: Int)(f: T => Extent): Seq[T] = { KNearestRDD.kNearest[T](self, x, y, k)(f) }
+
+  def kNearest(p: (Double, Double), k: Int)(f: T => Extent): Seq[T] = { KNearestRDD.kNearest[T](self, p, k)(f) }
```
lossyrob: Should have overload for `Point`

spark/src/main/scala/geotrellis/spark/knn/Implicits.scala
```diff
@@ -0,0 +1,10 @@
+package geotrellis.spark.knn
+
+import geotrellis.vector.Point
+import org.apache.spark.rdd.RDD
+
+object Implicits extends Implicits
+
+trait Implicits {
+  implicit class withKNearestMethods[T](val self: RDD[T]) extends KNearestMethods[T]
```
lossyrob: `T` is a bit generic. I can see, because we're saying, hey if you can give us a function that gives an extent then great, we can use this function...so perhaps that should stand. Unfortunate for compile times that this implicit class will be checked for every single call.

I think we should make implicit methods for each of the known cases. E.g. I want to be able to call

```scala
(rdd: RDD[PointFeature[D]]).kNearestNeighbors(point, 5)
(rdd: RDD[Point]).kNearestNeighbors(point, 5)
// etc...
```

Also, we wanted an overload to call for a sequence of points (k nearest neighbors to the cluster centers):

```scala
val clusterCenters: Traversable[Point] = ???
val pointFeatures: RDD[PointFeature[T]] = ???
val result: Map[Point, Seq[PointFeature[T]]] = pointFeatures.kNearestNeighbors(clusterCenters, 5)

// for each key in result, there is a Seq of at most 5 points, which are the 5 closest neighbors to that cluster center
```

jpolchlo: Yeah, I agree, but I was following the example set by the SpatialIndex
implementation, which is generic over T and relies on the transformation T
=> Extent.  I thought it should be SpatialIndex[T: Geometry], but it
wasn't, so I didn't argue.  Providing the additional overloads should mimic
the proper functionality.  Of course, the better solution is just to fix
SpatialIndex.

spark/src/main/scala/geotrellis/spark/knn/KNearestRDD.scala
```diff
@@ -0,0 +1,30 @@
...
+object KNearestRDD {
+  case class Ord[T] (ex: Extent, f: T => Extent) extends Ordering[T] with Serializable {
+    def compare(a: T, b: T) = (ex distance (f(a))) compare (ex distance (f(b)))
+  }
+
+  def kNearest[T](rdd: RDD[T], x: Double, y: Double, k: Int)(f: T => Extent): Seq[T] =
+    kNearest(rdd, new Extent(x, y, x, y), k)(f)
+
+  def kNearest[T](rdd: RDD[T], p: (Double, Double), k: Int)(f: T => Extent): Seq[T] =
+    kNearest(rdd, new Extent(p._1, p._2, p._1, p._2), k)(f)
```
lossyrob: Can use points instead of extents.

spark/src/main/scala/geotrellis/spark/knn/Implicits.scala
lossyrob: After discussion, should be on `Geometry` and `Feature[G <: Geometry]`

spark/src/test/scala/geotrellis/spark/TestEnvironment.scala
```diff
@@ -75,6 +75,7 @@ trait TestEnvironment extends BeforeAndAfterAll
     if(Properties.envOrNone("GEOTRELLIS_USE_JAVA_SER") == None) {
       conf
         .set("spark.serializer", "org.apache.spark.serializer.KryoSerializer")
+        // .set("spark.serializer", "org.apache.spark.serializer.TestKryoSerializer")
```
lossyrob: Remove

spark/src/main/scala/geotrellis/spark/knn/KNearestRDD.scala
```diff
@@ -0,0 +1,118 @@
+package geotrellis.spark.knn
+
+//import scala.collection.mutable.{PriorityQueue}
```
lossyrob: Remove.

Implicits should be ordered in 3 groupings, as follows:

```scala
// Imports of this project, in alphabetical order
import geotrellis.raster._
import geotrellis.spark._

// Imports of 3rd party libraries, in in alphabetical order
import org.apache.spark._

// Imports of java + scala, in alphabetical order
import java.util.PriorityQueue
import scala.collections.JavaConversions._

```

spark/src/main/scala/geotrellis/spark/knn/KNearestRDD.scala
```diff
@@ -0,0 +1,118 @@
...
+}
+
+object BoundedPriorityQueue {
+  /**
+   * Creates a new BoundedPriorityQueue instance.
+   * @param maximum the max number of elements
+   * @return a new bounded priority queue instance
+   */
+  def apply[A: Ordering](maximum: Int): BoundedPriorityQueue[A] = {
+    //val ordering = implicitly[Ordering[A]].reverse
```
lossyrob: Remove if unused.

spark/src/main/scala/geotrellis/spark/knn/KNearestRDD.scala
```diff
@@ -0,0 +1,118 @@
...
+   * Determines the k-nearest neighbors of an RDD of objects which can be
+   * coerced into Extents.
+   */
+  def kNearest[T](rdd: RDD[T], ex: Extent, k: Int)(f: T => Extent): Seq[T] = {
+    implicit val ord = new Ord[T](ex, f)
+
+    rdd.takeOrdered(k)
+  }
+
+  // def kNearest[G, H](rdd: RDD[G], centers: Traversable[H], k: Int)(g: G => Extent, h: H => Extent): Map[H, Seq[G]] = {
```
lossyrob: Remove if unused.

spark/src/main/scala/geotrellis/spark/knn/KNearestMethods.scala
```diff
@@ -0,0 +1,66 @@
...
+import geotrellis.vector.{Extent, Point, Geometry, Feature}
+
+import org.apache.spark.rdd.RDD
+
+trait KNearestMethods[T] extends MethodExtensions[RDD[T]] {
+  def kNearest(x: Double, y: Double, k: Int)(f: T => Extent): Seq[T] = { KNearestRDD.kNearest[T](self, x, y, k)(f) }
+
+  def kNearest(p: (Double, Double), k: Int)(f: T => Extent): Seq[T] = { KNearestRDD.kNearest[T](self, p, k)(f) }
+
+  def kNearest(p: Point, k: Int)(f: T => Extent): Seq[T] = { KNearestRDD.kNearest[T](self, p, k)(f) }
```
lossyrob: We don't need an f: T => Extent if it's a point. This is covered by the Geom methods.

spark/src/main/scala/geotrellis/spark/knn/KNearestMethods.scala
```diff
@@ -0,0 +1,66 @@
...
+  def kNearest(p: (Double, Double), k: Int): Seq[Feature[G, D]] = 
+    KNearestRDD.kNearest[Feature[G, D]](self, p, k){ g: Feature[G, D] => g.geom.envelope }
+
+  def kNearest(p: Point, k: Int): Seq[Feature[G, D]] = 
+    KNearestRDD.kNearest[Feature[G, D]](self, p, k){ g: Feature[G, D] => g.geom.envelope }
+
+  def kNearest(ex: Extent, k: Int): Seq[Feature[G, D]] = 
+    KNearestRDD.kNearest[Feature[G, D]](self, ex, k){ g: Feature[G, D] => g.geom.envelope }
+
+  def kNearest[H <: Geometry](centers: Traversable[H], k: Int): Seq[Seq[Feature[G, D]]] = {
```
lossyrob: This does not end up using the geometry's `distance` method, instead using the extent's, correct?

This will produce bad results, e.g. squiggly polygons with large bounding boxes.

PR #1478, opened 2016-05-10
Fix month and year calculations

shiraeeshi:
The operand 24 was missing in byMonth/byYear methods.


echeipesh:
Thank you for finding and fixing this.

PR #1480, opened 2016-05-13
[ADR] Streaming Writes

pomadchin:


lossyrob:
Nice, thanks for capturing this in an ADR

PR #1481, opened 2016-05-18
Use external build of Proj4J

dwins:
Since we've been discussing moving the GeoTrellis build of Proj4J out into a more accessible and independent library I wanted to get some code out that moves in that direction.  This branch shows what I was thinking we would need to do.  If you grab my [proj4j](https://github.com/dwins/proj4j) repo and issue `./sbt publishLocal` then this GeoTrellis branch should work.

Some questions remain - what should go in the LICENSE notice for the new repository? What will be the artifact id (currently under the "com.azavea" organization)?  Do we need to rename the Java packages within the project?  But apart from these organizational questions the code behavior seems ok already.


lossyrob:
Closing until we have a more serious effort to accomplish this extraction; marked this in the issue to provide a good start.

PR #1487, opened 2016-05-20
Handle entire layers of `NODATA`

fosskers:
## TODO
- [x] Guard on asking for a colour spread of 0 colours in `ColorRamp.spread`
- [x] Pass current tests
- [x] Write new tests to test the guarding
- [x] Test `geotrellis-admin` with a snapshot of `geotrellis` with this fix
- [x] Add tests to make sure empty `ColorRamp` methods don't throw exceptions

## Background

What we initially thought to be a problem with either `geotrellis-admin` or my ingesting process turned out to be an issue with how `ColorMap`s are created by `geotrellis-raster`. 

### Symptoms
The following is output from the `geotrellis-admin` server inside `docker-compose`:
```
adminserver_1  | [ERROR] [05/20/2016 16:52:04.572] [geotrellis-admin-server-akka.actor.default-dispatcher-5] [akka://geotrellis-admin-server/user/geotrellis-admin-service] Error during processing of request HttpRequest(GET,http://adminserver:8080/gt/tms/nlcd-tms-epsg3857-testing3/3/2/2?colorRamp=blue-to-yellow-to-red-heatmap&breaks=4d7ed1b2-7d99-48e7-bba9-ee963fcef49f&opacity=100,List(Host: adminserver:8080, Connection: close, User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:46.0) Gecko/20100101 Firefox/46.0, Accept-Language: en-US, en;q=0.5, Accept-Encoding: gzip, deflate, DNT: 1, Referer: http://localhost:8090/, Cookie: CSRF-Token-RRKAZ=HPp-TrMarTjw5tv1ojV7FeVCpf0wciEF; firstVisit=1463754555964),Empty,HTTP/1.0)
adminserver_1  | java.lang.ArrayIndexOutOfBoundsException: 0
adminserver_1  |     at geotrellis.raster.render.ColorRamp$.spread(ColorRamp.scala:131)
adminserver_1  |     at geotrellis.raster.render.ColorRamp.stops(ColorRamp.scala:11)
adminserver_1  |     at geotrellis.raster.render.ColorMap$.apply(ColorMap.scala:94)
adminserver_1  |     at geotrellis.raster.render.ColorRamp.toColorMap(ColorRamp.scala:82)
adminserver_1  |     at geotrellis.admin.server.AdminRoutes$$anonfun$serveTile$1$$anonfun$apply$13$$anonfun$apply$14$$anonfun$apply$15.apply(AdminActor.scala:274)
adminserver_1  |     at geotrellis.admin.server.AdminRoutes$$anonfun$serveTile$1$$anonfun$apply$13$$anonfun$apply$14$$anonfun$apply$15.apply(AdminActor.scala:272)
...
```
I was seeing this any time I zoomed in or out while trying to read tiles from my recently ingested layer `nlcd-tms-epsg3857-testing3`.  Note that for this layer, while the values are int8, [we seem to have agreed](https://github.com/azavea/azavea-data-hub/issues/4) that the `NODATA` value for this data set is 0 instead of `Int.MinValue`. 

I have only ingested about 1/50th of the full dataset, and incidently we think the data I did ingest was entirely `NODATA`. 

### Diagnosis

From `geotrellis.admin.server.AdminRoutes.breaks`:

```scala
   /** Calculate and store class breaks for a layer, and yield a UUID that                                 
     * references the saved breaks. Necessary for a `/tms/...` call.                                      
     */
   def breaks = pathPrefix(Segment / IntNumber) { (layer, numBreaks) =>
     import DefaultJsonProtocol._
     complete {
       val data = reader.read[SpatialKey, Tile, TileLayerMetadata[SpatialKey]](layerId(layer))

       val breaks: Array[Double] = data.classBreaksDouble(numBreaks)
       val uuid: String = java.util.UUID.randomUUID.toString
       breaksStore(uuid)(breaks)
 
       JsObject("classBreaks" -> JsString(uuid))
     }
   } 
```
It turns out that `classBreaksDouble` yields an empty `Array` when `data` here is entirely `NODATA`. 

When serving a tile, the server attempts to contruct a `ColorMap` and then render a PNG. The empty class breaks from above propagate to here (`geotrellis.raster.render.ColorRamp`):
```scala
   def spread(colors: Vector[Int], n: Int): Vector[Int] = {
     if (colors.length == n) return colors

     val colors2 = new Array[Int](n)
     colors2(0) = colors(0)  // <-- This throws an exception because n = 0.

     val b = n - 1
     val color = colors.length - 1
     var i = 1
     while (i < n) {
       colors2(i) = colors(math.round(i.toDouble * color / b).toInt)
       i += 1
     }
     colors2.toVector
   }
```
The line indicated throws an exception.

### Considerations
**Assertion**: A layer of all `NODATA` exists in the set of valid layers at a given resolution (or as the identity element of the monoid of layers at some resolution), thus a user should be able to view a layer of entirely `NODATA` without being inundated with exceptions.

**Question**: Is it better to catch the empty class breaks like I'm doing, or alter the behaviour of the `classBreaks` method? For instance, to return a breaks Array of the length the requested, populated entirely with 0s.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/render/ColorRamp.scala
+++ b/raster/src/main/scala/geotrellis/raster/render/ColorRamp.scala
@@ -138,7 +138,7 @@ object ColorRamp {
      * will still perform as expected and refuse to colour NODATA locations,
      * thanks to the `noDataColor` field in `ColorMap.Options`.
      */
-    if (n == 0) return Vector.empty[Int]
+    if (n < 1) return Vector.empty[Int]
 
     val colors2 = new Array[Int](n)
     colors2(0) = colors(0)
--- a/raster/src/main/scala/geotrellis/raster/render/ColorRamp.scala
+++ b/raster/src/main/scala/geotrellis/raster/render/ColorRamp.scala
@@ -128,8 +128,7 @@ object ColorRamp {
     /* The requested spread matches the original */
     if (colors.length == n) return colors
 
-    /* 2016 May 20 @ 14:21
-     * In the case of a Layer comprised entirely of NODATA, a call to
+    /* In the case of a Layer comprised entirely of NODATA, a call to
      * `classBreaksDouble` will yield an empty `Array`. That propagates here,
      * where an attempt to index below results in a bounds exception.
      *
--- a/raster/src/main/scala/geotrellis/raster/render/ColorRamp.scala
+++ b/raster/src/main/scala/geotrellis/raster/render/ColorRamp.scala
@@ -125,7 +125,20 @@ object ColorRamp {
     * @param n       Length of list to return
     */
   def spread(colors: Vector[Int], n: Int): Vector[Int] = {
+    /* The requested spread matches the original */
     if (colors.length == n) return colors
+
+    /* 2016 May 20 @ 14:21
+     * In the case of a Layer comprised entirely of NODATA, a call to
+     * `classBreaksDouble` will yield an empty `Array`. That propagates here,
+     * where an attempt to index below results in a bounds exception.
+     *
+     * Guarding on this condition and returning an empty `Vector` instead
+     * is still valid; a `ColorMap` that contains an empty `colors` field
+     * will still perform as expected and refuse to colour NODATA locations,
+     * thanks to the `noDataColor` field in `ColorMap.Options`.
+     */
+    if (n == 0) return Vector.empty[Int]
 
     val colors2 = new Array[Int](n)
     colors2(0) = colors(0)
--- a/raster/src/main/scala/geotrellis/raster/render/ColorRamp.scala
+++ b/raster/src/main/scala/geotrellis/raster/render/ColorRamp.scala
@@ -155,51 +155,58 @@ object ColorRamp {
 
   /** RGBA interpolation logic */
 
-  def chooseColors(c: Vector[Int], numColors: Int): Vector[Int] =
-    getColorSequence(numColors) { (masker: Int => Int, count: Int) =>
-      val hues = c.map(masker)
-      val mult = c.length - 1
-      val denom = count - 1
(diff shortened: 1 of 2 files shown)


raster/src/main/scala/geotrellis/raster/render/ColorRamp.scala
```diff
@@ -125,8 +125,21 @@ object ColorRamp {
     * @param n       Length of list to return
     */
   def spread(colors: Vector[Int], n: Int): Vector[Int] = {
+    /* The requested spread matches the original */
     if (colors.length == n) return colors
 
+    /* 2016 May 20 @ 14:21
```
lossyrob: No need to time stamp...this goes stale once someone edits the comment, and then requiring updating of it would be a maintenance burden...git takes care of this.

fosskers: A personal habit. I like to timestamp comments, not necessarily the change that was made.

fosskers: Want me to keep them out anyway?

lossyrob: Eh, my personal opinion is that it adds noise, and it's not clear why it's there, and there's not a standard for the codebase around doing it, so I'd rather it not be there. @echeipesh do you have any opinions on the matter?

echeipesh: I'd prefer to keep them out. As you say, git should take care of keeping track of changes and times and provides a place to dive through if that info is needed.

fosskers: Alright, taking it out.

lossyrob:
👍

PR #1493, opened 2016-05-23
Added support for int32raw cell types in CellType.fromString

### LINKED ISSUE #1471: int32raw support in CellType.fromString
It looks like raw int32 tiles don't have support in `CellType.fromString`. It's a one line fix which should probably make it into 0.10.1 in case people attempt to read a metadata file with 'int32raw' celltype

jpolchlo:
Closes #1471


lossyrob:
@jpolchlo fyi: the "closes" statement goes in the PR description, I moved it

PR #1494, opened 2016-05-23
Add ETL option to specify upper zoom limit for raster layer ingestion

mbertrand:
This PR addresses https://github.com/geotrellis/geotrellis/issues/1484 by adding an optional "maxZoom" argument to ETLConf (with a default value of 0), and modifies both the Etl and TileLayerMetadata classes to handle that option.

In testing with the Chattanooga demo however, setting maxZoom levels above 10 often led to the ingestion process being killed due to lack of memory.  Increasing VM memory to 6 GB, and the memory available to the ingestion process to 4 GB, allowed the ingestion process to succeed up to zoom level 12.

### DIFF
--- a/spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala
+++ b/spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala
@@ -127,11 +127,10 @@ case class Etl(args: Seq[String], @transient modules: Seq[TypedModule] = Etl.def
       case PerTileReproject =>
         val reprojected = rdd.reproject(destCrs)
         val (zoom: Int, md: TileLayerMetadata[K]) = scheme match {
-          case Left(layoutScheme) =>
-            TileLayerMetadata.fromRdd(reprojected, maxZoom match {
-              case Some(zoom) =>  ZoomedLayoutScheme(destCrs, conf.tileSize())
-              case _ => layoutScheme
-            }, maxZoom)
+          case Left(layoutScheme) => maxZoom match {
+              case Some(zoom) =>  TileLayerMetadata.fromRdd(reprojected, ZoomedLayoutScheme(destCrs, conf.tileSize()), maxZoom)
+              case _ => TileLayerMetadata.fromRdd(reprojected, layoutScheme)
+            }
           case Right(layoutDefinition) =>
             TileLayerMetadata.fromRdd(reprojected, layoutDefinition)
         }
@@ -140,10 +139,10 @@ case class Etl(args: Seq[String], @transient modules: Seq[TypedModule] = Etl.def
         zoom -> ContextRDD(reprojected.tileToLayout[K](amd, tilerOptions), amd)
 
       case BufferedReproject =>
-        val (_, md) = TileLayerMetadata.fromRdd(rdd, maxZoom match {
-          case Some(zoom) =>  ZoomedLayoutScheme(destCrs, conf.tileSize())
-          case _ => FloatingLayoutScheme(conf.tileSize())
-        }, maxZoom)
+        val (_, md) = maxZoom match {
+          case Some(zoom) =>  TileLayerMetadata.fromRdd(rdd, ZoomedLayoutScheme(destCrs, conf.tileSize()), maxZoom)
+          case _ => TileLayerMetadata.fromRdd(rdd, FloatingLayoutScheme(conf.tileSize()))
+        }
         val amd = adjustCellType(md)
         // Keep the same number of partitions after tiling.
         val tilerOptions = Tiler.Options(resampleMethod = method, partitioner = new HashPartitioner(rdd.partitions.length))
--- a/spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala
+++ b/spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala
@@ -128,7 +128,7 @@ case class Etl(args: Seq[String], @transient modules: Seq[TypedModule] = Etl.def
         val reprojected = rdd.reproject(destCrs)
         val (zoom: Int, md: TileLayerMetadata[K]) = scheme match {
           case Left(layoutScheme) => maxZoom match {
-              case Some(zoom) =>  TileLayerMetadata.fromRdd(reprojected, ZoomedLayoutScheme(destCrs, conf.tileSize()), maxZoom)
+              case Some(zoom) =>  TileLayerMetadata.fromRdd(reprojected, ZoomedLayoutScheme(destCrs, conf.tileSize()), zoom)
               case _ => TileLayerMetadata.fromRdd(reprojected, layoutScheme)
             }
           case Right(layoutDefinition) =>
@@ -140,7 +140,7 @@ case class Etl(args: Seq[String], @transient modules: Seq[TypedModule] = Etl.def
 
       case BufferedReproject =>
         val (_, md) = maxZoom match {
-          case Some(zoom) =>  TileLayerMetadata.fromRdd(rdd, ZoomedLayoutScheme(destCrs, conf.tileSize()), maxZoom)
+          case Some(zoom) =>  TileLayerMetadata.fromRdd(rdd, ZoomedLayoutScheme(destCrs, conf.tileSize()), zoom)
           case _ => TileLayerMetadata.fromRdd(rdd, FloatingLayoutScheme(conf.tileSize()))
         }
         val amd = adjustCellType(md)
--- a/spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala
+++ b/spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala
@@ -113,6 +113,7 @@ case class Etl(args: Seq[String], @transient modules: Seq[TypedModule] = Etl.def
   )(implicit sc: SparkContext): (Int, RDD[(K, V)] with Metadata[TileLayerMetadata[K]]) = {
     val targetCellType = conf.cellType.get
     val destCrs = conf.crs()
+    val maxZoom = conf.maxZoom()
 
     def adjustCellType(md: TileLayerMetadata[K]) =
       md.copy(cellType = targetCellType.getOrElse(md.cellType))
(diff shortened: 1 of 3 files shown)


spark/src/main/scala/geotrellis/spark/TileLayerMetadata.scala
```diff
@@ -157,20 +157,23 @@ object TileLayerMetadata {
...
+
   def fromRdd[
     K: GetComponent[?, ProjectedExtent]: (? => TilerKeyMethods[K, K2]),
     V <: CellGrid,
     K2: SpatialComponent: Boundable
-  ](rdd: RDD[(K, V)], scheme: LayoutScheme): (Int, TileLayerMetadata[K2]) = {
+  ](rdd: RDD[(K, V)], scheme: LayoutScheme, maxZoom: Int = 0): (Int, TileLayerMetadata[K2]) = {
     val (extent, cellType, cellSize, bounds, crs) = collectMetadataWithCRS(rdd)
-    val LayoutLevel(zoom, layout) = scheme.levelFor(extent, cellSize)
+    val LayoutLevel(zoom, layout) = (if (maxZoom > 0) scheme.asInstanceOf[ZoomedLayoutScheme].levelForZoom(maxZoom)
```
lossyrob: Probably should use an `Option[Int]` type for the maxZoom, so if someone calls the argument with a maxZoom of 0, we can deal with that error differently than if the argument was not there at all.

We wouldn't want to put this `asInstanceOf` in here (`asInstanceOf` is considered Evil in Scala and we should only use it as a last resort).

The maxZoom parameter shouldn't be used with anything but a `tms` layout, which could be detected in the `EtlConfig` code and an error would be issued at that point.

Here we could avoid the `asInstanceOf` by matching

  ```scala
val LayoutLevel(zoom, layout) = {
  maxZoom match {
    case Some(zoom) =>
      scheme match {
         case zoomedLayoutScheme: ZoomedLayoutScheme =>
            zoomedLayoutScheme.levelFor(extent, cellSize)
         case _ => // throw a descriptive error
      }
    case None =>
      scheme.levelFor(extent, cellSize)
  }
}
```

Comment (mbertrand) [inline spark/src/main/scala/geotrellis/spark/TileLayerMetadata.scala]: @lossyrob Thanks for the feedback and sorry for the evil Scala!  I've updated the code.

File: spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala [hunk 2]
```diff
@@ -55,7 +55,12 @@ case class Etl(args: Seq[String], @transient modules: Seq[TypedModule] = Etl.def
   def scheme: Either[LayoutScheme, LayoutDefinition] = {
     if (conf.layoutScheme.isDefined) {
       val scheme = conf.layoutScheme()(conf.crs(), conf.tileSize())
-      logger.info(scheme.toString)
+      if (conf.maxZoom.isDefined) {
+        scheme match {
+          case  zoomedLayoutScheme: ZoomedLayoutScheme => logger.info("maxZoom and layout scheme compatible")
+          case _ => throw new RuntimeException("maxZoom option should only be used with tms layout scheme")
+        }
+      }
```
lossyrob: This works, but I'm wondering if there's perhaps a better way to handle the conflict.

@echeipesh you would know, can you suggest another way or else we can just keep this as is.

spark/src/main/scala/geotrellis/spark/TileLayerMetadata.scala
```diff
@@ -157,20 +157,43 @@ object TileLayerMetadata {
...
+  ](rdd: RDD[(K, V)], scheme: LayoutScheme, maxZoom: Option[Int] = None): (Int, TileLayerMetadata[K2]) = {
     val (extent, cellType, cellSize, bounds, crs) = collectMetadataWithCRS(rdd)
-    val LayoutLevel(zoom, layout) = scheme.levelFor(extent, cellSize)
+    val LayoutLevel(zoom, layout) = {
+      maxZoom match {
+        case Some(zoom) =>
+          scheme match {
+            case zoomedLayoutScheme: ZoomedLayoutScheme =>
+              zoomedLayoutScheme.levelForZoom(maxZoom.get)
+            case _ => throw new RuntimeException("ZoomedLayoutScheme required when setting a max zoom level")
```
lossyrob: Instead of throwing here, I would like to enforce this with the type system. This is a situation where I would prefer some code duplication for type safety. So keeping the original fromRDD's, and adding

```scala
def fromRdd[
    K: GetComponent[?, ProjectedExtent]: (? => TilerKeyMethods[K, K2]),
    V <: CellGrid,
    K2: SpatialComponent: Boundable
  ](rdd: RDD[(K, V)], scheme: ZoomedLayoutScheme, maxZoom: Int): (Int, TileLayerMetadata[K2]) = {
    val (extent, cellType, cellSize, bounds, crs) = collectMetadataWithCRS(rdd)
    val LayoutLevel(zoom, layout) =
      zoomedLayoutScheme.levelForZoom(maxZoom)

    val GridBounds(colMin, rowMin, colMax, rowMax) = layout.mapTransform(extent)
    val kb = bounds.setSpatialBounds(KeyBounds(layout.mapTransform(extent)))
    (zoom, TileLayerMetadata(cellType, layout, extent, crs, kb))
  }
```

(and one for the other type of `fromRdd` call.

lossyrob:
This is looking good! Just added some more comments.

Before we merge these changes, you'll have to sign an Eclipse CLA. Can you sign one with an eclipse account tied to the same email address you commit to github with, and let me know when you have that? You can follow the right hand side instructions of here: https://www.eclipse.org/legal/CLA.php

Thanks!

mbertrand:
Thanks @lossyrob, I made some more updates to the code and signed the Eclipse CLA.

spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala
echeipesh: I would suggest to move the validation to `EtlConf` object and use the `validateOpt` helper from scallop: https://github.com/scallop/scallop/wiki/Arguments-validation#custom-validation

mbertrand: Thanks @echeipesh,
I've been trying to implement this but can't get it to work properly, I think because layoutScheme is returning a  LayoutSchemaProvider instead of a LayoutScheme (Zoomed or Floating) within the validate function.  Any ideas on how to get the actual LayoutScheme class here?

```
  validateOpt (layoutScheme, maxZoom) {
    case (Some(schema), Some(zoom)) =>  schema match {
        case scheme: ZoomedLayoutScheme => Right(Unit)
        case _ => Left("maxZoom should only be used with ZoomedLayoutScheme (tms)")
      }
    case (None, Some(zoom)) => Left("maxZoom not be used when a LayoutScheme is not provided")
    case _ => Right(Unit)
  }
```

echeipesh: Ah, that's a good point. That makes it more difficult than it should be. I think rather than trying to get fancy with config parsing lets stick with your original approach.

echeipesh:
@mbertrand I am failing to verify the Eclipse CLA through: https://projects.eclipse.org/user/cla/validate
I would guess you used an email other than the one attached to these commits.

spark/src/main/scala/geotrellis/spark/TileLayerMetadata.scala
```diff
@@ -165,6 +165,21 @@ object TileLayerMetadata {
   }
 
   def fromRdd[
+    K: (? => TilerKeyMethods[K, K2]) ,
+    V <: CellGrid,
+    K2: SpatialComponent: Boundable
+  ](rdd: RDD[(K, V)], crs: CRS, scheme: ZoomedLayoutScheme, maxZoom: Option[Int] = None):
```
lossyrob: Unfortunately Scala does not allow default arguments on a method that is overloaded. The rules around this are strange, and sometimes the compiler does not catch this by default. I think it might only happen on 2.11, but if you run these commands, you'll get these outputs:

```console
./sbt
geotrellis > project spark
spark > +package
[info] Compiling 230 Scala sources to /Users/rob/proj/gt/geotrellis-codereview/spark/target/scala-2.11/classes...
[error] /Users/rob/proj/gt/geotrellis-codereview/spark/src/main/scala/geotrellis/spark/TileLayerMetadata.scala:66: in object TileLayerMetadata, multiple overloaded alternatives of method fromRdd define default arguments.
[error] object TileLayerMetadata {
[error]        ^
[error] one error found
[error] (spark/compile:compileIncremental) Compilation failed
[error] Total time: 75 s, completed Jun 7, 2016 11:30:43 AM
spark > 
```

That means we have to create overloads for the default options. And since we have to do that anyway, might be good to not require the option

e.g.

```scala
  def fromRdd[
    K: (? => TilerKeyMethods[K, K2]) ,
    V <: CellGrid,
    K2: SpatialComponent: Boundable
  ](rdd: RDD[(K, V)], crs: CRS, scheme: ZoomedLayoutScheme, maxZoom: Option[Int] = None):
  (Int, TileLayerMetadata[K2]) = {
    val (extent, cellType, cellSize, bounds) = collectMetadata(rdd)
    val LayoutLevel(zoom, layout) = maxZoom match {
      case Some(zoom) => scheme.levelForZoom(maxZoom.get)
      case _ => scheme.levelFor(extent, cellSize)
    }
    val kb = bounds.setSpatialBounds(KeyBounds(layout.mapTransform(extent)))
    (zoom, TileLayerMetadata(cellType, layout, extent, crs, kb))
  }
```

becomes something like

```scala
def fromRdd[
    K: (? => TilerKeyMethods[K, K2]) ,
    V <: CellGrid,
    K2: SpatialComponent: Boundable
  ](rdd: RDD[(K, V)], crs: CRS, scheme: ZoomedLayoutScheme):
  (Int, TileLayerMetadata[K2]) =
   _fromRdd[K, V, K2](rdd, crs, scheme, None)

 def fromRdd[
    K: (? => TilerKeyMethods[K, K2]) ,
    V <: CellGrid,
    K2: SpatialComponent: Boundable
  ](rdd: RDD[(K, V)], crs: CRS, scheme: ZoomedLayoutScheme, maxZoom: Int):
  (Int, TileLayerMetadata[K2]) =
   _fromRdd[K, V, K2](rdd, crs, scheme, Some(maxZoom))

 private def _fromRdd[
    K: (? => TilerKeyMethods[K, K2]) ,
    V <: CellGrid,
    K2: SpatialComponent: Boundable
  ](rdd: RDD[(K, V)], crs: CRS, scheme: ZoomedLayoutScheme, maxZoom: Option[Int]):
  (Int, TileLayerMetadata[K2]) = {
    val (extent, cellType, cellSize, bounds) = collectMetadata(rdd)
    val LayoutLevel(zoom, layout) = maxZoom match {
      case Some(zoom) => scheme.levelForZoom(maxZoom.get)
      case _ => scheme.levelFor(extent, cellSize)
    }
    val kb = bounds.setSpatialBounds(KeyBounds(layout.mapTransform(extent)))
    (zoom, TileLayerMetadata(cellType, layout, extent, crs, kb))
  }
```

You might be able to restructure something that takes a `LayoutLevel` explicitly which might look a bit more pretty.

mbertrand: @lossyrob sorry for the delay, I just updated the branch with your suggested changes.

lossyrob:
@mbertrand thanks! This is a great change :+1:

PR #1495, opened 2016-05-23
Add Support for GeoTools SimpleFeature

jamesmcclain:
In this pull request, support is added for converting GeoTools `SimpleFeature`s into Geotrellis `Feature` objects and vice-versa.

### Still Needs ###
   - [x] Improved API
      - toSimpleFeature (Geometry)
      - toSimpleFeature (Feature)
      - toGeometry
      - toFeature
   - [x] Depends on #1502
   - [x] Suitable shapefiles for tests
   - [x] More tests


geotools/build.sbt
```diff
@@ -0,0 +1,30 @@
+import Dependencies._
+
+name := "geotrellis-geotools"
+
+libraryDependencies ++= Seq(
+  "org.geotools" % "gt-main" % Version.geotools,
+  "org.geotools" % "gt-coverage" % Version.geotools,
+  "org.geotools" % "gt-geotiff" % Version.geotools,
+  "org.geotools" % "gt-epsg-hsql" % Version.geotools,
+  "org.apache.spark" %% "spark-core" % Version.spark % "provided",
```
echeipesh: This doesn't seem to be used.

jamesmcclain: Indeed, the first few commits of this PR were taken from my other GeoTools-related PR.  When the `GridCoverage2D` stuff that Rob is working on goes in, I will rebase and this will go away.

lossyrob:
This API should be refactored to follow the `ToGridCoverage2DMethods` design pattern, to provide this functionality as implicit methods, and to provide better usability.

Client side view of the API should be e.g.

```scala
val point: Point = ???
val crs: CRS = ???
val data: Map[(String, Any)] = ???

point.toSimpleFeature()
point.toSimpleFeature(crs)
point.toSimpleFeature(crs, data)
point.toSimpleFeature(data)

val simpleFeature: SimpleFeature = ???
val point: Point = simpleFeature.toGeometry[Point]
val geom: Geometry = simpleFeature.toGeometry[Geometry]
val pointFeature: PointFeature[Map[String, Object]] = simpleFeature.toFeature[Point]

// this opens the door for an implicit that converts the map to a case class, e.g.
case class Foo(x: Int, y: String)
implicit def mapToFoo(map: Map[String, Object]): Foo = ???

val pointFeature: PointFeature[Foo] = simpleFeature.toFeature[Point, Foo](mapToFoo) // mapToFoo implicit param
```

jamesmcclain:
Okay, I will see if I can put those changes in soon.

jamesmcclain:

geotools/src/main/scala/geotrellis/geotools/SimpleFeatureToFeature.scala
```diff
@@ -0,0 +1,70 @@
...
+      case mp: jts.MultiPoint => MultiPoint(mp)
+      case ml: jts.MultiLineString => MultiLine(ml)
+      case mp: jts.MultiPolygon => MultiPolygon(mp)
+      case gc: jts.GeometryCollection => GeometryCollection(gc)
+      case  g: jts.Geometry => throw new Exception(s"Unhandled JTS Geometry $g")
+      case _ => throw new Exception("Non-Geometry")
+    }
+  }
+
+  def apply(simpleFeature: SimpleFeature): Feature[Geometry, immutable.Map[String, Object]] = {
```
lossyrob: We should have this typed on `[G <: Geometry: ClassTag]` then use a combination of

https://github.com/geotrellis/geotrellis/blob/master/vector/src/main/scala/geotrellis/vector/Geometry.scala#L81

and

https://github.com/geotrellis/geotrellis/blob/master/vector/src/main/scala/geotrellis/vector/Geometry.scala#L105

To return the correct type.

Then the `asInstanceOf` cast below gets pushed down into the moment we translate from a JTS geometry.

We could call `SimpleFeatureToFeature[Geometry]` to keep the most generic type

geotools/src/main/scala/geotrellis/geotools/package.scala
```diff
@@ -20,4 +23,11 @@ package object geotools {
 
   implicit class withGridCoverage2DConversionMethods(val self: GridCoverage2D) extends MethodExtensions[GridCoverage2D]
       with GridCoverage2DConversionMethods
+
+  implicit class withGeometryToSimpleFeatureMethods[G <: Geometry](val self: G) extends MethodExtensions[G]
+      with GeometryToSimpleFeatureMethods[G]
```
lossyrob: Need methods for `Feature[G, D]` => `SimpleFeature`

jamesmcclain:

geotools/src/main/scala/geotrellis/geotools/SimpleFeatureToFeature.scala
```diff
@@ -0,0 +1,63 @@
...
+  private def jtsToGeotrellis[G <: Geometry : ClassTag](geometry: Object): Option[G] = {
+    geometry match {
+      case g: jts.Geometry => Some(Geometry(g).asInstanceOf[G])
+      case g => throw new Exception(s"Input $g is not a jts.Geometry")
+    }
+  }
+
+  def apply[G <: Geometry : ClassTag](simpleFeature: SimpleFeature): Feature[G, immutable.Map[String, Object]] = {
+    val properties = simpleFeature.getProperties.asScala
+    val map = mutable.Map.empty[String, Object]
```
moradology: These references to `Object` should perhaps be `AnyRef`s, as that's the scala type which matches 1:1 onto java objects. References to `Any` in this context are potentially suspect. It looks like `java.lang.Double` and scala's `Double` both work for `Any`, while `SimpleFeature`s require the former

geotools/src/main/scala/geotrellis/geotools/SimpleFeatureToFeatureMethods.scala
```diff
@@ -0,0 +1,48 @@
...
+    val feature = SimpleFeatureToFeature(self)
+    val ct = implicitly[ClassTag[G]]
+
+    feature.geom match {
+      case ct(g) => Feature(feature.geom, feature.data)
+      case _ => throw new Exception(s"${feature.geom} is not a ${ct.toString}")
+    }
+  }
+
+  def toFeature[G <: Geometry : ClassTag, T]()(implicit transmute: Map[String, Any] => T): Feature[G, T] = {
```
moradology: This is the use of `Any` which might (errantly) include scala's `Double` as a legitimate type

geotools/src/main/scala/geotrellis/geotools/SimpleFeatureToFeature.scala
```diff
@@ -0,0 +1,63 @@
...
+object SimpleFeatureToFeature {
+
+  private def jtsToGeotrellis[G <: Geometry : ClassTag](geometry: Object): Option[G] = {
+    geometry match {
+      case g: jts.Geometry => Some(Geometry(g).asInstanceOf[G])
+      case g => throw new Exception(s"Input $g is not a jts.Geometry")
+    }
+  }
+
+  def apply[G <: Geometry : ClassTag](simpleFeature: SimpleFeature): Feature[G, immutable.Map[String, Object]] = {
```
moradology: I wonder about our ability to pull off a roundtrip without an inverse of the `transmute` function defined for the `Feature` -> `SimpleFeature` translation

jamesmcclain:
`Map[String, Object]` has been changed to `Map[String, AnyRef]`

lossyrob:
+1

PR #1498, opened 2016-05-25
Add ability to specify number of streaming buckets

moradology:
A paramater was unused in for streaming histograms, this should rectify that


moradology:
closes #1498

Issue #1499, opened 2016-05-26
Streaming histogram classbreaks could interpolate lower values


moradology:
It seems to me that there's no real harm (some negligible speed decrease) in interpolating values below the lowest seen by our `StreamingHistogram` while generating `percentileBreaks`. Here's some code that does that:

```scala
/**
    * For each q in qs, all between 0 and 1, find a number
    * (approximately) at the qth percentile.
    *
    * @param qs   A sequence of quantiles from which to generate breaks in tile values
    */
  def interpolatedPercentileBreaks(qs: Seq[Double]): Seq[Double] =
    if(buckets.size == 1) {
      qs.map(z => buckets.head.label)
    } else {
      val data = cdfIntervals
      if(!data.hasNext) {
        Seq()
      } else {
        val result = MutableListBuffer[Double]()
        var curr = data.next

        def getValue(q: Double): Double = {
          val (d1, pct1) = curr._1 // Prior bin
          val (d2, pct2) = curr._2 // Next bin

          // If q is less than the prior bin's pctile (first few q, perhaps)
          if (q == pct1) {
            d1
          } else if (q > pct1) {
            val x = (q - pct1) / (pct2 - pct1)
            (1 - x) * d1 + x * d2
          } else { // (q < pct1)
            val x = (pct1 - q) / (pct2 - pct1)
            d1 - (x * d1)
          }
        }

        // For each quantile
        for(q <- qs) {
          // If 0th qtile, -Infinity
          if (q == 0.0) { result += minValue().getOrElse(Double.NegativeInfinity) }
          // If 100th qtile, Infinity
          else if (q == 1.0) { result += maxValue().getOrElse(Double.PositiveInfinity) }
          // If between 0 and 100, do some math
          else {
            // First, ensure the correct interval is ready, then get the value
            while(data.hasNext && q >= curr._2._2) { curr = data.next }
            result += getValue(q)
          }
        }

        result.toSeq
      }
    }
```

Any thoughts on this approach?

moradology:
I suggest it because the repeated lower-end values seems very strange to the uninitiated and because the rendered values won't be different (unless a different breaks boundary type is selected - Exact or GreaterThan, say, which maybe aren't useful anyway(?))

jamesmcclain:
Looks fine to me.

moradology:
This is resolved, closing

PR #1500, opened 2016-05-26
Add logic to ensure use of minval/avoid repetition of breaks

### LINKED ISSUE #1462: Regression in classBreaks?
With 0.10.0-RC3 (and previous versions) a `classBreaks` call on test data returned
```
[346600,449400,542600,613000,688800,792800,902800,1148600,1571800,2000002]
```
With 0.10.0 the same call on the same data returns
```
[-2147483648,-2147483648,-2147483648,-2147483648,116075,434547,610418,788548,1172822,2000002]
```
Looks like it's including NODATA values?

The call is here: https://github.com/OpenTreeMap/otm-modeling/blob/6d8c63c4c756f07adb7f6995c776d4e7e18cd8bb/tile/src/main/scala/TileService.scala#L97

(`model` has type `TileLayerRDD[SpatialKey]`)

moradology:
Fixes #1462 and ameliorates #1499

This set of changes makes sure that any `ClassBreaks` produced will start with the minimum value and interpolate the values in between according to the bins observed by the streaming histogram.

### DIFF
--- a/raster-test/src/test/scala/geotrellis/raster/histogram/StreamingHistogramSpec.scala
+++ b/raster-test/src/test/scala/geotrellis/raster/histogram/StreamingHistogramSpec.scala
@@ -156,12 +156,15 @@ class StreamingHistogramSpec extends FunSpec with Matchers {
       hist.quantileBreaks(5) should be (Seq(1.0, 1.0, 1.0, 1.0, 1.0))
     }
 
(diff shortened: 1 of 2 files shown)


jamesmcclain:
Logically sound.

fosskers:
This patch applied to `us-census-population-density`, which previously wouldn't render properly:

population-density2

fosskers:
Previous state:

population-density

PR #1501, opened 2016-05-31
SparkContext temporal GeoTiff format args

echeipesh:
To read temporal GeoTiffs we need to specify the tag and date format, which is not standard. Without this fix there is no way to do this outside using `TemporalGeoTiffInputFormat`. This PR adds the two required arguments to the extension methods on `SparkContext` with default values.


pomadchin:
May make [these lines](https://github.com/pomadchin/geotrellis-integration-tests-tool/blob/master/src/main/scala/geotrellis/util/SparkSupport.scala#L29-L34) easier.

PR #1502, opened 2016-06-01
Conversions between GeoTools GridCoverage2D and GeoTrellis Raster types

lossyrob:
Supersedes #1444


jamesmcclain:
+1  This looks good to me.  There is quite a bit of test coverage and I was able to confirm that this produces `GridCoverage2D`s which are compatible with GeoWave and viewable in the GeoServer.

I think that a valuable contribution would have been a concrete suggestion for compactifying [some of the larger chunks of code](https://github.com/geotrellis/geotrellis/pull/1502/files#diff-04a7bf59a70ae9e4bbdf78707a1b84e0R252), but unfortunately -- to the best of my knowledge -- the visual/textual similarity between different blocks of code is not necessarily easily addressed in a statically-typed language without using higher-order functions.

geotools/src/main/scala/geotrellis/geotools/GridCoverage2DConversionMethods.scala
```diff
@@ -0,0 +1,59 @@
...
+    Raster(tile, extent)
+  }
+
+  def toProjectedRaster(bandIndex: Int): ProjectedRaster[Tile] =
+    GridCoverage2DConverters.getCrs(self) match {
+      case Some(crs) =>
+        ProjectedRaster(toRaster(bandIndex), crs)
+      case None =>
+        // Default LatLng
+        ProjectedRaster(toRaster(bandIndex), LatLng)
```
jamesmcclain: I wonder if it is worth emitting a warning here and/or giving a big warning in the source code that a CRS is being assumed.

geotools/src/main/scala/geotrellis/geotools/GridCoverage2DConverters.scala
```diff
@@ -0,0 +1,893 @@
...
+import scala.collection.JavaConverters._
+
+/**
+  * Houses methods that aide the translate
+  * between GridCoverage2D and related GeoTools types,
+  * and GeoTrellis types.
+  */
+object GridCoverage2DConverters {
+  /**
+    * Given a [[GridCoverage2D]] and an index, this function
```
jamesmcclain: These ScalaDocs will produce warnings, I think.

jamesmcclain:
I think that `geotools` needs to be added to [this list](https://github.com/lossyrob/geotrellis/blob/a9108606e655909b708385bac6de4344ccaec9ed/build.sbt#L58-L68) in order for the documentation to be buildable with the sbt `doc` command.

lossyrob:
@jamesmcclain good call. Can you do that in you do that in your SimpleFeature PR?

jamesmcclain:
Indeed.

Issue #1477, opened 2016-05-09
User-friendly errors for Tile vs MultibandTile


timothymschier:
When using ```Reader[SpatialKey, Tile]``` to read a MultibandTile source, an Avro exception is shown. This isn't friendly for new users, who may not be able to spot that they should be using ```Reader[SpatialKey, MultibandTile]``` instead.
A suggested improvement is throwing an error based on the metadata ```valueClass``` value.
Similarly ```keyClass``` could be used to help with index keys.

moradology:
Example stack trace:
```scala
org.apache.avro.AvroTypeException: Found geotrellis.raster.ByteArrayTile, expecting geotrellis.raster.ArrayMultibandTile, missing required field bands
	at org.apache.avro.io.ResolvingDecoder.doAction(ResolvingDecoder.java:292)
	at org.apache.avro.io.parsing.Parser.advance(Parser.java:88)
	at org.apache.avro.io.ResolvingDecoder.readFieldOrder(ResolvingDecoder.java:130)
	at org.apache.avro.generic.GenericDatumReader.readRecord(GenericDatumReader.java:223)
	at org.apache.avro.generic.GenericDatumReader.readWithoutConversion(GenericDatumReader.java:174)
	at org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:152)
	at org.apache.avro.generic.GenericDatumReader.readField(GenericDatumReader.java:240)
	at org.apache.avro.generic.GenericDatumReader.readRecord(GenericDatumReader.java:230)
	at org.apache.avro.generic.GenericDatumReader.readWithoutConversion(GenericDatumReader.java:174)
	at org.apache.avro.generic.GenericDatumReader.readArray(GenericDatumReader.java:274)
	at org.apache.avro.generic.GenericDatumReader.readWithoutConversion(GenericDatumReader.java:176)
	at org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:152)
	at org.apache.avro.generic.GenericDatumReader.readField(GenericDatumReader.java:240)
	at org.apache.avro.generic.GenericDatumReader.readRecord(GenericDatumReader.java:230)
	at org.apache.avro.generic.GenericDatumReader.readWithoutConversion(GenericDatumReader.java:174)
	at org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:152)
	at org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:144)
[...]

PR #1505, opened 2016-06-01
Add detail to avro exception message

moradology:
This PR adds a little bit of information necessary to properly understand why the Avro decoding process fails in some circumstances and is a response to #1477

The new error looks something like:
```scala
org.apache.avro.AvroTypeException: Found geotrellis.raster.ByteArrayTile, expecting geotrellis.raster.ArrayMultibandTile, missing required field bands. If attempting to read multiband tiles, use a Reader[K, MultibandTile] rather than Reader[K, Tile]
```

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/io/avro/AvroEncoder.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/avro/AvroEncoder.scala
@@ -53,7 +53,13 @@ object AvroEncoder {
     val reader = new GenericDatumReader[GenericRecord](writerSchema, schema)
     val decoder = DecoderFactory.get().binaryDecoder(decompress(bytes), null)
     val rec = reader.read(null.asInstanceOf[GenericRecord], decoder)
-    format.decode(rec)
+    try {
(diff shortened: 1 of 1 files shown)


lossyrob:
@moradology what is going on with this PR?

moradology:
I investigated a few paths one might take to address this issue and ended up deciding the best path at this point is also the simplest. It seems as though we might want to consider parameterizing the names of generic avro codecs - this would give us a greater capacity to intelligently handle certain errors and to provide useful information in the case of failure.

For instance, the name here (https://github.com/geotrellis/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/io/avro/codecs/KeyValueRecordCodec.scala#L13) could be parameterized on the names of the underlying Key and Value codecs

Here's an example of the exception thrown after applying this PR: `org.apache.avro.AvroTypeException: Found geotrellis.raster.ByteArrayTile, expecting geotrellis.raster.ArrayMultibandTile, missing required field bands. Check to ensure that the proper type parameters for this Reader have been provided.`

PR #1511, opened 2016-06-07
Remove the Kryo registration requirement

jpolchlo:
This PR removes the requirement that classes be registered with the KryoSerializer within any specs that derive from TestEnvironment.  This is meant to avoid having to register classes of the form, e.g.,
```
classOf[geotrellis.spark.SpatialKey$$anonfun$ordering$1]
```
These internal names are fragile and can cause problems during integration testing that do not arise during local testing.


spark/src/test/scala/geotrellis/spark/TestRegistrator.scala
```diff
@@ -13,36 +13,16 @@ class TestRegistrator extends NormalKryoRegistrator {
     super.registerClasses(kryo)
     if (Properties.envOrNone("GEOTRELLIS_KRYO_REGREQ") != None) {
```
lossyrob: This should be removed, no?

lossyrob:
@jpolchlo tip: run `scripts/buildall.sh` locally (which goes faster than Travis) to catch things like the latest travis error before pushing up

lossyrob:
Merge order issue; can you update to master? Cassandra subproject has a registrator that needs to be removed @jpolchlo

PR #1512, opened 2016-06-07
ColorMap => String

fosskers:
## TODO
- [x] Basic functionality
- [x] Move `breaksString` into trait and inherit
- [x] Special definition for `IntCachedColorMap`
- [x] Tests

## Motivation
When using the `render` output module in the ETL process, one must provide a `breaks` argument, or all tiles will be rendered grayscale. Example:

```console
--output render -O encoding=png path=file:///home/colin/tiles/{name}/{z}-{x}-{y}.png breaks="23:cc00ccff;30:aa00aaff;120:ff0000ff"
```
The problem is that for datasets which don't have well-defined colour breaks ahead of time like `nlcd`, there's no way to know what your break string should be for input into ETL. A `ColorMap` object can be created easily given `RDD.colorBreaks` and then rendered to a PNG within Scala code, but there is no way to know what `String` would represent that `ColorMap` from the outside. This chicken-and-egg problem requires a greater fix to how ETL accepts breaks information from the outside, but this is a start.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/render/ColorMap.scala
+++ b/raster/src/main/scala/geotrellis/raster/render/ColorMap.scala
@@ -264,7 +264,7 @@ class IntColorMap(breaksToColors: Map[Int, Int], val options: Options = Options.
     new IntCachedColorMap(orderedColors, ch, options)
   }
 
-  lazy val breaksString: String = {
+  def breaksString: String = {
     breaksToColors
       .toStream
       .map({ case (k,v) => s"${k}:${Integer.toHexString(v)}"})
@@ -381,7 +381,7 @@ class DoubleColorMap(breaksToColors: Map[Double, Int], val options: Options = Op
   def withBoundaryType(classBoundaryType: ClassBoundaryType): ColorMap =
     new DoubleColorMap(breaksToColors, options.copy(classBoundaryType = classBoundaryType))
 
-  lazy val breaksString: String = {
+  def breaksString: String = {
     breaksToColors
       .toStream
       .map({ case (k,v) => s"${k}:${Integer.toHexString(v)}"})
--- a/raster/src/main/scala/geotrellis/raster/render/ColorMap.scala
+++ b/raster/src/main/scala/geotrellis/raster/render/ColorMap.scala
@@ -206,8 +206,6 @@ class IntColorMap(breaksToColors: Map[Int, Int], val options: Options = Options.
 
   private lazy val orderedColors: Vector[Int] = orderedBreaks.map(breaksToColors(_))
   lazy val colors = orderedColors
-  lazy val breaksMap = breaksToColors
-  lazy val breaksMapDouble = Map.empty[Double,Int]
 
   private val zCheck: (Int, Int) => Boolean =
     options.classBoundaryType match {
@@ -282,9 +280,6 @@ class IntCachedColorMap(val colors: Vector[Int], h: Histogram[Int], val options:
     extends ColorMap {
   val noDataColor = options.noDataColor
 
-  lazy val breaksMap = Map.empty[Int,Int]
-  lazy val breaksMapDouble = Map.empty[Double,Int]
-
   def map(z: Int): Int = { if(isNoData(z)) noDataColor else h.itemCount(z).toInt }
 
   def mapDouble(z: Double): Int = map(d2i(z))
@@ -334,8 +329,6 @@ class DoubleColorMap(breaksToColors: Map[Double, Int], val options: Options = Op
 
   private val orderedColors: Vector[Int] = orderedBreaks.map(breaksToColors(_))
   lazy val colors = orderedColors
-  lazy val breaksMap = Map.empty[Int,Int]
-  lazy val breaksMapDouble = breaksToColors
 
   private val zCheck: (Double, Int) => Boolean =
     options.classBoundaryType match {
--- a/raster/src/main/scala/geotrellis/raster/render/ColorMap.scala
+++ b/raster/src/main/scala/geotrellis/raster/render/ColorMap.scala
@@ -143,6 +143,15 @@ object ColorMap {
     }.toOption
   }
 
+  /** Retrieve a "breaks string" from [[ColorMap]] data.
+    * The opposite of the [[fromString]] methods.
+    */
+  def breaksString[T <: AnyVal](breaksToColors: Map[T, Int]): String = {
+    breaksToColors
+      .toStream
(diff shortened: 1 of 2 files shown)


fosskers:
To test:

```console
sbt
project raster
console
import geotrellis.raster.render.ColorMap
val m = ColorMap.fromString("23:cc00ccff;30:aa00aaff;120:ff0000ff").get
ColorMap.breaksString(m.breaksMap)
```

raster/src/main/scala/geotrellis/raster/render/ColorMap.scala
```diff
@@ -143,6 +143,15 @@ object ColorMap {
...
+  /** Retrieve a "breaks string" from [[ColorMap]] data.
+    * The opposite of the [[fromString]] methods.
+    */
+  def breaksString[T <: AnyVal](breaksToColors: Map[T, Int]): String = {
+    breaksToColors
+      .toStream
+      .map({ case (k,v) => s"${k}:${Integer.toHexString(v)}"})
+      .mkString(";")
+  }
 }
```
fosskers: I don't want to have to write two functions just to distinguish between the `Int` and `Double` keys in the map, here. The function's implementation would be exactly the same.

What I really want to write is:

```scala
def breakString[T <: Num](...): String = { ... }
```
but Scala doesn't have a specialized number typeclass.

raster/src/main/scala/geotrellis/raster/render/ColorMap.scala
```diff
@@ -202,6 +215,8 @@ class IntColorMap(breaksToColors: Map[Int, Int], val options: Options = Options.
 
   private lazy val orderedColors: Vector[Int] = orderedBreaks.map(breaksToColors(_))
   lazy val colors = orderedColors
+  lazy val breaksMap = breaksToColors
+  lazy val breaksMapDouble = Map.empty[Double,Int]
```
lossyrob: Why do we need a breaks map? Can't we just implement this functionality as the `toString` method?

fosskers:
Really, the best thing would be for `ColorMap` to be polymorphic on its internal number type:
```scala
trait ColorMap[T] extends Serializable { ... }
```
And then duplicate code between `IntColorMap` and `DoubleColorMap` could be removed. Then, the `ColorMap` trait could ask for a `def breaksMap: Map[T,Int]`, and `breaksString` could be moved into the trait defined as:
```scala
def breaksString: String = {
  breaksMap  // was `breaksToColors` passed in as an argument
       .toStream
       .map({ case (k,v) => s"${k}:${Integer.toHexString(v)}"})                                           
       .mkString(";")
}
```
The huge downside of this would be that client code would have to care about the parameterized `ColorMap[T]`, unless we used some cool type aliasing so nothing breaks.
Thoughts, @lossyrob @echeipesh ?

raster/src/main/scala/geotrellis/raster/render/ColorMap.scala
fosskers: Temporary hack, see below.

fosskers:
Another trip-up: the user-level API doesn't expose `IntColorMap`, etc, only `ColorMap`. The colouring function has to be visible to the trait. You could put it in the companion object (as I've done initially here) but then the arguments to it must also be visible to the trait. `IntCachedColorMap` throws a wrench into that, as it has no `Map` internally, but still extends `ColorMap`.

lossyrob:
Your hitting up against the need for us not to abstract over `Int` and `Double` typed things due to performance implications. We jump through hoops to have Double and Int versions of things because if we try to make it generic, things slow down in very painful ways - boxing is the enemy, and ColorMaps are performance critical code that must not box.

I'm not sure "it not being good form in Haskell" is a good reason to avoid using the method that's there, defined as returning a [string representation of the object](http://docs.oracle.com/javase/6/docs/api/java/lang/Object.html#toString%28%29). One could argue that the string representation of a color map that you are concocting is actually a good representation for a user trying to read what the color map is; however, even if we wanted to differentiate between a "show" `toString` and some other form (`def breaksString: String`), we could still use implementations in the concrete classes to take care of the Double vs Int bits and avoid the generic `breaksString` method you have.

fosskers:
> haskell

That was more of an explanation as to my natural want to avoid that pattern, not necessarily that we should do it that way "cause Haskell".

re: my last comment. `override toString` could work for the inheritance problems.

lossyrob:
Gotcha. Yeah I think the argument could be made though that we would want to avoid using `toString` because it's too generic of a thing, and instead use the `breakString` with inheritance.

fosskers:
Meaning the actual definition of `breaksString` will be duplicated between the `*ColorMap` classes for performance reasons? I'll do it that way if that's the prevailing pattern.

lossyrob:
More to do with why we can't do

`trait ColorMap[T] extends Serializable { ... }`

and be generic on T.

But code duplication in this case might be tough to avoid, since the code to deal with doubles and ints will look the same (but not with the cached version of int, which you mentioned above).

raster/src/main/scala/geotrellis/raster/render/ColorMap.scala
```diff
@@ -259,6 +263,13 @@ class IntColorMap(breaksToColors: Map[Int, Int], val options: Options = Options.
     h.foreachValue(z => ch.setItem(z, map(z)))
     new IntCachedColorMap(orderedColors, ch, options)
   }
+
+  lazy val breaksString: String = {
```
lossyrob: Why lazy val for this and the other one?

fosskers: For the same reason that `orderedColors` is lazy, so that these potentially unneeded values aren't calculated upon `ColorMap` instantiation.

lossyrob: I wasn't thinking vals...why not defs? Is there ever a reason to hold them in memory after calculation (e.g. its an expensive operation that will be reused many times to justify the instance footprint)?

fosskers: I can't see it being called over and over for one instance of a `ColorMap`. I suppose the GC behaviour for these are:

* `val`: Eagerly evaluate, keep in memory
* `lazy val`: Lazily evaluate, and then keep in memory
* `def`: Lazily evaluate, don't keep in memory

Then perhaps it's best as a `def`.

lossyrob: Agreed.
+1 after this change.

fosskers:
Good to go, @lossyrob .

PR #1538, opened 2016-06-09
Add split by cols/rows to SplitMethods

echeipesh:
This PR adds `split(cols: Int, rows: Int)` overload to split methods. Existing methods required an input of `TileLayout`. In order to be able to perform this operation we impose the restriction that `SplitMethods` must operate on `Grid` types.

`Grid` type is new, it is super-class of `CellGrid`. Specifically we want to be able to split `RasterExtent` which does not have an associated `CellTypes`.

Now that type `Grid` exists there is an outstanding question of what interfaces should be loosened from `CellGrid` to `Grid`. This PR does not answer that question.


raster/src/main/scala/geotrellis/raster/Raster.scala
```diff
@@ -64,7 +64,7 @@ case class Raster[+T <: CellGrid](tile: T, extent: Extent) extends Product2[T, E
 
   def cols: Int = tile.cols
   def rows: Int = tile.rows
-  def dimensions: (Int, Int) = tile.dimensions
```
lossyrob: Why get rid of dimensions?

raster/src/main/scala/geotrellis/raster/RasterExtent.scala
```diff
@@ -67,18 +67,12 @@ case class GeoAttrsError(msg: String) extends Exception(msg)
   * RasterExtent.
   */
 case class RasterExtent(extent: Extent, cellwidth: Double, cellheight: Double, cols: Int, rows: Int)
-    extends GridDefinition {
+    extends GridDefinition with Grid {
```
lossyrob: Shouldn't a `GridDefinition` be a grid?

lossyrob: Weird a `GridDefinition` isn't a grid. Is this the best naming? Where a `Grid` can only contain Int cols and rows?

raster/src/main/scala/geotrellis/raster/Raster.scala
echeipesh: You get those from `CellGrid`, no need to overwrite here.

echeipesh:
@lossyrob: I thought about your comments on `GridDefinition` and realized that `GridDefinition` was just a trait for `GridExtent`. Merging those two and making the `GridExtent` -> `RasterExtent` hierarchy more explicit makes that code less confusing. There is still weird forward reference to `RasterExtent` with `toAllignedRasterExtent` method, but I think it's acceptable.

raster/src/main/scala/geotrellis/raster/GridExtent.scala
```diff
@@ -2,29 +2,115 @@ package geotrellis.raster
...
+  }
+}
+
+
+object GridExtent {
+  def apply(extent: Extent, cellSize: CellSize): GridExtent =
+    new GridExtent(extent, cellSize.width, cellSize.height)
+
+  def apply(extent: Extent, cellwidth: Double, cellheight: Double): GridExtent =
+    new GridExtent(extent, cellwidth, cellheight)
```
lossyrob: We lose an unapply here by taking away it's case classy-ness. Also proper equality. I get that you want no case class for inheritance - we should provide that functionality ourselves.

lossyrob:
Not super psyched about changing the API after the "API freezing" 0.10 release, but this is a small change and a good change, and a good one to rip the scratch the new floors with. +1 after the case class functionality reimplementation.

PR #1539, opened 2016-06-09
[WIP] Stream Reading for GeoTiffs

MyNameIsBouf:


lossyrob:
@MyNameIsBouf There are commits in this PR that don't have the correct email address.

Also, please add a description that conveys what the feature is trying to do, so that if someone who is unfamiliar with the project looks at this PR they will know what is going on and potentially be able to contribute to the conversation. Linking to the issue is good too.

raster-test/src/test/scala/geotrellis/raster/io/geotiff/reader/GeoTiffReaderSpec.scala
```diff
@@ -517,17 +517,13 @@ class GeoTiffReaderSpec extends FunSpec
...
 }
 
+
+
 class PackBitsGeoTiffReaderSpec extends FunSpec
     with RasterMatchers
     with GeoTiffTestUtils {
 
   describe("Reading geotiffs with PACKBITS compression") {
-    it("must read a single band bit raster as a multiband") {
```
lossyrob: Is this a duplicate of a test above?

raster-test/src/test/scala/geotrellis/raster/io/geotiff/reader/SinglebandGeoTiffReaderSpec.scala
```diff
@@ -68,7 +68,7 @@ class SinglebandGeoTiffReaderSpec extends FunSpec
       assertEqual(actual, expected)
     }
 
-    it("should be able to collect points from a large sparse compressed geotiff that couldn't fit all into memory") {
+    it("should be able to collect points from a large sparse Compressed geotiff that couldn't fit all into memory") {
```
lossyrob: ?

raster/src/main/scala/geotrellis/raster/io/geotiff/reader/GeoTiffStreamReader.scala
```diff
@@ -0,0 +1,303 @@
...
+
+import monocle.syntax.apply._
+
+import scala.io._
+import scala.collection.mutable
+import java.nio.{ByteBuffer, ByteOrder}
+import spire.syntax.cfor._
+
+
+object GeoTiffStreamReader {
```
lossyrob: We shouldn't have a mostly duplicated object for this.

We should be reading tiff tags from a byte buffer, and that byte buffer should be backed by a file map when reading the tifftags from a path (and not as part of the GeoTiff reading process).

Streaming reads are not useful if we're reading in the whole raster. It will be useful when we are reading a window out, but we're not there yet.

raster/src/main/scala/geotrellis/raster/io/geotiff/reader/TiffTagsReader.scala
```diff
@@ -14,16 +14,13 @@ import java.nio.{ ByteBuffer, ByteOrder }
...
-
-    // Set byteBuffer position
-    byteBuffer.position(0)
+  def read(path: String): TiffTags = read(Filesystem.streamByteBuffer(path))
+  
+  def read(byteBuffer: ByteBuffer): TiffTags = {
 
     // set byte ordering
-    (byteBuffer.get.toChar, byteBuffer.get.toChar) match {
+    (byteBuffer.reader.toChar, byteBuffer.reader.toChar) match {
```
lossyrob: Can you explain this change?

util/src/main/scala/geotrellis/util/Filesystem.scala
```diff
@@ -53,6 +53,24 @@ object Filesystem {
   }
 
   /**
+    * Read the contents of a file into a ByteBuffer.
+    *
+    * @param   path The path to the file to be read
+    * @return       A ByteBuffer that contains the file
+    */
+  def streamByteBuffer(path: String): ByteBuffer = {
```
lossyrob: I think "stream" here is not the right nomenclature. What exactly is this byte buffer that is being returned? what is the type of `channel`...is this really a file map buffer? Let's make sure to correctly name it.

raster/src/main/scala/geotrellis/raster/io/geotiff/reader/GeoTiffStreamReader.scala
MyNameIsBouf: Okay, I'll remove it for the time being and rework it later when it needs to be added.

raster/src/main/scala/geotrellis/raster/io/geotiff/reader/TiffTagsReader.scala
MyNameIsBouf: That was code that was left over from my original implementation. I though I had removed it. Sorry about that.

PR #1541, opened 2016-06-10
Reading TiffTags via MappedByteBuffer

MyNameIsBouf:


raster/src/main/scala/geotrellis/raster/io/geotiff/reader/TiffTagsReader.scala
```diff
@@ -14,14 +14,11 @@ import java.nio.{ ByteBuffer, ByteOrder }
 
 
 object TiffTagsReader {
-  def read(path: String): TiffTags = read(Filesystem.slurp(path))
-
-  def read(bytes: Array[Byte]): TiffTags = {
```
lossyrob: Should have overload that takes an Array[Byte]

```scala
def read(bytes: Array[Byte]): TiffTags =
  read(ByteBuffer.wrap(bytes, 0, bytes.size))
```

util/src/main/scala/geotrellis/util/Filesystem.scala
```diff
@@ -53,6 +53,24 @@ object Filesystem {
   }
 
   /**
+    * Read the contents of a file into a MappedByteBuffer.
+    *
+    * @param   path The path to the file to be read
+    * @return       A MappedByteBuffer containing the mapped file contents
+    */
+  def toMappedByteBuffer(path: String): ByteBuffer = {
```
lossyrob: Should call this method as the first part of `slurp`

lossyrob:
👍

PR #1551, opened 2016-06-15
Refactor proj wrapper to allow other referencing impls

dwins:
This updates the proj wrapper to factor all the calls that actually use proj4j into one `Backend` object, making it simple to replace that with other implementations (an alternative based on [Apache SIS](http://sis.apache.org/) is provided.)  However there are a few places in the geotrellis codebase where we appear to rely on proj.4 parameter strings (notably the GeoTiff parser, although there also seems to be some spark code which is putting parameter strings into JSON to serialize CRS's.)  I updated these to compile but I fully expect them to fail if the SIS backend is engaged.

### DIFF
--- a/proj4/build.sbt
+++ b/proj4/build.sbt
@@ -3,6 +3,8 @@ import Dependencies._
 name := "geotrellis-proj4"
 
 libraryDependencies ++= Seq(
+  apacheSIS,
+  apacheSISEPSG,
   openCSV,
   scalatest   % "test",
   scalacheck  % "test")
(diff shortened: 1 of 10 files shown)


dwins:
Also, the [remarks](http://sis.apache.org/epsg.html#maven) on the SIS website regarding the EPSG database are relevant for anyone trying to use it in GeoTrellis. This means that each node in a Spark cluster which might be doing reprojection work would need this setup applied.

lossyrob:
Closing, referenced in the appropriate issue so that we can track for future reference.

PR #1553, opened 2016-06-17
ETL Refactor

pomadchin:
Move ETL to use json configuration

- [x] classes to read to / from json
- [x] refactor etl input / output plugins to work with new inputs
- [x] update documentation
- [x] deprecate scallop
- [x] test on a real ingest (chatta-demo ingest / emr)
- [x] refactor landsat demo to work with this pr
- [x] refactor chatta demo to work with this pr
- [x] output into a separate file
- [x] improve json validation (our exceptions improvements and schema validation(?))
- [x] @echeipesh comments


pomadchin:
`datasets.json`:
```json
[  
   {  
      "name":"test",
      "ingestType":{  
         "format":"geotiff",
         "inputCredentials":"inputCredentials name",
         "output":"hadoop",
         "outputCredentials":"outputCredentials name",
         "input":"hadoop"
      },
      "path":{  
         "input":"input",
         "output":"output"
      },
      "cache":"NONE",
      "ingestOptions":{  
         "breaks":"0:ffffe5ff;0.1:f7fcb9ff;0.2:d9f0a3ff;0.3:addd8eff;0.4:78c679ff;0.5:41ab5dff;0.6:238443ff;0.7:006837ff;1:004529ff",
         "reprojectMethod":"buffered",
         "cellSize":"256.0,256.0",
         "encoding":"geotiff",
         "tileSize":256,
         "layoutExtent":{  
            "xmin":1.0,
            "ymin":2.0,
            "xmax":3.0,
            "ymax":4.0
         },
         "resolutionThreshold":0.1,
         "pyramid":true,
         "resampleMethod":"nearest-neighbor",
         "keyIndexMethod":{  
            "type":"zorder"
         },
         "layoutScheme":"tms",
         "cellType":"int8",
         "crs":"+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs "
      }
   }
]
```

lossyrob:
Does this allow for additional backends to be supplied via code in client projects?

pomadchin:
@lossyrob good point, would add

pomadchin:
@lossyrob yep, https://github.com/pomadchin/geotrellis/blob/feature/etl-json/spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala#L27

pomadchin:
should include #1494

RickMohr:
Will there be validation for the input JSON? If I type `resampleMetod` by mistake it would be fantastic to get an "Unknown property - resampleMetod" error rather than a silent baffling result.

(There are JSON schema validators for most languages.)

pomadchin:
@RickMohr a good point, sure, more over our json validation should be improved too. thx for that notice

docs/spark-etl/spark-etl-run-examples.md
```diff
@@ -14,68 +14,204 @@ spark-submit \
...
---class geotrellis.spark.etl.{SinglebandIngest | MultibandIngest} \
---master local[*] \
---driver-memory 2G \
-$JAR \
---input hadoop --format {geotiff | temporal-geotiff} --cache NONE -I path="file:///Data/nlcd/tiles" \
---output accumulo -O instance="gis-accumulo-instance" table="nlcd-table" user="root" password="password" zookeeper="zoo1:2181,zoo2:2181,zoo3:2181" \
---layer nlcd-tms --crs EPSG:3857 --pyramid --layoutScheme {tms | floating}
+### Ingest tiles from local fs or hdfs into s3 storage command
+
+`datasets.json`:
```
echeipesh: Would it make sense to have input dataset and output config to be in separate files? This way you can have multiple sources and only one destination like `--output azavea-datahub.json`

pomadchin: @echeipesh a very good point

pomadchin:
@RickMohr have you got any comments? (i decided to use https://github.com/daveclayton/json-schema-validator, looks like it has larger community and supported right now)

spark-etl/src/main/scala/geotrellis/spark/etl/config/EtlConf.scala
```diff
@@ -0,0 +1,68 @@
...
+import geotrellis.spark.etl.EtlJob
+import geotrellis.spark.etl.config.json._
+
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.Path
+import org.apache.spark.SparkContext
+import spray.json.DefaultJsonProtocol._
+import spray.json._
+
+class EtlConf(val credentials: Credentials, val datasets: List[Input], val output: Output) {
```
echeipesh: I think its a cool idea to be able to have multiple inputs into an ETL job but `EtlJob` only takes one input, so it doesn't seem like having a list of them here is used?

spark-etl/src/main/scala/geotrellis/spark/etl/config/EtlConf.scala
```diff
@@ -0,0 +1,68 @@
...
+  def getEtlJobs = datasets map { ds => EtlJob(ds, output, credentials.getInput(ds), credentialsOutput) }
+}
+
+object EtlConf {
+  val help = """
+               |geotrellis-etl
+               |
+               |Usage: geotrellis-etl [options]
+               |
+               |  --datasets <value>
```
echeipesh: should be `input` just to be consistent with what was and `output`

spark-etl/src/main/scala/geotrellis/spark/etl/config/Credentials.scala
```diff
@@ -0,0 +1,24 @@
+package geotrellis.spark.etl.config
+
+case class Credentials(accumulo: List[Accumulo], cassandra: List[Cassandra], s3: List[S3], hadoop: List[Hadoop]) {
```
echeipesh: The job of this class seems to be to parse out the backend specific parameters, this is something that should be handled by constructor or companion object of each particular back-end.

echeipesh:
Ok, my biggest comment is on handling of `credentials.json`. In reality it serves two related purposes, to provide credentials and to provide some back-end specific configuration, that's great. We just need to make it a lookup table that we can call out to from the `input.json` Currently it calls out the type of the back-end as part of the JSON structure, which is kind of strange. Ideally this is a name -> config table like;
```json
{
  "backend-profiles": [
    {
      "name": "accumulo-gis",
      "type": "Accumulo",
      "zookeepers": "zookeepers",
      "instance": "instance",
      "user": "user",
      "password": "password"
    },
    {
      "name": "cassandra-gis",
      "type": "Cassandra"
      "allowRemoteDCsForLocalConsistencyLevel": false,
      "localDc": "datacenter1",
      "usedHostsPerRemoteDc": 0,
      "hosts": "hosts",
      "replicationStrategy": "SimpleStrategy",
      "replicationFactor": 1,
      "user": "user",
      "password": "password"
    }
  ]
}
```

Now related problem is that an input needs to be able to refine those configurations by adding some kind of path/table/schema selection and configure an InputPlugin that will actually fetch me some bytes. It can look like:
`input.json`:
```json
   {
      "name":"nlcd-tms",
      "format":"{geotiff | temporal-geotiff}",
      "backend":{
        "type":"file"
        "profile": "datahub-local",
        "path": "file:///Data/nlcd/tiles",
      },
      "NoData": 0,
      "cache":"NONE",
   }
```
Here we're going to ask for everything that affects reading the tiles: what format they're in, if we should cache them, if we should force a NoData while while reading them (because nobody cares to set it correctly when writing the files) and most importantly the `backend` to use. The backend js object will select the credentials/config profile and add some information that identifies a dataset in there.

I think you could get a little fancy by basically merging the profile with the `backend` block, which would allow you to shift the configuration from the profile into the input section, but lets say that's not justified here.

Then final comment is that the output is actually what determines the index and layout used, so those sections need to be moved there:

```json
{
  "backend": {
      "type": "S3",
      "profile": "datahub-s3"
      "prefix": "/catalog"
  },
  "reprojectMethod":"buffered",
  "pyramid":true,
  "resampleMethod":"nearest-neighbor",
  "keyIndexMethod":{
    "type":"zorder"
  },
  "layoutScheme": { // just map the fields directly to zoomed and floating layout scheme objects
     "type": "zoomed",
     "tileSize": 256,
     "crs":"EPSG:3857"
  },
}
```

Hopefully by dealing with backend profiles this way you can even generate a case class that you can pass to the input module like `AccumuloBackend` that will have all the fields set. You could parse them by partially reading the JSON object to read the "type" and then using the correct format to read it fully and shove them into a `Map[String, Backend]` or something like that.

I like the idea of the CustomBackend you have and that fits pretty nicely into this flow as well.

spark-etl/src/main/resources/credentials-schema.json
```diff
@@ -0,0 +1,96 @@
...
+            "type": "string"
+          }
+        },
+        "required": [
+          "name",
+          "zookeepers",
+          "instance",
+          "user",
+          "password"
+        ]
```
RickMohr: Consider adding
```
"additionalProperties": false
```
(Explained e.g. at https://spacetelescope.github.io/understanding-json-schema/reference/object.html)

RickMohr:
Thanks for adding validation!

spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala
```diff
@@ -1,69 +1,69 @@
...
+      /* perform the reprojection and mosaicing step to fit tiles to LayoutScheme specified */
+      val (zoom, tiled) = etl.tile(sourceTiles)
+      /* save and optionally pyramid the mosaiced layer */
+      etl.save[K, V](LayerId(etl.input.name, zoom), tiled)
+    }
   }
 }
 
-case class Etl(args: Seq[String], @transient modules: Seq[TypedModule] = Etl.defaultModules) {
+case class Etl(@transient etlJob: EtlJob, @transient modules: Seq[TypedModule] = Etl.defaultModules) {
```
lossyrob: Why are these marked as transient?

I think the consequence of this would be to have null values assigned to the public vals on serialization and deserialization.

pomadchin: @lossyrob uff, would make it serializable ._.

lossyrob:
@pomadchin can you update? I think we're ready for this to be merged, yeah?

pomadchin:
@lossyrob yes, sure

pomadchin:
merged master, though need to tests on any demo, how / if hbase modules work

spark-etl/src/main/scala/geotrellis/spark/etl/config/BackendProfiles.scala
```diff
@@ -0,0 +1,47 @@
...
+
+import geotrellis.spark.io.accumulo.AccumuloInstance
+import geotrellis.spark.io.hbase.HBaseInstance
+import geotrellis.spark.io.cassandra.{BaseCassandraInstance, Cassandra}
+
+import org.apache.accumulo.core.client.security.tokens.PasswordToken
+
+sealed trait BackendProfile {
+  val name: String
+  def `type`: String
```
echeipesh: This can actually be `BackendType` instead of a string

spark-etl/src/main/scala/geotrellis/spark/etl/config/BackendType.scala
```diff
@@ -0,0 +1,58 @@
...
+  val name = "hadoop"
+}
+
+case object FileType extends BackendType {
+  val name = "file"
+}
+
+case class UserDefinedBackendType(name: String) extends BackendType
+
+case class UserDefinedBackendInputType(name: String) extends BackendInputType
```
echeipesh: this is unused and seems like a an artifact of earlier layout

spark-etl/src/main/scala/geotrellis/spark/etl/config/IngestOutputType.scala
```diff
@@ -0,0 +1,6 @@
+package geotrellis.spark.etl.config
+
+case class IngestOutputType(
```
echeipesh: unused class

pomadchin: @echeipesh good catch, it's from previous revisions; thanks!

spark-etl/src/main/scala/geotrellis/spark/etl/config/Backend.scala
```diff
@@ -0,0 +1,3 @@
+package geotrellis.spark.etl.config
+
+case class Backend(`type`: BackendType, path: String, profile: Option[String] = None)
```
echeipesh: This really begs for `profile` to be an `Option[BackendProfile]`, see comment in the parser for details.

spark-etl/src/main/scala/geotrellis/spark/etl/config/EtlConf.scala
```diff
@@ -0,0 +1,134 @@
...
+            })
+          }
+        }.toMap
+      }
+
+    val inputsParsed = input.parseJson.convertTo[List[Input]]
+
+    val outputParsed = output.parseJson.convertTo[Output]
+
+    inputsParsed.map { inputParsed =>
```
echeipesh: This is basically trying to do a lookup from profile name to profile instance. At this point you have parsed the profiles already. It would be just as easy to give that as a constructor to `JsonFormat[Inout]` instance defined here, which would be used for parsing. This would also allow you to upgrade `Backend` to have the actual profile.

spark-etl/src/main/scala/geotrellis/spark/etl/config/Output.scala
```diff
@@ -0,0 +1,59 @@
...
+  private def _getKeyIndexMethod: KeyIndexMethod[_] = (keyIndexMethod.`type`, keyIndexMethod.temporalResolution) match {
+    case ("rowmajor", None)    => RowMajorKeyIndexMethod
+    case ("hilbert", None)     => HilbertKeyIndexMethod
+    case ("hilbert", Some(tr)) => HilbertKeyIndexMethod(tr.toInt)
+    case ("zorder", None)      => ZCurveKeyIndexMethod
+    case ("zorder", Some(tr))  => ZCurveKeyIndexMethod.byMilliseconds(tr)
+    case _                     => throw new Exception("unsupported keyIndexMethod definition")
+  }
+
+  def getKeyIndexMethod[K] = _getKeyIndexMethod.asInstanceOf[KeyIndexMethod[K]]
```
echeipesh: Code style, no need for private def here you should be able to:

```scala
def getKeyIndexMethod[K] = {
(keyIndexMethod.`type`, keyIndexMethod.temporalResolution) match {
     case ("rowmajor", None)    => RowMajorKeyIndexMethod
     case ("hilbert", None)     => HilbertKeyIndexMethod
     case ("hilbert", Some(tr)) => HilbertKeyIndexMethod(tr.toInt)
     case ("zorder", None)      => ZCurveKeyIndexMethod
     case ("zorder", Some(tr))  => ZCurveKeyIndexMethod.byMilliseconds(tr)
     case _                     => throw new Exception("unsupported keyIndexMethod definition")
}.asInstanceof[KeyIndexMethod[K]]
```

echeipesh: Actually I would really expect scala type inference to figure it out without `asInstanceof`, could be wrong there.

spark-etl/src/main/scala/geotrellis/spark/etl/config/Input.scala
```diff
@@ -0,0 +1,13 @@
...
+import org.apache.spark.storage.StorageLevel
+
+case class Input(
+  name: String,
+  format: String,
+  backend: Backend,
+  cache: Option[StorageLevel] = None,
+  noData: Option[Double] = None
+) {
+  def params = getParams(backend.`type`, backend.path)
```
echeipesh: This branches on backend type. Should `Backend` have a type hierarchy? In that case this could be more clearly explained as overloading an abstract method.

spark-etl/src/main/scala/geotrellis/spark/etl/config/IngestType.scala
```diff
@@ -0,0 +1,7 @@
+package geotrellis.spark.etl.config
+
+case class IngestType(
```
echeipesh: Unused

spark-etl/src/main/scala/geotrellis/spark/etl/config/BackendProfiles.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  def getInstance = AccumuloInstance(instance, zookeepers, user, token)
+}
+case class HBaseProfile(name: String, master: String, zookeepers: String) extends BackendProfile {
+  def `type` = HBaseType.name
+
+  def getInstance = HBaseInstance(zookeepers.split(","), master)
+}
+
+case class BackendProfiles(backendProfiles: BackendProfile*)
```
echeipesh: I'm unclear if this is useful, only seems to be used in tests?

spark-etl/src/main/scala/geotrellis/spark/etl/config/package.scala
```diff
@@ -0,0 +1,25 @@
...
+      val S3UrlRx(_, _, bucket, prefix) = p
+      Map("bucket" -> bucket, "key" -> prefix)
+    }
+    case AccumuloType | HBaseType => Map("table" -> p)
+    case CassandraType            => {
+      val List(keyspace, table) = p.split("\\.").toList
+      Map("keyspace" -> keyspace, "table" -> table)
+    }
+    case HadoopType | FileType          => Map("path" -> p)
+    case UserDefinedBackendType(s)      => Map("input" -> s)
```
echeipesh: `UserDefinedBackendType` doesn't get to use the `path` ?

echeipesh:
Most of the comments in round two are around making a `Backend` type hierarchy to clean up all those match statements and upgrading `Strings` to instances of objects they name.

A really cool thing this allows is getting rid of the `params` all-together. Plugins could get just instances of `Backend` and `BackendProfile` and cast them to the type they must be. This would get rid of all the string parsing. The whole `params: Maps[String, String]` is an artifact parsing command line arguments that don't really have a deep structure that JSON does. This is a lot of code changes so probably it can be done as round 2 PR into ETL.

spark-etl/src/main/scala/geotrellis/spark/etl/EtlJob.scala
```diff
@@ -0,0 +1,15 @@
+package geotrellis.spark.etl
+
+import geotrellis.spark.etl.config._
+
+case class EtlJob(conf: EtlConf) extends Serializable {
```
echeipesh: Once backends generate their own params, what is the purpose of this class? We could just pass EtlConf at that point, but does that lose meaning?

spark-etl/src/main/scala/geotrellis/spark/etl/config/package.scala
pomadchin: i wanted to set there more semantics; like in case of hadoop and file we have "path", but in case of some custom backend (see landsat emr demo pr) it can be just any kind of input (as mentioned: params to load landsat tiles)

pomadchin:
includes rather important hbase deps fix (`"org.codehaus.jackson"  % "jackson-core-asl" % "1.8.3" intransitive()`)

pomadchin:
without las dirty fix: 

```scala
Exception in thread "main" java.lang.ClassCastException: geotrellis.spark.io.index.ZCurveKeyIndexMethod$ cannot be cast to geotrellis.spark.io.index.KeyIndexMethod
    at geotrellis.spark.etl.config.Output.getKeyIndexMethod(Output.scala:45)
    at geotrellis.spark.etl.accumulo.SpatialAccumuloOutput.writer(SpatialAccumuloOutput.scala:13)
    at geotrellis.spark.etl.OutputPlugin$class.apply(OutputPlugin.scala:19)
    at geotrellis.spark.etl.accumulo.SpatialAccumuloOutput.apply(SpatialAccumuloOutput.scala:11)
    at geotrellis.spark.etl.Etl.savePyramid$1(Etl.scala:178)
    at geotrellis.spark.etl.Etl.save(Etl.scala:192)
    at geotrellis.spark.etl.Etl$$anonfun$ingest$1.apply(Etl.scala:47)
    at geotrellis.spark.etl.Etl$$anonfun$ingest$1.apply(Etl.scala:39)
    at scala.collection.immutable.List.foreach(List.scala:318)
    at geotrellis.spark.etl.Etl$.ingest(Etl.scala:39)
    at geotrellis.chatta.ChattaIngest$delayedInit$body.apply(ChattaIngest.scala:15)
    at scala.Function0$class.apply$mcV$sp(Function0.scala:40)
    at scala.runtime.AbstractFunction0.apply$mcV$sp(AbstractFunction0.scala:12)
    at scala.App$$anonfun$main$1.apply(App.scala:71)
    at scala.App$$anonfun$main$1.apply(App.scala:71)
    at scala.collection.immutable.List.foreach(List.scala:318)
    at scala.collection.generic.TraversableForwarder$class.foreach(TraversableForwarder.scala:32)
    at scala.App$class.main(App.scala:71)
    at geotrellis.chatta.ChattaIngest$.main(ChattaIngest.scala:12)
    at geotrellis.chatta.ChattaIngest.main(ChattaIngest.scala)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:497)
    at org.apache.spark.deploy.SparkSubmit$.org$apache$spark$deploy$SparkSubmit$$runMain(SparkSubmit.scala:674)
    at org.apache.spark.deploy.SparkSubmit$.doRunMain$1(SparkSubmit.scala:180)
    at org.apache.spark.deploy.SparkSubmit$.submit(SparkSubmit.scala:205)
    at org.apache.spark.deploy.SparkSubmit$.main(SparkSubmit.scala:120)
    at org.apache.spark.deploy.SparkSubmit.main(SparkSubmit.scala)
```

echeipesh:
💯 Thank you for pushing this for so long, looks really good.

PR #1556, opened 2016-06-21
Better HDFS Support

echeipesh:
This PR improves HDFS layer writing and random value reading support.

`HadoopValuesReader` now opens up `MapFile.Readers` for each map file in the layer. These readers cache the index of available keys so they are able to provide quick lookups. This replaces and improves previous method of using `FileInputFormat` to query for a single record.

`HadoopRDDWriter` has multiple improvements:
- Accomplishes it's task with a single shuffle step
 - This is possible because instead of estimating the blocks required by counting we simply roll over to a new file when the record being written is about to surpass the block boundary.
 - Instead of using a groupBy we sort our records on shuffle which uses IO index to partition the records
- Writes happen through mapping the partition iterator, this is possible because the partition comes pre-sorted and greatly reduces memory pressure as records can be garbage collected after they are written.

Current testing shows:
- HDFS ingest speeds exceed those seen in Accumulo
- tile fetching latency is visually acceptable
- that jobs using the `GroupedCumulativeIterator` approach are able to complete in settings where .groupBy/sort/write jobs are killed for memory violations. 

Future improvements that on the mind:
- Save bloom filter index for layer MapFiles to reduce memory requirements for `HadoopValueReader` and to produce quicker lookup in most cases.
- In `HadoopValueReader` when fetching a record, cache all the tiles that share that index before filtering down to a single tile. These records are very likely to be asked for next and should be stored in LRU cache.
- Devise a method to compare KeyIndex instances such that if the RDD to be saved is already partitioned by compatible index (where all key boundaries are shared) to avoid the extra shuffle.

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopRDDWriter.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopRDDWriter.scala
@@ -43,7 +43,7 @@ object HadoopRDDWriter extends LazyLogging {
           MapFile.Writer.keyClass(classOf[LongWritable]),
           MapFile.Writer.valueClass(classOf[BytesWritable]),
           MapFile.Writer.compression(SequenceFile.CompressionType.NONE))
-      writer.setIndexInterval(1)
+      writer.setIndexInterval(32)
       writer
     }
 
--- a/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopRDDWriter.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopRDDWriter.scala
@@ -64,7 +64,7 @@ object HadoopRDDWriter extends LazyLogging {
     rdd: RDD[(K, V)],
     path: Path,
     keyIndex: KeyIndex[K],
-    indexInterval: Int = 16
+    indexInterval: Int = 4
   ): Unit = {
     implicit val sc = rdd.sparkContext
     val fs = path.getFileSystem(sc.hadoopConfiguration)
--- a/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopRDDWriter.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopRDDWriter.scala
@@ -61,7 +61,8 @@ object HadoopRDDWriter extends LazyLogging {
       bytesRemaining -= recordSize
     }
 
-    def close(): Unit = writer.close()
+    def close(): Unit =
+      if (null != writer) writer.close()
   }
 
   def write[K: AvroRecordCodec: ClassTag, V: AvroRecordCodec: ClassTag](
--- a/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopRDDWriter.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopRDDWriter.scala
@@ -43,7 +43,7 @@ object HadoopRDDWriter extends LazyLogging {
           MapFile.Writer.keyClass(classOf[LongWritable]),
           MapFile.Writer.valueClass(classOf[BytesWritable]),
           MapFile.Writer.compression(SequenceFile.CompressionType.NONE))
-      writer.setIndexInterval(32)
+      writer.setIndexInterval(16)
       writer
     }
 
--- a/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopRDDWriter.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopRDDWriter.scala
@@ -26,15 +26,12 @@ object HadoopRDDWriter extends LazyLogging {
     * opens a new file to continue writing. This allows to split partition into block-sized
     * chunks without foreknowledge of how bit it is.
     */
-  class MultiMapWriter(layerPath: String, partition: Int, blockSize: Long) {
+  class MultiMapWriter(layerPath: String, partition: Int, blockSize: Long, indexInterval: Int) {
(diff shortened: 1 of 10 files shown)


spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopRDDWriter.scala
```diff
@@ -43,7 +43,7 @@ object HadoopRDDWriter extends LazyLogging {
           MapFile.Writer.keyClass(classOf[LongWritable]),
           MapFile.Writer.valueClass(classOf[BytesWritable]),
           MapFile.Writer.compression(SequenceFile.CompressionType.NONE))
-      writer.setIndexInterval(1)
+      writer.setIndexInterval(32)
```
pomadchin: how it would effect query time? (just curious)

echeipesh: Layer query time would not be effected at all by this. When we're reading off ranges we're already seeking through the file, so not having as many points would have minimal impact if any. This is going to have more of an impact on random access through value reader. Spinning up a cluster to figure that part out. For reference the default interval value is 128

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopValueReader.scala
```diff
@@ -11,10 +11,12 @@ import geotrellis.spark.util.KryoWrapper
...
 import org.apache.hadoop.io._
+import org.apache.hadoop.conf.Configuration
 import org.apache.spark.SparkContext
 import spray.json._
 import spray.json.DefaultJsonProtocol._
 
+import scala.collection.immutable._
 import scala.reflect.ClassTag
 
 class HadoopValueReader(val attributeStore: HadoopAttributeStore)
```
lossyrob: We need to remove spark context

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopRDDWriter.scala
```diff
@@ -9,68 +11,85 @@ import geotrellis.spark.io.avro.codecs._
...
 
 import scala.reflect._
 
 object HadoopRDDWriter extends LazyLogging {
 
-  def write[K: AvroRecordCodec, V: AvroRecordCodec](
+  /**
+    * When record being written would exceed the block size of the current MapFile
+    * opens a new file to continue writing. This allows to split partition into block-sized
+    * chunks without foreknowledge of how bit it is.
```
lossyrob: `bit` => `big`

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopRDDWriter.scala
```diff
@@ -9,68 +11,85 @@ import geotrellis.spark.io.avro.codecs._
...
+    def close(): Unit =
+      if (null != writer) writer.close()
+  }
+
+  def write[K: AvroRecordCodec: ClassTag, V: AvroRecordCodec: ClassTag](
     rdd: RDD[(K, V)],
     path: Path,
     keyIndex: KeyIndex[K],
-    tileSize: Int = 256*256*8,
-    compressionFactor: Double = 1.3
```
lossyrob: 👍

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopRDDWriter.scala
```diff
@@ -9,68 +11,85 @@ import geotrellis.spark.io.avro.codecs._
...
+        var writer = new MultiMapWriter(layerPath, pid, blockSize, indexInterval)
+        for ( (index, pairs) <- GroupConsecutiveIterator(iter)(r => keyIndex.toIndex(r._1))) {
+          writer.write(
+            new LongWritable(index),
+            new BytesWritable(AvroEncoder.toBinary(pairs.toVector)(codec)))
+        }
+        writer.close()
+        // TODO: collect statistics on written records and return those
+        Iterator.empty
+      }.count
```
lossyrob: :clap:

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopValueReader.scala
```diff
@@ -7,67 +7,78 @@ import geotrellis.spark.io.avro.codecs._
...
-
+    val writerSchema = attributeStore.readSchema(layerId)
     val codec = KeyValueRecordCodec[K, V]
-    val kwSchema = KryoWrapper(schema) //Avro Schema is not Serializable
 
-    def read(key: K): V = {
-      val keyBounds = KeyBounds[K](key, key)
-      val filterDefinition = keyIndex.indexRanges(keyBounds).toArray
-      inputConf.setSerialized(FilterMapFileInputFormat.FILTER_INFO_KEY, filterDefinition)
+    // Map from last key in each reader to that reader
```
lossyrob: Feel like this doc string is talking about an older version

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopValueReader.scala
```diff
@@ -7,67 +7,78 @@ import geotrellis.spark.io.avro.codecs._
...
+      FilterMapFileInputFormat.layerRanges(header.path, conf)
 
-      val _codec = codec
-      val _kwSchema = kwSchema
+    def read(key: K): V = {
+      val index: Long = keyIndex.toIndex(key)
+      val valueWritable: BytesWritable =
+      ranges
+          .find{ row =>
+            index >= row._2 && index <= row._3
```
lossyrob: Should this be `index < row._3`?

lossyrob:
+1 after comments addressed

PR #1559, opened 2016-06-27
Cropped Windowed GeoTiff Reading

jbouffard:
This pull request adds the following:

1. A new implementation of CompressedBytes called SegmentBytes that allows for reading in the GeoTiff's segments into either a ByteBuffer or an Array[Array[Byte]].

2. GeoTiffs can now be read in via a stream.

3. GeoTiffs now have their own crop method which can be used in conjunction with the stream reading.

4. Reading in a windowed GeoTiff given an Extent.


raster/src/main/scala/geotrellis/raster/io/geotiff/CompressedBytesArray.scala
```diff
@@ -0,0 +1,71 @@
...
+import geotrellis.raster.io.geotiff._
+import geotrellis.raster.io.geotiff.reader._
+import geotrellis.raster.io.geotiff.tags._
+import geotrellis.raster.io.geotiff.util._
+
+import java.nio.ByteBuffer
+import monocle.syntax.apply._
+import spire.syntax.cfor._
+
+class CompressedBytesArray(compressedBytes: Array[Array[Byte]]) extends CompressedBytes {
```
lossyrob: Naming convention goes from left being the most specific, right being the most general.
Not everything follows this convention in the base library, but we should follow it consistently in this library (and I think we generally do with a couple of exceptions perhaps).

`ArrayCompressedBytes`
`ByteBufferCompressedBytes`

would fit better.

Also perhaps a different name would be more descriptive.

`SegmentBytes`
`ArraySegmentBytes`
`BufferSegmentBytes`

Thoughts?

lossyrob:
Should have unit tests around the implementations.

raster/src/main/scala/geotrellis/raster/io/geotiff/CompressedBytesArray.scala
MyNameIsBouf: I agree, those names were just place holders until I could think of something better. I like the names you proposed, and will rename the files and classes with their respective new names.

MyNameIsBouf: Sorry about that, I had begun finishing up the unit tests, but I felt that I was late to make a PR so I just added my code without the tests. I'll be adding the tests soon.

lossyrob: No worries, just figured I'd mention it explicitly, I know this is a WIP PR :)

lossyrob:
@jbouffard can you add a description of the feature in the PR description? Something that if people were to find this PR after the fact, they could understand what the feature being merged in is all about.

jbouffard:
@lossyrob Yeah, I just added one.

jbouffard:
@lossyrob I looked at where the checked failed, and I don't think it had to do with my code. This is the error message that caused it to fail:

> The future returned an exception of type: java.sql.BatchUpdateException, with message: Batch entry 0 insert into "cities" ("name","geom")  values ('Allentown,PA',?) was aborted.  Call getNextException to see the cause.. (PostgisSpec.scala:112)

lossyrob:
💯

PR #1560, opened 2016-06-27
Added documentation to the GeoTiff* files

jbouffard:
I added documentation to most methods and a few vals in the GeoTiff* files. There are a few more things that I want to add, but I think I'll need some things explained before I do.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
+++ b/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
@@ -12,6 +12,19 @@ import java.util.BitSet
 import spire.syntax.cfor._
 
 object GeoTiffMultibandTile {
+  /**
+   * Creates a new instances of GeoTiffMultibandTile.
+   *
+   * @param compressedBytes: An Array[Array[Byte]] that represents the segments in the GeoTiff
+   * @param decompressor: A [[Decompressor]] of the GeoTiff
+   * @param segmentLayout: The [[GeoTiffSegmentLayout]] of the GeoTiff
+   * @param compression: The [[Compression]] type of the data
+   * @param bandCount: The number of bands in the GeoTiff
+   * @param hasPixelInterleave: Does the GeoTiff have pixel interleave
+   * @param cellType: The [[CellType]] of the segments
+   * @param bandType: The data storage format of the band. Defaults to None
+   * @return An implementation of GeoTiffMultibandTile based off of band or cell type
+   */
   def apply(
     compressedBytes: Array[Array[Byte]],
     decompressor: Decompressor,
@@ -130,6 +143,12 @@ abstract class GeoTiffMultibandTile(
   val segmentCount = compressedBytes.size
   private val isTiled = segmentLayout.isTiled
 
+  /**
+   * Returns the corresponding GeoTiffTile from the inputted band index.
+   *
+   * @param bandIndex: The band's index number
+   * @return The corresponding [[GeoTiffTile]]
+   */
   def band(bandIndex: Int): GeoTiffTile = {
     if(bandIndex >= bandCount) { throw new IllegalArgumentException(s"Band $bandIndex does not exist") }
     if(hasPixelInterleave) {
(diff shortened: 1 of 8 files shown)


MyNameIsBouf:
@lossyrob  These are the methods which I'm not really getting right now: [1](https://github.com/geotrellis/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffTile.scala#L215), [2](https://github.com/geotrellis/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffTile.scala#L230), [3](https://github.com/geotrellis/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffTile.scala#L245), and [4](https://github.com/geotrellis/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffTile.scala#L271).

I know they have something to do with macros, but I don't understand it well enough to explain it in documentation.

lossyrob:
@MyNameIsBouf 

These functions implement the interface methods of `IterableTile` and `MappableTile[Tile]`, which are extended by the `Tile` trait (https://github.com/geotrellis/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/Tile.scala#L34)

MappleTile extends MacroMappableTile (https://github.com/geotrellis/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/MappableTile.scala#L8), and the implementation of function parameters that take more than 2 args is called via a macro (https://github.com/geotrellis/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/MappableTile.scala#L15). The IterableTile does a similar thing.

The macro methods create implementations of `map` and `foreach` that are called with "Visitor" or "Mapper" traits. These traits are new-ed up at compile time by macros like this one (https://github.com/geotrellis/geotrellis/blob/master/macros/src/main/scala/geotrellis/macros/TileMacros.scala#L34).

We don't need this crazy Macro business for function parameters with 2 or less parameters because of specialization. See how Function2 has specialized type params (https://github.com/scala/scala/blob/2.12.x/src/library/scala/Function2.scala#L29), but Function3 does not (https://github.com/scala/scala/blob/2.12.x/src/library/scala/Function3.scala#L16). Specialization is a weird and complex topic; here's an article dealing with specialization: http://axel22.github.io/2013/11/03/specialization-quirks.html

Macros are another heavy topic. Here's some docs on them: http://docs.scala-lang.org/overviews/macros/overview.html

Let me know if you have followup questions after trying to digest some of that stuff.

jbouffard:
@lossyrob Travis failed, but after looking at what caused the problem and double checking my code, I think there's actually nothing wrong with it. The failed test had something to do with PostgreSQL, which doesn't have anything to do with what I've done.

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
```diff
@@ -211,9 +230,16 @@ abstract class GeoTiffMultibandTile(
     }
   }
 
+  /** Converts all of the bands into a collection of Vecot[Tile] */
```
lossyrob: Typo

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
```diff
@@ -299,6 +339,13 @@ abstract class GeoTiffMultibandTile(
     mapDouble(fn)
   }
 
+  /**
+   * Map over a MultibandTile starting at the given band.
+   *
+   * @param b0: The starting band
```
lossyrob: "Starting band" reads strange here, since it's the only band in question.

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffSegment.scala
```diff
@@ -6,18 +6,28 @@ import geotrellis.raster._
...
 
+/**
+ * Base trait of GeoTiffSegment
+ */
 trait GeoTiffSegment {
   def size: Int
   def getInt(i: Int): Int
   def getDouble(i: Int): Double
 
+  // represents all of the bytes in the segment
```
lossyrob: Docstring comments vs normal comment, needs `/** */`

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffSegmentCollection.scala
```diff
@@ -6,6 +6,7 @@ import geotrellis.raster.io.geotiff.compression._
 trait GeoTiffSegmentCollection {
   type T >: Null <: GeoTiffSegment
 
+  // represents all of the segments in the geotiff
```
lossyrob: Docstring comment vs normal comment

lossyrob:
@jbouffard make sure to mark PR's that are ready for review (and merging if it gets a +1) as not `[WIP]`

jbouffard:
@lossyrob Okay. I made the changes to both the documentation and to the title.

lossyrob:
:+1:

PR #1563, opened 2016-06-28
Vector Tiles

fosskers:
### TODO
- [x] Move to `scalapb`
- [x] Higher level types with Geotrellis `Geometry` types
- [x] Decoding
  - [x] Decode bytes into protobuf-level objects
  - [x] `Command` conversion
  - [x] Protobuf backend for high-level `trait`s
  - [x] Geometry decoding
  - [x] Lazy `Geometry` Streams
  - [x] Lazy `Feature` construction 
- [x] Encoding
  - [x] Geometry encoding (`toCommands` instances)
  - [x] VectorTile encoding
- [x] API Improvements
  - [x] Handle Geotrellis `LayoutDefinition`s on decode
  - [x] Handle Geotrellis `LayoutDefinition`s on encode
  - [x] Fix tests that broke with API change
- [x] Commit new test `.mvt` files
- [x] Port tests
- [x] Be fast
- [x] Benchmark decoding and encoding
- [x] Add/remove extra project files (build scripts, etc)
- [x] Fix Exception output (`InvalidCommand`, etc)

### Motivation
Invented by [Mapbox](https://www.mapbox.com/), they are a combination of the ideas of finite-sized tiles and vector geometries. Mapbox maintains the official implementation spec for VectorTile codecs.

VectorTiles are advantageous over raster tiles in that:

1. They are typically smaller to store
2. They can be easily transformed (rotated, etc.) in real time
3. They allow for continuous (as opposed to step-wise) zoom in Slippy Maps.

Raw VectorTile data is stored in the protobuf format. Any codec implementing
[the spec](https://github.com/mapbox/vector-tile-spec/tree/master/2.1) must decode and encode data according to [this *.proto* schema](https://github.com/mapbox/vector-tile-spec/blob/master/2.1/vector_tile.proto).

### Post-PR Next Steps
- Layer IO via Spark
- Parsing Feature-level metadata into useful `case class`es
- Rework `LayoutDefinition` to not assume Rasters. (inner TileLayout does)
- Massage Geotrellis to accomodate `VectorTileRDD`
- Inform Mapbox about the codecs


vectortile/src/main/scala/geotrellis/vectortile/Command.scala
```diff
@@ -0,0 +1,143 @@
...
+  /** Convert a list of parsed Commands back into their original Command
+    * and Z-encoded Parameter Integer forms.
+    */
+  def uncommands(cmds: Array[Command]): Array[Int] = {
+    cmds.flatMap({
+      case MoveTo(ds) => unparseCmd(1, ds.length) +: params(ds)  // (+:) is bad!
+      case LineTo(ds) => unparseCmd(2, ds.length) +: params(ds)  // (+:) is bad!
+      case ClosePath  => Array(unparseCmd(7, 1))
+    })
+  }
```
fosskers: @lossyrob `flatMap`ping here is the clean thing, but very likely not the fast thing.

vectortile/src/main/scala/geotrellis/vectortile/Command.scala
```diff
@@ -0,0 +1,143 @@
...
+          res.update(i, (unzig(ps(i*2)), unzig(ps(i*2+1))))
+
+          i += 1
+        }
+
+        LineTo(res) +: commands(rest)
+      }
+      case (7,_) => ClosePath +: commands(ns.tail)
+    }
+  }
```
fosskers: @lossyrob `+:` seems to copy the source array, which probably murders performance. Recursion was also the most intuitive here, but is probably also bad.

fosskers:
@lossyrob Something I need to get back into on Monday: We'll be storing RDDs of high-level `VectorTile`s, right? Would such a `case class` need all its data fields available at the parameter level? A la the `a b c` in: `case class Foo(a: A, b: B, c: C)`. If we're doing funky lazy parsing stuff within the class body, I can see there being awkwardness when dealing with RDDs / Avro serialized versions of these data types.

fosskers:
Unless it's okay to Avro serialize:
```scala
 case class ProtobufLayer(
   name: String,
   extent: Int,
   rawLayer: vt.Tile.Layer  // a raw protobuf layer
 )
```
and have these within RDDs. Then any geometry manipulation on them will be lazy and internal as usual, as if they had come straight from protobuf bytes.

fosskers:
It turns out the `vt.Tile.*` classes (the autogen'd protobuf classes) are all `Serializable`.

fosskers:
@lossyrob I did some more thinking on Streams vs custom laziness:

Consider the following scenarios, where P and MP are Polygon and MultiPolygon respectively:

  [P MP P MP P MP]  -- A list of alternating raw Polygon features.

  (1) The user wants to find a particular Polygon, which unknown to them
      is the second one in the list. They have to parse the first P,
      do *something* to the first MP, parse the second P and match on it,
      then stop.

      With Streams, the original list now looks like: [ MP P MP ]
      With custom laziness, it looks like: [ MP MP P MP ]

      The custom laziness wins for speed here, since we were able to
      cancel the parsing of the first MP early, and the Streams
      approach fully parsed the first MP.

  (2) The user wants to perform another operation, this time across all
      Ps. Both approaches must thus map over the entire list.

      With Streams, the original list is now empty: []
      With custom laziness, it looks like: [ MP MP MP ]

      It's harder to tell who wins here, since while the Streams
      had to waste time fully parsing each MP, the custom laziness
      had to reparse old MPs it had already looked through.

  (3) The user wants to perform another operation on all Ps. The streams
      can go ahead since everything has been parsed. The custom approach
      must reparse all the MPs to check for Ps, since it wouldn't know
      there weren't any left.

  (4) The user wants to perform an operation on MPs this time. The streams
      can go ahead since the MPs are already parsed. The custom approach
      has to reparse the MPs for the fourth time.

  My takeaway: the custom approach is better for one-off operations on a
  particular geometry type. The stream approach quickly overtakes the other
  if you plan multiple operations over the same geometries.

fosskers:
A good shower made me realize your three-stage idea wins out, since we'd never have to reparse MPs until Step 4 above. @echeipesh warnings of over-engineering echo in my ear, mind you.

fosskers:
Suggestion from @pomadchin: think about `scalaz-streams`.

pomadchin:
@fosskers if we can make native we should try native though

scripts/buildall.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+./sbt -J-Xmx2G "project vectortile" test || { exit 1; }
```
lossyrob: need to also add to a `.travis/build-and-test-set` file, or else it won't be tested by travis

vectortile/data/DataGen/java-vector-tile.java
```diff
@@ -0,0 +1,230 @@
+/* THIS WILL NOT RUN
+ * Below, there are a few examples of how the test files were generated using
+ * java-vector-tile.
+ * https://github.com/ElectronicChartCentre/java-vector-tile
+ */
```
lossyrob: Probably should add a license header to this, with the proper copyright attribution.

fosskers: That was from the previous guy, Kevin. I intend to untrack it.

fosskers:
Simple usage:

```scala
import geotrellis.vectortile.protobuf.{ Protobuf, ProtobufTile }

// Read file, and parse into high-level `VectorTile` object.
val tile = ProtobufTile(Protobuf.decodeIO("vectortile/data/roads.mvt"))

// Fetch the layer names
val names: Set[String] = tile.layers.keys

// Fetch all single Polygons in a layer
val polys: Stream[Polygon] = tile.layers.get("landuse").get.polygons.map(_.geom)
```

fosskers:
Usage now looks like:
```scala
val layout: LayoutDefinition = ...
val keysBytes: Seq[(SpatialKey, Array[Byte])] = ...
val tiles: Seq[(SpatialKey, VectorTIle)] = keysBytes.map({ case (key, bytes) => ProtobufTile.fromBytes(bytes, key, layout) })
```

fosskers:
@lossyrob when reviewing this PR, it might be helpful to browse through the `geotrellis-vectortile` Scaladocs first, to get a sense of what is exposed.

fosskers:
@lossyrob I've included some "Post-PR next steps" at the bottom of the PR description.

vectortile/src/main/scala/geotrellis/vectortile/VectorTile.scala
```diff
@@ -0,0 +1,73 @@
...
+// --- //
+
+/** A high-level representation of a Vector Tile. At its simplest, a Tile is
+  * just a collection of Layers. We opt to expose each Layer name at the Tile
+  * level, as the keys of a [[Map]]. This way, if the layer names are known by
+  * the user ahead of time, they can search through the Tile quickly.
+  *
+  * Traditionally, VectorTiles are encoded as Protobuf data, which this library
+  * provides a codec for. However, by making this top-level type a trait, we
+  * are able to define alternative backends (GeoJson, for instance).
```
lossyrob: Mentioning GeoJson might give users the impression that that is a complete feature

vectortile/src/main/scala/geotrellis/vectortile/VectorTile.scala
```diff
@@ -0,0 +1,73 @@
...
+  /** The layer's name. */
+  def name: String
+
+  /** The width/height of this Layer's coordinate grid. By default this is 4096,
+    * as per the VectorTile specification.
+    *
+    * Not to be confused with a [[geotrellis.vector.Extent]], which represents
+    * some projected area on a map.
+    */
+  def extent: Int
```
lossyrob: I think we should rename this, as avoid Extent/extent confusion. tileWidth?

vectortile/src/main/scala/geotrellis/vectortile/package.scala
```diff
@@ -0,0 +1,89 @@
+package geotrellis
+
+/** Invented by [[https://www.mapbox.com/ Mapbox]], VectorTiles are a combination of the
```
lossyrob: If we attribute Mapbox, can we mention that this is an open specification. Just thinking ahead at having to deal with IP questions about it.

vectortile/src/main/scala/geotrellis/vectortile/package.scala
```diff
@@ -0,0 +1,89 @@
...
+  *   - If a VectorTile `geometry` list marked as `POINT` has only one pair
+  *     of coordinates, it will be decoded as a Geotrellis `Point`. If it has
+  *     more than one pair, it will be decoded as a `MultiPoint`. Likewise for
+  *     the `LINESTRING` and `POLYGON` types. A complaint has been made about
+  *     the spec regarding this, and future versions may include a difference
+  *     between single and multi geometries.
+  *
+  * @author cwoodbury@azavea.com
+  * @version 2.1
+  */
```
lossyrob: There are great ScalaDocs, but I see there's no markdown docs...we've generally been putting this sort of doc in markdown instead of in ScalaDoc. I'm not sure how to avoid duplication between markdown and scaladocs, but I would want to see something for the markdown...would linking to the scaladocs work? That might be a bit weird.

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
```diff
@@ -0,0 +1,293 @@
...
+  /** Encode this VectorTile back into its original form of Protobuf bytes. */
+  def toBytes: Array[Byte] = toProtobuf.toByteArray
+}
+
+object ProtobufTile {
+  /** Create a ProtobufTile masked as its parent trait. */
+  def fromPBTile(
+    tile: vt.Tile,
+    key: SpatialKey,
+    layout: LayoutDefinition
```
lossyrob: We actually just want to pass in the Extent. SpatialKey/LayoutDefinition are Sparky things, and vector tiles shouldn't depend on sparky things (but be used by sparky things in the layer versions of them)

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
```diff
@@ -0,0 +1,293 @@
...
+  }
+}
+
+/**
+  * A [[Layer]] decoded from Protobuf data. All of its Features are decoded
+  * lazily, making for very fast extraction of single features/geometries.
+  *
+  */
+// TODO The Layer storing `key` and `layout` is likely very wasteful
+// it terms of memory.
```
lossyrob: Does this TODO still make sense?

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
```diff
@@ -0,0 +1,293 @@
...
+  def features: Seq[Feature[Geometry, Map[String, Value]]] = {
+    points.append(multiPoints)
+     .append(lines)
+     .append(multiLines)
+     .append(polygons)
+     .append(multiPolygons)
+  }
+
+  /* Unconsumed raw Features */
+  // TODO Should this be a def?
```
lossyrob: perhaps a lazy val?

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
```diff
@@ -0,0 +1,293 @@
...
+  /**
+   * Construct Feature-specific metadata from the key/value lists of
+   * the parent layer.
+   */
+  private def getMeta(keys: Seq[String], vals: Seq[vt.Tile.Value], tags: Seq[Int]): Map[String, Value] = {
+    val pairs = new ListBuffer[(String, Value)]
+    var i = 0
+
+    while (i < tags.length) {
+      val k: String = keys(tags(i))
```
lossyrob: Index call on Seq, bad for performance...only should ever called on IndexedSeq (like Vector) or Array

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
```diff
@@ -0,0 +1,293 @@
...
+}
+
+/**
+  * A [[Layer]] decoded from Protobuf data. All of its Features are decoded
+  * lazily, making for very fast extraction of single features/geometries.
+  *
+  */
+// TODO The Layer storing `key` and `layout` is likely very wasteful
+// it terms of memory.
+case class ProtobufLayer(
```
lossyrob: We should follow the "one file to one class/object/trait combo" rule here and other places.

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
lossyrob: Could do

```scala
tags.grouped(2).map { case List(k, v) =>
  pairs.append(keys(k), values(v))
}
```

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
fosskers: I've already made this change in the next PR.

vectortile/src/main/scala/geotrellis/vectortile/VectorTile.scala
fosskers: Agreed. `tileWidth` is clear.

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
```diff
@@ -0,0 +1,293 @@
...
+    val points = new ListBuffer[vt.Tile.Feature]
+    val lines = new ListBuffer[vt.Tile.Feature]
+    val polys = new ListBuffer[vt.Tile.Feature]
+
+    features.foreach { f =>
+      f.getType match {
+        case POINT => points.append(f)
+        case LINESTRING => lines.append(f)
+        case POLYGON => polys.append(f)
+        case _ => Unit // `UNKNOWN` or `Unrecognized`.
```
lossyrob: I wonder if there's some reporting we should do here. Perhaps this should log?

vectortile/src/main/scala/geotrellis/vectortile/package.scala
fosskers: I'll add it.

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
```diff
@@ -0,0 +1,293 @@
...
+  private lazy val polyStream = geomStream[Polygon, MultiPolygon](polyFs)
+
+  /* OPTIMIZATION NOTES
+   * `Stream.flatMap` maintains laziness. A common pattern here to "fold away"
+   * results you don't want are to use [[Option]]. However, flatMap here
+   * expects an [[Iterable]], and employs an implicit conversion from [[Option]]
+   * to get it.
+   *
+   * By calling directly what that implicit eventually calls at the bottom of
+   * its call stack, we save some operations.
```
lossyrob: Thanks for explaining that.

vectortile/src/main/scala/geotrellis/vectortile/package.scala
fosskers: If we hard linked to the official ones on `geotrellis.io`, it might not be too bad.

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
fosskers: That's moot now, I'll remove it.

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
fosskers: I'm going to be changing the internal composition of `ProtobufLayer` around to hopefully avoid some of these Kryo issues. These raw features will be relegated to a temporary `.apply` call.

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
fosskers: `Seq` is what the `scalapb` protobuf decoder yields, and luckily these metadata key/value maps usually aren't that large (less than 10 pairs). Is the change still worth it?

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
```diff
@@ -0,0 +1,293 @@
...
+        case _ => Unit // `UNKNOWN` or `Unrecognized`.
+      }
+    }
+
+    (points, lines, polys)
+  }
+
+  /** Encode this Layer back into a mid-level Protobuf object. */
+  def toProtobuf: vt.Tile.Layer = {
+    // TODO Where should these be?
```
lossyrob: Not sure what this TODO is asking...where the implicit values should live?

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
fosskers: Roger.

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
fosskers: `UNKNOWN` is a legal value that Protobuf geometries can have, we're just electing to ignore them. I agree that the user should be notified somehow, if it was their intention to actually use them.

Or we just put our feet down and mention in the docs "No UNKNOWNs!"

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
fosskers: Haha because otherwise `case (Left(p), meta) => new ::(Feature(p, meta), Nil)` is some serious witchcraft.

vectortile/src/main/scala/geotrellis/vectortile/protobuf/internal/Command.scala
```diff
@@ -0,0 +1,151 @@
...
+      case MoveTo(ds) => unparseCmd(1, ds.length) +=: params(ds)
+      case LineTo(ds) => unparseCmd(2, ds.length) +=: params(ds)
+      case ClosePath  => ListBuffer(unparseCmd(7, 1))
+    }).toSeq
+  }
+
+  /** Divide a Command Integer into its Command ID and parameter count. */
+  private def parseCmd(n: Int): (Int,Int) = {
+    val cmd = n & 7
+    val count = n >> 3  // or >> ?
```
lossyrob: what is the or comment mean?

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
fosskers: I suppose I'm asking if that's the right way to be calling the implicits in.

vectortile/src/main/scala/geotrellis/vectortile/protobuf/internal/package.scala
```diff
@@ -0,0 +1,375 @@
...
+    * vtX = X + (R * tileX)
+    * vtY = Y - (R * tileY)
+    * }}}
+    *
+    */
+  def toProjection(point: (Int, Int), topLeft: Point, resolution: Double): Point = {
+    Point(
+      topLeft.x + (resolution * point._1),
+      topLeft.y - (resolution * point._2)
+    )
```
lossyrob: Do we need to use `Point` here? Would `(Double, Double)` suffice? There is some cost for using `Point`, because it creates a JTS geometry, and if we only need to store x,y double values we should avoid that overhead.

vectortile/src/main/scala/geotrellis/vectortile/protobuf/internal/package.scala
```diff
@@ -0,0 +1,375 @@
...
+  /**
+    * Translate [[Point]] coordinates within a CRS to those of a fixed
+    * VectorTile grid. The reverse of [[toProjection]].
+    *
+    * @param point The [[Point]] in CRS space.
+    * @param topLeft The CRS coordinates of the top-left corner of this Tile.
+    * @param resolution How much of the CRS's units are covered by a single VT grid coordinate.
+    * @return Grid coordinates in VectorTile space.
+    */
+  // TODO Is `toInt` good enough for rounding?
```
lossyrob: If we're ok with always rounding down (which I think is totally fine), then yes.

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
lossyrob: Ah. Yeah it wouldn't make too much of a difference, but I'd still make the change so that we can comfortably say "we don't call .apply on a non-index seq" to a linter or scrupulous eye, or someone new to scala that might imitate.

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
lossyrob: Gotcha, I guess that makes sense to just ignore the valid but Unknown values. We should doc that we just skip anything that isn't a geometry or is unknown to us, in case you haven't already done that.

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
lossyrob: Yup, that's the way to do it.

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
fosskers: I'm all for reducing apply-magic.

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
fosskers: At the back of my head I'm always imagining other Azavea teams being the first adopters, and having difficulting coming up with a reason why they'd want an `UNKNOWN` in the first place. If they did, at least we could sit down with them in person and discuss it.

vectortile/src/main/scala/geotrellis/vectortile/protobuf/internal/Command.scala
fosskers: `>>` and `>>>` are logical and arithmetic shift in Scala, but I can't recall which is which. I think initially I was using `>>>`, which turned out to be wrong, and the comment didn't get cleared.

vectortile/src/main/scala/geotrellis/vectortile/protobuf/internal/package.scala
fosskers: We get that `topLeft` point for free from the initially passed-in `Extent` via `Extent.northWest`.

fosskers: Or did you mean the return type?

vectortile/src/main/scala/geotrellis/vectortile/protobuf/internal/package.scala
fosskers: On the grand scheme of things it would only ever result in any given Point being off in one of its coordinates by 1 out of 4096. In terms of floating point operations, `floor` is faster than rounding logical, no?

vectortile/src/main/scala/geotrellis/vectortile/protobuf/ProtobufTile.scala
fosskers: `(keys(k), vals(v))` <- we would still be calling `apply` here.

vectortile/src/main/scala/geotrellis/vectortile/protobuf/internal/package.scala
lossyrob: Seems like toProjection is used a lot in constructing the other geom types, so the return type could be (Double, Double) to avoid things like a massive Polygon having to create all sorts of intermediate point objects.

fosskers: To reiterate what I said in the standup for the record: `Polygon`s are defined in terms of `Line`, and `Line` are defined in terms of `Point`, so avoiding returning `Point` from `toProjection` doesn't actually save any work.

fosskers:
@lossyrob this is good to go.

fosskers:
@lossyrob `getMeta` fix made, and README added.

lossyrob:
@fosskers just needs an update and we're good to go.

fosskers:
@lossyrob updated to master.

lossyrob:
💯

PR #1564, opened 2016-06-28
doc-examples subproject; example for tiling to GeoTiff

lossyrob:
Added a subproject that we can use for compiling documentation examples.

Added an example for starting with an `RDD[(ProjectedExtent, Tile)]` and ending up with a GeoTiff.


lossyrob:
Travis failure due to jai download fail

pomadchin:
Nice idea to have  docs-examples subproject, +1.

PR #1568, opened 2016-06-29
Cassandra driver downgrade

pomadchin:
We need to move 3.0.0 -> 2.1.10.2 due to guava compatibilities issues

https://github.com/geotrellis/geotrellis-landsat-emr-demo/pull/6

- [x] check it on a cluster

### DIFF
--- a/cassandra/build.sbt
+++ b/cassandra/build.sbt
@@ -2,7 +2,8 @@ import Dependencies._
 
 name := "geotrellis-cassandra"
 libraryDependencies ++= Seq(
-  "com.datastax.cassandra" % "cassandra-driver-core" % "3.0.0"
+  // 2.1.10.2 version used to provide guava compatibility with hadoop stack
+  "com.datastax.cassandra" % "cassandra-driver-core" % "2.1.10.2"
     excludeAll (ExclusionRule("org.jboss.netty"), ExclusionRule("io.netty"), ExclusionRule("org.slf4j"), ExclusionRule("io.spray"), ExclusionRule("com.typesafe.akka"))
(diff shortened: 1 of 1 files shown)


lossyrob:
Is this due specifically to an EMR dependency, or a version of guava that is brought in by Spark?

pomadchin:
@lossyrob [there](https://github.com/geotrellis/geotrellis-landsat-emr-demo/pull/6) provided logs of deps hell; can confirm that even original hadoop 2.7.2 distribution depends on `guava-11.0.2` as well. But i still have questions: why geotrellis-integration project worked well on geodocker cluster, however i had a similar issue in the [chatta-demo](https://github.com/pomadchin/geotrellis-chatta-demo/blob/feature/cassandra-support/geotrellis/build.sbt#L39)

pomadchin:
@lossyrob there provided logs of deps hell; can confirm that even original hadoop 2.7.x distribution depends on guava-11.0.2 as well. So yes, troubles caused by weird spark classpath on emr, it includes just everything.

pomadchin:
We can solve this issue by excluding latest datastax driver, 1.2.x and 3.x are compatible [usage example](https://github.com/pomadchin/geotrellis-chatta-demo/blob/feature/cassandra-support/geotrellis/build.sbt)

Think that probably it is not necessary to downgrade driver version, we even can upgrade it up to `3.0.2`.

pomadchin:
It works, but no support for Cassandra 3.x; workaround with driver excluding and manually fixing dependency, reasons of possible issues would be described in docs. Can be reopened if that would be necessary.

PR #1577, opened 2016-07-04
Added example for focal operation on multiband layer.

lossyrob:
Also add an example for querying a region by a multipolygon, masking by that geometry, computing max NDVI over time and saving as GeoTiff.


echeipesh:
+1, really glad we're adding to these. I need to keep in mind to do the same when questions come up.

PR #1592, opened 2016-07-15
TileFeature additions

pomadchin:
* TileFeature should extend CellGrid
* Additional avro codec to work with MultibandTiles


lossyrob:
+1

PR #1600, opened 2016-07-17
Added mapBands to MultibandTile, updated doc example using it.

lossyrob:
There was a spark example in which it was a bit awkward to create a new multiband tile out of a multiband tile by applying a focal median operation over each band. This API addition makes it less awkward.


moradology:
LGTM - +1

PR #1606, opened 2016-07-27
Collections API

pomadchin:
https://github.com/geotrellis/geotrellis/issues/1605

Backends support:
- [x] File
- [x] Hadoop
- [x] S3
- [x] Accumulo
- [x] Cassandra
- [x] HBase
- [x] we can improve performance by reads  parallelizing


pomadchin:
* geotrellis.slick.PostgisSpec test failed

pomadchin:
* geotrellis.slick.PostgisSpec

spark/src/main/scala/geotrellis/spark/io/CollectionLayerReader.scala
```diff
@@ -0,0 +1,76 @@
...
+import scala.reflect._
+
+abstract class CollectionLayerReader[ID] { self =>
+  val defaultNumPartitions = CollectionLayerReader.defaultNumPartitions
+
+  def read[
+    K: AvroRecordCodec: Boundable: JsonFormat: ClassTag,
+    V: AvroRecordCodec: ClassTag,
+    M: JsonFormat: GetComponent[?, Bounds[K]]
+  ](id: ID, rasterQuery: LayerQuery[K, M], numPartitions: Int, indexFilterOnly: Boolean): Seq[(K, V)] with Metadata[M]
```
echeipesh: this read should not have `numPartitions` argument. All the reads should be using threads to distributed the work as much as makes sense, but by definition you only ever get one partition, the collection.

accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloCollectionReader.scala
```diff
@@ -0,0 +1,43 @@
...
+    val codec = KeyValueRecordCodec[K, V]
+    val includeKey = (key: K) => queryKeyBounds.includeKey(key)
+
+    val ranges = queryKeyBounds.flatMap(decomposeBounds)
+
+    ranges flatMap { range: AccumuloRange =>
+      val scanner = instance.connector.createScanner(table, new Authorizations())
+      scanner.setRange(range)
+      scanner.fetchColumnFamily(columnFamily)
+      scanner.iterator.map { case entry =>
```
echeipesh: I know scanner will have a thread pool available to run multiple requests, does it do the thing that is very useful here and process multiple ranges async or are they essentially sequential from range to range but async for each range?

pomadchin: async for each range; would fix that issue with updating that pr up to master

echeipesh:
It looks like a lot of code is shared between the collection readers and rdd readers in S3, File, Hadoop, and Cassandra. HBase and Accumulo are exceptions because they have `InputFormat`s but even then we could replace those easily with custom readers and possible gain some control.

Reading a collection seems exactly like reading a partition. Maybe we can abstract all of that  code into something that reads just the ranges and the splitting/filtering of those ranges can be up to the whichever reader. I'm mostly concerned because this async code is very fiddly and error-prone. I can see us having to refactor it again when we learn something new.

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopCollectionReader.scala
```diff
@@ -0,0 +1,93 @@
...
+                if (indexFilterOnly) Some(items, iter)
+                else Some(items.filter { row => includeKey(row._1) }, iter)
+              }
+            } else {
+              None
+            }
+          }
+      }
+
+      nondeterminism.njoin(maxOpen = threads, maxQueued = threads) { range map read }.runFoldMap(identity).unsafePerformSync
```
echeipesh: `readers` need to be closed, I'm unsure on what consequence, but probably nothing good.

pomadchin: are not they closing with cache expiration?

echeipesh: I can see how that would mess with the caching though, if you had multiple async calls n to a Collection reader. If that's the case it almost makes it seem like CollectionsReader needs to be closable, as it will invariably represent some resources.

Main use case for these is to be called from endpoints to work on "small" areas, so async `read` calls are totally in scope of design.

echeipesh: Same story goes for HBase and Accumulo collection reads, they're based on some version of range/batch writers.

pomadchin:
Includes important HBase fix.

spark/src/main/scala/geotrellis/spark/io/CollectionLayerReader.scala
```diff
@@ -0,0 +1,80 @@
...
+  def query[
+    K: AvroRecordCodec: Boundable: JsonFormat: ClassTag,
+    V: AvroRecordCodec: ClassTag,
+    M: JsonFormat: GetComponent[?, Bounds[K]]
+  ](layerId: ID): BoundLayerQuery[K, M, Seq[(K, V)] with Metadata[M]] =
+    new BoundLayerQuery(new LayerQuery, read[K, V, M](layerId, _))
+}
+
+object CollectionLayerReader {
+  def njoin[K: AvroRecordCodec: Boundable, V: AvroRecordCodec](
```
echeipesh: We as for `AvroRecordCodec` but we do not use them. If the `readFunc` is `Long => Array[Byte]` we can spin the iterator here and do the decoding. You'll need an extra `refinementFilter: K => Boolean` to capture the `filterIndexOnly` param.

pomadchin: @echeipesh good catch, can you review context bounds?

s3/src/main/scala/geotrellis/spark/io/s3/S3RDDReader.scala
```diff
@@ -33,7 +34,7 @@ trait S3RDDReader {
     filterIndexOnly: Boolean,
     writerSchema: Option[Schema] = None,
     numPartitions: Option[Int] = None,
-    threads: Int = ConfigFactory.load().getInt("geotrellis.s3.threads.rdd.read")
+    threads: Int = ConfigFactory.load().getThreads("geotrellis.s3.threads.rdd.read")
```
echeipesh: `S3RDDReader` can now use  `CollectionLayerReader.njoin` ?

echeipesh:
This is the list of places where it looks like we should use `CollectionLayerReader.njoin`:
 
- [x] CassandraCollectionReader
- [ ] CassandraRDDReader
- [x] S3CollectionReader
- [ ] S3RDDReader
- [ ] HadoopCollectionReader
- [ ] HadoopRDDReader
- [x] FileCollectionReader
- [ ] FileRDDReader

They can not be used in:
- AccumuloCollectionReader
- AccumuloRDDReader
- HBaseCollectionReader
- HBaseRDDReader

.. because one or another those backends read ranges rather than (K,V). I think that's fine, we don't need to abstract over that.

pomadchin:
Cannot be used in HadoopRDDReader as we use MR job

spark/src/main/scala/geotrellis/spark/io/CollectionLayerReader.scala
pomadchin: we use `AvroCodec` to decode bytes (we explicitly use `AvroEncoder.fromBinary`)

echeipesh:
+1, submitted a meta-PR that tidies up the njoin signature a little.

PR #1607, opened 2016-08-02
Parallelize reads

pomadchin:
- [x] s3
- [x] cassandra
- [x] file

### DIFF
--- a/s3/src/main/scala/geotrellis/spark/io/s3/S3RDDReader.scala
+++ b/s3/src/main/scala/geotrellis/spark/io/s3/S3RDDReader.scala
@@ -6,7 +6,7 @@ import geotrellis.spark.io.index.{IndexRanges, MergeQueue}
 import geotrellis.spark.io.avro.{AvroEncoder, AvroRecordCodec}
 import geotrellis.spark.util.KryoWrapper
 
-import scalaz.concurrent.Task
+import scalaz.concurrent.{Strategy, Task}
 import scalaz.std.vector._
 import scalaz.stream.{Process, nondeterminism}
 import com.amazonaws.services.s3.model.AmazonS3Exception
@@ -14,6 +14,8 @@ import org.apache.avro.Schema
 import org.apache.commons.io.IOUtils
 import org.apache.spark.SparkContext
 import org.apache.spark.rdd.RDD
+
+import java.util.concurrent.Executors
 
 trait S3RDDReader {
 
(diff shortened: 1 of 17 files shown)


fosskers:
@lossyrob Grisha has opted here not to explicitly use `Task`s with an explicitely defined thread pool. Apparently `nondeterminism` handles the specifics of that.

He was seeing similar issues to what we had before, where Tasks weren't appearing to complete and using `pool.shutdown()` was causing problems. We might want to consider back-porting his solution here to Lattice.

echeipesh:
Without dedicated pool it will end up using `scala.concurrent.ExecutionContext.Implicits.global` which is comes with this info:
```
The implicit global ExecutionContext. Import global when you want to provide the global ExecutionContext implicitly.

The default ExecutionContext implementation is backed by a work-stealing thread pool. By default, the thread pool uses a target number of worker threads equal to the number of available processors.
```

I think that works out alright for spark tasks

s3/src/main/scala/geotrellis/spark/io/s3/S3RDDReader.scala
```diff
@@ -45,37 +45,42 @@ trait S3RDDReader {
...
+                  case e: AmazonS3Exception if e.getStatusCode == 404 => Some(Vector.empty, iter)
+                }
+              } else {
+                None
               }
             }
+          }
 
-          tileSeq.flatten
+          nondeterminism.njoin(maxOpen = 8, maxQueued = 8) { range map read }.runFoldMap(identity).unsafePerformSync
```
echeipesh: Notice `8` is magic number here but `32` is for Cassandra and File. Did it benchmark to be best that way?

lossyrob: It would be best if we could configure this somehow, and not have magic numbers impossibly un-magicked

echeipesh: How do you feel about using typesafe config for this? This is probably a system wide property rather than per-action property.

pomadchin:
@echeipesh in terms of scalaz, it would use `DefaultStrategy`, which is FixedThread pool with threads = available processors amount. Thx for a pointer, though I am still curious (as an example of what i am talking about S3RDDWriter), by creating a batch of tasks using our custom thread pool, how tasks would be scheduled with `nondeterminism.njoin` as it uses `DefaultStrategy` and `DefaultExecutor`?

```scala
def njoin[A](maxOpen: Int, maxQueued: Int)(source: Process[Task, Process[Task, A]])(implicit S: Strategy): Process[Task, A]
```

s3/src/main/scala/geotrellis/spark/io/s3/S3RDDReader.scala
pomadchin: just experimented how tests would run faster on a local machine; though they are still magical. probably it makes sense to make them configurable, as speed of them depends on certain machine(s) configuration.

fosskers:
This continues to be an informative PR.

PR #1608, opened 2016-08-04
`Projections, Extents, and Layout Definitions` doc

fosskers:
*The other day, on gitter:*

Me: Alright, I'm digging into the relationship between projections, layout definitions, and extents.

@echeipesh : That sounds like a title of an excellent doc page I wish we had.

### DIFF
--- a/docs/concepts/projections-extents-layouts.md
+++ b/docs/concepts/projections-extents-layouts.md
@@ -1,0 +1,162 @@
+Projections, Extents, and Layout Definitions
+============================================
+
+Geographical Information Systems, like any specialized field, has a wealth
+of jargon and unique concepts. When represented in software, these concepts
+can sometimes be skewed or expanded from their original forms. Three of
+these critical to effective use of Geotrellis are *Projections*, *Extents*,
+and *Layout Definitions*. We give a thorough definition of each here, while
+referencing the Geotrellis objects and source files backing them.
+
+This document aims to be informative to new and experienced GIS users alike.
+
+Projections
+-----------
+
+**Data structures:** `CRS`, `LatLng`, `WebMercator`, `ConusAlbers`
+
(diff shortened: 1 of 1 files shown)


fosskers:
Not confident of the line: 

> "min" and "max" here are CRS specific, as the location of the point `(0,0)` varies between different CRS.

lossyrob:
Would be useful to comment on the diff line you're referring to.

docs/concepts/projections-extents-layouts.md
```diff
@@ -0,0 +1,162 @@
...
+`geotrellis.raster.{ GridExtent, RasterExtent }`,
+`geotrellis.raster.reproject.ReprojectRasterExtent`
+
+**What is an extent?**
+
+An `Extent` is a rectangular section of a 2D projection of the Earth. It is
+represented by two coordinate pairs that are its "min" and "max" corners in
+some Coorindate Reference System. "min" and "max" here are CRS
+specific, as the location of the point `(0,0)` varies between different CRS.
+An Extent can also be referred to as a *Bounding Box*.
```
lossyrob: The whole extent is CRS specific; anything relating to the Extent will have to be in that CRS, and users of the extent should know the CRS of the extent if they are trying to combine it with other things, either by knowing implicitly that all the geometries they are working with are in the same CRS, or explicitly by holding onto CRS instances, and reprojecting where necessary.

fosskers:
https://github.com/geotrellis/geotrellis/pull/1608/files#diff-345d6013e8e8878b87ef9caff9208ccaR78

docs/concepts/projections-extents-layouts.md
fosskers: Yup, that's mentioned in the next paragraph.

PR #1617, opened 2016-08-31
Stream Reading GeoTiffs From S3

### LINKED ISSUE #1278: Byte Range reads via HTTP in GeoTiff reader (i.e. /vsicurl/ in GeoTrellis)
Using `/vsicurl/`, GDAL can read remote HTTP rasters using the [Range HTTP request header](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.35). This allows it to only read remote data that it needs to satisfy a request, for instance, a windowed read request. We should implement this by utilizing a 3rd party library for doing byte range reads of HTTP binary data, or, if no acceptable solution already exists, writing the networking logic that would allow this. This in combination with #1277 will allow us to work on very large remote GeoTiff data without having to download the whole image for each windowed read.

Depends on #1277

jbouffard:
This is a fork from off of #1559 

This fork adds the ability to read GeoTiff files from S3 via streaming. This will allow for faster speeds when accessing files from the S3 server; and for the ability to read in files that would have normally been to large to load under the previous implementation.

Things to Do:
- [x] Do range reads into a GeoTiff on S3
- [x] Implement an S3ByteBuffer 
- [x] Read in GeoTiffs from S3

Fixes #1278


s3/src/test/scala/geotrellis/spark/io/s3/S3GeoTiffReadingSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+import geotrellis.raster.testkit._
+import geotrellis.raster.io.geotiff._
+import geotrellis.raster.io.geotiff.reader._
+import geotrellis.raster.io.geotiff.tags._
+
+import java.nio.{ByteBuffer, ByteOrder}
+import org.scalatest._
+import com.amazonaws.services.s3.model._
+
+class S3GeoTiffReadingSepc extends FunSpec
```
fosskers: `Sepc` -> `Spec`

util/src/main/scala/geotrellis/util/ByteReader.scala
```diff
@@ -0,0 +1,41 @@
+package geotrellis.util
+
+import java.nio.{Buffer, ByteBuffer, ByteOrder}
+import scala.language.implicitConversions
+
+trait ByteReader {
```
fosskers: This trait and its methods needs docstrings.

util/src/main/scala/geotrellis/util/StreamBytes.scala
```diff
@@ -0,0 +1,22 @@
+package geotrellis.util
+
+import java.io.InputStream
+import java.nio.ByteBuffer
+
+abstract class StreamBytes(chunkSize: Int) {
```
fosskers: This also needs docstrings, since its hard to tell at a glance if this is a "noun" or a "verb" class.

raster/src/main/scala/geotrellis/raster/io/geotiff/reader/TiffTagsReader.scala
```diff
@@ -106,7 +113,7 @@ object TiffTagsReader {
         byteBuffer.readDoublesTag(tiffTags, tagMetadata)
     }
 
-  implicit class ByteBufferTagReaderWrapper(val byteBuffer: ByteBuffer) extends AnyVal {
+  implicit class ByteReaderTagReaderWrapper(val byteBuffer: ByteReader) extends AnyVal {
```
fosskers: Any idea why this extends `AnyVal`?

s3/src/main/scala/geotrellis/spark/io/s3/util/S3ByteReader.scala
```diff
@@ -0,0 +1,102 @@
...
+import geotrellis.util._
+import geotrellis.spark.io.s3._
+import com.amazonaws.services.s3.model._
+import scala.language.implicitConversions
+
+import spire.syntax.cfor._
+
+import java.nio.{ByteOrder, ByteBuffer, Buffer}
+
+class S3ByteReader(s3StreamBytes: S3StreamBytes) extends ByteReader {
```
fosskers: Docstrings here too.

s3/src/main/scala/geotrellis/spark/io/s3/util/S3StreamBytes.scala
```diff
@@ -0,0 +1,44 @@
+package geotrellis.spark.io.s3.util
+
+import geotrellis.util.StreamBytes
+import geotrellis.spark.io.s3._
+import org.apache.commons.io.IOUtils
+
+import com.amazonaws.services.s3.model._
+
+class S3StreamBytes(request: GetObjectRequest,
```
fosskers: Docstrings here too.

raster/src/main/scala/geotrellis/raster/io/geotiff/reader/TiffTagsReader.scala
jbouffard: No idea, that was written a pretty long time ago, I think. If I had to guess, it's because of the wide range of types that need to accessed and evaluated to read in TiffTags.

s3/src/main/scala/geotrellis/spark/io/s3/util/S3ByteReader.scala
jbouffard: Okay, I'll add those after everything else has been completed.

util/src/main/scala/geotrellis/util/StreamBytes.scala
jbouffard: Yeah, I realized that after you pointed that out yesterday. I'll be adding Docstrings for sure to all of my code.

s3/src/test/scala/geotrellis/spark/io/s3/S3GeoTiffReadingSpec.scala
jbouffard: Thanks. I missed that.

raster/src/main/scala/geotrellis/raster/io/geotiff/reader/TiffTagsReader.scala
fosskers: Extending `AnyVal` here would only affect the class as a structure, not what its methods could do. 

http://www.scala-lang.org/api/current/#scala.AnyVal

Sounds like it could be an optimization for runtime behaviour. @lossyrob ?

lossyrob: Implicit classes that extend anyVal and follow ceretain rules will incur no runtime cost. 

See here for more details: http://docs.scala-lang.org/overviews/core/value-classes.html

fosskers: Oh nice, compile-time unboxing with type safety. I've seen an analogue to that in other languages but didn't know Scala could do it.

docs/raster/S3Streaming.md
```diff
@@ -0,0 +1,118 @@
...
+
+**Note** This feature is still be worked on and is subject to change at any time. I will update this document when any changes have been made to this feature as soon as I can. To check the latest version, see the link to the pull request at the bottom of this document.
+- - -
+### How to Use this Feature
+
+In order for streaming from S3 to work, two new types have been created. These are: ```StreamBytes``` and ```Bytereader```, respectively. the first deals with getting bytes in the form of ```Map[Long, Array[Byte]``` called a, "chunk". The ```Long``` of this chunk is the position in the file where the ```Array[Byte]``` starts. For example, if a new chunk needs to be created at the 100th byte of the file then the chunk that is created will have a value of 100. These chunks can vary in size and a new one is produced each time a call is made to another part of the file which is not contained within the current chunk.
+
```
fosskers: Markdown tip: Inlined code highlighting only needs a single backtick.

docs/raster/S3Streaming.md
```diff
@@ -0,0 +1,118 @@
...
+**Note** This feature is still be worked on and is subject to change at any time. I will update this document when any changes have been made to this feature as soon as I can. To check the latest version, see the link to the pull request at the bottom of this document.
+- - -
+### How to Use this Feature
+
+In order for streaming from S3 to work, two new types have been created. These are: ```StreamBytes``` and ```Bytereader```, respectively. the first deals with getting bytes in the form of ```Map[Long, Array[Byte]``` called a, "chunk". The ```Long``` of this chunk is the position in the file where the ```Array[Byte]``` starts. For example, if a new chunk needs to be created at the 100th byte of the file then the chunk that is created will have a value of 100. These chunks can vary in size and a new one is produced each time a call is made to another part of the file which is not contained within the current chunk.
```
fosskers: `Bytereader` -> `ByteReader`

docs/raster/S3Streaming.md
```diff
@@ -0,0 +1,118 @@
...
+
+```ByteReader``` is what reads the bytes from the chunks. As of right now, only ```java.nio.ByteBuffer```s and the ```StreamByteReader``` type can access the chunk.
+
+#### Creating ```StreamBytes```
+There is only one way to construct an instance of ```StreamBytes``` as of this writing, and that is through the ```S3StreamBytes``` class. Found in ```geotrellis.spark.io.s3.util``` package, this class will be able to get chunks from a desired GeoTifff on S3. Bellow is a basic guide on creating this class:
```
fosskers: `Bellow` -> `Below`

docs/raster/S3Streaming.md
```diff
@@ -0,0 +1,118 @@
...
+ import geotrellis.util.StreamByteReader
+ 
+ val reader = StreamByteReader(s3Bytes)
+```
+
+#### Streaming a GeoTiff From S3
+Now that we have a ```ByteReader``` all that needs to be done now is to supply it to the ```GeoTiffReader``` with any additional reading parameters that might be needed.
+
+#### Streaming the Entire GeoTiff
+Because GeoTiffs can be read in lazily now, files of any size should be able to be read in fully either from S3 or locally. However, precautions should be taken when doing so. It appears that when a file's size passes a certain point, the time it takes to perform operations on it degrades rapidly. This aspect of whole file streaming has not yet been looked into closely, so it is uncertain at what point performance begins to fall and to what extent. Therefore, it is best to use caution when streaming the entire file.
```
fosskers: Could this have something to do with the `skip` operation? Could the streamer be linearly reading through early bytes to access later ones, even if all you want are the later ones?

s3/src/main/scala/geotrellis/spark/io/s3/util/S3StreamBytes.scala
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * This class extends [[StreamBytes]] by reading chunks out of a GeoTiff on the
+ * AWS S3 server.
+ *
+ * @param request: A [[GetObjectRequest]] of the desired GeoTiff.
+ * @param client: The [[AmazonS3Client]] that retrieves the data.
+ * @param chunkSize: An Int that specifies how many bytes should be read in at a time.
+ * @return A new instance of S3StreamBytes.
+ */
+class S3StreamBytes(
```
fosskers: Since it turned out this was a "verby" class, I recommend a name change to `S3ByteStreamer`. Likewise for the parent  `StreamBytes`.

s3/src/test/scala/geotrellis/spark/io/s3/S3GeoTiffReadingSpec.scala
```diff
@@ -0,0 +1,116 @@
...
+import geotrellis.spark.io.s3._
+import geotrellis.raster.testkit._
+import geotrellis.raster.io.geotiff._
+import geotrellis.raster.io.geotiff.reader._
+
+import java.nio.{ByteBuffer, ByteOrder}
+import com.amazonaws.services.s3.model._
+import org.scalatest._
+
+class S3GeoTiffReadingSepc extends FunSpec
```
fosskers: `Sepc` -> `Spec`

util/src/main/scala/geotrellis/util/StreamBytes.scala
```diff
@@ -0,0 +1,26 @@
+package geotrellis.util
+
+/**
+ * This trait defines methods for breaking up a source of bytes into
+ * Map[Long, Array[Byte]] called a, "chunk". Where the Long is where within
+ * the file the chunk begins and the Array[Byte] containing the actual bytes.
+ */
+trait StreamBytes {
```
fosskers: Perhaps `ByteStreamer`, as mentioned above.

jbouffard:
@lossyrob This PR is ready for review whenever you have the chance.

I know that you didn't want to add GeoTiff reading to the s3 project, but I think adding an object or two could help with the API. Something like:

```scala
object S3GeoTiffReader {
  readSingleband(r: GetObjectRequest, client: AmazonS3Client,
  chunkSize: Int): SinglebandGeoTiff = {
  val s3StreamBytes = S3StreamBytes(r, client, chunkSize)
  val byteReader = StreamByteReader(s3StreamBytes)

  GeoTiffReader.readSingleband(byteReader, false, true)
}
```

It might make things a bit easier for people so they don't have to go through as many steps to stream in the file.

docs/raster/S3Streaming.md
jbouffard: I figured out how to do reading without using skip for the actual S3Client. Here's what it is [now](https://github.com/jbouffard/geotrellis/blob/wgt-s3/s3/src/main/scala/geotrellis/spark/io/s3/S3Client.scala#L173). I think it might have to do with the sheer size of the file and how many bytes it has to read through. Since to access the data the bytes need to be read in via a loop.

jbouffard: Sorry, I sent you a link to the wrong method. Here's what it actually is: https://github.com/jbouffard/geotrellis/blob/wgt-s3/s3/src/main/scala/geotrellis/spark/io/s3/S3Client.scala#L183

docs/raster/S3Streaming.md
```diff
@@ -0,0 +1,118 @@
...
+ import com.amazonaws.services.s3.model._
+ 
+ val client: S3Client = S3Client.default
+ val bucket: String = "my-bucket"
+ val key: String = "path/to/my/geotiff.tif"
+ 
+ // how many bytes that should be streamed at a time
+ val chunkSize: Int = 256000
+ 
+ val s3bytes = S3BytesStreamer(bucket, key, client, chunkSize)
```
pomadchin: `S3BytesStreamer` not works with `S3Client` type, but works with `AmazonS3Client`, i think it is possible to use more generic `S3Client` type in `S3BytesStreamer` constructor.

s3/src/main/scala/geotrellis/spark/io/s3/util/S3StreamBytes.scala
```diff
@@ -0,0 +1,63 @@
...
+ * AWS S3 server.
+ *
+ * @param request: A [[GetObjectRequest]] of the desired GeoTiff.
+ * @param client: The [[AmazonS3Client]] that retrieves the data.
+ * @param chunkSize: An Int that specifies how many bytes should be read in at a time.
+ * @return A new instance of S3BytesStreamer.
+ */
+class S3BytesStreamer(
+  request: GetObjectRequest,
+  client: AmazonS3Client,
```
pomadchin: `S3Client` type can be used there

PR #1621, opened 2016-09-05
Create GeoMesa suproject

pomadchin:
- [x] layer reader
- [x] layer writer
- [x] tests coverage
- [ ] GeoMesa 1.2.7


geomesa/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaInputFormat.scala
```diff
@@ -0,0 +1,335 @@
+/***********************************************************************
+  * Copyright (c) 2013-2016 Commonwealth Computer Research, Inc.
+  * All rights reserved. This program and the accompanying materials
+  * are made available under the terms of the Apache License, Version 2.0
+  * which accompanies this distribution and is available at
+  * http://www.opensource.org/licenses/apache2.0.php.
+  *************************************************************************/
+
+package org.locationtech.geomesa.jobs.mapreduce
```
pomadchin: should be removed after moving to geomesa 1.2.7 https://github.com/locationtech/geomesa/pull/1077

geomesa/src/test/scala/geotrellis/geotools/GeoMesaSimpleFeatureType.scala
```diff
@@ -0,0 +1,38 @@
...
+      case "MultiPoint" => sftb.add(GeometryToGeoMesaSimpleFeature.whereField, classOf[jts.MultiPoint])
+      case "MultiLine" => sftb.add(GeometryToGeoMesaSimpleFeature.whereField, classOf[jts.MultiLineString])
+      case "MultiPolygon" => sftb.add(GeometryToGeoMesaSimpleFeature.whereField, classOf[jts.MultiPolygon])
+      case g => throw new Exception(s"Unhandled GeoTrellis Geometry $g")
+    }
+    sftb.setDefaultGeometry(whereField)
+    if(temporal) sftb.add(whenField, classOf[java.util.Date])
+    val sft = sftb.buildFeatureType
+    if(temporal) sft.getUserData.put(Constants.SF_PROPERTY_START_TIME, whenField) // when field is date
+    sft.getUserData.put("geomesa.mixed.geometries", java.lang.Boolean.TRUE) // allow GeoMesa to index points and extents together
```
echeipesh: Looks like these tags are specific to Geomesa, so the object should be in `geotrellis.geomsea` package.

geomesa/src/test/scala/geotrellis/geotools/GeoMesaSimpleFeatureType.scala
```diff
@@ -0,0 +1,38 @@
...
+import org.locationtech.geomesa.accumulo.index.Constants
+
+import scala.reflect._
+
+object GeoMesaSimpleFeatureType {
+
+  val whenField  = "when"
+  val whereField = "where"
+
+  def apply[G <: Geometry: ClassTag](featureName: String, crs: Option[GCRS] = Some(WebMercator), temporal: Boolean = false) = {
```
echeipesh: need return type for API readability

geomesa/src/test/scala/geotrellis/geotools/GeoMesaSimpleFeatureType.scala
```diff
@@ -0,0 +1,38 @@
...
+    val sftb = (new SimpleFeatureTypeBuilder).minOccurs(1).maxOccurs(1).nillable(false)
+
+    sftb.setName(featureName)
+    crs.foreach { crs => sftb.setSRS(s"EPSG:${crs.epsgCode.get}") }
```
echeipesh: Is there a fallback if the CRS is not EPSG code? Not sure how common that is with vectors but in rasters we've seen that happen.

pomadchin: James sets crs using `org.geotools.referencing.CRS`, we can try it that way; potentially we can fix [these lines](https://github.com/geotrellis/geotrellis/blob/master/geotools/src/main/scala/geotrellis/geotools/GeometryToSimpleFeature.scala#L45) as well

geomesa/src/test/scala/geotrellis/geotools/GeoMesaSimpleFeatureType.scala
```diff
@@ -0,0 +1,38 @@
...
+    classTag[G].toString.split("\\.").last match {
```
echeipesh: `classTag[G].runtimeClass.getName`

echeipesh: Shame you can't easily match on the `ClassTag` instance.

geomesa/src/main/scala/geotrellis/spark/io/geomesa/GeoMesaAttributeStore.scala
```diff
@@ -0,0 +1,30 @@
+package geotrellis.spark.io.geomesa
+
+import geotrellis.spark.LayerId
+
+import org.geotools.data.DataStoreFinder
+import org.locationtech.geomesa.accumulo.data.AccumuloDataStore
+
+import scala.collection.JavaConverters._
+
+case class GeoMesaAttributeStore(instanceName: String, zookeepers: String, user: String, password: String, useMock: Boolean = false) extends Serializable {
```
echeipesh: This is not an `AttributeStore` and doesn't really do anything like AttributeStore interface. lets just make it an `apply`. So maybe there is a need of an object that is like `GeoMesaConnector` that keeps the map as fields?

geomesa/src/main/scala/geotrellis/spark/io/geomesa/GeoMesaLayerReader.scala
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.spark.rdd.RDD
+import org.apache.spark.SparkContext
+import org.geotools.data._
+import org.locationtech.geomesa.jobs.mapreduce.GeoMesaInputFormat
+import org.locationtech.geomesa.jobs.GeoMesaConfigurator
+import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
+
+import scala.reflect.ClassTag
+
+class GeoMesaLayerReader(val attributeStore: GeoMesaAttributeStore, table: String)(implicit sc: SparkContext) extends Serializable {
```
echeipesh: This should just take `DataStore` as constructor argument and provide `apply` overloads to help with construction. Also layer writer is generally for rasters. I would call this `GeoWaveFeatureReader`, @lossyrob thoughts?

geomesa/src/main/scala/geotrellis/spark/io/geomesa/GeoMesaAttributeStore.scala
pomadchin: Yes thought about it, mb `GeoMesaInstance`? and yes, probably we need to have just a Map of fields, to get connector in runtime and to abstract over backend types

geomesa/src/main/scala/geotrellis/spark/io/geomesa/GeoMesaLayerReader.scala
pomadchin: @echeipesh @lossyrob  +1 ; are we omitting word `Layer`?
- `GeoMesaFeatureReader`
- `GeoMesaFeatureValueReader`
- `GeoMesaFeatureWriter`

geomesa/src/main/scala/geotrellis/spark/io/geomesa/GeoMesaLayerWriter.scala
```diff
@@ -0,0 +1,41 @@
...
+import scala.collection.JavaConversions._
+
+class GeoMesaLayerWriter(val attributeStore: GeoMesaAttributeStore, table: String)(implicit sc: SparkContext) extends Serializable {
+  def write[G <: Geometry, D: ? => Seq[(String, Any)]]
+    (layerId: LayerId, rdd: RDD[Feature[G, D]])
+    (implicit ev: Feature[G, D] => FeatureToGeoMesaSimpleFeatureMethods[G, D]): Unit = {
+
+    rdd
+      .zipWithIndex // it is necessary to provide unique id for ingest batch
+      .map { case (f, index) => val sf = f.toSimpleFeature(s"${attributeStore.layerIdString(layerId)}_${index}", layerId.name); sf.getFeatureType -> sf }.groupByKey
```
echeipesh: Could you explain what is going on here? Where is this requirement for unique id coming from and how is it reflected in what is stored?

pomadchin: @echeipesh the practise during one ingest batch features should have different ids; but they could be same during different batches; would ask this question in a geomesa channel to clarify it

geomesa/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/GeoMesaInputFormat.scala
lossyrob: Can you please create an issue for this?

geomesa/src/test/scala/geotrellis/geomesa/geotools/GeoMesaSimpleFeatureType.scala
```diff
@@ -0,0 +1,41 @@
...
+      case "geotrellis.vector.MultiPolygon" => sftb.add(GeometryToGeoMesaSimpleFeature.whereField, classOf[jts.MultiPolygon])
+      case g => throw new Exception(s"Unhandled GeoTrellis Geometry $g")
+    }
+    sftb.setDefaultGeometry(whereField)
+    if(temporal) sftb.add(whenField, classOf[java.util.Date])
+    val sft = sftb.buildFeatureType
+    if(temporal) sft.getUserData.put(Constants.SF_PROPERTY_START_TIME, whenField) // when field is date
+    sft.getUserData.put(SimpleFeatureTypes.MIXED_GEOMETRIES, java.lang.Boolean.TRUE) // allow GeoMesa to index points and extents together
+    sft.getUserData.put(Hints.USE_PROVIDED_FID, java.lang.Boolean.FALSE) // generate feature ids
+    sft
```
lossyrob: I think this might be missing some things, e.g. visibility and periodicity:

https://github.com/locationtech/geomesa/blob/7fda8d02c195359165d9a30915153b398f2f0e91/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala#L270

https://github.com/locationtech/geomesa/blob/7fda8d02c195359165d9a30915153b398f2f0e91/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/Conversions.scala#L276

lossyrob: Then again we don't really have those concepts in GeoTrellis layers, so this might be a good thing to make a note/issue out of and add once we put those concepts into GeoTrellis (perhaps through sfcurve usage?)

geomesa/src/main/scala/geotrellis/geomesa/geotools/FeatureToGeoMesaSimpleFeatureMethods.scala
```diff
@@ -0,0 +1,11 @@
+package geotrellis.geomesa.geotools
+
+import geotrellis.proj4.CRS
+import geotrellis.util.MethodExtensions
+import geotrellis.vector.{Feature, Geometry}
+import org.opengis.feature.simple.SimpleFeature
+
+trait FeatureToGeoMesaSimpleFeatureMethods[G <: Geometry, T] extends MethodExtensions[Feature[G, T]] {
+  def toSimpleFeature(featureName: String, featureId: Option[String] = Some(null), crs: Option[CRS] = None)(implicit transmute: T => Seq[(String, Any)]): SimpleFeature =
```
elahrvivaz: doesn't `Some(null)` kind of break the `Option` contract?

geomesa/src/main/scala/geotrellis/geomesa/geotools/GeometryToGeoMesaSimpleFeature.scala
```diff
@@ -0,0 +1,60 @@
...
+    }
+    sftb.setDefaultGeometry(whereField)
+    data.foreach({ case (key, value) => sftb
+      .minOccurs(1).maxOccurs(1).nillable(false)
+      .add(key, value.getClass)
+    })
+
+    val sft = sftb.buildFeatureType
+    if(data.map(_._1).contains(whenField)) sft.getUserData.put(Constants.SF_PROPERTY_START_TIME, whenField) // when field is date
+    sft.getUserData.put(SimpleFeatureTypes.MIXED_GEOMETRIES, java.lang.Boolean.TRUE) // allow GeoMesa to index points and extents together
```
elahrvivaz: FYI you actually don't need this (although it won't hurt) - it's only if you specify the geometry binding as `Geometry`, which you don't do in your match above.

geomesa/src/main/scala/geotrellis/geomesa/geotools/GeometryToGeoMesaSimpleFeature.scala
```diff
@@ -0,0 +1,60 @@
...
+    sft.getUserData.put(Hints.USE_PROVIDED_FID, java.lang.Boolean.FALSE) // generate feature ids
```
elahrvivaz: This won't get used actually - you have to set (or not set) the hint on a per-feature basis. If you don't explicitly set the hint then generating IDs is the default behavior.

geomesa/src/main/scala/geotrellis/geomesa/geotools/GeometryToGeoMesaSimpleFeature.scala
```diff
@@ -0,0 +1,60 @@
...
+      case Line(ln) => sfb.add(ln)
+      case Polygon(pg) => sfb.add(pg)
+      case MultiPoint(mp) => sfb.add(mp)
+      case MultiLine(ml) => sfb.add(ml)
+      case MultiPolygon(mp) => sfb.add(mp)
+      case g: Geometry => throw new Exception(s"Unhandled Geotrellis Geometry $g")
+    }
+    data.foreach({ case (key, value) => sfb.add(value) })
+
+    sfb.buildFeature(featureId.orNull)
```
elahrvivaz: as an optimization, you might want to cache the simple feature types - it's probably going to be slow to generate it for each simple feature you create.

geomesa/src/main/scala/geotrellis/spark/io/geomesa/GeoMesaFeatureReader.scala
```diff
@@ -0,0 +1,54 @@
...
+
+class GeoMesaFeatureReader(val instance: GeoMesaInstance)(implicit sc: SparkContext) extends Serializable {
+  def readSimpleFeatures(
+    featureName: String,
+    query: Query,
+    simpleFeatureType: SimpleFeatureType,
+    numPartitions: Option[Int] = None
+  ): RDD[SimpleFeature] = {
+    val dataStore = instance.accumuloDataStore
+    dataStore.createSchema(simpleFeatureType)
```
elahrvivaz: this is a no-op if the schema already exists - but if the schema doesn't exist then there's not going to be any data in it... maybe that is ok here.

geomesa/src/main/scala/geotrellis/spark/io/geomesa/GeoMesaFeatureWriter.scala
```diff
@@ -0,0 +1,45 @@
...
+        val dataStore = instance.accumuloDataStore
+        partition.foreach { case (sft, sf) =>
+          // register feature types and write features
+          dataStore.createSchema(sft)
+          val featureWriter = dataStore.getFeatureWriterAppend(sft.getTypeName, Transaction.AUTO_COMMIT)
+          val attrNames = featureWriter.getFeatureType.getAttributeDescriptors.map(_.getLocalName)
+          try {
+            sf.foreach { rawFeature =>
+              val newFeature = featureWriter.next()
+              attrNames.foreach(an => newFeature.setAttribute(an, rawFeature.getAttribute(an)))
```
elahrvivaz: as a minor optimization, you could do
```
(0 until sft.getAttributeCount).foreach(i => newFeature.setAttribute(i, rawFeature.getAttribute(i))
```
lookup by index is always faster than lookup by attribute name

geomesa/src/test/scala/geotrellis/geomesa/geotools/GeoMesaSimpleFeatureType.scala
```diff
@@ -0,0 +1,41 @@
...
+      case "geotrellis.vector.MultiLine" => sftb.add(GeometryToGeoMesaSimpleFeature.whereField, classOf[jts.MultiLineString])
+      case "geotrellis.vector.MultiPolygon" => sftb.add(GeometryToGeoMesaSimpleFeature.whereField, classOf[jts.MultiPolygon])
+      case g => throw new Exception(s"Unhandled GeoTrellis Geometry $g")
+    }
+    sftb.setDefaultGeometry(whereField)
+    if(temporal) sftb.add(whenField, classOf[java.util.Date])
+    val sft = sftb.buildFeatureType
+    if(temporal) sft.getUserData.put(Constants.SF_PROPERTY_START_TIME, whenField) // when field is date
+    sft.getUserData.put(SimpleFeatureTypes.MIXED_GEOMETRIES, java.lang.Boolean.TRUE) // allow GeoMesa to index points and extents together
+    sft.getUserData.put(Hints.USE_PROVIDED_FID, java.lang.Boolean.FALSE) // generate feature ids
```
elahrvivaz: same as noted above - these 2 options won't really have any effect here.

lossyrob:
Thanks for the review @elahrvivaz. @pomadchin are all the review comments addressed?

pomadchin:
@lossyrob yes, walked through all comments

lossyrob:
Looks like there's some optimizations and other comments that still need to be addressed...need to either make the changes or reply to comments about why the suggested change shouldn't happen.

geomesa/src/test/scala/geotrellis/geomesa/geotools/GeoMesaSimpleFeatureType.scala
pomadchin: lines removed and created an issue https://github.com/geotrellis/geotrellis/issues/1654

geomesa/src/main/scala/geotrellis/spark/io/geomesa/GeoMesaFeatureWriter.scala
pomadchin: thx! implemented [there](https://github.com/geotrellis/geotrellis/pull/1621/files#diff-161e9579ddafc0cea10c58d73f5d7018R31)

geomesa/src/main/scala/geotrellis/geomesa/geotools/GeometryToGeoMesaSimpleFeature.scala
pomadchin: Lines removed

geomesa/src/main/scala/geotrellis/geomesa/geotools/GeometryToGeoMesaSimpleFeature.scala
pomadchin: cache [implemented](https://github.com/geotrellis/geotrellis/pull/1621/files#diff-cf90f45d0a4555af2e74583c673e4247R18)

geomesa/src/main/scala/geotrellis/spark/io/geomesa/GeoMesaFeatureReader.scala
pomadchin: [added](https://github.com/geotrellis/geotrellis/pull/1621/files#diff-7f28b41f7fe580fc9964aaf10ce13064R27) an existence check

geomesa/src/main/scala/geotrellis/geomesa/geotools/GeometryToGeoMesaSimpleFeature.scala
pomadchin: Thx, [implemented](https://github.com/geotrellis/geotrellis/pull/1621/files/#diff-cf90f45d0a4555af2e74583c673e4247R18).

geomesa/src/main/scala/geotrellis/geomesa/geotools/GeometryToGeoMesaSimpleFeature.scala
pomadchin: Thx, lines removed

geomesa/src/main/scala/geotrellis/geomesa/geotools/FeatureToGeoMesaSimpleFeatureMethods.scala
lossyrob: Weird, is this PR diff just not updating properly?

pomadchin: ._. looks like these comments attached to a certain sha

lossyrob: Oh. Yeah you can comment commits rather than prs, might of been what happened here.

lossyrob: Looks like it's just the new review functionality for github - it no longer removes comments when a new change comes in that addresses the issue. Which makes it harder to see if issues are addressed, which is pretty crappy of GitHub.

PR #1630, opened 2016-09-15
HBase upgrade

pomadchin:
Update up to HBase 1.2.3, fix Travis CI tests.


pomadchin:
random fail reason:`geotrellis.slick.PostgisSpec`

jamesmcclain:
Do you think that we should cache the downloaded hbase tarball?

pomadchin:
@jamesmcclain do you think to add `cache` instruction would be enough?

PR #1631, opened 2016-09-16
Streaming API

pomadchin:
Beta Streaming API (right now it demonstrates how we can port RDD API on Spark Streaming, cleanup required)
## Context

At first we wanted to try to use Streaming to control memory usage during the ingest process, to make backups (checkouts) and to ingest tiles by iterations. Definitely it can be used too ingest new coming tiles / features and controlling outside of the Spark stream according to interval, window, rate settings.
## API

Beautiful Spark Streaming API makes easy wrapping / reusing of GeoTrellis RDD functions, as `DStream` implements

``` scala
def foreachRDD(foreachFunc: (RDD[T]) ⇒ Unit): Unit
def transform[U](transformFunc: (RDD[T]) ⇒ RDD[U])(implicit arg0: ClassTag[U]): DStream[U]
def transformWith[U, V](other: DStream[U], transformFunc: (RDD[T], RDD[U]) ⇒ RDD[V])(implicit arg0: ClassTag[U], arg1: ClassTag[V]): DStream[V]
```

And other higher ordered functions , which makes possible GeoTrellis `RDD` functions usage.
## Spark Streaming extended details

Basic terms and important Streaming setting to control performance: 
- `Batch interval` - the interval at which the streaming API will update (socket / folder / receiver period lookup) data
- `Window` - data between times
- `receiver.maxRate` - maximum rate (number of records per second) at which each receiver will receive data
- `Streaming Events` - some events that can be set (exmpl. on microbatch complete), description would be provided in ADR document

WARN: using events we cant understand was completed or not first batch of input data computation.

More information would be published in the ADR document.


pomadchin:
@lossyrob @echeipesh and others who intersted, need your opinion on what functions should have streaming wrapper

PR #1632, opened 2016-09-20
More Joda serializers

fosskers:
### Motivation
Kryo and JodaTime do not get along. We use the [kryo-serializers](https://github.com/magro/kryo-serializers) library to provide extra serializers for Joda types. This PR bumps the version we depend on, and registers an additional serializer necessary for internal projects.

### DIFF
--- a/spark/build.sbt
+++ b/spark/build.sbt
@@ -4,7 +4,7 @@ name := "geotrellis-spark"
 libraryDependencies ++= Seq(
   "org.apache.spark" %% "spark-core" % Version.spark % "provided",
   "org.apache.hadoop" % "hadoop-client" % Version.hadoop % "provided",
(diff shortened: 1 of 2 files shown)


pomadchin:
Have you checked it on a n node cluster?

fosskers:
No, only locally.

pomadchin:
@fosskers spark replaces our kryo deps by its deps from its class path: https://github.com/geotrellis/geotrellis/pull/1286 please double check that replacement at least on a current lc emr demo O:

fosskers:
This PR will also die if we deprecate Java 7, since we can unregister all the Joda serializers.

pomadchin:
I don't like that `"de.javakaffee" % "kryo-serializers" % "0.38"` depends on kryo `3.0.3`, as we exclude it and replace by spark compatible `2.21`; but kryo `3.x` binary incompatible with `2.x`, that may cause problems (or may not, that would be great) on a real cluster.

fosskers:
This can be ignored, @pomadchin added the changes to his Spark 2 PR.

PR #1633, opened 2016-09-20
Example: ShardingKeyIndex[K]

fosskers:
### TODO
- [x] `ShardingKeyIndex` class
- [x] `JsonFormat` and registrator
- [x] Unit tests
- [x] Markdown docs

### Motivation
This PR provides an example of how to implement a new `KeyIndex` from scratch.

### DIFF
--- a/doc-examples/src/main/scala/geotrellis/doc/examples/spark/ShardingKeyIndex.scala
+++ b/doc-examples/src/main/scala/geotrellis/doc/examples/spark/ShardingKeyIndex.scala
@@ -68,8 +68,12 @@ class ShardingKeyIndex[K](val inner: KeyIndex[K], val shardCount: Int) extends K
  * A standard JsonFormat for [[ShardingKeyIndex]], parameterized on the key type ''K''.
  */
 class ShardingKeyIndexFormat[K: JsonFormat: ClassTag] extends RootJsonFormat[ShardingKeyIndex[K]] {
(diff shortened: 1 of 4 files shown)


fosskers:
@lossyrob unit tests pass, with help from @pomadchin and @moradology . My shard prefix algorithm is probably ripe for critique.

doc-examples/src/main/scala/geotrellis/doc/examples/spark/ShardingKeyIndex.scala
```diff
@@ -0,0 +1,128 @@
...
+ * ==Assumptions==
+ *   - The given shard count will be between 1 and 8.
+ *   - The ''inner'' index will produce a value less than 2^60 for any
+ *     given key.
+ */
+class ShardingKeyIndex[K](val inner: KeyIndex[K], val shardCount: Int) extends KeyIndex[K] {
+
+  /* Necessary for extending `KeyIndex` */
+  def keyBounds: KeyBounds[K] =
+    inner.keyBounds
```
fosskers: This might be incorrect.

lossyrob: No, since the Keys are actually completely decoupled from the index side of things, and the shard doesn't effect anything but the index, the keybounds remains valid. So we're good.

Issue #1575, opened 2016-07-01
The toSpatial Method gives metadata of type TileLayerMetadata[SpaceTimeKey]


jamesmcclain:
The [toSpatial](https://github.com/geotrellis/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/filter/SpaceTimeRasterToSpatialMethods.scala#L31) method which takes an RDD of `SpaceTimeKey`s to an RDD of `SpatialKey`s produces an RDD whose associated metadata is of type `TileLayerMetadata[SpaceTimeKey]` (although the KeyBounds appear to adhere to type `SpatialKey`).

pomadchin:
There is a question should we require all custom metadata types to depend on key type or not.

jamesmcclain:
Life is a mystery.

fosskers:
https://github.com/geotrellis/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/filter/ToSpatial.scala#L32

The above has type `KeyBounds[K]`, so I don't think it automatically conforms to `SpatialKey`, @jamesmcclain . Yeah, there's nothing here that forces the metadata to recognize the change in key type.

fosskers:
@pomadchin , do you mean function signatures should be altered to be `RDD[(K, V)] with Metadata[K, M]`?

fosskers:
There is hope in sight.

pomadchin:
@fosskers function signature is: 

```scala
def apply[
    K: SpatialComponent: TemporalComponent,
    V,
    M: Component[?, Bounds[K]]
  ](rdd: RDD[(K, V)] with Metadata[M], instant: Long): RDD[(SpatialKey, V)] with Metadata[M] = ???
```

That means that in the current function implementation compiler has not information about `M`, so probably: 

```scala
def apply[
    K: SpatialComponent: TemporalComponent,
    V,
    M[K]: Component[?, Bounds[K]]
  ](rdd: RDD[(K, V)] with Metadata[M[K]], instant: Long): RDD[(SpatialKey, V)] with Metadata[M[SpatialKey]] = ???
``` 

But we need to have a type class to make `toSpatial` for metadata.

fosskers:
That's what I was experimenting with yesterday, although now I'm debating whether an entire it's worth it to invent a new typeclass just for this one use case.

pomadchin:
[...]

PR #1643, opened 2016-09-29
Custom Functor Typeclass

fosskers:
### TODO
- [x] `Functor` trait
- [x] Instance for `TileLayerMetadata[K]` (maps over `KeyBounds`)
- [x] Remove `SpatialComponent` constraint on `TIleLayerMetadata.apply`
- [x] Get `.map` to inject properly (thanks to @pomadchin )
- [x] ~~Add test to confirm #1575 fix~~ (no point, the compiler guarantees it)

### Motivation
#1575 describes how `TileLayerRDD.toSpatial` doesn't reflect the change in key type in its metadata. After much deliberation, we recognized our attempts to address this as attempts to write a `Functor` instance for `TileLayerMetadata[K]` which maps over its `KeyBounds`. Unfortunately, Scalaz's `Functor` trait is a bit too restrictive, in that types that use it can't impose additional constraints on their inner type. Compare Scalaz:

```scala
trait Functor[F[_]] extends InvariantFunctor[F] { self =>

  /** Lift `f` into `F` and apply to `F[A]`. */
  def map[A, B](fa: F[A])(f: A => B): F[B]
...
}
```
to our implementation:

```scala
 trait Functor[F[_], A] extends MethodExtensions[F[A]]{
   /** Lift `f` into `F` and apply to `F[A]`. */
   def map[B](f: A => B): F[B]
 }
```
where `A` is lifted into `Functor`'s type signature. This allows us to write code like this:
```scala
def foo[K: A: B: C](implicit f: Functor[TileLayerMetadata, K]): ??? = {
...
  tlm.map(k => ... ) // `k` has all the power of its constraints A, B, and C.
...
}
```


fosskers:
This may or may not be an actual Functor, depending on how well we can convince ourselves that the second Functor law holds:

```haskell
fmap (g . h) = (fmap g) . (fmap h)
```

If you only had one implicit `Functor[F, A]` in scope, then:

```scala
val m: F[A] = ???
val f: A => B = ???
val g: B => C = ???

val n: F[C] = m.map(f).map(g)
```
wouldn't compile. But if you *did* have both `Functor[F, A]` and `Functor[F, B]` (which may both use the same underlying implicit class, see `class TlmFunctor[A]`), then the above would compile. Furthermore, so would:

```scala
val n: F[C] = m.map(k => g(f(k)))
```
and so the Functor law holds, and it is indeed a Functor.

spark/src/main/scala/geotrellis/spark/filter/Implicits.scala
```diff
@@ -16,9 +16,8 @@ trait Implicits {
       extends TileLayerRDDFilterMethods[K, V, M]
 
   implicit class withSpaceTimeToSpatialMethods[
-    K: SpatialComponent: TemporalComponent,
+    K: SpatialComponent: TemporalComponent: λ[α => M[α] => Functor[M, α]]: λ[α => Component[M[α], Bounds[α]]],
```
lossyrob: This should be type aliased w/ a broken down explained in a comment, pretty tough to parse and also understand why they just can't be context bounds on M

spark/src/main/scala/geotrellis/spark/package.scala
```diff
@@ -62,6 +61,14 @@ package object spark
       new ContextRDD(rdd, metadata)
   }
 
+  implicit class TileLayerMetadataFunctor[A](val self: TileLayerMetadata[A]) extends Functor[TileLayerMetadata, A] {
```
lossyrob: Comments to explain why this is needed.

spark/src/main/scala/geotrellis/spark/filter/Implicits.scala
fosskers: It's definitely space voodoo the first time you see it, yeah. For technical reasons they can't be bounds on M, so I'll explain that.

PR #1647, opened 2016-10-03
ETL Improvements and other issues fixes

pomadchin:
* https://github.com/geotrellis/geotrellis/issues/1547
* https://github.com/geotrellis/geotrellis/issues/1642
* https://github.com/geotrellis/geotrellis/issues/1641
* https://github.com/geotrellis/geotrellis/issues/1276
* https://github.com/geotrellis/geotrellis/issues/1408


pomadchin:
@RickMohr and @lossyrob what do you think about clipping by extent?

RickMohr:
If clipping by extent and also transforming coordinates, should the clipping frame be specified in the source or target coordinate system? I think it would be more useful in the target coordinate system. In either case some doc would be important to avoid confusion.

pomadchin:
@RickMohr in the current implementation - the source coordinate system; though I agree with you, let's make it in target coordinate system and need to update docs, thanks!

RickMohr:
Clipping looks good, thanks!

pomadchin:
:0 sorry for that @lossyrob

PR #1650, opened 2016-10-05
Add MGRS encode/decode support to proj4

jpolchlo:
This PR provides a port of proj4js/mgrs.  This code provides an object offering functions to transform long/lat coordinates to their containing Military Grid Reference System (MGRS) grid cell, and from cell specification strings to their long/lat bounding boxes or the center location of their respective bounding boxes.

The original code can be found [here](https://github.com/proj4js/mgrs), and has a very permissive license.


proj4/src/main/scala/geotrellis/proj4/mgrs/MGRS.scala
```diff
@@ -0,0 +1,485 @@
...
+    encode(LLtoUTM(long, lat), accuracy)
+  }
+
+  /**
+   * Converts an MGRS grid specifier to a bounding box in long/lat coordinates.  The return value gives
+   * the bounding box as (left, bottom, right, top) positions.  If the accuracy is zero (i.e., no numerical
+   * location characters following the zone number, zone letter, and grid square identifiers), the bounding
+   * box will be degenerate; use mgrsToLongLat in that case.
+   */
+  def mgrsToBBox(mgrs: String): (Double, Double, Double, Double) = {
```
echeipesh: This is obviously an `Extent` but we can't have `proj4` subproject depends on `vector` project. Which kind of means that this should be moved to `vector` even through this is dealing with a projecting. I favor the latter @lossyrob thoughts?

Issue #1659, opened 2016-10-06
OSM => VectorTiles :: (3) RDD[OSMElement] to RDD[Feature[G, D]]


fosskers:
*This is Part 3 of a series of issues documenting the process of creating a world's worth of VectorTiles from OSM Planet data. Please use these issues to discuss solutions.*

OSM data has the concepts of `Node`, `Way`, and `Relation`, which don't directly correspond to GeoTrellis geometries.

**Questions:** 

1. What is necessary for such a conversion?
2. What should `D` be?

fosskers:
This isn't a straight mapping. OSM `Way`s are (conceptually) made up of `Node`s, but the XML itself only contains references to those nodes to save on memory. Different Ways can also share Nodes using this "pointer" system. Here is a full node:

```xml
<node id="25496583" lat="51.5173639" lon="-0.140043" version="1" changeset="203496" user="80n" uid="1238" visible="true" timestamp="2007-01-28T11:40:26Z">
  <tag k="highway" v="traffic_signals"/>
</node>
```

And here is a Way. The `ref` attribute corresponds to a `node`'s `id` attribute.
```xml
<way id="5090250" visible="true" timestamp="2009-01-19T19:07:25Z" version="8" changeset="816806" user="Blumpsy" uid="64226">
  <nd ref="822403"/>
  <nd ref="21533912"/>
  <nd ref="821601"/>
  <nd ref="21533910"/>
  <nd ref="135791608"/>
  <nd ref="333725784"/>
  <nd ref="333725781"/>
  <nd ref="333725774"/>
  <nd ref="333725776"/>
  <nd ref="823771"/>
  <tag k="highway" v="residential"/>
  <tag k="name" v="Clipstone Street"/>
  <tag k="oneway" v="yes"/>
</way>
```

fosskers:
Furthermore, Ways correspond to both `Line` and `Polygon` in GeoTrellis. The full analogue to a Polygon is called an `Area`, where its first and last node are the same. In terms of the data structure, however, both are stored as a `Way`.

fosskers:
The `Relation` element might not have an analogue in GeoTrellis.

fosskers:
`D` as `TagMap` seems to be a good fit.

lossyrob:
I think that `Relation` should end up as specific tags in the TagMap, so for all geometries that participate in that relation, they get the relation tags (clearly marked as tags coming from the relation)

fosskers:
[...]

PR #1663, opened 2016-10-07
RDD[Element] => RDD[Feature[Geometry, Tree[ElementData]]]

fosskers:
### TODO
- [x] Mock Scala types for parsed OSM data (may change in future)
- [x] Interpret open `Way`s as GeoTrellis `Line`
- [x] Interpret closed `Way`s as GeoTrellis `Line` if `highway=*` or `barrier=*` is given, unless `area=yes` is given
- [x] Interpret all other closed `Way`s as GeoTrellis `Polygon`
- [x] Interpret any left over `Node`s (not used in any `Way`) as single GeoTrellis `Point`s
- [x] Fix Node order issue in Ways
- ~~Keep used `Node`s, and have their `TagMap` reference the `Way`s they're used in~~ (not necessary with new `Tree` approach to storing child `ElementData`)
- [ ] Handle `Relation`s
  - [x] MultiPolygons
    - [x] Merge needed `Line`s into `Polygon`s
    - [x] Associate "inner" and "outer" polygons
    - [x] Handle combined metadata correctly
  - [ ] Other relationships relegated to `TapMap`s of child Elements
- [ ] Unit tests

### Motivation
#1659

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -73,9 +73,10 @@ lazy val root = Project("geotrellis", file(".")).
     hbase,
     geowave,
     geomesa,
-    geotools,    
+    geotools,
     slick,
-    vectortile
+    vectortile,
+    osm
   ).
   settings(commonSettings: _*).
   settings(
@@ -92,6 +93,9 @@ lazy val root = Project("geotrellis", file(".")).
 
 lazy val macros = Project("macros", file("macros")).
   settings(commonSettings: _*)
+
+lazy val osm = Project("osm", file("osm"))
+  .settings(commonSettings: _*)
 
 lazy val vectortile = Project("vectortile", file("vectortile"))
   .dependsOn(vector, spark)
--- a/build.sbt
+++ b/build.sbt
@@ -95,7 +95,7 @@ lazy val macros = Project("macros", file("macros"))
   .settings(commonSettings: _*)
 
 lazy val osm = Project("osm", file("osm"))
-  .dependsOn(util, vector)
+  .dependsOn(util, vector, sparkTestkit % "test->test", spark % "provided;test->test")
   .settings(commonSettings: _*)
 
 lazy val vectortile = Project("vectortile", file("vectortile"))
--- a/build.sbt
+++ b/build.sbt
@@ -91,10 +91,11 @@ lazy val root = Project("geotrellis", file(".")).
   )
   .settings(unidocSettings: _*)
 
-lazy val macros = Project("macros", file("macros")).
-  settings(commonSettings: _*)
+lazy val macros = Project("macros", file("macros"))
+  .settings(commonSettings: _*)
 
 lazy val osm = Project("osm", file("osm"))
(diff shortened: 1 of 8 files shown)


fosskers:
Some notes from my exploration of an 11MB OSM XML file, from a bounding box around North Vancouver:

- The file is ~70% `Node`, ~29% `Way`s, and < 1% `Relation`s
- Many of the `Way`s are streets, and are named in their tags.
- Multipolys looks like this:

```xml
 <relation id="1001354" visible="true" version="1" changeset="5062643" timestamp="2010-06-23T22:34:26Z" user="WBSKI" uid="14351">
   <member type="way" ref="63451236" role="outer"/>
   <member type="way" ref="63451237" role="inner"/>
   <member type="way" ref="63451238" role="inner"/>
   <member type="way" ref="63451239" role="inner"/>
   <tag k="type" v="multipolygon"/>
 </relation>
```

fosskers:
For my own notes:
```xml
 <relation id="1643967" visible="true" version="2" changeset="32655074" timestamp="2015-07-15T15:47:52Z" user="pathym" uid="689176">
  <member type="way" ref="104017200" role=""/>
  <member type="way" ref="104017171" role=""/>
  <member type="way" ref="360740836" role=""/>
  <tag k="name" v="Keith Rd"/>
  <tag k="network" v="lcn"/>
  <tag k="ref" v="Keith"/>
  <tag k="route" v="bicycle"/>
  <tag k="type" v="route"/>
 </relation>

 <relation id="1524228" visible="true" version="5" changeset="36415698" timestamp="2016-01-06T23:52:47Z" user="MetVanRider123acme" uid="2800822">
  <member type="way" ref="65253226" role="outer"/>
  <member type="way" ref="166619861" role="outer"/>
  <member type="node" ref="2089388397" role="admin_centre"/>
  <tag k="addr:province" v="British Columbia"/>
  <tag k="addr:province:fr" v="Colombie-Britannique"/>
  <tag k="admin_level" v="8"/>
  <tag k="attribution" v="Natural Resources Canada"/>
  <tag k="boundary" v="administrative"/>
  <tag k="name" v="North Vancouver City"/>
  <tag k="source" v="CanTopo"/>
  <tag k="type" v="boundary"/>
 </relation>
```

fosskers:
I pulled down the data for [Baarle-Nassau](https://en.wikipedia.org/wiki/Baarle-Nassau), a town in the Netherlands which has exclaves within Belgian enclaves.

Peering at the actual OSM data, they have many over-sized Relations, and a number of Super-Relations (relations within relations).  There are also some conflicts with Relation types, in that `boundary`, `multilinestring` and `land_area` seem to be used interchangeably. According to the OSM wiki, only `boundary` is blessed as an "established" relation type.

My initial main take-away is that `type=boundary` should be handled with the same logic as `multipolygon`.

fosskers:
This work has been moved to: https://github.com/geotrellis/vectorpipe

PR #1667, opened 2016-10-08
Feature/geotiff photometric interpretation

metasim:
Added GeoTiff read/write support of `TIFFTAG_PHOTOMETRIC` via `GeoTiffOptions`.


lossyrob:
+1 after travis

PR #1696, opened 2016-10-24
Update Contributor list

fosskers:
Updated to include core members joined since 2015.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,18 +200,25 @@ This is a list of features contained in the GeoTrellis library. It is broken up 
 
 ## Contributors
 
- - Josh Marcus
- - Erik Osheim
- - Rob Emanuele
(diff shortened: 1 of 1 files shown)


lossyrob:
Now that we have a bunch of contributors, is it worth trying to keep this list up? This doesn't have everybody.

Maybe have a definition of "core"? Like, people who have contributed to x amount of +/-?

fosskers:
>  is it worth trying to keep this list up?

Perhaps not, the `Contributors` tab does a pretty good job of this already. Core contributors are also easily checkable from the GeoTrellis organization page.

lossyrob:
True. I'm down for deleting this section.

fosskers:
Done!

PR #1697, opened 2016-10-24
Ingest tiles without CRS in metadata

pomadchin:
Fixes https://github.com/geotrellis/geotrellis/issues/1414

Double checked that we can read tiffs without crs (crs would be set to `LatLng`), added additional setting `inputCrs` for `S3` and `Hadoop` `InputFormats`.


spark-etl/src/main/scala/geotrellis/spark/etl/config/json/ConfigFormats.scala
```diff
@@ -200,7 +200,8 @@ trait ConfigFormats {
       "backend" -> bf.write(i.backend),
       "cache"   -> i.cache.toJson,
       "noData"  -> i.noData.toJson,
-      "clip"    -> i.clip.toJson
+      "clip"    -> i.clip.toJson,
+      "input"   -> i.crs.toJson
```
lossyrob: Shouldn't this be `"crs"`?

Issue #1683, opened 2016-10-20
Fix Voronoi implementation


lossyrob:
Our current implementation does not handle some edge cases. JTS has a more robust implementation. We should either fix ours or move to theirs.

PR #1699, opened 2016-10-25
Repaired non-robust Delaunay triangulator.

jpolchlo:
This PR replaces the custom, non-robust Delaunay triangulator with wrapped version of the JTS triangulator.  The old triangulator did not implement robust predicates for collinearity/cocircularity which could cause failures in some real-world data sets.

The JTS triangulator was wrapped to allow a more idiomatic Scala usage.  Namely, allowing for PointFeatures to be used more easily, which would facilitate the use of a triangulation for the purposes of, say, interpolating a field of numbers defined over a sample of points.  This wrapped version also replaces the quad-edge structure used by JTS for a lighter-weight half-edge structure.  Documentation for that structure can be found in docs/vector/voronoi.md.

This PR closes issue #1683.


lossyrob:
@jpolchlo you made benchmarks for the old voronio, right? Is it possible to rerun those benchmarks, against the two different versions, to see if we're losing a lot of speed?

jpolchlo:
You and I sat down to look at benchmarks, but I didn't write them myself,
so I'm not 100% on how to proceed.  With some pointers, I'm sure I could do
it. But I wonder what you have in mind here, since the version that got
replaced will fail on a certain class of inputs, and surely, if I were to
make my original code robust, it will necessarily slow down based on the
additional operations needed to ensure correctness.

lossyrob:
@jpolchlo the idea is to have a sense of how much things have slowed down. And to have an idea of how efficient this operation is...I remember that you were calculating complex triangulations on the nickel dataset in about 200ms. If this turns out to be a 2x vs a 10x slowdown, that would be good to understand. If it's a serious slowdown, I'd want to either try to optimize before release or at least put in an issue for future optimization.

jpolchlo:
Roger, boss.  I'll see what I can figure out.

vector/src/main/scala/geotrellis/vector/voronoi/Delaunay.scala
```diff
@@ -333,3 +571,96 @@ case class Delaunay(verts: Array[Point]) {
...
+      if (a < b) {
+        val line = Line(a,b)
+        Rasterizer.foreachCellByLineString(line, re){ (c,r) => tile.set(c,r,1) }
+      }
+    }}
+  }}
+  val cm = ColorMap(scala.collection.immutable.Map(1 -> 0x000000ff, 255 -> 0xffffffff))
+  tile.renderPng(cm).write("delaunay.png")
+}
+*/
```
lossyrob: What's going on with all the commented code here?

vector-test/src/test/scala/spec/geotrellis/vector/voronoi/DelaunaySpec.scala
```diff
@@ -102,13 +111,42 @@ class DelaunaySpec extends FunSpec with Matchers {
...
+      // rasterizeDT(dt)
 
       (dt.triangles.forall{ case ((ai,bi,ci),_) =>
         val otherPts = (0 until numpts).filter{ i: Int => i != ai && i != bi && i != ci }
         otherPts.forall{ i => ! localInCircle((ai,bi,ci),i) }
       }) should be (true)
     }
+
+    it("should work on a real dataset with many collinear points") {
+      val allcities = ResourceReader.asString("populous_cities.txt")
```
lossyrob: We don't normally read test data out of resources packed in the JAR - is there an advantage you see of doing it this way, other then adding to test files similar to other tests? We would put it here https://github.com/geotrellis/geotrellis/tree/master/vector-test/data

util/src/main/scala/geotrellis/util/ResourceReader.scala
```diff
@@ -0,0 +1,10 @@
+package geotrellis.util
+
+import java.io._
+
+object ResourceReader {
+  def asString(filename: String): String = {
```
lossyrob: This is for test functionality, not sure it's worth putting in the core utils JAR

jpolchlo: Had talked to Eugene about this.  He suggested putting it in util, but it's not super valuable, so I'll nuke it.

vector/src/main/scala/geotrellis/vector/voronoi/Delaunay.scala
```diff
@@ -102,234 +107,19 @@ object Predicates {
...
-    }
-  }
-
-  /**
-   * Provides a handle on the bounding loop of a Delaunay triangulation.  All of the half-edges
-   * reachable by applying next to boundary have no bounding face information.  That is
-   * boundary(.next)*.face == None.
-   */
-  val boundary = triangulate(0, vIx.length-1)
+  val triangles = _triangles.map{ poly => Polygon.jtsToPolygon(poly.asInstanceOf[JTSPolygon]) }
```
lossyrob: can just use `Polygon(poly.asInstanceOf[JTSPolygon]`

vector-test/src/test/scala/spec/geotrellis/vector/voronoi/DelaunaySpec.scala
```diff
@@ -1,114 +0,0 @@
...
-import scala.math.pow
-import org.apache.commons.math3.linear._
-
-import geotrellis.raster._
-import geotrellis.raster.rasterize._
-import geotrellis.raster.render._
-
-import org.scalatest.{FunSpec, Matchers}
-
-class DelaunaySpec extends FunSpec with Matchers {
```
lossyrob: Should we really scrap all these tests? I think still testing our JTS backed Delaunay class makes sense.

vector/src/main/scala/geotrellis/vector/voronoi/Delaunay.scala
```diff
@@ -102,234 +107,19 @@ object Predicates {
...
-  private def distinctBy[T](eq: (T,T) => Boolean)(l: List[T]): List[T] = {
-    l match {
-      case Nil => Nil
-      case x::Nil => x::Nil
-      case x::y::xs if eq(x,y) => distinctBy(eq)(x::xs)
-      case x::y::xs => x :: distinctBy(eq)(y::xs)
-    }
+  private val _triangles = {
+    val tris = subd.getTriangles(gf)
+    for ( i <- 0 until tris.getNumGeometries ) yield tris.getGeometryN(i)
```
lossyrob: use a `cfor` loop for performance...also do we need the private val?

```scala
import spire.syntax.cfor._

val triangles: Seq[Polygon] = {
  val arr = Array.ofDim[Polygon](tris.getNumGeometries)
  val len = arr.length
   cfor(0)(_ < len, _ + 1) { i =>
    arr(i) = Polygon(tris.getGeometryN(i).asInstanceOf[JTSPolygon])
  }
  arr
}
```

vector-test/src/test/scala/spec/geotrellis/vector/voronoi/DelaunaySpec.scala
jpolchlo: As much as I would like to provide some unit tests here, I'm just at a loss for what should be tested.  It seems that most anything that I could test for tests JTS's feature set, and not the features of this package.  I'm not sure if that's appropriate.  As it is, there is an indirect test of Delaunay and Voronoi in the EuclideanDistanceTileSpec.  I'd be happy for some suggestions about what should be tested for in these classes!

jpolchlo: After some discussion, we've opted to re-insert some tests essentially as regression tests.

lossyrob:
@jpolchlo "This PR closes issue #1683." doesn't kick off the auto close. Has to be "Fixes #1683".

PR #1700, opened 2016-10-25
Additional ETL Improvements

pomadchin:
- refactored configs, to allow etl config parsing functions usage in client projects
- refactored EtlConf case class (removed unused vals)


lossyrob:
What's an example of what this enables? do we need documentation additions for this?

pomadchin:
@lossyrob it allows at least to re-use json reading function, nothing too critical, but probably it makes sense to use it in integration tests project; 

The important change is `class EtlConf(val input: Input, val output: Output, val inputProfile: Option[BackendProfile] = None, val outputProfile: Option[BackendProfile] = None)` => `class EtlConf(val input: Input, val output: Output)`

spark-etl/src/main/scala/geotrellis/spark/etl/config/DefaultEtlConf.scala
```diff
@@ -0,0 +1,87 @@
+package geotrellis.spark.etl.config
+
+import geotrellis.spark.etl.config.json._
+
+import org.apache.spark.SparkContext
+import com.github.fge.jackson.JsonLoader
+import spray.json._
+
+trait DefaultEtlConf extends ConfigParse {
```
lossyrob: The name default is confusing. Should it be BaseEtlConfig? Perhaps ask the room, if I'm the only one who thinks so then perhaps we keep it as is

pomadchin: @lossyrob I think that the word base is better, though for some reason I chose Default (guess it's spray json influence )

PR #1702, opened 2016-10-25
Add interpretAs and withNoData methods to Tile

echeipesh:
Fixes: https://github.com/geotrellis/geotrellis/issues/1674

Allows for more flexibility when working around NoData values.

Additionally this expands the capacity of the `ConstantTile` to represent constant tiles other than tiles with `ConstantNoDataHandling` to satisfy the conversions implied by the new API.

TODO:
 - [x] Unit tests for interpret 
 - [x] GeoTiffMultibandTile.interpretAs
 - [x] `Tile.withNoData`
 - [x] Tile concrete classes `noDataValue` constructor overload
 - [x] New API review / discussion

### DIFF
--- a/raster-test/src/test/scala/geotrellis/raster/ArrayTileSpec.scala
+++ b/raster-test/src/test/scala/geotrellis/raster/ArrayTileSpec.scala
@@ -91,18 +91,32 @@ class ArrayTileSpec extends FunSpec
      */
 
     def checkFloatInterpretAs(tile: Tile, udCt: Double => CellType, constCt: CellType) = {
-      tile.foreachDouble { v =>
-        withClue(s"ND=$v") {
-          assert(tile.interpretAs(udCt(v)).interpretAs(constCt).asRawTile equals tile)
-        }
+      for {
+        r <- 0 until tile.rows
+        c <- 0 until tile.cols
+      } {
+        val v = tile.get(c, r)
+        val udTile = tile.interpretAs(udCt(v))
+        val constTile = udTile.interpretAs(constCt)
+        val res = constTile.asRawTile
+        withClue(s"ND=$v") { assert(res equals tile) }
+        val cell = udTile.getDouble(c,r)
+        withClue(s"udTile($c, $r), ND=$v") { assert(isNoData(cell)) }
       }
     }
 
     def checkIntInterpretAs(tile: Tile, udCt: Int => CellType, constCt: CellType) = {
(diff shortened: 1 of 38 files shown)


lossyrob:
Is this WIP until the RFC gets worked out?

echeipesh:
I'd be happy to bundle them together, but they don't have to be. This PR can be merged as is and addresses the issue that spawned it.

echeipesh:
@lossyrob @moradology Ready for review again after `.withNoData` changes. 

One final through is that what is currently `.interpretAs` might be better named `.withCellType`. That name seems consistent with behavior:
 - If we're not  changing underlying data type this just re-wraps the array with new CellType.
 - Otherwise we coerce to the desired cell type and label it with new NoDataHandling

lossyrob:
@echeipesh I think `withCellType` is a bit ambiguous, and might be read as doing the same thing `convert` does. But I've though about this less, so if you are strong +1 on that name I'll follow your lead.

moradology:
I also think `withCellType` seems more ambiguous. Have we considered, as an alternative to `interpretAs`, `reinterpret`?

echeipesh:
I prefer `interpretAs` to `reinterpret`. I thought that `withCellType` has nice symmetry to `withNoData` but lets discount it since its got two votes against it. I'm still not totally happy with `interpretAs`, I'd like to capture the intuition is that this is equivalent of calling `_.toDouble` or `_.toInt` on cell by cell basis.

- `toCellType(IntCells)`
- `asCellType(IntCells)`
- `convertRaw(IntCells)` 
- `convertFromRaw(IntCells)` 
- `reassign(IntCells)`
- `coerce(IntCells)`

fosskers:
Semantic ambiguity could be addressed in docstrings.

fosskers:
+1 `interpretAs`. I think it's the most semantically clear.

pomadchin:
I like the word `coercion` but it usually means an _implicit_ type conversion the word _cast_ used for explicit (correct me if I'm wrong). So probably `interpretAs` or even `cast` (but `cast` looks similar to `convert`). But I see no problems `withCellType` name as well, as for me it means more a `type` _interpretation_ rather than `converting` :)) 
-  +1 `interpretAs`

echeipesh:
Thank you everybody for chiming in on the naming!
For eons let it written, let it be said: `Tile.interpretAs`

Issue #1272, opened 2016-01-06
x, y, z coordinates in Vector - optionally drop the z on deserialization


moradology:
It isn't uncommon for coordinates in GeoJson to be specified with a Z value - GeoTrellis can't/won't handle the third dimension, but it would be useful to have an option which specifies automatically dropping the third dimension on JSON deserialization.

lossyrob:
Thinking through this API

```scala
string.parseGeoJson[MultiPolygon](options = GeoJson.Options(ignoreExtraDimensions = true))
```

lossyrob:
Fix to this would fix  #1676

PR #1704, opened 2016-10-25
Add support for 3d points in GeoJson

jpolchlo:
This PR allows GeoTrellis to read GeoJson with 3-dimensional coordinate information by adopting the approach of throwing out the z-coordinate.  Smarter handling of elevation data will require a more thorough approach, requiring a new Geometry subclass with z-coordinate capabilities, which would necessarily percolate out into the rest of the library.

This PR closes issues #1272.


lossyrob:
@jpolchlo need to say exactly "Fixes #1272" in order for the automatic issue closing to work

PR #1705, opened 2016-10-25
RDDWriters resources allocation improvements

pomadchin:
We need to be sure that partition is not empty, as creating a connection / session is expensive.

Thx @fosskers for a pointer long time ago.


fosskers:
With this change, performance should improve when `RDD`s are "over-partitioned".

PR #1709, opened 2016-10-25
RFC: Tile.withNoData and NoData Constructor

echeipesh:
Requires: https://github.com/geotrellis/geotrellis/pull/1702

##  Adding  .withNoData methods on Tile

Because `Tile.withNoData` has to be pushed up it must be generic on the no data value:

```scala
def withNoData[T: Numeric](v: T): Tile = 
  IntArrayTile(array, cols, rows, implicitly[Numeric[T]].toInt(v).toShort) 
```

This would be useful if you wanted to write code setting NoData value to some "safe"/common value like `0` without having to match on the `CellType` and construct a type.

However, that may result in coercion when assigning NoData value:

```scala
val tile: Tile = IntArrayTile(array, cols, rows)
tile.withNoData(3.5) // IntUserDefinedNoDataArrayTile(3)
tile.withNoData(None) //IntCellType
tile.asRawTile // IntCellType
```

I think the coercion is intuitive and "correct" so this API seems like it should be flushed out.

## NoData Constructors

Working with our CellType constructs is a bit of a pain and in places where it is obvious what we mean and type safety is assured this apply overload provides a really intuitive interface:

```scala
  def apply(arr: Array[Int], cols: Int, rows: Int, noData: Option[Int]): IntArrayTile = 
   noData match {
    case Some(nd) if isNoData(nd) =>
      new IntConstantNoDataArrayTile(arr, cols, rows) 
    case Some(nd) =>  
     new IntUserDefinedNoDataArrayTile(arr, cols, rows, IntUserDefinedNoDataCellType(nd)) 
    case None =>
      new IntRawArrayTile(arr, cols, rows) 
  }  
```

Putting this as PR for discussions because there are tradeoffs involved.

### DIFF
--- a/raster-test/src/test/scala/geotrellis/raster/ArrayTileSpec.scala
+++ b/raster-test/src/test/scala/geotrellis/raster/ArrayTileSpec.scala
@@ -91,18 +91,32 @@ class ArrayTileSpec extends FunSpec
      */
 
     def checkFloatInterpretAs(tile: Tile, udCt: Double => CellType, constCt: CellType) = {
-      tile.foreachDouble { v =>
-        withClue(s"ND=$v") {
-          assert(tile.interpretAs(udCt(v)).interpretAs(constCt).asRawTile equals tile)
-        }
+      for {
+        r <- 0 until tile.rows
+        c <- 0 until tile.cols
+      } {
+        val v = tile.get(c, r)
+        val udTile = tile.interpretAs(udCt(v))
+        val constTile = udTile.interpretAs(constCt)
+        val res = constTile.asRawTile
+        withClue(s"ND=$v") { assert(res equals tile) }
+        val cell = udTile.getDouble(c,r)
+        withClue(s"udTile($c, $r), ND=$v") { assert(isNoData(cell)) }
       }
     }
 
     def checkIntInterpretAs(tile: Tile, udCt: Int => CellType, constCt: CellType) = {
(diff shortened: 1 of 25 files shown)


lossyrob:
@echeipesh can you specify the tradeoff here with accepting these changes (the negative effects)?

echeipesh:
The tradeoff is in losing some type safety when setting NoData value as shown in the blurb above. Because the `Tile` interface does not have any information about CellType we can't produce a compile time error.

This should not produce any performance impact because `.withNoData` method would create a tile through our current constructors, just different parameters.

lossyrob:
I am not seeing how this isn't type safe...the fact that `3.5` gets coerced to 3?

echeipesh:
Yep just that. It can get weirder too. If you tried to set NoData of 450 on `ByteCells` tile it will end up being -62 because of the `.toByte` call. To be clear I think this is covered by "user be-ware" clause.

lossyrob:
Yeah, I don't think it's that bad. It's the same thing if you do a `set` call on a mutable tile with specific cell types.

echeipesh:
That's good point, in that case I'll make it part of https://github.com/geotrellis/geotrellis/pull/1702 so we can check the expanded NoData API as a whole

echeipesh:
Replaced by: https://github.com/geotrellis/geotrellis/pull/1702

PR #1711, opened 2016-10-26
Hdfs Range Reads

jbouffard:
This PR adds a rangeRead method to HdfsUtils that allows for reading ranges of bytes.

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/io/hadoop/HdfsUtils.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/hadoop/HdfsUtils.scala
@@ -178,6 +178,21 @@ object HdfsUtils extends LazyLogging {
     bytes
   }
 
+  def readRange(path: Path, start: Long, length: Int, conf: Configuration): Array[Byte] = {
+    val fs: FileSystem = path.getFileSystem(conf)
+    val end: Long = start + length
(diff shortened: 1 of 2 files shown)


jbouffard:
@lossyrob Ready for review.

lossyrob:
@jbouffard there was one build that passed, but another that failed with

```
geotrellis.spark.io.hadoop.HdfsUtilsSpec *** ABORTED ***
[info]   java.io.FileNotFoundException was thrown inside describe("HdfsUtils"), construction cannot continue: "raster-test/data/geotiff-test-files/ls8_int32.tif (No such file or directory)" (HdfsUtilsSpec.scala:14)
```

Not sure how one passed and one didn't, maybe caching?

lossyrob:
@jbouffard will this allow us to implement a streaming Hdfs byte reader, so that we can place large geotiffs on S3 and eventually do a windowed ingest from them? Is that functionality planned for a follow up PR?

jbouffard:
@lossyrob Yes, this is similar to the `readRange` method that we currently have for the `S3Client` [link](https://github.com/geotrellis/geotrellis/blob/master/s3/src/main/scala/geotrellis/spark/io/s3/S3Client.scala#L183). While I don't think we ever officially planned anything, I assumed that's what I'd be doing after I finish up the tasks for 1.0.

PR #1715, opened 2016-10-26
Update to GeoJson documentation

### LINKED ISSUE #1432: Add some documentation about using the JsonFeatureCollectionMap
In general, JSON is fairly well documented. A notable blind spot is the `JsonFeatureCollectionMap`.

jpolchlo:
Fixes #1432 

Expands the documentation for GeoJson support.

### DIFF
--- a/docs/vector/geojson-support.md
+++ b/docs/vector/geojson-support.md
@@ -1,59 +1,89 @@
-# Parsing geojson in geotrellis.vector.io.json
+# Parsing GeoTrellis objects from GeoJson #
 
-Say you’re looking to go from a string to a feature collection
-object. This task involves a number of steps and requires
-keeping track of a few implicits.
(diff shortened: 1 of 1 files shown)


fosskers:
Good to go?

docs/vector/geojson-support.md
```diff
@@ -1,59 +1,89 @@
...
+```scala
+val collection = fc.parseGeoJson[JsonFeatureCollectionMap]
 ```
+it will be necessary to extract the desired features from `collection`.  In order to maintain type safety, these results are pulled using accessors such as `.getAllPoints`, `.getAllMultiLineFeatures`, and so on.  Each geometry and feature type requires the use of a different method call.
 
-2. Once the string (which must be valid geojson!) is parsed, it is
-time to query your JsonFeatureCollection for its points:
+As in the case of serialization, to extract the feature data from this example string, we must create a case class with an integer member named `SomeProp` and register it using `jsonFormat1`.
+```scala
+case class SomeProp(SomeProp: Int)
```
lossyrob: It's non-idiomatic in JSON to have upper case property names. Can we change this here and in the JSON to `someProp`?

PR #1720, opened 2016-10-27
Changes to Interpolation

lossyrob:
- Fixes the Kriging interpolation implicit structure to fit the rest of the library
- Added functionality to IDW interpolation to have it match what GDAL does
- Made implicit methods for IDW interpolation
- Got rid of VectorToRaster, which was holding the older version of IDW and some point-counting thing that serves better as a documentation example then actual library functionality.


raster/src/main/scala/geotrellis/raster/interpolation/InverseDistanceWeighted.scala
```diff
@@ -18,44 +33,57 @@ object InverseDistanceWeighting {
...
       val ewr2 = equalWeightRadius * equalWeightRadius
 
-      def idw(points: Seq[PointFeature[D]], x: Double, y: Double)(filter: Double => Boolean) = {
+      def idw(points: Traversable[PointFeature[D]], x: Double, y: Double, hasRadius: Boolean) = {
         var sum = 0.0
         var count = 0
         var weightSum = 0.0
         var sampleSum = 0.0
         var sampleCount = 0
+        val halfPow = weightingPower / 2.0
```
lossyrob: Using half power because working with squared distances

vector/src/main/scala/geotrellis/vector/SpatialIndex.scala
```diff
@@ -44,7 +44,7 @@ object SpatialIndex {
     si
   }
 
-  def fromExtents[T](items: Iterable[T])(f: T => Extent): SpatialIndex[T] = {
+  def fromExtents[T](items: Traversable[T])(f: T => Extent): SpatialIndex[T] = {
```
echeipesh: It's more general but interested why upgrade to `Traversable` from `Iterable`?

PR #1733, opened 2016-10-27
Update GeoJson CRS API

echeipesh:
Fixes: https://github.com/geotrellis/geotrellis/issues/1475

I considered a solution that would get rid of `WithCrs` object but it would require a dependency on shapeless that we do not currently have an can't bring in. Overall this is cleanup and refinement PR:

- Renamed `CRS` to `JsonCRS`
- Chaned `LinkedCRS.href` to URI
- Created `JsonFormat[JsonCRS]`
- Created `JsonFormat[NamedCRS]`
- Created `JsonFormat[LinkedCRS]`
- Modified `JsonFormat[LinkedCRS]` to delegate to those above on read/write
- Added `JsonCRS.toCRS: Option[CRS]`
 - For `NamedCRS` we can expect this to work for EPSG code
 - For `LinkedCRS` we can expect this to work for URLs pointing to Proj4 strings


moradology:
This looks good - I wonder if there's not some way to carry out `toCRS` (from `NamedCRS`) during deserialization. Consider it; +1 elsewise

echeipesh:
I specifically wanted to avoid calling `toCRS` during deserialization because it can fail for any number of reasons. The user can have a choice how to deal with an object that represents the JSON structure.

PR #1737, opened 2016-10-28
Pyramid down function

pomadchin:
As a possible issue https://github.com/geotrellis/geotrellis/issues/1717  solution

Though I agree with @echeipesh ; mb we don't need this function?


lossyrob:
I think the pyramid down functionality can just be taken care of by a correct call to `tileToLayout` with the layout definition of the zoom level.

e.g. because this sets the "maxZoom", wouldn't this just tile something down to a higher zoom level? e.g. if I had tiles that fit zoom 15 but wanted zoom 18, wouldn't this just work? https://github.com/geotrellis/geotrellis/blob/master/spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala#L127-L128

pomadchin:
@lossyrob oops, double checked and it worked; the reason why I implemented this func in fact was a hardcoded maxZoom value: https://github.com/geotrellis/geotrellis-landsat-emr-demo/blob/master/ingest/src/main/scala/demo/LandsatIngest.scala#L31

PR #1743, opened 2016-10-28
Crop Iterator for GeoTiffs

jbouffard:
This PR adds a `CropIterator` abstract class and its two subclasses: `SinglebandCropIterator` and `MultibandCropIterator`. These new classes are extensions of `Iterator` that will take a GeoTiff and split it into multiple GeoTiffs based on inputted dimensions. The returned values will be these sub-tiles.


jbouffard:

lossyrob:
LGTM after brief review, merging. I'm going to work with this a bit so if there's other comments either I'll address in another PR or request via other channel

PR #1747, opened 2016-10-28
[WIP] Core Concepts - Tile Layers

fosskers:
### TODO
- [x] Diagram
- [x] More `Tile Layers` info
- [x] Typeclass info

type-composition


rossbernet:
so this is no longer WIP correct?

PR #1748, opened 2016-10-28
Additional Text Relating to Performance

jamesmcclain:
Still Needs
- [x] Discussion of Mutable Types
- [x] ~~Discussion of Primitive Types (pasted-in from Nathan's writing?)~~ Tile Heirarchy
- [x] Disucssion of Spark

### DIFF
--- a/docs/architecture/high-performance-scala.md
+++ b/docs/architecture/high-performance-scala.md
@@ -1,7 +1,17 @@
 # JVM Primitives
 ------
+
+
 # Mutable Data
 ------
+
+
+# Spark
+------
+
+
+# Macros
+-----
 
 > NOTE: Because scala macros require a separate stage of compilation, they've
 > been broken out into their own package in GeoTrellis. Otherwise, the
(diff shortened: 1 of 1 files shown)


docs/architecture/high-performance-scala.md
```diff
@@ -1,8 +1,18 @@
 # JVM Primitives
 ------
```
fosskers: The `#` and `---` combination creates some funky double-underlining when rendered by github. Either `##` or `#` alone for these should work.

fosskers:
You could also talk about the `Tile` type hierarchy that we use, to avoid boxing.

docs/architecture/high-performance-scala.md
```diff
@@ -90,9 +100,45 @@ readability. They've been introduced merely to overcome shortcomings in
...
 
+## Loops ##
 
-# Macros
------
-# Spark
-------
\ No newline at end of file
+A commonly-seen feature throughout the codebase is the use of `cfor`- or `while`-loops where it would seem that an ordinary `for`-loop would be sufficient.
+We avoid using "`for`-loops" in most cases because they are more than just loops in Scala.
```
pomadchin: Dunno how to correct to name `for` a sa `loop`; as it is a `for comprehension`, that's why it is slow; http://docs.scala-lang.org/tutorials/FAQ/yield.html

jamesmcclain: Yeah, I was trying to decide how to approach it, as well, that is why I put it in quotation marks.

jamesmcclain:
> You could also talk about the Tile type hierarchy that we use to avoid boxing.

I am hoping that @moradology 's `CellType` discussion would covered most of that (though I have not seen it).  Can anyone confirm or deny?

fosskers:
It does a bit, yeah.

fosskers:
I just confirmed; mention of the tile hierarchy and why we use it isn't mentioned in the Cell Type explanation.

jamesmcclain:
Thanks for all of the useful feedback ... incorporating.

pomadchin:
Can we add information about `TileCombiners`, and to mention that we generate some boilerplate: https://github.com/geotrellis/geotrellis/blob/master/project/Boilerplate.scala ?

jamesmcclain:
> Can we add information about TileCombiners, and to mention that we generate some boilerplate: https://github.com/geotrellis/geotrellis/blob/master/project/Boilerplate.scala ?

Certainly, thanks for the input.  I will add that when there is time.

jamesmcclain:
I believe that all prior comments have been addressed.

PR #1753, opened 2016-10-31
Reading Big GeoTiffs

jbouffard:
This PR adds the ability for GeoTrellis to read Big, GeoTiff files. It is important to note that this **will only work with GeoTiffs**. Regular BigTiff files are currently not supported even though their smaller counterparts are. This is an issue that will be addressed and fixed in the near future.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/io/geotiff/reader/TiffTagsReader.scala
+++ b/raster/src/main/scala/geotrellis/raster/io/geotiff/reader/TiffTagsReader.scala
@@ -29,17 +29,28 @@ object TiffTagsReader {
 
     // Validate GeoTiff identification number
     val geoTiffIdNumber = byteReader.getChar
-    if ( geoTiffIdNumber != 42)
-      throw new MalformedGeoTiffException(s"bad identification number (must be 42, was $geoTiffIdNumber)")
+    if ( geoTiffIdNumber != 42 || geoTiffIdNumber != 43)
+      throw new MalformedGeoTiffException(s"bad identification number (must be 42 or 43, was $geoTiffIdNumber)")
 
-    val tagsStartPosition = byteReader.getInt
+    //val tagsStartPosition =
+    if (geoTiffIdNumber == 42) {
+      val smallOffset = byteReader.getInt
+      read(byteReader, smallOffset)
+    } else {
+      byteReader.position(8)
+      val bigOffset = byteReader.getLong
+      read(byteReader, bigOffset)
+    }
 
-    read(byteReader, tagsStartPosition)
+    //read(byteReader, tagsStartPosition)
   }
 
-  def read(byteReader: ByteReader, tagsStartPosition: Int): TiffTags = {
+  def read[T](byteReader: ByteReader, tagsStartPosition: T): TiffTags = {
 
-    byteReader.position(tagsStartPosition)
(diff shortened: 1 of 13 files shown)


jbouffard:
I should also mention that I added three new files to the `geotiff-test-files.zip` [link](https://github.com/jbouffard/geotrellis/blob/4b316edd2229655d9def58c3850ee5bdabce6959/raster-test/data/geotiff-test-files.zip). Because of that, it is now *just* over the max file size of GitHub, though, it still uploaded fine. It's 50.8 MB while the max is 50 MB.

raster/src/main/scala/geotrellis/raster/io/geotiff/tags/codes/TiffFieldType.scala
```diff
@@ -14,5 +14,7 @@ object TiffFieldType {
   val SignedFractionalsFieldType = 10
   val FloatsFieldType = 11
   val DoublesFieldType = 12
-
+  val LongsFieldType = 16
```
jbouffard: This is 64 bits

jamesmcclain: I would suggest something like

```scala
val LongsFieldType = 16 // This is 64-bits
```

Just to make sure it is super-obvious that this field is not present in normal TIFFs (that is, "long" means 64-bits in the Java world, whereas others might expect "long long").

jbouffard:
@jamesmcclain Unfortuantely, even when I created a correct version of `all-ones.tif` and then made a BigTiff version of that with `gdal_translate` it still failed.

jamesmcclain:
> @jamesmcclain Unfortuantely, even when I created a correct version of all-ones.tif and then made a BigTiff version of that with gdal_translate it still failed.

Ah, okay.  If we can just confirm that the actual bytes in the offset field of entry 34737 match those of the too-large value, that would be good enough for me.

jbouffard:
@jamesmcclain I'm going to take this PR out of WIP, since it seems to work except for this one file; but I'll compare the bytes of the two GeoTiffs in a bit.

jbouffard:

jamesmcclain:
> @jamesmcclain I'm going to take this PR out of WIP, since it seems to work except for this one file; but I'll compare the bytes of the two GeoTiffs in a bit.

Okay, sounds like a winner.

raster/src/main/scala/geotrellis/raster/io/geotiff/reader/TiffTagsReader.scala
```diff
@@ -29,19 +29,31 @@ object TiffTagsReader {
 
     // Validate GeoTiff identification number
     val geoTiffIdNumber = byteReader.getChar
-    if ( geoTiffIdNumber != 42)
-      throw new MalformedGeoTiffException(s"bad identification number (must be 42, was $geoTiffIdNumber)")
-
-    val tagsStartPosition = byteReader.getInt
-
-    read(byteReader, tagsStartPosition)
+    if ( geoTiffIdNumber != 42 && geoTiffIdNumber != 43)
```
lossyrob: formatting: space after `if (`

jbouffard:
@lossyrob Format change has been made.

PR #1758, opened 2016-10-31
[WIP] Added geotiff window reading for hadoop source RDDs

lossyrob:


pomadchin:
Don't forget to add `maxTileSize` in a `json` validation schema: https://github.com/geotrellis/geotrellis/blob/master/spark-etl/src/main/resources/input-schema.json 

Otherwise it would not be possible to use it in etl, due to https://github.com/geotrellis/geotrellis/blob/master/spark-etl/src/main/resources/input-schema.json#L64

lossyrob:
Superseded by #1763

Issue #1509, opened 2016-06-03
Alternative to .color(ColorMap) for transforming values


RickMohr:
It's useful to be able to transform raster values using a Scala `Map` which transforms input values to output values -- different from `.color(ColorMap)` in (at least) these ways:
- Control the output cell type (e.g. you could map doubles to doubles or doubles to ints)
- NODATA (as defined by the input cell type) maps to NODATA (as defined by the output cell type)

Ideally it would be available for both tiles and RDD's, and benefit from any optimizations implemented for #1489.

fosskers:
Leaving a note here that #1489 was optimized by #1648 .

fosskers:
@lossyrob there was talk in the standup this morning about tabling this issue until we've had a chance to really investigate `spire`. This would be fairly trivially solved with a polymorphic `Tile[T]`. If that doesn't end up working, I'll go for the more manual route as originally planned.

fosskers:
A discussion with @echeipesh revealed the following:

- My PR doesn't need to extend `Tile`'s API; conversion from `Int => Double` and `Double => Int` is already possible with proper use of `map` on a Floating tile, or `mapDouble` on an Integer tile.
- [Eugene's PR](https://github.com/geotrellis/geotrellis/pull/1766)  is actually what adresses @RickMohr 's intial desire to control the output cell type
- The Functional workflow one expects to be able to do is:

```scala
val t: Tile = ...
val f: Int => Int = ...
val newCellType: CellType = ...

t.map(f).convert(newCellType)
```
and indeed, this is already possible but not performant. If in the near future we implemented lazy Tile operation semantics, then this issue is solved for free, and no additional `map` overloads are necessary.

fosskers:
#1760 has been merged, which addresses the general ability to efficiently map over a tile via a `Map` with arbitrary Int/Double targets and sources. 

Fine-grained control over `CellType` is forth-coming.

PR #1760, opened 2016-10-31
Break Map

fosskers:
### TODO
- [x] `@specialized` BreakMap class
- [x] Decide whether or not to use subclasses
- [x] Reconcile `Tile.mapWith` API
- [x] (optional) Make `ColorMap` use `BreakMap`
- [x] Benchmarks (ongoing)
- [x] Docstrings for `BreakMap` usage with `map`
- [x] Revert changes to the `Tile` API 
- [x] Fix tests

### Motivation
Addresses (but does not close) #1509 .

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/render/BreakMap.scala
+++ b/raster/src/main/scala/geotrellis/raster/render/BreakMap.scala
@@ -42,11 +42,11 @@ object MapStrategy {
   * with breaks and a break strategy.
   *
   * {{{
-  * val bm: BreakMap = ...
+  * val bm: BreakMap[Int, Int] = ...
   * val t: Tile = ...
   *
   * // Map all the cells of `t` to a target bin value in O(klogn).
-  * val newT: Tile = t.mapWith(vm)
+  * val newT: Tile = t.map(vm)
   * }}}
   *
   * '''Note:''' `A` and `B` are specialized on `Int` and `Double`.
@@ -119,15 +119,15 @@ class BreakMap[
 
 /** Helper methods for constructing BreakMaps. */
 object BreakMap {
-  def i2i(m: Map[Int, Int]): BreakMap[Int, Int] =
-    new BreakMap(m, MapStrategy.int, { i => isNoData(i) })
+  def i2i(m: Map[Int, Int], s: MapStrategy[Int] = MapStrategy.int): BreakMap[Int, Int] =
+    new BreakMap(m, s, { i => isNoData(i) })
 
-  def i2d(m: Map[Int, Double]): BreakMap[Int, Double] =
-    new BreakMap(m, MapStrategy.double, { i => isNoData(i) })
+  def i2d(m: Map[Int, Double], s: MapStrategy[Double] = MapStrategy.double): BreakMap[Int, Double] =
+    new BreakMap(m, s, { i => isNoData(i) })
 
-  def d2d(m: Map[Double, Double]): BreakMap[Double, Double] =
-    new BreakMap(m, MapStrategy.double, { d => isNoData(d) })
+  def d2d(m: Map[Double, Double], s: MapStrategy[Double]=MapStrategy.double): BreakMap[Double, Double] =
+    new BreakMap(m, s, { d => isNoData(d) })
 
-  def d2i(m: Map[Double, Int]): BreakMap[Double, Int] =
-    new BreakMap(m, MapStrategy.int, { d => isNoData(d) })
+  def d2i(m: Map[Double, Int], s: MapStrategy[Int] = MapStrategy.int): BreakMap[Double, Int] =
+    new BreakMap(m, s, { d => isNoData(d) })
 }
--- a/raster/src/main/scala/geotrellis/raster/render/BreakMap.scala
+++ b/raster/src/main/scala/geotrellis/raster/render/BreakMap.scala
@@ -24,31 +24,18 @@ case object Exact extends ClassBoundaryType
   *
   * '''Note:''' Specialized for `Int` and `Double`.
   */
-class MapStrategy[@specialized(Int, Double) A, @specialized(Int, Double) B](
+class MapStrategy[@specialized(Int, Double) A](
   val boundary: ClassBoundaryType,
-  val noDataValue: B,
-  val fallbackValue: B,
-  val strict: Boolean,
-  val noDataCheck: A => Boolean
+  val noDataValue: A,
+  val fallbackValue: A,
+  val strict: Boolean
 )
 
 /** Helper methods for constructing a [[MapStrategy]]. */
 object MapStrategy {
-  def i2i: MapStrategy[Int, Int] = new MapStrategy(
-    LessThanOrEqualTo, 0x00000000, 0x00000000, false, { i => isNoData(i) }
-  )
+  def int: MapStrategy[Int] = new MapStrategy(LessThanOrEqualTo, 0x00000000, 0x00000000, false)
 
-  def i2d: MapStrategy[Int, Double] = new MapStrategy(
-    LessThanOrEqualTo, Double.NaN, Double.NaN, false, { i => isNoData(i) }
-  )
-
-  def d2d: MapStrategy[Double, Double] = new MapStrategy(
-    LessThanOrEqualTo, Double.NaN, Double.NaN, false, { d => isNoData(d) }
-  )
-
-  def d2i: MapStrategy[Double, Int] = new MapStrategy(
-    LessThanOrEqualTo, 0x00000000, 0x00000000, false, { d => isNoData(d) }
-  )
+  def double: MapStrategy[Double] = new MapStrategy(LessThanOrEqualTo, Double.NaN, Double.NaN, false)
 }
 
 /** A `Map` which provides specific Binary Search-based ''map'' behaviour
@@ -67,7 +54,7 @@ object MapStrategy {
 class BreakMap[
   @specialized(Int, Double) A: Order,
   @specialized(Int, Double) B: Order
-](breakMap: Map[A, B], strategy: MapStrategy[A, B]) {
+](breakMap: Map[A, B], strategy: MapStrategy[B], noDataCheck: A => Boolean) {
 
   /* A Binary Tree of the mappable values */
   private lazy val vmTree: BTree[(A, B)] = {
@@ -114,7 +101,7 @@ class BreakMap[
   }
 
   def map(z: A): B = {
-    if (strategy.noDataCheck(z)) {
+    if (noDataCheck(z)) {
       strategy.noDataValue
     } else {
       vmTree.searchWith(z, branchPred) match {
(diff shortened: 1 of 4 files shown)


fosskers:
So far, we are able to keep a `@specialized` BreakMap implementation/usage and only pay ~3% in speed. Recall the ~50% increase gained by `BTree` in the first place.

I only *barely* dipped into `spire` for this to work: Spire's `Order` typeclass is necessary to write the Binary Search branch predicate generically. `Order` provides `<` etc for numeric types, which are built-in and not typeclass-bound in vanilla Scala.

Otherwise, the rest of the generic coding is vanilla `@specialized` use.

fosskers:
Current thoughts: instead of trying to force some sort of inheritance structure between the monomorphic `ColorMap` and polymorphic `BreakMap`, `ColorMap`s should just *use* a `BreakMap` internally to reduce the duplicate code currently sitting in `IntColorMap` and `DoubleColorMap`.

fosskers:
Naming discussion:

The point of this data structure is to provide a mechanism for mapping over a `Tile`, where instead of passing a function, we pass what amounts to a `Map[A, B]` specialized for number types. Using class-break boundary conditions, we leverage an internal `BTree` to quickly find what bin the source value should be mapped to. This generalizes the current behaviour of `ColorMap` and `Tile.color`.

**Question:** What should this data structure be called?

Initially we had `ValueMap`, but thinking that didn't have enough semantic clarity I switched it to `BreakMap`.

echeipesh:
I think the GIS term for this is reclassify, which I really like. How about `ValueReclassify` or just `Reclassify` ?

You'd get signatures like `Reclassify[Int, Int]`.

fosskers:
The use-case is:

```scala
val t: Tile = ...
val m: Reclassify = ...

Tile.mapWith(m)
```
Does the signature `Tile.mapWith: Reclassify => Tile` feel good semantically?

lossyrob:
I think `Reclassify` is a general term for what is going on, but also that could be said for our `map`.

+1 `BreakMap`

I'm into the method being named as an overload of `map`:

```scala
val t: Tile = ???
val m: BreakMap = ???

t.map(m)
```

echeipesh:
Yes, `Tile.mapWith: Reclassify => Tile` does not jive, +1 `BreakMap` based on that.

fosskers:
> overload of `map`

I'm not the biggest fan of overloading, I think it increases complexity / makes it harder to grock APIs. An example from my personal past being Matlab work, where every function is so heavily overloaded that typos / oversights often still run and produce debugging misery.

lossyrob:
My thought on overloadin `map` is
- we already are doing it for `map(f: Int => Int)` and  `map(f: (Int, Int, Int) => Int)`
- it's pretty clear what the context of the map is based on the object you're passing in
- I don't see a method name that would better explain what we're doing (mapWith doesn't help me understand what is going on, just seems like differentiating for the sake of differentiating)

What do other people think?

fosskers:
Updated benchmark results with `ColorMap` backed by `BreakMap`:

```
 1280x1280 pixel Tile, 50 Breaks, 5 trials each, same JVM

 `Tile.color` ORIGINAL LINEAR
 ----------------------------
 Rendering => 176.53 ms
 Rendering => 180.56 ms
 Rendering => 170.98 ms
 Rendering => 170.94 ms
 Rendering => 170.47 ms

 Mean: 173.89 ms

 `Tile.color` WITH BTREE
 -----------------------
 Rendering => 120.25 ms
 Rendering => 117.32 ms
 Rendering => 121.19 ms
 Rendering => 117.35 ms
 Rendering => 116.86 ms

 Mean: 118.6 ms

 Speedup: ~47%

 (The above two are old results. The `WITH BTREE` test clocked a few ms slower today.)

 `Tile.color` BACKED BY `BREAKMAP`
 ---------------------------------
 Rendering => 133.06 ms
 Rendering => 123.85 ms
 Rendering => 120.52 ms
 Rendering => 122.59 ms
 Rendering => 119.76 ms

 Mean: 123.96 ms

 Slowdown: ~3.5%

 `Tile.mapWith` (uses `Tile.map`)
 --------------------------------
 Rendering => 128.96 ms
 Rendering => 126.02 ms
 Rendering => 125.55 ms
 Rendering => 123.38 ms
 Rendering => 121.14 ms

 Mean: 125.01 ms
```

echeipesh:
Re: `tile.mapWith`

Conceptually `BreakMap` is a function, in fact if `BreakMap.map` was renamed to `BreakMap.apply` it should be able to extend `Function2` without losing specialization. At that point using `tile.map` seems pretty natural.

fosskers:
Would target cell types be set properly?

fosskers:
Clarification @echeipesh : do you mean if `BreakMap` extended `Function`, we'd be able to use the preexisting `Tile.map`?

lossyrob:
We wouldn't be able to, due to `[A, B]` stuff in the map...

fosskers:
Why not `class BreakMap[... A, ... B] extends Function1[A, B]`?

fosskers:
Right, since `Tile.map: Int => Int`. Hmm...

fosskers:
Our hands might be forced by the monomorphic `Tile`:
```scala
def map(bm: BreakMap[Int, Int]): Tile

def mapDouble(bm: BreakMap[Double, Double]): Tile

def mapd2i(bm: BreakMap[Double, Int]): Tile

def mapi2d(bm: BreakMap[Int, Double]): Tile
```

lossyrob:
We don't need the separate names, since the behavior of the map function depends on the values of the type params.

If there's "same type after erasue" issues, just use dummy implicits:

```scala
def map(m: BreakMap[Int, Int], targetCellType: CellType): Tile = ???
def map(m: BreakMap[Int, Double], targetCellType: CellType)(implicit d: DummyImplicit): Tile = ???
def map(m: BreakMap[Double, Int], targetCellType: CellType)(implicit d: DummyImplicit, d2: DummyImplicit): Tile = ???
def map(m: BreakMap[Double, Double], targetCellType: CellType)(implicit d: DummyImplicit, d2: DummyImplicit, d3: DummyImplicit): Tile = ???
```

fosskers:
Exposed API:

```scala
trait Tile extends ... {
  def map(f: Int => Int): Tile  /* unchanged */

  def map(f: Int => Double, tct: CellType): Tile /* new */

  def map(f: Double => Int, tct: CellType): Tile /* new */

  def mapDouble(f: Double => Double): Tile /* unchanged */
}
```

fosskers:
Needs docstrings to explain the speedup you can get via `BreakMap`.

lossyrob:
Overloading `map` like that will confuse the compiler when you try to pass things in. My preference is to keep map and mapDouble the same (and avoid conflict with the other PR) and just create `map`s that specifically take BreakMap

raster/src/main/scala/geotrellis/raster/render/BreakMap.scala
```diff
@@ -0,0 +1,133 @@
...
+/** A strategy for mapping values via a [[BreakMap]].
+  *
+  * '''Note:''' Specialized for `Int` and `Double`.
+  */
+class MapStrategy[@specialized(Int, Double) A](
+  val boundary: ClassBoundaryType,
+  val noDataValue: A,
+  val fallbackValue: A,
+  val strict: Boolean
+)
```
fosskers: `MapStrategy` might deserve its own file, although it's strictly subordinate to `BreakMap`.

raster/src/main/scala/geotrellis/raster/render/BreakMap.scala
```diff
@@ -0,0 +1,133 @@
...
+
+/** Helper methods for constructing a [[MapStrategy]]. */
+object MapStrategy {
+  def int: MapStrategy[Int] = new MapStrategy(LessThanOrEqualTo, 0x00000000, 0x00000000, false)
```
fosskers: Is zero correct, or `Int.MinValue`? Zero is the old default from `ColorMap`.

raster/src/main/scala/geotrellis/raster/render/BreakMap.scala
```diff
@@ -0,0 +1,133 @@
...
+  noDataCheck: A => Boolean
+) extends (A => B) {
+
+  /* A Binary Tree of the mappable values */
+  private lazy val vmTree: BTree[(A, B)] = {
+    val a: Array[(A, B)] = breakMap.toArray
+
+    Sorting.quickSort(a)
+
+    BTree.fromSortedSeq(a.toIndexedSeq).get
```
fosskers: `quickSort` needs an `Array`, but we could probably go faster if we avoided `toIndexedSeq` (makes it a `Vector`) and had another `BTree` constructor that took `Array`.

Within `fromSortedSeq`, we need the ability to `slice` and have O(1) access to elements.

raster/src/main/scala/geotrellis/raster/render/ColorMap.scala
```diff
@@ -204,85 +206,60 @@ class IntColorMap(breaksToColors: Map[Int, Int], val options: Options = Options.
...
+    new IntColorMap(
+      breaksToColors,
+      new Options(new MapStrategy(
+        options.strategy.boundary,
+        color,
+        options.strategy.fallbackValue,
+        options.strategy.strict
+      ))
+    )
+  }
```
fosskers: These are ugly. It would be much cleaner with `MapStrategy.copy`, except that would require it to be a `case class`. Case classes generate more bytecode, but if their instantiation cost isn't terrible, it might be worth it to reduce code smell here.

fosskers:
I'm also feeling like we should rip out `Options` in favour of using straight `MapStrategy`.

raster/src/main/scala/geotrellis/raster/render/BreakMap.scala
fosskers: On the other hand, this might be negligible since `ColorMap` / `BreakMap` instantiation is usually a one-time cost.

raster/src/main/scala/geotrellis/raster/render/BreakMap.scala
```diff
@@ -0,0 +1,133 @@
...
+        case None if strategy.strict => sys.error(s"Value ${z} did not have an associated value.")
+        case _ => strategy.fallbackValue
+      }
+    }
+  }
+}
+
+/** Helper methods for constructing BreakMaps. */
+object BreakMap {
+  def i2i(m: Map[Int, Int], s: MapStrategy[Int] = MapStrategy.int): BreakMap[Int, Int] =
```
fosskers: The purpose of these functions is pretty clear from their type signatures, and I argue that having overloaded `apply`s with `DummyImplicit` is actually harder to grok. Does the average Scala dev know about the type erasure problems, and that this is a hack to get around them?

pomadchin: Basically I don't like c-styled function names (i2i, i2d, etc), but I remember that somewhere in our codebase we have smth similar (i believe it was related to function call performance); 
So +1 for overloads using `DummyImplicit`s, it's a usual Scala way to resolve such overloads. ¯\_(ツ)_/¯ 

Think we need one more opinion at least.

echeipesh: +1 to `DummyImplicits`, that gives the least surprising application.

fosskers: Although I'm opposed to overloads in general, after thinking about it, overloaded `apply` isn't that terrible so long as the return type matches the parent class, which it would here.

I'll switch to `apply`.

PR #1763, opened 2016-11-01
Windowed Reading GeoTiffs from S3 and Hdfs

jbouffard:
This PR was based on another, preexisting PR: #1758

Two new features are added to Geotrellis with this PR: Reading windowed GeoTiffs from S3 and Hdfs. These features will allow for the breaking up of large GeoTiffs into smaller tiles over a distributed system. By using this method, the computation time for each GeoTiff that needs to analyzed can be shortened considerably.

To do:
- [x] Unit Testing for Windowed Reading for Hdfs
- [x] Implement Windowed Reading for S3
  - [x] `SpatialGeoTiff`s
  - [x] `TemporalGeoTiff`s
- [x] Unit Testing for Windowed Reading for S3


lossyrob:
@jbouffard can you make sure to include the changes @pomadchin mentions here: https://github.com/geotrellis/geotrellis/pull/1758#issuecomment-257545148

jbouffard:
@lossyrob Yeah, I was just about to do that.

raster/src/main/scala/geotrellis/raster/io/geotiff/ArraySegmentBytes.scala
```diff
@@ -51,12 +51,9 @@ object ArraySegmentBytes {
             val result = Array.ofDim[Array[Byte]](offsets.size)
 
             cfor(0)(_ < offsets.size, _ + 1) { i =>
-              byteReader.position(offsets(i).toInt)
-              result(i) = byteReader.getSignedByteArray(byteCounts(i).toInt)
+              result(i) = byteReader.getSignedByteArray(byteCounts(i), offsets(i))
```
echeipesh: Stale doc string, `byteBuffer` no longer a prameter

echeipesh: `oldOffset` is unused. Looks like the previous version of this file used it to return position, I would guess thats not a good idea and this variable needs to be removed now.

echeipesh: This maybe a conversation that already happened but why are there `ByteReaderExtensions` when we control the trait we're extending? Why shouldn't `getSignedByteArray` and others be methods on `ByteReader` trait directly?

raster/src/main/scala/geotrellis/raster/io/geotiff/LazySegmentBytes.scala
```diff
@@ -17,9 +17,9 @@ import spire.syntax.cfor._
  * @param byteReader: A ByteReader that contains bytes of the GeoTiff
  * @param storageMethod: The [[StorageMethod]] of the GeoTiff
  * @param tifftags: The [[TiffTags]] of the GeoTiff
- * @return A new instance of BufferSegmentBytes
+ * @return A new instance of LazySegmentBytes
```
echeipesh: `@param` section is out of sync with code

raster/src/main/scala/geotrellis/raster/io/geotiff/util/ByteReaderExtensions.scala
```diff
@@ -173,33 +173,22 @@ trait ByteReaderExtensions {
     }
 
     final def getSignedByteArray(length: Long, valueOffset: Long): Array[Byte] = {
-      val arr = Array.ofDim[Byte](length.toInt)
-
+      // NOTE: We don't support lengths greater than Int.MaxValue yet (or ever).
```
echeipesh: This comment should be in a doc string

echeipesh: Also if so, why is the `length` parameter `Int`?

raster/src/main/scala/geotrellis/raster/io/geotiff/ArraySegmentBytes.scala
jbouffard: `ByteReader` is meant to be a replacement for `ByteBuffer`, so I wanted the two to have the same methods (or at least have them be as close as possible). Since none of the methods in `ByteReaderExtensions` are in `ByteBuffer`, they were made into their own separate thing.

raster/src/main/scala/geotrellis/raster/io/geotiff/LazySegmentBytes.scala
jbouffard: Fixed

raster/src/main/scala/geotrellis/raster/io/geotiff/util/ByteReaderExtensions.scala
jbouffard: Fixed it. Why it was converted to `Int`? That's because you can't make an `Array.ofDim` using a `Long`.

PR #1764, opened 2016-11-01
Added save action to ETL

lossyrob:
This feature allows us to include client code in the save process for ETL. Lets us do things like:

```scala
    EtlConf(args) foreach { conf =>
      val etl = Etl(conf)

      val sourceTiles = etl.load[ProjectedExtent, Tile]
      val (zoom, tiled) = etl.tile[ProjectedExtent, Tile, SpatialKey](sourceTiles)

      etl.save[SpatialKey, Tile](LayerId(etl.input.name, zoom), tiled, { (attributeStore, layerWriter, layerId, rdd) =>
        // Save off histogram computed from zoom 9, store in zoom 0's attributes.
        if(layerId.zoom == 9) {
          val histogram = rdd.histogram(512)
          attributeStore.write(
            layerId.copy(zoom = 0),
            "histogram",
            histogram: Histogram[Double])
        }
      })
    }
```

which is otherwise very difficult to do in the ETL context.

__Also in this PR__

Changing the layerExists functionality on AttributeStore to check specifically for the metadata attribute, so that you can store attributes in layer IDs without that layer being considered as existing.

### DIFF
--- a/spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala
+++ b/spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala
@@ -26,11 +26,12 @@ import scala.reflect.runtime.universe._
 object Etl {
   val defaultModules = Array(s3.S3Module, hadoop.HadoopModule, file.FileModule, accumulo.AccumuloModule, cassandra.CassandraModule, hbase.HBaseModule)
 
-  type PostSaveHook[K, V, M] = (AttributeStore, Writer[LayerId, RDD[(K, V)] with Metadata[M]], LayerId, RDD[(K, V)] with Metadata[M]) => Unit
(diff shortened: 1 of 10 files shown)


spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala
```diff
@@ -156,15 +168,21 @@ case class Etl(conf: EtlConf, @transient modules: Seq[TypedModule] = Etl.default
     * Saves provided RDD to an output module specified by the ETL arguments.
     * This step may perform two to one pyramiding until zoom level 1 is reached.
     *
-    * @param id     Layout ID to b
-    * @param rdd Tiled raster RDD with TileLayerMetadata
+    * @param id        Layout ID to b
+    * @param rdd       Tiled raster RDD with TileLayerMetadata
+    * @param postSave  Function to allow saving additional attributes or layers per layer saved.
```
echeipesh: wrong comment name

lossyrob:
@echeipesh your comment was addressed. GTG?

PR #1765, opened 2016-11-02
Add a reprojection option to specify target rasterextent

pomadchin:
Fixes: https://github.com/geotrellis/geotrellis/issues/1670


raster/src/main/scala/geotrellis/raster/reproject/ReprojectRasterExtent.scala
```diff
@@ -32,7 +32,7 @@ object ReprojectRasterExtent {
     val threshold = math.min(xThreshold, yThreshold)
 
     // Densify the extent to get a more accurate reprojection
-    val denseGeom = Polygon(Densifier.densify(extent.toPolygon.jtsGeom, threshold).asInstanceOf[com.vividsolutions.jts.geom.Polygon])
+    val denseGeom = Polygon(Densifier.densify(extent.toPolygon().jtsGeom, threshold).asInstanceOf[com.vividsolutions.jts.geom.Polygon])
```
lossyrob: why `()`?

pomadchin: @lossyrob as it is (ofc it is more a convention question: http://docs.scala-lang.org/style/naming-conventions.html#parentheses): 

```scala
/** Return this extent as a polygon */
  def toPolygon(): Polygon =
    Polygon( Line((xmin, ymin), (xmin, ymax), (xmax, ymax), (xmax, ymin), (xmin, ymin)) )
```

and not 

```scala
def toPolygon = { ... }
```

I can roll it back btw :)

pomadchin:
@lossyrob want to hear your opinion on functions behaviour, and especially windowed case

PR #1766, opened 2016-11-02
Add target CellType overload to Tile

echeipesh:
Fixes: https://github.com/geotrellis/geotrellis/issues/1410

An acceptable solution is to provide a way to request a target `CellType` in general transformations like `map` and `combine` such that they can be used directly with operation functions. Propagating this extra parameter to the extension methods is too large an impact for an advanced feature.

```scala
trait UseCaseSomething {   
   def tile1: Tile 
   def tile2: Tile  
   tile1.map { x => x + 1 }
    tile1.map(IntCellType) { x => x + 1 }
   tile1.map(IntCellType) (BreakMap.forIntegers(m))
   tile1.map(FloatCellType) (BreakMap.forDouble(m))
     tile1.combine(tile2) { (v1, v2) => v1 + v2 }
    tile1.combine(tile2, IntCellType) { (v1, v2) => v1 + v2 } } 
    tile1.combine(tile2, IntCellType)(Add)
}
```

TODO:
 - [ ] Update tests for `Tile` / `GeoTiffTile` / `ConstantTile` / `CroppedTile` / `CompositeTile`
 - [ ] Fix `ConstantTile` map/combine needs to use `cellType`
 - [ ] `GeoTiff` tile needs to have `map(f, ct)` on `GeoTiffSegments`
 - [ ] Add `CellType` overloads to `MultibandTile`
 - [ ] Add tests to related tiles
 - [ ] Check impacted doc-strings for consistency with arg-list

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/ArrayTile.scala
+++ b/raster/src/main/scala/geotrellis/raster/ArrayTile.scala
@@ -127,8 +127,8 @@ trait ArrayTile extends Tile with Serializable {
     * @param   f  A function from Int to Int, executed at each point of the tile
     * @return     The result, a [[Tile]]
     */
-  def map(f: Int=>Int): Tile = {
-    val output = ArrayTile.alloc(cellType, cols, rows)
+  def map(ct: CellType)(f: Int=>Int): Tile = {
+    val output = ArrayTile.alloc(ct, cols, rows)
     var i = 0
     val len = size
     while (i < len) {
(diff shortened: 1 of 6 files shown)


fosskers:
We need to collaborate on this, since [my PR](https://github.com/geotrellis/geotrellis/pull/1760#issuecomment-257721026) is exposing a `map`-related API as well.

raster/src/main/scala/geotrellis/raster/ArrayTile.scala
```diff
@@ -127,8 +127,8 @@ trait ArrayTile extends Tile with Serializable {
     * @param   f  A function from Int to Int, executed at each point of the tile
     * @return     The result, a [[Tile]]
     */
-  def map(f: Int=>Int): Tile = {
-    val output = ArrayTile.alloc(cellType, cols, rows)
+  def map(ct: CellType)(f: Int=>Int): Tile = {
+    val output = ArrayTile.alloc(ct, cols, rows)
```
lossyrob: We could do things like:

```scala
scala> case class Foo(x: Int) {
     |   def map(f: Int => Int, y: Int = 0) = f(x + y)
     | }
defined class Foo

scala> val x = Foo(2)
x: Foo = Foo(2)

scala> x.map { z => z + 1 }
res0: Int = 3

scala> x.map({ z => z + 1 }, 4)
res1: Int = 7

```

echeipesh:
This does not appear to be a way to go about solving the issue. It breaks the `Tile` API considerably. Tt requires deep changes in GeoTiff tiles and it's not entirely clear how it should work with macro methods. Also the API is just not that great. Future efforts to solve this will probably be around bringing `LazyTile` back in some form.

PR #1774, opened 2016-11-03
Add convenience overloads to GeoTiff companion object

### LINKED ISSUE #1757: GeoTiff Companion Object Methods
A common pattern in scala to use companion object to provide constructor overloads for the base class. This is also convenient place to provide read constructors and in general decoders. An example can be seen here: https://github.com/geotrellis/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/io/geotiff/SinglebandGeoTiff.scala#L45

This API design/improvement issue.

In this particular case `SinglebandGeoTiff` is part of a type hierarchy with `MultibandGeoTiff` where they both extend `GeoTiff` interface. Because this distinction is forced by type structure of GeoTrellis and not really a distinction that users have in mind when working with GeoTiff files we would like to extend the companion overload pattern to the `GeoTiff` super-type. That is:

When working with GeoTiff (either singleband or multiband) there should be an apply on `GeoTiff` companion object that converts the arguments to either `SinglebandGeoTiff` or `MultibandGeoTiff`. When types are not enough to specify what is intended a method will need to have a unique name, such as for reading.

For instance a String can be converted to a GeoTiff by reading the file:

`GeoTiff.readSingleband(path: String): SinglebandGeoTiff` has the same parameter signature as `GeoTiff.readMultiband(path: String): MultibandGeoTiff`, so therefore needs the name.

Also [`ProjectedRaster[T]`](https://github.com/geotrellis/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/ProjectedRaster.scala) has enough information to be converted to a either a `SinglebandGeoTiff` or `MultibandGeotiff`. But in this case can be handled with an apply overload.

miafg:
Add a method to GeoTiff.scala so that users can create GeoTiff objects without knowing whether they want a multiband or singleband geotiff. Also added functions to allow users to read multiband or singleband files without having a Multiband or Singleband Geotiff object.

Fixes #1757


raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiff.scala
```diff
@@ -52,6 +52,19 @@ trait GeoTiff[T <: CellGrid] extends GeoTiffData {
...
+    MultibandGeoTiff(path)
+
+  def readSingleBand(path: String): SinglebandGeoTiff =
+    SinglebandGeoTiff(path)
+
+  def apply(path: String): Either[SinglebandGeoTiff, MultibandGeoTiff] =
+    if (MultibandGeoTiff(path).tile.bandCount == 1) {
+      Left(SinglebandGeoTiff(path))
+    } else {
+      Right(MultibandGeoTiff(path))
```
echeipesh: Here you're potentially reading the file twice

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiff.scala
```diff
@@ -52,6 +52,19 @@ trait GeoTiff[T <: CellGrid] extends GeoTiffData {
...
+    }
+
```
echeipesh: We also should have overloads to build a GeoTiff from [`ProjectedTile`](https://github.com/geotrellis/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/ProjectedRaster.scala)

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiff.scala
```diff
@@ -52,6 +52,19 @@ trait GeoTiff[T <: CellGrid] extends GeoTiffData {
  * Companion object to GeoTiff
  */
 object GeoTiff {
+  def readMultiBand(path: String): MultibandGeoTiff =
+    MultibandGeoTiff(path)
+
+  def readSingleBand(path: String): SinglebandGeoTiff =
```
echeipesh: should be `readSingleband` (lower case b) to be consistent with methods on the readers

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiff.scala
```diff
@@ -52,6 +52,21 @@ trait GeoTiff[T <: CellGrid] extends GeoTiffData {
...
+  def readMultiband(path: String): MultibandGeoTiff =
+    MultibandGeoTiff(path)
+
+  def readSingleband(path: String): SinglebandGeoTiff =
+    SinglebandGeoTiff(path)
+
+  def apply(path: String): Either[SinglebandGeoTiff, MultibandGeoTiff] = {
+    val multiband = MultibandGeoTiff(path)
+    if (multiband.tile.bandCount == 1) {
+      Left(SinglebandGeoTiff(path))
```
pomadchin: You don't actually need to read tiff from the source path again as you already read geotiff into memory using `MultibanfGeotiff(path)`:

```scala
new SinglebandGeoTiff(tile = multiband.tile.band(0), multiband.extent, multiband.crs, multiband.tags, multiband.options)
```

PR #1784, opened 2016-11-08
Multiband histogram methods

echeipesh:
Adds method extensions for `histogram` and related stats functions to `MultibandTile` and `RDD[(K, MultibandTile)]`.

Because `MultibandTile` is essentially a `Vector[Tile]` the extension methods have little choice than return an `Array` of their `Tile` counterpart.

Fixes: https://github.com/geotrellis/geotrellis/issues/1773

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/summary/MultibandTileSummaryMethods.scala
+++ b/raster/src/main/scala/geotrellis/raster/summary/MultibandTileSummaryMethods.scala
@@ -1,0 +1,73 @@
+package geotrellis.raster.summary
+
+import geotrellis.raster._
+import geotrellis.raster.histogram.{Histogram, StreamingHistogram}
+import geotrellis.util.MethodExtensions
+
+
+/**
+  * Trait containing [[MultibandTile]] extension methods for summaries.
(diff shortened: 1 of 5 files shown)


raster/src/main/scala/geotrellis/raster/summary/MultibandTileSummaryMethods.scala
```diff
@@ -0,0 +1,73 @@
...
+    *           rounded to integers when making the Histogram.
+    */
+  def histogram: Array[Histogram[Int]] =
+    self.bands.map(_.histogram).toArray
+
+  /**
+    * Create a histogram from double values in a raster.
+    */
+  def histogramDouble(): Array[Histogram[Double]] =
+    self.bands.map(_.histogramDouble(StreamingHistogram.DEFAULT_NUM_BUCKETS)).toArray
```
lossyrob: This should just pass to `histogramDouble(StreamingHistogram.DEFAULT_NUM_BUCKETS)` - it's a good practice to only let the last overload do any actual work, so that code changes will only have to ever happen there.

PR #1788, opened 2016-11-10
SBT Upgrade

pomadchin:
May improve compilation speed ([source](https://twitter.com/retronym/status/795838877118279681))


pomadchin:
@lossyrob Any reasons why we can't move on .13 in terms on 1.0? (just a question) :D

lossyrob:
@pomadchin to answer your question, because we publish via SBT, I don't want any suprises publish time, and we might as well wait

PR #1793, opened 2016-11-11
Fixed bug in merge logic.

lossyrob:
There was a bug which took effect when we used the extent-based merge methods on cell types that had no NoData handling. This is caused by the interpolation returning NoData values when we interpolate outside the bounds of the original tile, so was happening when we were interpolating values of a larger raster with the values of a smaller one (e.g. during tileToLayout and Pyramiding)


jamesmcclain:
It looks as though the intent is to treat the value zero as neutral for purposes of merging using raw cell types?  I can detect no problems: all raw types are accounted for on lines 103 and 118 of `SinglebandTileMergeMethods.scala` and there are no non-raw types there, the logic itself appears to be correct.

lossyrob:
@jamesmcclain yes. I'll take that as approval? You can go to the "Files changed" tab, "Review Changes" and then "Approve" if so. Thankya

jamesmcclain:
I approve!

Issue #1795, opened 2016-11-11
Fix VectorJoin


rossbernet:
The current implementation of the VectorJoin forces a broadcast of the `right` partition after pulling it to the driver. This is not what we expect and needs to be changed to avoid using the driver memory through cartisian join on rtrees.

Also there exists another version of a VectorJoin on `RDD[(K, V)]` that builds the rtree on the `K`, which must be mappable to `Envelope` and then uses the index to produce a "smart" cartisian product through an RDD.flatMap or something.

jamesmcclain:
A few additional notes:

Using the [PointsToDem code](https://github.com/jamesmcclain/PointsToDem/tree/feature/vector-join-experiments) as a proxy for the typical vector join workload (using this command: `$SPARK_HOME/bin/spark-submit --master 'local[*]' --class geotrellis.dem.Dem dem/target/scala-2.11/dem-assembly-0.jar $algo 30`), I got the following results when testing [a number of different vector join approaches](https://github.com/jamesmcclain/geotrellis/tree/fix/vector-join).

(Expand/reduce is not a vector join approach, it is there for comparison.)

Algorithm|Millis
---|---
Expand/Reduce|18062
Vector Join (current master)|24073
Vector Join (produce list of keys per Geometry)|152538
Vector Join (Eugene's original w/ optimal `LayoutDefinition`)|143234

Conclusion: the current vector join implementation is pretty good.  It is computationally optimal but sub-optimal in terms of communication.  I cannot find anything in the list of standard Spark "transformations" and "actions" that provides a way forward in terms of reducing communication overhead (there may or may not be be something in the developer APIs).

PR #1796, opened 2016-11-11
VectorJoin Fixes

jamesmcclain:
This changeset includes various fixes and improvements to vector join.

Connects #1795

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/join/VectorJoin.scala
+++ b/spark/src/main/scala/geotrellis/spark/join/VectorJoin.scala
@@ -21,11 +21,11 @@ object VectorJoin {
     L: ClassTag : ? => Geometry,
     R: ClassTag : ? => Geometry
   ](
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/geotrellis/spark/join/VectorJoin.scala
```diff
@@ -36,31 +36,29 @@ object VectorJoin {
 
       Iterator(rtree)
     }, preservesPartitioning = true)
-      .zipWithIndex
-      .map({ case (v, k) => (k, v) })
       .cache
```
lossyrob: This shouldn't cache, no longer necessary (and also we as library developers should avoid caching - that's a client concern that could have negative performance impact based on your setup)

jamesmcclain: Okay, will remove.

PR #1797, opened 2016-11-11
DelayedConvert feature

lossyrob:
This feature solves a very specific problem: Sometimes you want to `map` over a Tile, or `combine` a MulitbandTile's bands, or the like, and have the resulting Tile be of a different `CellType` then the original tile. For instance, when doing NDVI, if you do this on a `IntConstantNoDataCellType` multiband raster:

```scala
multibandTile.combineDouble(3, 4) { (r, ir) =>
  (ir - r) / (ir + r)
}
```

You'll get garbage values, since the ratio produces results from [-1, 1], but our result tile will be `IntConstantNoDataCelltype`, which will not hold the floating point values. What you have to do now is convert it to a floating point valued raster first:

```scala
multibandTile.convert(DoubleConstantNoDataCellType).combineDouble(3, 4) { (r, ir) =>
  (ir - r) / (ir + r)
}
```

This is awful for performance, because not only do you have to iterate over band 3 and 4 to set each value to a double value before you do the combine, but this will actually iterate over bands you don't even care about. What you can do to get around the latter issue is this:

```scala
multibandTile.subsetBands(3, 4).convert(DoubleConstantNoDataCellType).combineDouble(0, 1) { (r, ir) =>
  (ir - r) / (ir + r)
}
```

but that doesn't not solve the latter problem.

This solves this particular brand of problem by creating a `delayedConversion` method, which returns a `Tile` or `MultibandTile` that is exactly like the parent tile, except that for any method that produces a `Tile` or `MultibandTile`, that resulting tile will be an `ArrayTile` or `ArrayMultibandTile` with the target cell type. Our NDVI calculation now becomes:

```scala
multibandTile.delayedConversion(DoubleConstantNoDataCellType).combineDouble(3, 4) { (r, ir) =>
  (ir - r) / (ir + r)
}
```

And now we do the thing we want: iterate over band 3 and 4 only, and set the double NDVI value into a `DoubleConstantNoDataCellType`.

__NOTE__

Also contained in this PR is some ArrayMultibandTile optimizations.

Connects https://github.com/geotrellis/geotrellis/issues/1410

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/DelayedConversionMultibandTile.scala
+++ b/raster/src/main/scala/geotrellis/raster/DelayedConversionMultibandTile.scala
@@ -1,0 +1,306 @@
+package geotrellis.raster
+
+import geotrellis.raster.resample._
+import geotrellis.vector.Extent
+
+import spire.syntax.cfor._
+
+/**
+  * [DelayedConversionTile]] represents a tile that wraps an inner tile,
+  * and for any operation that returns a Tile, returns an ArrayTile with
+  * a cell type of the target cell type.
+  */
+class DelayedConversionMultibandTile(inner: MultibandTile, override val targetCellType: CellType)
+  extends MultibandTile with MacroMultibandCombiners {
+
+  private def validateBand(i: Int) = assert(i < bandCount, s"Band index out of bounds. Band Count: $bandCount Requested Band Index: $i")
+
+  val cols = inner.cols
+  val rows = inner.rows
+
+  def cellType: CellType =
+    inner.cellType
+
+  def convert(cellType: CellType): MultibandTile =
+    inner.convert(cellType)
+
+  def withNoData(noDataValue: Option[Double]): MultibandTile =
+    inner.withNoData(noDataValue)
+
+  def interpretAs(newCellType: CellType): MultibandTile =
+    withNoData(None).convert(newCellType)
+
+  def bandCount: Int = inner.bandCount
+  def band(bandIndex: Int): Tile = inner.band(bandIndex)
+  def bands: Vector[Tile] = inner.bands
+
+  def subsetBands(bandSequence: Seq[Int]): MultibandTile =
+    inner.subsetBands(bandSequence)
+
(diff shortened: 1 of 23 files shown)


lossyrob:
@echeipesh ready for review

raster/src/main/scala/geotrellis/raster/DelayedConversionMultibandTile.scala
```diff
@@ -0,0 +1,313 @@
...
+    })
+
+    (0 until bandCount).foreach({ b =>
+      val bandTile = band(b).delayedConversion(targetCellType)
+
+      if (set.contains(b))
+        newBands(b) = bandTile.map({ z => f(b, z) })
+      else if (targetCellType.isFloatingPoint)
+        newBands(b) = bandTile.mapDouble({ z => z })
+      else
```
echeipesh: Useless else, this is already fixed in `ArrayMultibandTile` in this PR, just need to update here.

raster/src/main/scala/geotrellis/raster/DelayedConversionMultibandTile.scala
```diff
@@ -0,0 +1,313 @@
...
+
+    (0 until bandCount).foreach({ b =>
+      val bandTile = band(b).delayedConversion(targetCellType)
+
+      if (set.contains(b))
+        newBands(b) = bandTile.mapDouble({ z => f(b, z) })
+      else if (targetCellType.isFloatingPoint)
+        newBands(b) = bandTile.mapDouble({ z => z})
+      else
+        newBands(b) = bandTile.map({ z => z })
```
echeipesh: ditto above

echeipesh:
+1 after review

raster/src/main/scala/geotrellis/raster/DelayedConversionMultibandTile.scala
lossyrob: They need to change cell type, so I changed these to use `convert` to be more explicit

echeipesh: Ah, that's right.

PR #1799, opened 2016-11-11
Fix math error in CDF interpolation

echeipesh:
Verified using by equalizing Landsat imagery using histogram of the layer recorded at ingest.

image


jamesmcclain:
Superseded by geotrellis/geotrellis#1803

PR #1800, opened 2016-11-14
More test for foreachCellByPolygon

zaikin:


jamesmcclain:
It looks as though the test that you added passes on OpenJDK  but not on the Oracle JDK.  We will investigate.

zaikin:
It seems like you haven’t run this test with openjdk...

jamesmcclain:
> It seems like you haven’t run this test with openjdk...

You are right, I had forgotten that the tests were split like that.

PR #1806, opened 2016-11-15
Ensure code in tutorial runs

jpolchlo:
This PR alters the kernel density tutorial so that the code can be copied and pasted into a console and run directly.


fosskers:
Fixed merge conflicts manually.

PR #1779, opened 2016-11-04
Finishing off Core Concepts

fosskers:
### TODO
- [x] Rework Projections, Extents, and Layout Definitions
- [x] Reorganize section order
- [x]  Modernize `Raster` section
- [x] Vector Data
- [x] Key and Key Indexes
- [x] Tiles
- [x] Port remaining older docs
  - [x] Rasters
  - [x] Vectors
  - [x] Spark

Connects azavea/geotrellis#9

PR #1831, opened 2016-11-26
[TO VIEW CORRECT COMMITS - DO NOT MERGE] Docs/core concepts.md

lossyrob:
https://github.com/locationtech/geotrellis/pull/1779

@fosskers having a hard time with this cherry-pick. There's conflicts that I am nervous I'm resolving incorrectly, and think probably you should do this. The process is to checkout a new branch off of master, then cherry pick each of the commits in this list via e.g.

`git cherry-pick 02c5f6e` (that would cherry pick the first one)

I got stumped at this commit, which deletes a lot of content, has a commit message that makes it seem like it should be adding a new markdown file, and doesn't: https://github.com/locationtech/geotrellis/pull/1779/commits/d4d1e94c594a0bd27b7978585a17cfc49f23d51c

While your at it, can you fix the incorrect link to "botanic.azavea.internal" etc in core-concepts? search for "Check out our macros docs for more on our use of macros like isData and isNoData." Looks like @moradology's doing but might just be worth fixing it up in this PR.


fosskers:
Sure, I can handle this.

fosskers:
@lossyrob This can be closed, we have a new PR: https://github.com/locationtech/geotrellis/pull/1872

PR #1836, opened 2016-11-27
[WIP] Conforming Delaunay

lossyrob:
This fixes #1821 after the LocationTech move.


lossyrob:
@jamesmcclain is this really WIP? What else needs to be done here?

PR #1838, opened 2016-11-27
Add MGRS encode/decode support to proj4

lossyrob:
Fixes #1650 after locationtech move.


lossyrob:
TODO: Open up a CQ for the codebase this is ported from.

lossyrob:
Merging now for RC; will create CQ as part of official release process.

PR #1849, opened 2016-11-28
[WIP] PointCloud Support (HDFS and S3 InputFormats, general PDAL support)

pomadchin:
Fixes #1816 after locationtech move

Be sure, that you have installed [PDAL JNI bindings](https://github.com/PDAL/PDAL/pull/1371), and correctly installed `LD_LIBRARY_PATH`.

If you use MAC OS, please install `LD_LIBRARY_PATH` as well (it is related to GeoTrellis projects only),


pointcloud/src/main/scala/geotrellis/spark/pointcloud/tiling/CutPointCloud.scala
```diff
@@ -0,0 +1,66 @@
...
+        val extent = inKey.extent
+
+        mapTransform(extent)
+          .coords
+          .map  { spatialComponent =>
+            val outKey    = inKey.translate(spatialComponent)
+            val outExtent = mapTransform(outKey)
+            val newBytes  = mutable.ArrayBuffer[Array[Byte]]()
+
+            cfor(0)(_ < pointCloud.length, _ + 1) { i =>
```
lossyrob: This is iterating over the points for each extent. We should be able to do this by iterating over the points once.

Sketch: have a Map[SpatialKey, Array[Byte]] that will hold the result.
Iterate over the points
use mapTransform(x, y) to get the spatial Key.
if the key exists in the map, add it to the array, or else initialize that key

lossyrob:
Closing, this is merged into the feature branch

Issue #1669, opened 2016-10-10
GeoWave Code Currently Non-Functional


jamesmcclain:
The GeoWave integration code (at least the raster portion of it) is currently non-functional.  Our code currently relies on GeoWave `0.9.3-SNAPSHOT` rather than a specific hash or release, and some incompatible changes were introduced at some recent point in the past.

Hopefully there will be some resolution before `0.9.3` is released an we can depend on that, or we can find a way to depend on a particular hash (`c127c16` is known to work).

A [GeoWave issue](https://github.com/ngageoint/geowave/issues/914) has been created.  The present issue is for keeping track of state of that issue with respect to our code.

The pull request https://github.com/geotrellis/geotrellis/pull/1665 fixes our tests so that they pass on GeoWave `c127c16`.

lossyrob:
If this is not fixed before the release, we'll have to mark this feature as experimental.

jamesmcclain:
That is logical.

jamesmcclain:
This appears to have been fixed by locationtech/geotrellis#1933

PR #1851, opened 2016-11-28
Disable Building of GeoWave

jamesmcclain:
The GeoWave subproject tests have been failing for a while, but now the GeoWave API has changed, so the code no longer builds.  Hopefully the disabling of this project will just be temporary?

See also: https://github.com/locationtech/geotrellis/issues/1669

screenshot from 2016-11-28 16 07 02


lossyrob:
Taken care of in  #1855

PR #1859, opened 2016-12-01
Added CollectNeighbors

lossyrob:
This feature allows you to connect generic values that are keyed on something with a spatial component.


echeipesh:
:+1:

Only comment is a suggestion: In cases like this I increasingly find it handy to nest the enumerating objects like this:

```scala
sealed trait Direction

object Direction {
  case object CenterDirection extends Direction
  case object TopDirection extends Direction
  case object TopRight extends Direction
  case object Right extends Direction
  case object BottomRight extends Direction
  case object Bottom extends Direction
  case object BottomLeft extends Direction
  case object Left extends Direction
  case object TopLeft extends Direction
}
```

Keeps the silly names down, namespaces the super generic name, which is what I think you're after, and still allows for `import geotrellis.spark.buffer.Direction._` when you need to use it a lot.

PR #1865, opened 2016-12-02
Remediate VectorJoin

### LINKED ISSUE #1795: Fix VectorJoin

jamesmcclain:
A new RDD type has been defined, which is similar to the existing `CartesianRDD` type, but which allows short-circuiting of communication when a partition can be demonstrated to be disjoint from another (the word  "disjoint" is defined by a predicate function).  This behavior is useful for performing vector joins.

   - [x] Further optimization possible

When joining an RDD of 10,000 elements (split into 32 partitions) to itself (giving the driver 32GB of memory, running locally on 32 cores), I get the following results:
```
vector join apply: count=88804   millis=17014
vector join naive: count=88804   millis=43075
```

10,000 (64 partitions):
```
vector join apply: count=88804   millis=25334
vector join naive: count=88804   millis=42647
```

10,000 (128 partitions):
```
vector join apply: count=88804   millis=32060
vector join naive: count=88804   millis=43319
```

40,000 (32 partitions):
```
vector join apply: count=357604          millis=83635
vector join naive: count=357604          millis=566942
```

40,000 (256 partitions):
```
vector join apply: count=357604          millis=171588
vector join naive: count=357604          millis=622096
```

Fixes #1795

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/join/VectorJoin.scala
+++ b/spark/src/main/scala/geotrellis/spark/join/VectorJoin.scala
@@ -32,14 +32,10 @@ object VectorJoin {
   /**
     * A function which calculates the envelope of a partition.
     *
-    * @param  i   The index of the partition
     * @param  gs  An iterator containing the contents of the RDD
-    * @return     An Iterator containing one index-envelope pair
+    * @return     An Iterator containing one envelope
     */
-  def calculateEnvelope[T : ? => Geometry](
-    i: Int,
-    gs: Iterator[T]
-  ): Iterator[(Int, Envelope)] = {
+  def calculateEnvelope[T : ? => Geometry](gs: Iterator[T]): Iterator[Envelope] = {
     val env = gs.foldLeft(new Envelope)({ (env: Envelope, g: T) =>
       val Extent(xmin, ymin, xmax, ymax) = g.envelope
       val env2 = new Envelope(xmin, xmax, ymin, ymax)
@@ -47,7 +43,7 @@ object VectorJoin {
       env
     })
 
-    Iterator((i, env))
+    Iterator(env)
   }
 
   /**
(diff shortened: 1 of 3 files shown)


spark/src/main/scala/geotrellis/spark/join/VectorJoin.scala
```diff
@@ -18,63 +18,79 @@ package geotrellis.spark.join
...
+  /**
+    * Perform the vector join operation over an RDD[L] and and RDD[R],
+    * where both L and R are viewble as Geometry.
+    *
+    * @param  left   An RDD[L], where L is viewable as a Geometry
+    * @param  right  An RDD[R], where R is viewable as a Geometry
+    * @param  pred   A predicate which answers whether an L and an R should be joined
+    * @return        An RDD of L-R pairs
+    */
+  def naive[
```
lossyrob: Any reason to keep around the naive implementation?

jamesmcclain: No, I have just removed it.  Will push up all or most review changes at once.

echeipesh:
:+1: will try to get this into Jupyter and play with it but this looks good.

spark/src/main/scala/org/apache/spark/rdd/FilteredCartesianRDD.scala
```diff
@@ -0,0 +1,87 @@
...
+  override def getPreferredLocations(split: Partition): Seq[String] = {
+    val currSplit = split.asInstanceOf[CartesianPartition]
+    (rdd1.preferredLocations(currSplit.s1) ++ rdd2.preferredLocations(currSplit.s2)).distinct
+  }
+
+  override def compute(split: Partition, context: TaskContext): Iterator[(T, U)] = {
+    val currSplit = split.asInstanceOf[CartesianPartition]
+    val part1 = currSplit.s1 // partition for (|meta)rdd1
+    val part2 = currSplit.s2 // partition for (|meta)rdd2
+    val meta1 = metardd1.iterator(part1, context).next
```
echeipesh: What makes sure that the `iterator` is usable here? AFAII it is the reference in `NarrowDependency` that makes certain partitions available in `compute`.  But its not clear how that would happen for `metardd1`. I think this is a problem that will reveal itself in the distributed case with multiple machines and it just happens to work in the one node case.

jamesmcclain: I believe that this works by the same mechanism as the two uses `iterator` below on lines 66 and 67: if the data are available they are fetched directly, [otherwise they are recomputed](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/rdd/RDD.scala#L283-L289).

jamesmcclain: Just to document the conversation that I had with @echeipesh A little bit: when One does a `mapPartitions` with `preservesPartitioning = true`, the child RDD inherts its partitions directly from its parent (see [here](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/rdd/MapPartitionsRDD.scala#L35) and [here](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/rdd/RDD.scala#L245-L260)).  We assume that the partitions stay valid for the life of the RDD, because if they did not then the [`CartesianRDD` type](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/rdd/CartesianRDD.scala#L35-L36) would probably not work.

PR #1873, opened 2016-12-05
Fractional-Pixel Polygon Rasterizer

jamesmcclain:
In this changeset, a polygon rasterizer which reports partial pixels and the respective fractions of those pixels that are covered by the polygon is presented.  The approach used seems to be the most efficient that is practical if sub-pixel geometries are to be supported.

   - [x] Bresenham or similar for generating partial pixels

 [This resource](https://gis.stackexchange.com/questions/59821/how-to-calculate-volume-from-a-raster-file) recommends the same method for volumetric calculation as we are envisioning.  I have not found any contradictory resources.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/rasterize/polygon/FractionalRasterizer.scala
+++ b/raster/src/main/scala/geotrellis/raster/rasterize/polygon/FractionalRasterizer.scala
@@ -65,97 +65,88 @@ object FractionalRasterizer {
     re: RasterExtent,
     poly: Polygon,
     set: mutable.Set[(Int, Int)],
-    fn: FractionCallback
+    cb: FractionCallback
   ): Unit = {
-    // Screen coordinates
     val (x0, y0, x1, y1) = edge
-    val xmin = min(x0, x1)
-    val ymin = min(y0, y1)
-    val xmax = max(x0, x1)
-    val ymax = max(y0, y1)
     val m = (y1 - y0) / (x1 - x0)
 
-    // Integral screen coordinates
-    val xminint = floor(xmin).toInt
-    val yminint = floor(ymin).toInt
-    val xmaxint =  ceil(xmax).toInt
-    val ymaxint =  ceil(ymax).toInt
+    // Grid coordinates
+    val colMin = floor(min(x0, x1)).toInt
+    val rowMin = floor(min(y0, y1)).toInt
+    val colMax =  ceil(max(x0, x1)).toInt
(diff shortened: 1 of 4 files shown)


raster/src/main/scala/geotrellis/raster/rasterize/polygon/FractionalRasterizer.scala
```diff
@@ -0,0 +1,185 @@
...
+
+  def foreachCellByPolygon(
+    poly: Polygon,
+    re: RasterExtent
+  )(fn: FractionCallback): Unit = {
+    val seen = mutable.Set.empty[(Int, Int)]
+    val option = Rasterizer.Options(includePartial = false, sampleType = PixelIsArea)
+
+    polygonToEdges(poly, re)
+      .par
```
lossyrob: We keep parallelization out of core geotrellis, since we promote parallelization via other means (e.g. spark). It's unclear if that's definitely the right choice, but that's the choice the rest of the library has made.

Perhaps there should be an option that defaulted to false for parallelization?

raster/src/main/scala/geotrellis/raster/rasterize/polygon/FractionalRasterizer.scala
```diff
@@ -0,0 +1,185 @@
...
+        }
+        y += 1
+      }
+    }
+  }
+
+  def foreachCellByPolygon(
+    poly: Polygon,
+    re: RasterExtent
+  )(fn: FractionCallback): Unit = {
```
lossyrob: We should use a trait instead of Function3, because Function3 isn't specialized. The way we get around this for tile methods is via macros. We would eventually want to macroize this as well, but that can be a TODO.

raster/src/main/scala/geotrellis/raster/rasterize/polygon/FractionalRasterizer.scala
```diff
@@ -0,0 +1,185 @@
...
+  private def renderEdge(
+    edge: Segment,
+    re: RasterExtent,
+    poly: Polygon,
+    set: mutable.Set[(Int, Int)],
+    fn: FractionCallback
+  ): Unit = {
+    // Screen coordinates
+    val (x0, y0, x1, y1) = edge
+    val xmin = min(x0, x1)
```
lossyrob: why call these screen coordinates? The convention in the library is that these are "grid coordinates", and we represent them with `col`, `row`, `rowMin`, etc. This differentiates them from `x`, `y`, `xmin` etc which represent map coordinates.

PR #1874, opened 2016-12-06
Cleanup for 1.0

lossyrob:
- Added more info in contributing.md
- Changed inconsistent naming of EPSG vs Epsg in code
- Gave a better error message for CRS write fails
- Modified about.html with 3rd party dependencies (autogenerated from Eclipse IP log)


lossyrob:
It looks like this didn't actually make it into 1.0

PR #1843, opened 2016-11-27
Updated the Intro to Spark ETL Doc

lossyrob:
Fixes #1820 after the locationtech repo move.

PR #1884, opened 2016-12-06
Updated the etl doc so that it reflects the new fields that have been…

jbouffard:
This PR is the same as #1843, only this one can be merged with the current master.


fosskers:
Unfortunately this PR unfixes a lot of things I had corrected previously in my giant `Core Concepts` PR.

PR #1901, opened 2016-12-11
Slight changes to PointCloudTriangulation

lossyrob:


lossyrob:
Proper signoff added in #1902

PR #1903, opened 2016-12-11
LightPoint -> Point3D, other TIN changes

lossyrob:
Optimized TIN rasterization.
Cleaned up code a bit.


raster/src/main/scala/geotrellis/raster/rasterize/triangles/TrianglesRasterizer.scala
```diff
@@ -42,7 +55,7 @@ object TrianglesRasterizer {
         val x = xStart + col * re.cellwidth
         val y = yStart + row * re.cellheight
         val screenCol = ((x - re.extent.xmin) / re.cellwidth).toInt
-        val screenRow = ((y - re.extent.ymin) / re.cellheight).toInt
+        val screenRow = ((re.extent.ymax -y) / re.cellheight).toInt
```
lossyrob: @jamesmcclain I believe this is a legitimate change (rows go down, map coords to up), but this breaks tests. Can you verify this is correct, considering increasing Y values go north and increasing row values go south? If so, can you change the tests?

PR #1905, opened 2016-12-12
Streaming/Windowed GeoTiff Reading

echeipesh:
The main culprit here is `StreamingSegmentBytes`. In order to do its job it needs to know ahead of time which segments you will read. At that point it can use the TiffTags to group them into contiguous chunks which can be fetched in on-disk order. This requires slight inversion of control between `segmentBytes` and `GeoTiffTile` where we allow the bytes to decide the order in which we see the segment. Overall it looks like this:

```scala
cfor(0)(_ < segmentCount, _ + 1) { segmentIndex =>
   val segment = getSegment(segmentIndex).bytes
   val segmentSize = segment.size
   val bandSegmentCount = segmentSize / bandCount
   val bandSegment = Array.ofDim[Byte](bandSegmentCount)
...
```
is replaced by
```scala
getSegments(0 until segmentCount).foreach { case (segmentIndex, geoTiffSegment) =>
    val segment = geoTiffSegment.bytes
    val segmentSize = segment.size
    val bandSegmentCount = segmentSize / bandCount
    val bandSegment = Array.ofDim[Byte](bandSegmentCount)
...
```

Because `getSegments` gives us an iterator we can zip them and use  them  for combine as well:
```scala
getSegments(0 until segmentCount)
  .zip(otherGeoTiff.getSegments(0 until segmentCount))
  .foreach { case ((segmentIndex, segment), (otherIndex, otherSegment)) =>
    require(segmentIndex == otherIndex, s"Segment index mismatch: $segmentIndex != $otherIndex")
    val newBytes = segment.mapWithIndex { (i, z) =>
      f(z, otherSegment.getInt(i))
    }
    arr(segmentIndex) = compressor.compress(newBytes, segmentIndex)
  }
```

Things that need done before this is done:
- [x] Test on S3 and verify performance on reading fully and windowed
- [x] Test on S3  file reading multiple crops from the same GeoTiff (verify that byteReader doesn't Reset)
- [x] Replace all usages of `getSegment` with `getSegments` in `GeoTiffMultiBandTile` (some a tricksy)
- [x] Update the `StreamingSegmentBytesSpec` for current code (restored `LazySegmentBytes`)


raster/src/main/scala/geotrellis/raster/io/geotiff/Float64GeoTiffSegmentCollection.scala
```diff
@@ -16,22 +16,18 @@
...
+  def noDataValue: Option[Double]
 
-  lazy val createSegment: Int => Float64GeoTiffSegment = noDataValue match {
+  lazy val decompressGeoTiffSegment = noDataValue match {
     case None =>
-      { i: Int => new Float64RawGeoTiffSegment(getDecompressedBytes(i)) }
-    case Some(nd) if (nd == Float.NaN) =>
-      { i: Int => new Float64ConstantNoDataGeoTiffSegment(getDecompressedBytes(i)) }
+      (i: Int, bytes: Array[Byte]) => new Float64RawGeoTiffSegment(decompressor.decompress(bytes, i))
+    case Some(nd) if nd.isNaN =>
```
echeipesh: Ditto on the NaN

raster/src/main/scala/geotrellis/raster/io/geotiff/Float32GeoTiffSegmentCollection.scala
```diff
@@ -16,22 +16,18 @@
...
+  def noDataValue: Option[Float]
 
-  lazy val createSegment: Int => Float32GeoTiffSegment = noDataValue match {
+  lazy val decompressGeoTiffSegment = noDataValue match {
     case None =>
-      { i: Int => new Float32RawGeoTiffSegment(getDecompressedBytes(i)) }
-    case Some(nd) if (nd == Float.NaN) =>
-      { i: Int => new Float32ConstantNoDataGeoTiffSegment(getDecompressedBytes(i)) }
+      (i: Int, bytes: Array[Byte]) => new Float32RawGeoTiffSegment(decompressor.decompress(bytes, i))
+    case Some(nd) if nd.isNaN =>
```
lossyrob: We always use `java.lang.Float.isNaN(nd)`, because for some reason it's significantly faster than `.isNaN`

raster/src/main/scala/geotrellis/raster/io/geotiff/Float64GeoTiffSegmentCollection.scala
lossyrob: We always use `java.lang.Float.isNaN(nd)`, because for some reason it's significantly faster than `.isNaN`

raster/src/main/scala/geotrellis/raster/io/geotiff/ArraySegmentBytes.scala
```diff
@@ -45,6 +41,10 @@ class ArraySegmentBytes(compressedBytes: Array[Array[Byte]]) extends SegmentByte
     * @return    An Array[Byte] that contains the bytes of the segment
     */
   def getSegment(i: Int) = compressedBytes(i)
```
echeipesh: Use of `getSegment` needs to be minimized wherever possible in favor of `getSegments`. This will make all the difference in working with streaming segment sources.

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
```diff
@@ -166,74 +168,36 @@ abstract class GeoTiffMultibandTile(
...
-            compressedBandBytes(segmentIndex) = compressor.compress(bandSegment, segmentIndex)
+          getSegments(0 until segmentCount).foreach { case (segmentIndex, geoTiffSegment) =>
+            val bytes = GeoTiffSegment.deinterleave(geoTiffSegment.bytes, bandCount, bytesPerSample)(bandIndex)
+            compressedBandBytes(segmentIndex) = compressor.compress(bytes, segmentIndex)
           }
 
           GeoTiffTile(new ArraySegmentBytes(compressedBandBytes), compressor.createDecompressor(), segmentLayout, compression, cellType, Some(bandType))
       }
     } else {
+      // TODO: rewrite this to center around getSegments
```
echeipesh: Rewrite this to center around getSegments.
Band interlaced segments are stored in order, first you see all the segments for band 0, then band 1, etc.

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
```diff
@@ -166,74 +168,36 @@ abstract class GeoTiffMultibandTile(
...
-            var j = 0
-            cfor(bandIndex * bytesPerSample)(_ < segmentSize, _ + bytesPerCell) { i =>
-              var b = 0
-              while(b < bytesPerSample) { bandSegment(j + b) = segment(i + b) ; b += 1 }
-              j += bytesPerSample
-            }
 
-            compressedBandBytes(segmentIndex) = compressor.compress(bandSegment, segmentIndex)
+          getSegments(0 until segmentCount).foreach { case (segmentIndex, geoTiffSegment) =>
+            val bytes = GeoTiffSegment.deinterleave(geoTiffSegment.bytes, bandCount, bytesPerSample)(bandIndex)
```
echeipesh: This is actually not very efficient because we know we only want one band, but these functions will de-interlace all present bands, allocating all that memory for nothing.

echeipesh: But this function needs to be used for `def bands`, which does exactly this, splits out all the bands into `Vector[Tile]`

echeipesh: Ideally there is an efficient way to deinterlace only the bands you want so it can also be used in `def subsetBands`.

echeipesh: `def toArrayTile` needs to delegate to `def bands`

echeipesh: combine functions need to be rewritten around `getSegments`

pomadchin:
Took into account all comments; `StreamingSegmentBytesSpec` is still needs to be done.
@echeipesh @lossyrob

raster/src/main/scala/geotrellis/raster/io/geotiff/SegmentBytes.scala
```diff
@@ -16,17 +16,18 @@
...
- * The base trait of SegmentBytes. It can be implemented either as
- * an Array[Array[Byte]] or as a ByteBuffer that is lazily read in.
- */
-trait SegmentBytes extends Traversable[Array[Byte]] {
+  * Represents the bytes of the segments with-in a GeoTiff.
+  *
+  * The base trait of SegmentBytes. It can be implemented either as
+  * an Array[Array[Byte]] or as a ByteBuffer that is lazily read in.
+  */
+trait SegmentBytes extends Seq[Array[Byte]] {
```
echeipesh: This had to changes from `Traversable` to `Seq` because the only way the former can tell its length is to iterate through the collection, which is obviously not great for streaming.

echeipesh:
Here is a handy test I'm using:

```scala
import geotrellis.spark.io.s3._
import geotrellis.spark.io._
import geotrellis.spark._
import geotrellis.raster._
import geotrellis.raster.io._
import geotrellis.raster.io.geotiff._
import geotrellis.proj4._
import geotrellis.vector._

val client = S3Client.DEFAULT

import geotrellis.spark.io.s3.util._
val rr = S3RangeReader("geotrellis-test", "rf-test/356f564e3a0dc9d15553c17cf4583f21-6.tif", client)
val tiff = MultibandGeoTiff.streaming(rr)
val sube = tiff.extent.center.buffer(tiff.extent.width * 0.01).envelope
tiff.crop(sube)
```

PR #1919, opened 2016-12-16
Bring home orphaned `raster` JsonFormats

fosskers:
The format for `CellSize` in particular was in an odd place. It's been moved to `raster`.


echeipesh:
This one is alright, there is no credible reason to depend on `etl` package so this essentially an API addition.

PR #1924, opened 2016-12-18
Set up publishing workflow

lossyrob:
This changes the build file to publish releases to Sonatype and snapshots to LocationTech.

This also sets up a docker container to build and publish releases through, along with a Makefile to make building the container and publish easy, with instructions in a README.


publish/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+RUN apt-get install git
+
+RUN useradd -ms /bin/bash geotrellis
+
+USER geotrellis
+
+RUN set -x \
+    && cd /home/geotrellis \
+    && git clone https://github.com/locationtech/geotrellis /home/geotrellis/geotrellis \
```
echeipesh: Are we specifically avoiding trying to publish from the current project directory and forcing publish from something thats already on github ?

PR #1930, opened 2016-12-20
GeoMesa update

pomadchin:
* GeoMesa 1.2.7 update
* sbt files syntax update


pomadchin:
@lossyrob looks like it is even for 1.0.1 ?

PR #1931, opened 2016-12-20
Added WKT Parser to improve WKT.getEpsgCode

### LINKED ISSUE #1775: Fix CRS from WKT String
Proj4 project promises to be able to translate from WKT string to an instance of `CRS` using [`CRS.fromWKT`](https://github.com/geotrellis/geotrellis/blob/master/proj4/src/main/scala/geotrellis/proj4/CRS.scala#L62-L66), this however is a lie. 

What the code tries to do is match WKT string against a registry of strings to find corresponding EPSG code. However this perform a string match, so any spaces or ordering is going to make the process fail. Also some WKT tags don't necessarily change the CRS if absent.

There is various levels of good to solve this issue, but at least: There should be tokenization for the match (this would remove spacing and ordering problem)
 
Preferably there may be an internal Proj4 structure that can be created from WKT tokens that would allow a "deep" match. 

Ref: https://github.com/geotrellis/geotrellis/blob/master/proj4/src/main/scala/geotrellis/proj4/io/wkt/WKT.scala

Ultimately this is the function to blame: https://github.com/geotrellis/geotrellis/blob/master/proj4/src/main/scala/geotrellis/proj4/io/wkt/WKT.scala#L60-L69

As part of the debugging we should add failing tests to this unite test file: https://github.com/geotrellis/geotrellis/blob/master/proj4/src/test/scala/geotrellis/proj4/io/wkt/WKTTest.scala

We should that equality is preserved when:
 - extra spaces
 - changed field order
 - optional fields like `EXTRA`

We should be able to parse this: http://spatialreference.org/ref/sr-org/7483/prettywkt/ as `EPSG:3857`

miafg:
Signed-off-by: miafg <miaf@sas.upenn.edu>

Fixes #1775 

The parser takes in a WKT string and creates a WktCS object that follows the WKT specifications in order to improve the lookup of WKT strings in the EPSG.properties file. Uses Scala parser combinator to parse the file. Additionally, the WKT object hierarchy could be useful in inspecting the WKT structure or creating a representation.


lossyrob:
Great job!

PR #1932, opened 2016-12-20
[WIP] Actually fast Delaunay triangulator

jpolchlo:
The JTS Delaunay triangulator is good and slow.  To achieve better runtimes, the following PR is submitted for your consideration.  It uses a half edge representation, implemented as a simple integer table.  Each row of the table captures an integer reference to a vertex (which will have their geometric particulars represented as JTS Coordinates), and integer references to other rows in the table that give the next edge in the current loop and the current edge's opposing half edge.  This representation is fast and Kryo serializable for use in Spark applications.

This PR also provides the means to build a collection of triangulations from a set of spatially partitioned point sets, extract the salient portion of those triangulations, and subsequently merge these boundary representations.  (The use of the extracted boundary meshes lowers the communication overhead when shuffling information between Spark partitions.)  This enables distributed computation of large triangulations.  The final triangulation comprises the per-partition triangulations overlayed with the triangles generated during stitching process (which likely will have a non-empty intersection).  The result is *not* a single, coherent mesh.  The stitched triangles supersede any triangles they overlap in the original triangulations.

- [x] Base triangulator
- [x] Extracting boundary meshes
- [x] Stitching boundary meshes
- [x] Testing the whole shebang
- [ ] Cleaning up code


jpolchlo:
Due to a failure to sign commits, this PR will be closed in favor of PR 1954.

PR #1936, opened 2016-12-21
Streaming API

pomadchin:
Fixes #1631 after the locationtech move


echeipesh:
Looking over this again reinforces two original conclusions:
 - we can't use spark streaming to make ingests more reliable
 - streaming API seems very compositional over RDD API, so there is probably no need to declare a dedicated GeoTrellis API for it.

The ADR is valuable and should be committed for future reference, but this PR should not merged until a more clear use case develops around streaming.

echeipesh:
Superseded by: https://github.com/locationtech/geotrellis/pull/2264

PR #1937, opened 2016-12-21
[WIP] Parquet support

pomadchin:
Fixes #1754 after locationtech repo move.


lossyrob:
How does this play into the work of #2051?

echeipesh:
#2051 will provide "deep" serialization to parquet that we get from native Spark API. More work will likely need to experiment with querying a parquet file based on those columns and optimizing the record structure, but it should not be related to this PR.

PR #1947, opened 2016-12-23
Collections mapalgebra local and mask functions

pomadchin:
- [x] local functions
- [x] temporal local functions
- [x] mask functions


lossyrob:
This adds a lot of functionality that "should" be there in 1.0. On that basis I think it can go into 1.0.1. @echeipesh thoughts?

PR #1948, opened 2016-12-26
Generic Functor usage

pomadchin:
Use more generic `scalaz.Functor` instead of our `Functor` abstraction.
Notice: I am not forcing this PR :D 


pomadchin:
@lossyrob It's a thing to discuss indeed as it uses scalaz, you are correct. Though, I remember that we talked about our [Functor](https://github.com/locationtech/geotrellis/blob/master/util/src/main/scala/geotrellis/util/Functor.scala) and this pr is a proof that we can use a common and more generic scalaz / cats `Functor` definition without defining our own type. I am not sure why we didn't try to do it before, but all in all it seems to me that a standard `Functor` type usage should be more clear for clients. 
I'll remove milestone label to continue talk / to close this pr and not to worry about it :)

fosskers:
Our custom `Functor` was originally introduced by #1643  to solve #1575 . After much initial experimenting there, we realized that we were trying to reinvent the `Functor` concept. Instead of doing that, we rolled an actual `Functor`, because at the time Scalaz's didn't seem to cut it, type-wise. Grisha's PR here suggests otherwise, so I must have been wrong before.

Either way, the addition of the `Functor` idea there trivially solved #1575. In terms of actually using Scalaz (or MeowMeowKitties) directly, I have some thoughts:

#### In General
- I'd much rather use a "community" `Functor` type than a home-spun one. This offloads maintenance overhead and improves interop with other libraries.

- We can elect to use Scalaz types only where necessary, avoiding some huge GeoTrellis API overhaul. More than once I've wanted to use a `State` monad in isolation.

#### On Difficulty
> [Scalaz] adds syntax and concepts to scala that many find make it even more difficult to use.

- We don't need to demand that our users use GT via Scalaz types. Use of it can remain as internal machinery, with a well-defined line-in-the-sand for GT contributors.  

- Scala is FP, and `Functor`, `Applicative` and `Monad` are facts of life by now. Anything with a `.map` is implicitely a `Functor`, `scala-parser-combinators` is 100% the Applicative pattern, and vanilla `for` blocks are sugar for binding Monadic operations. We already embrace the idea of typeclasses, and these are the canonical higher-kinded three. If their use would improve a code block's readability/usability/efficiency/length, then we should use them.

- Re: Line-in-the-sand. Internally, there's nothing compelling us to use `Cofree`, `Profunctor` or `Coyoneda` etc., if we don't want to. We *would* need to have an extended conversion about what else is kosher though. If we `Monad`, do we Monad Transformer where appropriate (say, in non-numerical higher-level operations)?

- We've already embraced lenses as a concept (both via Monocle and our GT `Component`s), which is arguably an "advanced" FP feature, although simple once you grock it (just like FAM). 

- GeoTrellis is already an advanced Scala library. We use pretty much all the tricks in the book, from project organization to type system tricks.

#### On Dependencies

- We already transitively depend on Scalaz via Monocle.

#### On Scalaz vs Cats

- We already depend on Scalaz transitively, so adding Cats on top of that opens the doors for confusion.
- All politics aside (of which I have no interest), Scalaz is more featureful than Cats.

fosskers:
If / when this is merged, we'll need to update the docs to remove references to our custom `Functor`.

PR #1949, opened 2016-12-28
Pipeline builder

pomadchin:


pointcloud/src/main/scala/geotrellis/pointcloud/pipeline/PipelineExpressions.scala
```diff
@@ -0,0 +1,487 @@
...
+  def json: Json =
+    Json.obj(
+      "pipeline" -> list
+        .map(
+          _.asJsonObject
+            .remove("class_type") // remove type
+            .filter { case (key, value) => !value.isNull } // cleanup options
+        ).asJson
+    )
+}
```
fosskers: It might be possible to avoid this class altogether.

```scala
 /* "Lightest" way */
 sealed trait PipelineExpr

 object Pipeline {
   type Pipeline = List[PipelineExpr]

   def json(list: Pipeline): Json = ... // same definition you have                                       
 }

 // Define a pipeline like:                                                                               
 val pipeline: Pipeline = List(
   foo,
   bar,
   baz
 )

 // Get the JSON                                                                                          
 val json: Json = Pipline.json(pipeline)
```
or
```scala
 /* Fancy way */
 sealed trait PipelineExpr {
   /* The next stage in the pipeline */
   def next: Option[PipelineExpr]

   def json: Json = ???
 }

 object PipelineExpr {
   val pipelineSemi: Semigroup[PipelineExpr] = new Semigroup[PipelineExpr] { ... }
 }

 // Define a pipeline like:                                                                               
 val json: Json = List(
   foo,
   bar,
   baz
 ).reduce(_ mappend _).json

// or like this
val json: Json = (foo |+| bar |+| baz).json
```

PR #1952, opened 2017-01-04
Fixed AvroEncoder So That it Will Now Work With the Python Bindings

jbouffard:
Changed the `toBinary` and `fromBinary` methods in `AvroEncoder` so that user has control on whether they when to compress and/or decompress an `Array[Byte]`, respectively.

I've published this branch locally and used it with `GeoPySpark`, and it worked.

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/io/avro/AvroEncoder.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/avro/AvroEncoder.scala
@@ -113,8 +113,8 @@ object AvroEncoder {
       val rec = reader.read(null.asInstanceOf[GenericRecord], decoder)
       format.decode(rec)
     } catch {
-        throw new AvroTypeException(e.getMessage + ". " +
       case e: AvroTypeException =>
+        throw new AvroTypeException(e.getMessage + ". " +
           "This can be caused by using a type parameter which doesn't match the object being deserialized.")
     }
   }
--- a/spark/src/main/scala/geotrellis/spark/io/avro/AvroEncoder.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/avro/AvroEncoder.scala
@@ -25,7 +25,7 @@ import org.apache.commons.io.IOUtils
 import org.apache.commons.io.output.ByteArrayOutputStream
 
 object AvroEncoder {
-  val deflater =new Deflater(Deflater.BEST_SPEED)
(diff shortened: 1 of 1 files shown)


jbouffard:

spark/src/main/scala/geotrellis/spark/io/avro/AvroEncoder.scala
```diff
@@ -100,8 +113,8 @@ object AvroEncoder {
       val rec = reader.read(null.asInstanceOf[GenericRecord], decoder)
       format.decode(rec)
     } catch {
-      case e: AvroTypeException =>
         throw new AvroTypeException(e.getMessage + ". " +
+      case e: AvroTypeException =>
```
pomadchin: looks like you have a typo here o:

jbouffard: Thanks for pointing that out! I'm not sure how that happened.

spark/src/main/scala/geotrellis/spark/io/avro/AvroEncoder.scala
```diff
@@ -38,14 +38,14 @@ object AvroEncoder {
     baos.toByteArray
   }
 
-  def  decompress(bytes: Array[Byte]): Array[Byte] = {
```
jamesmcclain: Sorry to be pedantic, but there is an extra space here.

jbouffard: It's fine. I just push up the change that should fix it.

jamesmcclain:
@jbouffard Please take a look at https://github.com/jbouffard/geopyspark/pull/1

spark/src/main/scala/geotrellis/spark/io/avro/AvroEncoder.scala
jamesmcclain: Actually, there was no extra space ... it was being removed there, not added.  Sorry about that.

lossyrob:
Note for release: this PR does not break API per se, because it only adds an optional parameter. It would break binary compatibility (I believe), so I'm marking as `api-changed` for discussion.

echeipesh:
Made non-breaking by: https://github.com/locationtech/geotrellis/pull/2059

PR #1955, opened 2017-01-05
[WIP] Restore custom Delaunay triangulator and apply to pointclouds

jpolchlo:
The JTS Delaunay triangulator is good and slow.  To achieve better runtimes, the following PR is submitted for your consideration.  It uses a half edge representation, implemented as a simple integer table.  Each row of the table captures an integer reference to a vertex (which will have their geometric particulars represented as JTS Coordinates), and integer references to other rows in the table that give the next edge in the current loop and the current edge's opposing half edge.  This representation is fast and Kryo serializable for use in Spark applications.

This PR also provides the means to build a collection of triangulations from a set of spatially partitioned point sets, extract the salient portion of those triangulations, and subsequently merge these boundary representations.  (The use of the extracted boundary meshes lowers the communication overhead when shuffling information between Spark partitions.)  This enables distributed computation of large triangulations.  The final triangulation comprises the per-partition triangulations overlayed with the triangles generated during stitching process (which likely will have a non-empty intersection).  The result is *not* a single, coherent mesh.  The stitched triangles supersede any triangles they overlap in the original triangulations.

- [x] Base triangulator
- [x] Extracting boundary meshes
- [x] Stitching boundary meshes
- [x] Testing the whole shebang
- [ ] Update Voronoi diagrams
- [ ] Update Euclidean distance tile
- [ ] Cleaning up code


pointcloud/src/main/scala/geotrellis/spark/pointcloud/triangulation/BoundaryDelaunay.scala
```diff
@@ -0,0 +1,287 @@
...
+
+import com.vividsolutions.jts.algorithm.distance.{DistanceToPoint, PointPairDistance}
+import com.vividsolutions.jts.geom.Coordinate
+
+import geotrellis.spark.buffer.Direction
+import geotrellis.spark.buffer.Direction._
+import geotrellis.vector.Extent
+import geotrellis.vector.triangulation.{DelaunayTriangulation, HalfEdgeTable, Predicates, TriangleMap}
+
+case class BoundaryDelaunay (dt: DelaunayTriangulation, boundingExtent: Extent) {
```
jpolchlo: Crap.  This will require carrying around the source `DelaunayTriangulation` object.  Will have to move construction over to the companion object.

vector/src/main/scala/geotrellis/vector/triangulation/DelaunayStitcher.scala
```diff
@@ -0,0 +1,276 @@
...
+    // right.prev.src is ___ base  ON            OK       ????     ADVANCE
+    //                             LEFT OF     REVERSE   REVERSE   REVERSE
+    //
+    // Linear sub-triangulations pose a special case.  Since, by virtue of
+    // using advance and reverse, e's source vertex must be a corner,
+    // e.dest and e.prev.src should always be *the same* vertex for linear
+    // triangulations.  Thus, both are right of, left of, or on the base
+    // candidate.  When both are left or right of base, we need no special
+    // handling, but when both are on base, we only need to advance if
+    // e.dest lies inside the line segment from base.src to base.dest.
```
lossyrob: great comment.

pomadchin:
Replaced by #1961

PR #1963, opened 2017-01-15
Address Polygon Rasterizer Issues

### LINKED ISSUE #1962: Bug in PolygonRasterizer
Hello,

I am fighting since a couple of hours with a " java.lang.ArrayIndexOutOfBoundsException" exception coming from the PolygonRasterizer.
After some dig out, I notice it is due to the line 
https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/rasterize/polygon/PolygonRasterizer.scala#L386

My rowRuns array as a length of 3  (using PixelIsPoint) method,
so when calling line 386 at the second i-loop iterations, it tries to access the fourth element of a 3 elements array... and crashes...

I understand that the cause of the error is likely coming from the runsPoint function itself,
or possibly because of this polygon being too complex and/or fishy to met the computation assumptions...
(I am not particularly interested in debugging this...)

However, I would still suggest to change line 384 from
` while (i < rowRuns.length) {`
to
` while (i < rowRuns.length-1) {`

which would make me miss a row in this fishy polygon, but would avoid the exception.

Thanks in advance,
Loic

jamesmcclain:
Still Needs
   - [x] Proper handling of horizontal segments for point-pixels
   - [x] More test cases
   - [x] Solidification
   -  Feedback from gentleman who reported bug?

Fixes #1962


jamesmcclain:
@quertenmont This is not completely finished, but it maybe complete enough to consider testing.  The polygons that you sent should work now.

quertenmont:
I'll try to give it a try in the coming days,
Thanks for your fast fix!

jamesmcclain:
You are welcome, sir.  Thank you for the bug report and test cases!

PR #1964, opened 2017-01-16
Added `withDefaultNoData` for all `DataType`s.

metasim:
Signed-off-by: Simeon H.K. fitch <simeon@metasim.net>

I have run into a number of situations where I'd like to enable the default `ConstantNoData` value for a `DataType` when the specific type isn't known at runtime. This PR is a draft implementation for the purposes of getting feedback. If this feature and implementation pass muster, I'll add unit tests.


metasim:
Need to sign with different address.

PR #1966, opened 2017-01-17
Added `withDefaultNoData` for all `DataType`s.

metasim:
Signed-off-by: Simeon H.K. fitch <fitch@astraea.io>

I have run into a number of situations where I'd like to enable the default ConstantNoData value for a DataType when the specific type isn't known at runtime. This PR is a draft implementation for the purposes of getting feedback. If this feature and implementation pass muster, I'll add unit tests.
 

- [ ] Create unit tests

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/CellType.scala
+++ b/raster/src/main/scala/geotrellis/raster/CellType.scala
@@ -44,6 +44,18 @@ sealed abstract class DataType extends Serializable { self: CellType =>
     */
   def withNoData(noDataValue: Option[Double]): CellType
 
+  /**
+   * Creates a [[CellType]] with the default `ConstantNoData` value. If instance is already one of the default `NoData`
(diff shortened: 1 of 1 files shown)


echeipesh:
This seems alright, I think we nearly wrote something like this originally but couldn't quite justify it so it was left out. It would be helpful to know what situations specifically prompted this for future understanding.

metasim:
@echeipesh This is where we've run into it: we have a directory structure of GeoTIFF images with different `CellType`s without the `NODATA` value set. These images (e.g. L8) have an implicit no-data value that is the same as GeoTrellis' default value. So, instead of having to pattern match over all the different cell types to set the default no-data value when one isn't set, it would be nice to have a polymorphic method that does that for us.

echeipesh:
steam checked it, looks good and once again thank you.

PR #1971, opened 2017-01-18
Avro Schema Creation for ProjectedExtent and TemporalProjectedExtent

jbouffard:
This PR adds two new codecs that can be encode/decoded in GeoTrellis: `ProjectedExtent` and `TemporalProjectedExtent`. The main reason for adding these two codecs is to allow for backend support between GeoTrellis and GeoPySpark.

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/io/avro/codecs/ProjectedExtentCodec.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/avro/codecs/ProjectedExtentCodec.scala
@@ -1,0 +1,38 @@
+package geotrellis.spark.io.avro.codecs
+
+import geotrellis.proj4.CRS
+import geotrellis.spark._
+import geotrellis.spark.io.avro._
+import geotrellis.spark.io.avro.codecs._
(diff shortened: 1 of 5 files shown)


jbouffard:

spark/src/main/scala/geotrellis/spark/io/avro/codecs/ProjectedExtentCodec.scala
```diff
@@ -0,0 +1,38 @@
...
+
+// --- //
+
+trait ProjectedExtentCodec {
+  implicit def projectedExtentCodec = new AvroRecordCodec[ProjectedExtent] {
+    def schema: Schema = SchemaBuilder
+      .record("ProjectedExtent").namespace("geotrellis.vector")
+      .fields()
+      .name("extent").`type`(extentCodec.schema).noDefault()
+      .name("epsg").`type`().intType().noDefault()
```
lossyrob: Can you update to use the proj4 string.
Or, if possible, check if it's an EPSG code, and then fall back to CRS?

PR #1982, opened 2017-01-24
AvroSerialization Now Uses Proj4Strings in the Encoding/Decoding of ProjectedExtents and TemproalProjectedExtents

jbouffard:
This PR fixes a previously merged in change, #1971 , so that the `Proj4String` of `ProjectedExtent` and `TemproalProjectedExtent` will be encoded/decoded instead of their EPSG code.


echeipesh:
This change is fine and it certainly allows you to support more projections. One thing that's convenient about EPSG code is that it's less ambitious and much more compact.

Do you think it makes sense to define the codec to store the EPSG code if its available or proj4 string otherwise?

jbouffard:
I think it does. I'll update this PR so that the serializer can use either value.

echeipesh:
Superseded by: https://github.com/locationtech/geotrellis/pull/2034

PR #1994, opened 2017-01-30
Add apply methods to s3rangereader

moradology:
A very common use case when using S3 is having to account for URIs and strings. We can use the structure of S3 URIs to make the API for grabbing an `S3RangeReader` a little less imposing.


moradology:
@jbouffard

moradology:
## TiffTags read API comparison

Before:
```scala
val s3Uri = new AmazonS3URI(uri)
val s3Client = new AmazonS3Client(new AWSAmazonS3Client(new DefaultAWSCredentialsProviderChain))
val s3RangeReader = S3RangeReader(s3Uri.getBucket, s3Uri.getKey, s3Client)
TiffTagsReader.read(s3RangeReader)
```

After:
```scala
val s3RangeReader = S3RangeReader(uri)
TiffTagsReader.read(s3RangeReader)
```

PR #1995, opened 2017-01-30
[WIP] Repairs to pointcloud Delaunay triangulation

jpolchlo:
This PR aims to fix triangulation problems that arise when attempting to perform distributed triangulations.  

These errors affected usage of StitchedDelaunay when applied to BoundaryDelaunay objects.  The trouble appears to be related to having an insufficient buffer of triangles in the construction of BoundaryDelaunay instances.  It used to be the case that a BoundaryDelaunay object contained all and only the triangles from some source triangulation with circumcircles that extended outside a specified Extent; the "and only" clause has now been relaxed.  The criterion is now that the triangle's circumcircle leaves the given extent, or the triangle shares an edge with a triangle that satisfies that condition (or the triangle has an edge which is on the boundary of the triangulation).


pointcloud/src/main/scala/geotrellis/pointcloud/spark/triangulation/BoundaryDelaunay.scala
```diff
@@ -209,12 +209,16 @@ case class BoundaryDelaunay (dt: DelaunayTriangulation, boundingExtent: Extent)
...
+
             if (circumcircleLeavesExtent(boundingExtent)(e)) {
               //println("         Triangle circle leaves extent")
-              val tri = copyConvertTriangle(e)
+      //        val tri = copyConvertTriangle(e)
+
               //print("         ")
               //navigator.showLoop(tri)
-              navigator.join(opp, tri)
+      //        navigator.join(opp, tri)
```
jpolchlo: The salient change in this PR is to comment lines 217 and 221 and move them to lines 212 and 213.

PR #1999, opened 2017-02-02
Spark-Enabled Cost-Distance

### LINKED ISSUE #1980: Cost-Distance for Spark
# A #

I am assuming that the goal is to implement something similar to or inspired by the algorithm given at the bottom of page 13 in [1].

The main characteristic of this algorithm that prevents it from being simply reorganized into a data-parallel shape is the (mostly false) dependence of all of the pixels on the top of the single priority queue.  That dependence is mostly false because the distance from any pixel to the set of starting geometry can be defined with respect to only one of the data.

# B #

For a completely "homogeneous medium" (using the language found in the article), the issue mentioned above can be solved by producing a (general) Voronoi diagram of the starting geometry.  Points (pixels) can be quickly located with respect to Voronoi cells, which provides a way to produce an algorithm  that fits into Spark's notion of data parallelism.

Of course, this is not sufficient as a solution if the medium is heterogeneous.

# C #

It is possible to estimate the skew in the radius of (what would have been) a homogeneous Voronoi cell caused by the heterogeneity of the medium.  That can be used to conservatively associate more than one geometry with a cell, rather than just one.  It can be hoped that the data will be such that this number will be bounded by a smallish number over all pixels.

~~(That was hand-wavey: it can probably be made more formal by considering the k-th order rather than first-order Voronoi diagram of the starting geometries.)~~

# D #

The foregoing discussion assumed that we must closely follow the existing algorithm.  There may also be room for a more original approach.

-----

1. Tomlin, Dana. "Propagating radial waves of travel cost in a grid." International Journal of Geographical Information Science 24.9 (2010): 1391-1413.

jamesmcclain:
- [x] Simplified cost-distance implementation
- [x] Implementation of distributed cost-distance algorithm
- [x] Unit Tests
- [x] Comments

Fixes #1980 

screenshot from 2017-02-06 13 14 25

screenshot from 2017-02-06 13 22 14

screenshot from 2017-02-14 14 10 36

screenshot from 2017-02-14 14 10 40


lossyrob:
party like it's PR 1999

jamesmcclain:
:tada: :tada: :tada: :tada: :tada: :tada: :tada:

spark/src/main/scala/geotrellis/spark/costdistance/MrGeoCostDistance.scala
```diff
@@ -0,0 +1,187 @@
...
+  * algorithm [1] is "heavily inspired" by the MrGeo implementation
+  * [2] but does not share any code with it.
+  *
+  * 1. Tomlin, Dana.
+  *    "Propagating radial waves of travel cost in a grid."
+  *    International Journal of Geographical Information Science 24.9 (2010): 1391-1413.
+  *
+  * 2. https://github.com/ngageoint/mrgeo/blob/0c6ed4a7e66bb0923ec5c570b102862aee9e885e/mrgeo-mapalgebra/mrgeo-mapalgebra-costdistance/src/main/scala/org/mrgeo/mapalgebra/CostDistanceMapOp.scala
+  */
+object MrGeoCostDistance {
```
lossyrob: I think the reference in the doc string is sufficient pointing back to Mr Geo, the type names should probably just be `CostDistance`

spark/src/main/scala/geotrellis/spark/costdistance/MrGeoCostDistance.scala
```diff
@@ -0,0 +1,187 @@
...
+      val costs: List[CostDistance.Cost] = points
+        .filter({ point => extent.contains(point) })
+        .map({ point =>
+          val col = rasterExtent.mapXToGrid(point.x)
+          val row = rasterExtent.mapYToGrid(point.y)
+          val friction = tile.getDouble(col, row)
+          (col, row, friction, 0.0)
+        })
+      accumulator.add((key, costs))
+    })
```
lossyrob: I see the loop version does a `count` to force the execution and cache the RDD. Is it possible to avoid this `foreach` with a similar method, where this accumulator is moved to the `map`, that RDD persisted, and the RDD counted to force execution? Perhaps not the best optimization, but your opinion on this would probably help me understand the logic better.

jamesmcclain: The `foreach` is strictly for the side-effects (to set the accumulator).

lossyrob: Gotcha, but it causes a double iteration of the RDD. The accumulator could be set in the map, and then the map executed with caching, so that the RDD transformation is cached and iterated over once, and the accumulator value is set.

jamesmcclain: Oh right

jamesmcclain:
All comments addressed.

raster/src/main/scala/geotrellis/raster/costdistance/CostDistance.scala
```diff
@@ -16,236 +16,190 @@
...
-        while(z < l) {
-          val p = cornerOffsets(z)
-          val c1 = p._1 + c
-          val r1 = p._2 + r
-          val cost1 = calcCost(c, r, c1, r1, cost)
-          if (cost1.isDefined) {
-            val cost2 = calcCost(c1, r1, dir(c, r), cost)
-            if (cost2.isDefined) {
-              curMinCost = math.min(curMinCost, source + cost1.get + cost2.get)
+  def compute(
```
lossyrob: Have you benchmarked the single-threaded case for this change? It would be good to get numbers to prove this method works faster/just as fast.

spark/src/main/scala/geotrellis/spark/costdistance/IterativeCostDistance.scala
```diff
@@ -0,0 +1,221 @@
...
+  /**
+    * Perform the cost-distance computation.
+    *
+    * @param  friction  The friction layer; pixels are in units of "seconds per meter"
+    * @param  points    The starting locations from-which to compute the cost of traveling
+    * @param  maxCost   The maximum cost before pruning a path (in units of "seconds")
+    */
+  def apply[K: (? => SpatialKey), V: (? => Tile)](
+    friction: RDD[(K, V)] with Metadata[TileLayerMetadata[K]],
+    points: Seq[Point],
```
lossyrob: Discussion point, not a change request: What would it take to create an option to pass in an RDD of points? I'm curious if there are use cases for this, and if it would be possible with the current algorithm.

jamesmcclain:
See jamesmcclain/CostDistance#11 and geotrellis/geotrellis-libya-weighted-overlay#4

spark/src/main/scala/geotrellis/spark/costdistance/IterativeCostDistance.scala
jamesmcclain: I think that there would be some mechanical changes needed in order to support an RDD of points, probably the friction layer and the points would need to be `join`ed (either logically or literally) as the first step.

As a practical matter, I do not think that supporting RDDs of points would add any real capability.  If someone really did have so many points that they do not fit into memory on one machine *and* some appreciable percentage of those points have unique projections into the raster, then it is likely that the raster layer is so large (in terms of pixels) that this algorithm would run out of driver memory before even reaching the second iteration.

raster/src/main/scala/geotrellis/raster/costdistance/CostDistance.scala
jamesmcclain: I have not run any benchmarks, but (informally) I did not detect any noticeable speed difference.

That having been said, it is likely that there is one, because the changes that I made were very self-consciously **deoptimizations**.

That was necessary because the previous algorithm seemed to avoid putting elements into the priority queue whenever it could.  That was laudable in the single-threaded case, but in order to maintain coherence in the case were points need to be transferred from some adjacent tile to the present one, I think that it makes sense to have one and only one place for points to enter (to ensure that they are being treated exactly the same as points that originated in the present tile).

After getting the basics working, I then took another pass an re-added some mild optimizations.  Generally speaking, I am pretty comfortable with were the single-threaded version is.

pomadchin:
[Benchmarks](https://gist.github.com/pomadchin/0039560b660dce98a36d321538685699). Let me know if that's not enough and i need to test more, but i think it is already obvious that it's significantly slower.

jamesmcclain:
No, I'll just pull the original one out of version control and use that for the single-tile case.

raster/src/main/scala/geotrellis/raster/costdistance/SimpleCostDistance.scala
```diff
@@ -0,0 +1,204 @@
...
+    points: Seq[(Int, Int)],
+    maxCost: Double = Double.PositiveInfinity,
+    resolution: Double = 1
+  ): DoubleArrayTile = {
+    val cols = frictionTile.cols
+    val rows = frictionTile.rows
+    val costTile = generateEmptyCostTile(cols, rows)
+    val q: Q = generateEmptyQueue(cols, rows)
+
+    var i = 0; while (i < points.length) {
```
pomadchin: Why don't to use just `cfor`? It's a common dep for the whole raster package. Just less ugly ^^'.

spark/src/main/scala/geotrellis/spark/costdistance/IterativeCostDistance.scala
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  private def geometryToKeys[K: (? => SpatialKey)](
+    md: TileLayerMetadata[K],
+    g: Geometry
+  ) = {
+    val keys = mutable.ArrayBuffer.empty[SpatialKey]
+    val bounds = md.layout.mapTransform(g.envelope)
+
+    var row = bounds.rowMin; while (row <= bounds.rowMax) {
```
pomadchin: Here `cfor` can be used too ^^'; i guess you just don't like it? :D

spark/src/main/scala/geotrellis/spark/costdistance/IterativeCostDistance.scala
```diff
@@ -0,0 +1,257 @@
...
+  class ChangesAccumulator extends AccumulatorV2[KeyCostPair, Changes] {
+    private val list: Changes = mutable.ArrayBuffer.empty
+
+    def copy: ChangesAccumulator = {
+      val other = new ChangesAccumulator
+      other.merge(this)
+      other
+    }
+    def add(pair: KeyCostPair): Unit = {
+      this.synchronized { list.append(pair) }
```
pomadchin: I totally forgot about it, but @echeipesh noticed these locks. @jamesmcclain, have you considered any java concurrent collection usage here?

jamesmcclain: The most appropriate structure that I found was `CopyOnWriteArrayList`.  The documentation for that class says the following:

> This is ordinarily too costly, but may be more efficient than alternatives when traversal operations vastly outnumber mutations, and is useful when you cannot or don't want to synchronize traversals, yet need to preclude interference among concurrent threads. The "snapshot" style iterator method uses a reference to the state of the array at the point that the iterator was created. This array never changes during the lifetime of the iterator, so interference is impossible and the iterator is guaranteed not to throw ConcurrentModificationException. The iterator will not reflect additions, removals, or changes to the list since the iterator was created. Element-changing operations on iterators themselves (remove, set, and add) are not supported. These methods throw UnsupportedOperationException.

I do not think that a synchronized data structure is warranted here either since there is only one place for new data to go (the end of the list).

PR #2001, opened 2017-02-06
Add AsyncIngest for sinks returning a Future

kthompson:
Currently, when we are ingesting items we typically want them to be able to write in parallel or close to parallel from within the sink. The changes within the PR allow us to be able to continue creating the Pyramid while writing layers to our layer store increasing performance of the task.

The code is mostly copied from the existing Ingest code and allows the sink method to return a Future and make sure that the RDD is resisted until the Future has completed. Then we return a big future which is comprised of all the sink Futures, which allows us to handle the completion of all of the futures.

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/ingest/AsyncIngest.scala
+++ b/spark/src/main/scala/geotrellis/spark/ingest/AsyncIngest.scala
@@ -1,0 +1,111 @@
+package geotrellis.spark.ingest
+
+import geotrellis.proj4._
+import geotrellis.raster._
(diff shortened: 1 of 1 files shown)


echeipesh:
@kthompson Interesting idea, thank you for the PR. While you can't really parallelize writing of RDDs through use of `Future`s this would certainly parallelize on writing layer metadata for reach level. I'm curious what kind of speedup you have seen on ingests using this method. I imagine it would have significant impact on smaller jobs?

echeipesh:
This seems like its gone cold for a bit. Closing until interests warm up.

PR #2007, opened 2017-02-09
Simplification of Delaunay triangulations

jpolchlo:
This PR aims to allow for detail-preserving simplification of Delaunay triangulations through vertex removal.  In this process, each vertex is "speculatively removed", generating a collection of triangles that would replace the fan of triangles incident to the target vertex.  This neighborhood is most likely "flatter" than the triangles it replaces.  To assess the error, each face in the set of replacement triangles generates an error quadric--a matrix which, when applied in a quadratic form with the vector representation of a point, describes the squared distance of that point to the plane containing the triangle.  All the triangles that would be created by removing point `p` have their quadrics summed (giving a sum of squared distances from a point to the planes of the new triangles) to produce an error metric which gives a means to determine the "badness" induced by removing vertex `p`.  Assessing this badness for each vertex, we remove the least bad vertex, and proceed in that fashion until the desired number of vertices are removed.

 - [X] Delaunay-preserving vertex removal
 - [x] Generation of quadric error matrices
 - [x] Per-vertex scoring w.r.t. local quadrics
 - [x] Iterative vertex removal
 - [x] Testing and debugging
 - [x] Clean up


pointcloud/build.sbt
```diff
@@ -23,3 +23,5 @@ fork in Test := true
 parallelExecution in Test := false
 
 javaOptions += s"-Djava.library.path=${Environment.ldLibraryPath}"
+
+testOptions in Test += Tests.Argument("-oDF")
```
pomadchin: Probably we need to throw it away; as it prints complete stack traces (potentially not good for travis)

pomadchin:
Merging, further improvements / bug fixes can be done in a separate PR.

Issue #1991, opened 2017-01-26
Extract `vectortile` subproject


fosskers:
In an attempt to better define what should be considered core GeoTrellis and what shouldn't, it became clear that the `vectortile` subproject ought to be moved to its own repo.  

It relies on a few core GT types, but otherwise is closer to "something that uses GeoTrellis" as opposed to "GeoTrellis" itself. 

However, removing it raises the question of how to handle its Avro codecs, which currently live in `geotrellis-spark`. They originally lived in `vectortile` itself, but were moved to avoid the heavy dependency. Note that the codecs for `vector` and `raster` live in `spark` in the same way. If `vectortile` is moved to its own repo, where is the best place for those avro codecs?

metasim:
If you put the Avro codecs in the companion object for the target types then you shouldn't (?) have backwards compatibility issues. There are arguments against doing this with respect to separation of concerns, API pollution, etc. but if this new library is fairly focused on one thing, maybe it's not a big deal.

I now note that `AvroRecordCodec` is in the spark module, which makes the approach above kinda suck since you may want to use vectortile without the spark baggage. Perhaps `AvroRecordCodec` should live somewhere else?

fosskers:
That was the original thought, yeah. `vectortile` itself has nothing to do with Spark, but still needs the baggage to get at the Codecs.

metasim:
How many other "codec patterns" are there in GT? JSON? Protobuf? etc.? Wondering if there's a way to pull `AvroRecordCodec` and friends out of `spark` without a) over-engineering via sub-project explosion, and b) dragging along unwanted dependencies when Avro isn't used. A new `geotrellis-io` sub-project smells like "a". The `util` sub-project would be a good, but for "b".

fosskers:
Is there anything necessarily sparky about the Avro codecs? If not, I don't see why they couldn't be factored out.

lossyrob:
Factoring them out makes sense, but contributes to the subproject explosion. However, if there was a `geotrellis-avro` or something of the like, it would be convenient - seeing how we're going to end up factoring out vectortile and pointcloud, perhaps we shouldn't be as afraid of teh subproject explosion as we have in the past.

fosskers:
The PR associated with this was closed, so this should be too. VT is to remain in core GeoTrellis.

PR #2009, opened 2017-02-10
Remove `vectortile` from Core GeoTrellis

fosskers:
### TODO
- [x] Move code into new repository
- [x] Remove mentions of `vectortile` from config/README
- [x] Fix Travis

### Motivation
We recently decided that the `vectortile` subproject is not core GeoTrellis per se, but is something built with it. As such, it should be moved out of this repo to help better define the edges of Core GeoTrellis.

See #1991 .


fosskers:
This will have to be redone if I make changes to `vectortile` while working on `vectorpipe`.

pomadchin:
@fosskers would it be hard to bump up scalapb dp? https://github.com/locationtech/geotrellis/blob/master/vectortile/build.sbt#L7

fosskers:
Not sure, I'll have to check. Is it messing things up?

pomadchin:
Yep, looks like api changes on minor upgrade ._.

fosskers:
#2038

echeipesh:
After some internal discussion `vectortile` package is going to remain in geotrellis project because it is enough of a core datatype that is related to the way we work with geospatial data in GeoTrellis.

Issue #1967, opened 2017-01-18
Links from search via "Read the Docs" result in "page does not exist" error.


metasim:
To reproduce:

1. Go to [GeoTrellis Manual](http://geotrellis.readthedocs.io/en/latest/)
2. Enter a search term:  
![screen shot 2017-01-18 at 10 11 56 am](https://cloud.githubusercontent.com/assets/131013/22069524/c103e416-dd66-11e6-992d-880cb3266766.png)
3. Click one of the result links:  
![screen shot 2017-01-18 at 10 12 14 am](https://cloud.githubusercontent.com/assets/131013/22069562/de8e6e84-dd66-11e6-95cc-34d01a9d06c2.png)
4. Admire the error message:  
![screen shot 2017-01-18 at 10 12 27 am](https://cloud.githubusercontent.com/assets/131013/22069595/f568dc7a-dd66-11e6-912e-a4bc40382e8f.png)

fosskers:
While we're quite sure "why" yet, we've noticed that URLs generated by RTDs search function produce old links. If you search for the same content shown in the search previews, they exist at different locations in the real rendered docs.

fosskers:
This *could* be due to the fact that the `1.0` and `stable` branches on RTD are quite a bit behind `latest`. Investigating now.

fosskers:
@metasim this is completely fixed now.

PR #2016, opened 2017-02-14
Port Docs to `.rst`

fosskers:
### TODO
- [x] Mass conversion via `pandoc`
- [x] Fix markup errors
- [x] Enable nested TOCs
- [x] Test fix for #1967 

### Motivation
#1967 . It turns out that RTD search just *doesn't* work for projects based on `mkdocs`, which we currently are.


fosskers:
RTD `latest` now points to this repo's `master` again, as it originally did. Being able to point latest at different branches is handy for testing. The setting ("default branch") can be found in RTD's `Admin` panel, under `Advanced Settings`.

PR #2017, opened 2017-02-16
Loggers serialization safety

pomadchin:
Even though loggers are serializable, we have to be sure that there would be no problems during multiple ser / deser. Spark internal loggers all marked with `@transient`, all Hadoop loggers are `final static`. There is still no unit tests related to that issue, though you can notice it during Spark memory shuffle and using `DISK_ONLY` or `MEMORY_AND_DISK` RDD persistence levels.

### DIFF
--- a/accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerUpdater.scala
+++ b/accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerUpdater.scala
@@ -23,7 +23,6 @@ import geotrellis.spark.io.avro.codecs._
 import geotrellis.spark.merge._
 import geotrellis.util._
 
-import com.typesafe.scalalogging.LazyLogging
 import org.apache.spark.SparkContext
 import org.apache.spark.rdd.RDD
 import spray.json._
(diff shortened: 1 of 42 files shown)


hjaekel:
We tested your patch with our code and the `NullPointerException` is gone, everything works fine now.

Just for the documentation, this was the exception we get when GeoTrellis is using the `com.typesafe.scalalogging.LazyLogging`:

    java.lang.NullPointerException
            at org.slf4j.impl.Log4jLoggerAdapter.isWarnEnabled(Log4jLoggerAdapter.java:391)
            at geotrellis.raster.ArrayTile$class.convert(ArrayTile.scala:46)
            at geotrellis.raster.UShortArrayTile.convert(UShortArrayTile.scala:24)

We could not reproduce this issue in unit tests, too. The NPEs just occur while executing bigger processes with limited memory.

lossyrob:
Thanks for looking into that @hjaekel.

👍

echeipesh:
I don't think this is a breaking change or even api-change. We replaced a protected field with a protected field.

And addition of `@transient` does not break binary compatibility: https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#jls-13.4.11

PR #2018, opened 2017-02-16
Bump slick-pg to 1.14.6

echeipesh:
Also exclude explicit postgresql dependency. slick-pg directly depends on `"org.postgresql" % "postgresql" % "9.4-1201-jdbc41"`. 

Ideally `slick-pg` would also reduce the `postgresql` to compile time dependency, which is not the case currently.


echeipesh:
Superseded by #2020

PR #2023, opened 2017-02-18
Docs: Core Concepts entry for ingest Catalogs

fosskers:
Users should understand the layout and usage of our ingest Catalogs, and this new section should help with that.


docs/guide/core-concepts.rst
```diff
@@ -980,6 +980,147 @@ that call ``geotrellis.vector`` home:
 -  ``geotrellis.vector.reproject`` defines methods for translating
    between projections
 
+Catalogs
+========
+
+We call the output of any ingest a *catalog*. It is a collection of
```
echeipesh: The output of an ingest is a layer. I think when we ingest multiple layers into the same location using the same attribute store we start to call this collection a catalog, but mostly in the vernacular rather than in code.

This is conscious choice because we found that the idea of "catalog" is fairly nebulous from the standpoint of interface despite being intuitive. So we stuck to making API about layers, making them, moving them, reading them, deleting them.

We use the word "catalog" in configs and output directors fairly often but it is important to make this distinction for the sake of understanding the API

docs/guide/core-concepts.rst
```diff
@@ -980,6 +980,147 @@ that call ``geotrellis.vector`` home:
...
+       ├── 5
+       ├── 6
+       ├── 7
+       ├── 8
+       └── 9
+
+   16 directories, 14 files
+
+The children of ``landsat`` are directories, but we used ``-L 2`` to hide
+their contents. They actually contain thousands of ``Tile`` files.
```
echeipesh: Each file contains `Array[(K, V)]`, common case is `Array[(SpatialKey, Tile)]`. Probably the most salient point here is that it could be a couple of tiles and they have their layout keys with them. This is a product of SFC encoding the keys btw.

Although in this case, encoding SpatialKeys they will be arrays of a single tuple. hard to say if the additional information here is illuminating or confusing.

docs/guide/core-concepts.rst
```diff
@@ -980,6 +980,147 @@ that call ``geotrellis.vector`` home:
...
+               "minKey": { "col": 56, "row": 24 },
+               "maxKey": { "col": 57, "row": 25 }
+             }
+           }
+         },
+         ... // more here
+       }
+     ]
+
+Of note is the ``header`` block, which tells GeoTrellis where to look for
```
echeipesh: Worth while to explicitly call out that it varies by method of storage "s3" "hdfs" etc

fosskers:
I triggered a new build of the docs on RTD, so these changes are now live.

PR #2024, opened 2017-02-20
Remove -Ymacro-expand:none flag

pomadchin:


lossyrob:
What's going on with this change?

pomadchin:
@lossyrob with this flag updated unidocs plugin can't access lenses in raster package: https://travis-ci.org/locationtech/geotrellis/jobs/203502868 
If it's quite important we can rollback unidocs plugin back.

lossyrob:
Isn't important, since it's just in the aggregate project. Just deserved a little bit of explaining :)

pomadchin:
@lossyrob ufff :D I was afraid i broke smth i could not catch, as I checked multiple times my local generated docs after that change and could not find difference.

Issue #2021, opened 2017-02-17
Pyramid.up throws `geotrellis.raster.GeoAttrsError: invalid cols: 0`


echeipesh:
This has happened when `Pyramid.up` is used on `RDD[(K, V)] with Metadata[M]` where the `TileLayout` in `Matadata[M]` is not compatible with the `layoutScheme`.

Typically this is the case when the `rdd` has been produced using [`FloatingLayoutScheme`](https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/tiling/FloatingLayoutScheme.scala) but `layoutScheme` is [`ZoomedLayoutScheme`](https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/tiling/ZoomedLayoutScheme.scala).

What usually happens is that the source `TileLayout` does not have enough columns or rows in the layout and at some point during the pyramid process the [division by 2](https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/tiling/ZoomedLayoutScheme.scala#L124) produces `0` for number of columns/rows in the next tile layout.

There should be a way for `ZoomedLayoutScheme` to throw early and meaningfully when it is used with inappropriate source tile layout.

There also needs to be some way to pyramid a layer in floating layout scheme. Perhaps there can be common sense implementations for [these methods](https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/tiling/FloatingLayoutScheme.scala#L41) that pyramid until they reach tile layout with 1 tile.

PR #2027, opened 2017-02-22
Minor issues fixes

### LINKED ISSUE #2012: Publish `geotrellis.spark.TestEnvironment`
When doing Spark tests in my own code it would be nice to use the same test environment setup that GeoTrellis uses for Spark, as defined by `geotrellis.spark.TestEnvironment`. There are at least two ways to do this:

1. Publish the `"org.locationtech.geotrellis" %% "geotrellis-spark" % "1.0.0" % "test"` artifact as a part of the standard release process. This should be as easy as adding `publishArtifact in (Test, packageBin) := true` to the sub-project.
2. Refactor `geotrellis.spark.TestEnvironment` so that it can live in `geotrellis-spark-testkit`, which does get published (I think).

### LINKED ISSUE #2022: renderPng(colorMap) treats Int rasters as Double
https://github.com/locationtech/geotrellis/blob/a00c35b928e96083188d91734252d66574e3b4a7/raster/src/main/scala/geotrellis/raster/render/PngRenderMethods.scala#L53-L62

### LINKED ISSUE #2026: Degenerate TileLayouts
Currently, the following are possible:

```scala
scala> TileLayout(0, 0, 10, 10)
res0: geotrellis.raster.TileLayout = TileLayout(0,0,10,10)

scala> TileLayout(-5, -5, 10, 10)
res1: geotrellis.raster.TileLayout = TileLayout(-5,-5,10,10)
```
where the results are degenerate, nonsensically sized `TileLayout`s. Changing this to make it impossible would mean a guarantee of sanity in code that expects positively-sized `TileLayout`s, with the disadvantage of bounds-checking every time one is created. 

In the absence of unsigned integer types on the JVM (the proper fix), I suppose the only approach to refusing negative values would be to throw an Exception?

pomadchin:
Resolves #2022
Resolves #2012
Resolves #2026

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/TileLayout.scala
+++ b/raster/src/main/scala/geotrellis/raster/TileLayout.scala
@@ -46,6 +46,13 @@ object TileLayout {
  * @param tileRows   number of pixel rows in each tile, North to South
  */
 case class TileLayout(layoutCols: Int, layoutRows: Int, tileCols: Int, tileRows: Int) {
+
+  assert(
+    layoutCols >= 0 && layoutRows >= 0 && tileCols >= 0 && tileRows >= 0,
+    s"TileLayout should contain cols and rows >= 0: " +
+    s"TileLayout(layoutCols = $layoutCols, layoutRows = $layoutRows, tileCols = $tileCols, tileRows = $tileRows)"
+  )
+
   def isTiled = layoutCols > 1 || layoutRows > 1
 
   /**
--- a/raster/src/main/scala/geotrellis/raster/TileLayout.scala
+++ b/raster/src/main/scala/geotrellis/raster/TileLayout.scala
@@ -47,10 +47,9 @@ object TileLayout {
  */
 case class TileLayout(layoutCols: Int, layoutRows: Int, tileCols: Int, tileRows: Int) {
 
-  assert(
+  require(
     layoutCols >= 0 && layoutRows >= 0 && tileCols >= 0 && tileRows >= 0,
-    s"TileLayout should contain cols and rows >= 0: " +
-    s"TileLayout(layoutCols = $layoutCols, layoutRows = $layoutRows, tileCols = $tileCols, tileRows = $tileRows)"
+    s"TileLayout should contain cols and rows >= 0: $toString"
   )
 
   def isTiled = layoutCols > 1 || layoutRows > 1
--- a/raster/src/main/scala/geotrellis/raster/TileLayout.scala
+++ b/raster/src/main/scala/geotrellis/raster/TileLayout.scala
@@ -48,8 +48,8 @@ object TileLayout {
 case class TileLayout(layoutCols: Int, layoutRows: Int, tileCols: Int, tileRows: Int) {
 
   require(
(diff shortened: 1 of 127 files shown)


raster/src/main/scala/geotrellis/raster/TileLayout.scala
```diff
@@ -46,6 +46,13 @@ object TileLayout {
  * @param tileRows   number of pixel rows in each tile, North to South
  */
 case class TileLayout(layoutCols: Int, layoutRows: Int, tileCols: Int, tileRows: Int) {
+
+  assert(
```
echeipesh: should use `require` in these cases, that gives more relevant exception.

raster/src/main/scala/geotrellis/raster/TileLayout.scala
```diff
@@ -46,6 +46,13 @@ object TileLayout {
...
+    layoutCols >= 0 && layoutRows >= 0 && tileCols >= 0 && tileRows >= 0,
+    s"TileLayout should contain cols and rows >= 0: " +
+    s"TileLayout(layoutCols = $layoutCols, layoutRows = $layoutRows, tileCols = $tileCols, tileRows = $tileRows)"
```
echeipesh: I think its fine to use `.toString` in these cases even though it doesn't have field labels its a little more tenable as a pattern.

pomadchin:
#2021 work would be done in a separate PR; only descriptive failure messages were introduced in this PR at least to reduce pain.

spark/src/main/scala/geotrellis/spark/tiling/ZoomedLayoutScheme.scala
```diff
@@ -116,15 +116,23 @@ class ZoomedLayoutScheme(val crs: CRS, val tileSize: Int, resolutionThreshold: D
 
   def zoomOut(level: LayoutLevel) = {
     val layout = level.layout
+
+    if(layout.layoutCols < 2 || layout.layoutRows < 2)
+      sys.error(
```
echeipesh: This is too late to give this exception. At this point a couple of pyramid levels would have already been saved and all that work would have been wasted.

Its possible to give this exception before we start the pyramid but now the nature of this error depends on work in the broken out PR. I would say lets remove this exception here and deal with the issue fully in separate PR.

raster/src/main/scala/geotrellis/raster/TileLayout.scala
```diff
@@ -46,6 +46,12 @@ object TileLayout {
  * @param tileRows   number of pixel rows in each tile, North to South
  */
 case class TileLayout(layoutCols: Int, layoutRows: Int, tileCols: Int, tileRows: Int) {
+
+  require(
+    layoutCols >= 0 && layoutRows >= 0 && tileCols >= 0 && tileRows >= 0,
```
echeipesh: `>` not `>=`, need at least one pixel or a tile in the tile/layout.

echeipesh:
This seems fine for `1.1`, moving `TestEnvironment` from test projects to `testkit` is addition, other stuff changes implementation.

PR #2029, opened 2017-02-23
Changed ZoomedLayoutScheme zooming functions

pomadchin:
Closes https://github.com/locationtech/geotrellis/issues/2021


spark/src/main/scala/geotrellis/spark/tiling/ZoomedLayoutScheme.scala
```diff
@@ -116,13 +116,15 @@ class ZoomedLayoutScheme(val crs: CRS, val tileSize: Int, resolutionThreshold: D
 
   def zoomOut(level: LayoutLevel) = {
     val layout = level.layout
+    val newZoom = level.zoom - 1
+    val newSize = math.pow(2, newZoom).toInt
```
echeipesh: Does this change allow to correctly pyramid a floating layer into a zoomed layout scheme? I think so, but there should be a small unit test for that.

If all of that works out fine I'm not sure there is any reason to change `FloatingLayoutScheme` as its primary purpose it to produce a layer with cell width matching the source rasters.

pomadchin: @echeipesh what do you mean by that? as floating layout scheme has no `def levelFor(extent: Extent, cellSize: CellSize)` function, I am not sure how to identify max zoom level for the floating layout scheme

pomadchin: For zoomed layout scheme i think we can create a new, called `Square`, +that would mean a good unit test for it :)

echeipesh: Not sure we need to keep the /2 *2 functionality, that was just poor implementation I think. The way this scheme is intended to be used is with square, power of two, layouts. The fact that it may work under some conditions for non-power of two is incidental and would not have produce a usably correct result if it completed. We're just fixing and testing an implementation here right now.

spark/src/test/scala/geotrellis/spark/pyramid/PyramidSpec.scala
```diff
@@ -129,5 +129,79 @@ class PyramidSpec extends FunSpec with Matchers with TestEnvironment {
...
+      val rdd =
+        createSpaceTimeTileLayerRDD(
+          Seq( (tile1, dt1), (tile2, dt2) ),
+          tileLayout
+        )
+
+      val layoutScheme = ZoomedLayoutScheme(LatLng, 2)
+
+      val LayoutLevel(_, layout) = FloatingLayoutScheme(512).levelFor(LatLng.worldExtent, CellSize(0.5, 0.5))
+      val LayoutLevel(zoom, _) = layoutScheme.levelForZoom(LatLng.worldExtent, 2)
```
echeipesh: Whats the point of this line ? `ZoomedLayoutScheme.levelForZoom` will always just give you the same zoomed you asked for, its the layout that is the real output of this function.

pomadchin: 🤦‍♂️

PR #2030, opened 2017-02-27
A dash of API sugar to simplify summoning of implicit codec.

metasim:
Allows `AvroRecordCodec[Tile]` instead of `implicitly[AvroRecordCodec[Tile]]`, a pattern used a lot in the avro2spark PoC.

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/io/avro/AvroRecordCodec.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/avro/AvroRecordCodec.scala
@@ -37,3 +37,7 @@ abstract class AvroRecordCodec[T: ClassTag] extends AvroCodec[T, GenericRecord] 
     implicitly[ClassTag[T]].unapply(other).isDefined
   }
 }
+
+object AvroRecordCodec {
+  def apply[T: AvroRecordCodec]: AvroRecordCodec[T] = implicitly
+}


pomadchin:
I like that idea, but mb to have a better name for it? like `AvroRecordCodec.derive` instead of `apply`. (yep, i know that such version looks more pretty, but probably we can add semantics there?)

metasim:
I see it more like a type-level factory method. Like the `apply` on a companion object for a case class. I want a `AvroRecordCodec[Tile]`, so I just ask the companion object for it, just like I would if I were dealing with a case class. So I think `apply` is appropriate. I can't remember where I first saw this technique, but was reminded of it [here](https://gist.github.com/skorikov/2ffd25d85ee7f28087fd7a05bd2ff506#file-typeclass-scala-L8), described in [this article](https://blog.codecentric.de/en/2017/02/ad-hoc-polymorphism-scala-mere-mortals/)

lossyrob:
In an interesting play on words, it feels like `implicitly` is more explicit than having an object apply do the work here: it hides a bit of the magic in an already magical function.

That being said, it does do the work of an object apply; it creates the thing that is the type of the companion object. If it's an established pattern that wouldn't throw people off, I'd say this change is cool. Should it be applied more broadly?

echeipesh:
It's consistent with object apply for sure. 👍 to making this code-base standard. Won't be able to apply to things outside our control like `JsonFormat`, but :man_shrugging: there.

fosskers:
Would using this elsewhere mean floods of:
```scala
object Foo {
￼   def apply[T: AvroRecordCodec]: AvroRecordCodec[T] = implicitly
￼ }
```

fosskers:
Apologies, pressed the wrong button.

metasim:
FWIW, it's also the pattern/convention used by [avro4s](https://github.com/sksamuel/avro4s). Just to be clear, I'm primarily interested in it just for this class; no stake in whether it's used more broadly.

PR #2031, opened 2017-02-28
Tile columns cannot be zero when rounding

snoe925:
These round to zero values cause zero column tiles in a LatLng to WebMercator reproject.  I could not figure out the root cause.  It looked to me like reasonable values were used, and that rounding was the issue.  An alternative fix might be to clip the data to the maximum extent of WebMercator in the projection.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/GridExtent.scala
+++ b/raster/src/main/scala/geotrellis/raster/GridExtent.scala
@@ -41,8 +41,8 @@ class GridExtent(val extent: Extent, val cellwidth: Double, val cellheight: Doub
     *       types.
     */
   def toRasterExtent(): RasterExtent = {
-    val targetCols = 1L.max(math.round(extent.width / cellwidth).toLong)
-    val targetRows = 1L.max(math.round(extent.height / cellheight).toLong)
+    val targetCols = math.max(1L, math.round(extent.width / cellwidth).toLong)
+    val targetRows = math.max(1L, math.round(extent.height / cellheight).toLong)
     if(targetCols > Int.MaxValue) {
       throw new GeoAttrsError(s"Cannot convert GridExtent into a RasterExtent: number of columns exceeds maximum integer value ($targetCols > ${Int.MaxValue})")
     }
--- a/raster/src/main/scala/geotrellis/raster/GridExtent.scala
+++ b/raster/src/main/scala/geotrellis/raster/GridExtent.scala
@@ -41,8 +41,8 @@ class GridExtent(val extent: Extent, val cellwidth: Double, val cellheight: Doub
     *       types.
     */
   def toRasterExtent(): RasterExtent = {
-    val targetCols = math.round(extent.width / cellwidth).toLong
-    val targetRows = math.round(extent.height / cellheight).toLong
+    val targetCols = 1L.max(math.round(extent.width / cellwidth).toLong)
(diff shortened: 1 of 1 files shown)


raster/src/main/scala/geotrellis/raster/GridExtent.scala
```diff
@@ -41,8 +41,8 @@ class GridExtent(val extent: Extent, val cellwidth: Double, val cellheight: Doub
     *       types.
     */
   def toRasterExtent(): RasterExtent = {
-    val targetCols = math.round(extent.width / cellwidth).toLong
-    val targetRows = math.round(extent.height / cellheight).toLong
+    val targetCols = 1L.max(math.round(extent.width / cellwidth).toLong)
+    val targetRows = 1L.max(math.round(extent.height / cellheight).toLong)
```
lossyrob: Thanks for the contribution!

This is a nitpick, but I think using `math.max` directly would be more optimal.

The `1L.max` is a method implicitly added via [RichLong](https://github.com/scala/scala/blob/05016d9035ab9b1c866bd9f12fdd0491f1ea0cbb/src/library/scala/runtime/RichLong.scala) (implicit [here](https://github.com/scala/scala/blob/05016d9035ab9b1c866bd9f12fdd0491f1ea0cbb/src/library/scala/Predef.scala#L593)), which extends [ScalaNumberProxy](https://github.com/scala/scala/blob/05016d9035ab9b1c866bd9f12fdd0491f1ea0cbb/src/library/scala/runtime/ScalaNumberProxy.scala#L24) via IntegralProxy[Long], which is an unspecialized class. So, I would assume this would box. However, looking at the byte code, I don't see that it does:

Scala:
```scala
  def withImplicit: Long =
    1L.max(2L)

  def withoutImplicit: Long =
    math.max(1L, 2L)
```
bytecode:
```
  public long withImplicit();
    Code:
       0: getstatic     #66                 // Field scala/runtime/RichLong$.MODULE$:Lscala/runtime/RichLong$;
       3: getstatic     #71                 // Field scala/Predef$.MODULE$:Lscala/Predef$;
       6: lconst_1
       7: invokevirtual #75                 // Method scala/Predef$.longWrapper:(J)J
      10: ldc2_w        #76                 // long 2l
      13: invokevirtual #81                 // Method scala/runtime/RichLong$.max$extension:(JJ)J
      16: lreturn

  public long withoutImplicit();
    Code:
       0: getstatic     #87                 // Field scala/math/package$.MODULE$:Lscala/math/package$;
       3: lconst_1
       4: ldc2_w        #76                 // long 2l
       7: invokevirtual #90                 // Method scala/math/package$.max:(JJ)J
      10: lreturn
```

Not sure how this actually doesn't box...(sorry 'thinking out loud' here)

However, we do pay the price of the `longWrapper` call, which instantiates a class. So, the `math.max` direct call would be faster than the `1l.max` call. By how much? Surely not a lot. But given one of two options, I'll always choose the faster one (and the one that produces less byte code) :)

Can you switch to that method?

snoe925: I will change it.

snoe925: The Eclipse CLA is signed as snoe925@gmail.com

lossyrob:
:+1:

PR #2034, opened 2017-02-28
Fix/2033

metasim:


lossyrob:
I'm +1 on this, @echeipesh your review was requested.

spark/src/main/scala/geotrellis/spark/io/avro/codecs/TemporalProjectedExtentCodec.scala
```diff
@@ -33,20 +33,34 @@ trait TemporalProjectedExtentCodec {
...
     def encode(temporalProjectedExtent: TemporalProjectedExtent, rec: GenericRecord): Unit = {
+      val crs = temporalProjectedExtent.crs
       rec.put("extent", extentCodec.encode(temporalProjectedExtent.extent))
-      rec.put("epsg", temporalProjectedExtent.crs.epsgCode.get)
+      if(crs.epsgCode.isDefined) {
+        rec.put("epsg", crs.epsgCode.get)
+      }
+      else {
+        rec.put("proj4", crs.toProj4String)
+      }
```
metasim: Another (breaking) option would be to replace `epsg` all together with `crs`, and prefix EPSG codes according [to this](https://github.com/locationtech/geotrellis/blob/master/proj4/src/main/scala/geotrellis/proj4/CRS.scala#L120), falling back to "proj4" strings otherwise.

lossyrob: With Avro schema migration capabilities (which allow for aliasing), this might not have to be a breaking change. See: https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/io/avro/codecs/KeyCodecs.scala#L52

pomadchin:
Think for consistency https://github.com/metasim/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/io/avro/codecs/ProjectedExtentCodec.scala should be modified too.

metasim:
@pomadchin Good catch. 

@echeipesh are you good with this?

echeipesh:
+1, nice that it was kept from being a breaking change. Shame that `ProjectedExtentCodec` can't directly use the CRS codec.

spark/src/main/scala/geotrellis/spark/io/avro/codecs/TemporalProjectedExtentCodec.scala
pomadchin: @metasim @lossyrob I'm wondering what means breaking here, as there was no this codec in `1.0.0` O:

metasim:
@echeipesh Created a shared CRS codec.

metasim:
D'oh!.... Think this should be handled some other way?

echeipesh:
@lossyrob As I'm reading this Avro schemes they fall under schema migration, so that isn't breaking: old records can be read using new versions. The new `CRSCodec` class is new so since we're adopting SemVer would place this change into `1.1.0`, which seems alright to me.

Probably the moral is that we'll have to release `1.1.0` soon after `1.0.1` as a lot of the PRs that have been merged in since `1.0.0` have added classes or methods.

lossyrob:
@metasim for future reference: The PR should be named describing the change, and the body of the PR should say "Fixes #XXXX" - this way they are properly linked in GitHub, and the issue actually closes once the PR is merged. This way (with Fixes comment in a commit, and issue number in PR name) took just a bit to track down, no big deal but just a heads up :)

PR #2036, opened 2017-03-01
Update build to work with IntelliJ

metasim:
This takes care of a pseudo-cyclic dependency between the `spark` test scope, `spark-testkit`,
and `spark` main (compile) scope. sbt is happy with this. IntelliJ requires that `spark-testkit`
be added to the `spark` module dependencies manually (via "Open Module Settings" context menu for "spark" module).

Using the [technique here](https://github.com/locationtech/geotrellis/compare/master...metasim:feature/build-in-intellij?expand=1#diff-fdc3abdfd754eeb24090dbd90aeec2ceR161), it could open the door to moving `raster-test`, `s3-test`, and `vector-test` back into their parent packages. The downside would be that in IntelliJ you'd have a few more cases where you'd have to manually add the testkit as a dependency.


build.sbt
```diff
@@ -103,101 +112,125 @@ lazy val root = Project("geotrellis", file(".")).
...
+  .settings(commonSettings)
+
+lazy val spark = project
+  .dependsOn(util, vectortile, raster, `raster-testkit` % "test")
+  .settings(commonSettings)
+  .settings(
+    // This takes care of a pseudo-cyclic dependency between the `spark` test scope, `spark-testkit`,
+    // and `spark` main (compile) scope. sbt is happy with this. IntelliJ requires that `spark-testkit`
+    // be added to the `spark` module dependencies manually (via "Open Module Settings" context menu for "spark" module).
+    unmanagedClasspath in Test ++= (fullClasspath in (LocalProject("spark-testkit"), Compile)).value
```
pomadchin: Cool if it works (!)
There is a question appears what settings should be in the root `build.sbt` and what in local projects `build.sbt` files.

lossyrob:
Is there somewhere in the docs that we can add the note you make on adding testkits to IntelliJ?

build.sbt
metasim: That is a good question. I could see it going either way with this particular setting. I'm ambivalent.

pomadchin:
Hm, for me everything worked without special manipulations. Great! 
Another just code-style question: I am usually trying to avoid ``` ` ``` symbols in code; though here it looks ok? It's not significant, just wondering.

metasim:
I only use the backtick symbols so that we can use the `project` macro which fills in the `id` and `base` attributes from the variable name. I think overall makes the build moderately more readable when you have directories with dashes in them, and makes things a little less "stringly typed". Feel free to change back if you're really adverse to them.

lossyrob:
@metasim merging your other PR created a conflict, can you update to master and push? Thanks.

PR #2038, opened 2017-03-01
VectorTile: Use latest ScalaPB

fosskers:
Bumping the dep wasn't enough, the IO code had to be regenerated as well.

Note: The generated code, while marked `DO NOT EDIT!`, has to be edited to include the license headers and the correct package name, namely `geotrellis.vectortile.protobuf.internal.vector_tile`.

### DIFF
--- a/vectortile/build.sbt
+++ b/vectortile/build.sbt
@@ -4,5 +4,5 @@ name := "geotrellis-vectortile"
 
 libraryDependencies ++= Seq(
   scalatest % "test",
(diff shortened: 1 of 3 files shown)


echeipesh:
This change to generated code seemed to have been handled safely for minor release

fosskers:
Also, the autogenerated code lives in an `*.internal` package, which by definition is never user facing and avoids SemVer issues.

PR #2039, opened 2017-03-02
Relax safeguards on delete methods

moradology:
Prior to this change, `delete` methods threw in any case where the reference to a layer happened to be faulty. This is based on metadata which, though always present in case of tile data (and necessary for its deletion), might not be present. Lacking this metadata, we should log information about the failure and delete any `_attribute` data we can.

### DIFF
--- a/cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraLayerDeleter.scala
+++ b/cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraLayerDeleter.scala
@@ -18,47 +18,48 @@ package geotrellis.spark.io.cassandra
 
 import geotrellis.spark.LayerId
 import geotrellis.spark.io._
+import geotrellis.util.LazyLogging
 
 import com.datastax.driver.core.querybuilder.QueryBuilder
 import com.datastax.driver.core.querybuilder.QueryBuilder.{eq => eqs}
 
 import scala.collection.JavaConversions._
 
-class CassandraLayerDeleter(val attributeStore: AttributeStore, instance: CassandraInstance) extends LayerDeleter[LayerId] {
+class CassandraLayerDeleter(val attributeStore: AttributeStore, instance: CassandraInstance) extends LazyLogging with LayerDeleter[LayerId] {
 
   def delete(id: LayerId): Unit = {
-    if (!attributeStore.layerExists(id)) throw new LayerNotFoundError(id)
-    val header = try {
-      attributeStore.readHeader[CassandraLayerHeader](id)
-    } catch {
-      case e: AttributeNotFoundError => throw new LayerDeleteError(id).initCause(e)
-    }
+    try {
+      val header = attributeStore.readHeader[CassandraLayerHeader](id)
+      instance.withSessionDo { session =>
+        val squery = QueryBuilder.select("key")
+          .from(header.keyspace, header.tileTable).allowFiltering()
+          .where(eqs("name", id.name))
+          .and(eqs("zoom", id.zoom))
 
-    instance.withSessionDo { session =>
-      val squery = QueryBuilder.select("key")
-        .from(header.keyspace, header.tileTable).allowFiltering()
-        .where(eqs("name", id.name))
-        .and(eqs("zoom", id.zoom))
+        val dquery = QueryBuilder.delete()
+          .from(header.keyspace, header.tileTable)
+          .where(eqs("key", QueryBuilder.bindMarker()))
+          .and(eqs("name", id.name))
+          .and(eqs("zoom", id.zoom))
 
-      val dquery = QueryBuilder.delete()
-        .from(header.keyspace, header.tileTable)
-        .where(eqs("key", QueryBuilder.bindMarker()))
-        .and(eqs("name", id.name))
-        .and(eqs("zoom", id.zoom))
+        val statement = session.prepare(dquery)
 
-      val statement = session.prepare(dquery)
-
-      session.execute(squery).all().map { entry =>
-        session.execute(statement.bind(entry.getLong("key").asInstanceOf[java.lang.Long]))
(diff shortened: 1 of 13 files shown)


s3/src/main/scala/geotrellis/spark/io/s3/S3AttributeStore.scala
```diff
@@ -98,7 +98,6 @@ class S3AttributeStore(val bucket: String, val prefix: String) extends BlobLayer
       .exists(_.getKey.endsWith(s"${AttributeStore.Fields.metadata}${SEP}${layerId.name}${SEP}${layerId.zoom}.json"))
 
   def delete(layerId: LayerId, attributeName: String): Unit = {
-    if(!layerExists(layerId)) throw new LayerNotFoundError(layerId)
```
pomadchin: Why do we need to remove it here? O: 
All backends have this check, mb to make it more smart?

s3/src/main/scala/geotrellis/spark/io/s3/S3LayerDeleter.scala
```diff
@@ -18,26 +18,29 @@ package geotrellis.spark.io.s3
...
+      val header = attributeStore.readHeader[S3LayerHeader](id)
+      val bucket = header.bucket
+      val prefix = header.key
+      val s3Client = getS3Client()
+
+      s3Client.deleteListing(bucket, s3Client.listObjects(bucket, prefix))
     } catch {
-      case e: AttributeNotFoundError => throw new LayerDeleteError(id).initCause(e)
+      case e: AttributeNotFoundError =>
+        logger.info(s"Metadata for $id was not found. Any associated layer data (if any) will require manual deletion")
```
pomadchin: I believe we already had similar way of wrapping errors (`SomeException.initcause() ...`), probably you remember it; if it throws such an exception it can be handled on a client side; or we should consider changing the behaviour of _all backends_?

pomadchin:
Hm, it really can be a problem, but mb it's not a gt issue that metadata is missing; there are also cases when metadata exists but there are no files to delete which sounds more like a gt issue.

Mb the initial problem of not completely correct `LayerNotFound` error throw in case of LayerDeleters, mb it should be thrown when there is no metadata and no files related to this layer at all? 

Or to add a special behavior for all layer deleters, when layer deleter deletes whatever can be related to a certain layer and silences all _filenotfound_ exceptions? Though for me it's still can be considered as a separate thing which clears directories / tables.

s3/src/main/scala/geotrellis/spark/io/s3/S3LayerDeleter.scala
pomadchin: Btw, can this change be done as: 

```
if (!attributeStore.layerExists(id)) { 
  logger.info(s"Metadata for $id was not found. Any associated layer data (if any) will require manual deletion")
  throw new LayerNotFoundError(id)
}
```

pomadchin:
Eh i didn't understood immediately that you want to improve logging :D

s3/src/main/scala/geotrellis/spark/io/s3/S3LayerDeleter.scala
```diff
@@ -18,26 +18,29 @@ package geotrellis.spark.io.s3
...
+      case e: Exception =>
+        throw e
```
pomadchin: can be omited

s3/src/main/scala/geotrellis/spark/io/s3/S3LayerDeleter.scala
pomadchin: Obfuscation for tests: `throw new LayerDeleteError(id).initCause(e)`

spark/src/main/scala/geotrellis/spark/io/file/FileLayerDeleter.scala
```diff
@@ -20,43 +20,35 @@ import geotrellis.spark._
...
-        // Delete the elements
-        val sourceLayerPath = new File(attributeStore.catalogPath, header.path)
-        if(sourceLayerPath.exists()) {
-          sourceLayerPath
-            .listFiles()
-            .foreach(_.delete())
-
-          sourceLayerPath.delete
-        }
+class FileLayerDeleter(val attributeStore: FileAttributeStore) extends LazyLogging with LayerDeleter[LayerId] {
```
moradology: This is worth taking a close look at - I don't think `new`ing a trait inside an apply is a good idea for readability.

cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraLayerDeleter.scala
```diff
@@ -18,47 +18,48 @@ package geotrellis.spark.io.cassandra
...
 import geotrellis.spark.io._
+import geotrellis.util.LazyLogging
 
 import com.datastax.driver.core.querybuilder.QueryBuilder
 import com.datastax.driver.core.querybuilder.QueryBuilder.{eq => eqs}
 
 import scala.collection.JavaConversions._
 
-class CassandraLayerDeleter(val attributeStore: AttributeStore, instance: CassandraInstance) extends LayerDeleter[LayerId] {
+class CassandraLayerDeleter(val attributeStore: CassandraAttributeStore, instance: CassandraInstance) extends LazyLogging with LayerDeleter[LayerId] {
```
pomadchin: Here can be used a generic `AttributeStore`, only `FileLayerDeleter` has a restricted attributeStore type here (but it's a local for-tests backend).

spark/src/main/scala/geotrellis/spark/io/file/FileLayerDeleter.scala
pomadchin: what do you mean?

PR #2043, opened 2017-03-07
Additional comments for FilteredCartesianRDD

echeipesh:
Expanding documentation for FilteredCartesianRDD which relies on knowledge about spark internal implementations.

### DIFF
--- a/spark/src/main/scala/org/apache/spark/rdd/FilteredCartesianRDD.scala
+++ b/spark/src/main/scala/org/apache/spark/rdd/FilteredCartesianRDD.scala
@@ -28,13 +28,37 @@ import scala.reflect.ClassTag
 import org.apache.spark._
 import org.apache.spark.util.Utils
 
-
+/** Performs a cartesian join of two RDDs using filter and refine pattern.
+  *
+  * During RDD declaration n*m partitions will be generated, one for each possible cartesian mapping.
+  * During RDD execution summary functions will be applied in a map-side reduce to `rrd1` and `rdd2`.
+  * These results will be collected and filtered using `metapred` for partitions with potential matches.
+  * Partition pairings with possible matches will be checked using `pred` in a refinement step.
+  *
+  * No shuffle from `rdd1` or `rdd2` will be performed by the filter step,
+  * but the records of metardds, produced using the summary functions, will be shuffled (as they must be).
+  * The metardds contain one item per partition (ex: a "bounding box" of records in parent rdd),
+  * so it is assumed that this shuffle will be low cost.
+  *
+  * For efficient execution it is assumed that potential matches exist for limited number
+  * of cartesian pairings, if no filtering is possible worst case scenario is full cartesian product.
+  *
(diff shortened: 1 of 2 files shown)


spark/src/main/scala/org/apache/spark/rdd/FilteredCartesianRDD.scala
```diff
@@ -28,13 +28,33 @@ import scala.reflect.ClassTag
 import org.apache.spark._
 import org.apache.spark.util.Utils
 
-
+/** Performs a cartesian join of two RDDs using filter and refine pattern.
+  *
+  * During RDD declaration n*m partitions will be generated, one for each possible cartesian mapping.
+  * During RDD execution summary functions will be applied in a map-side reduce to `rrd1` and `rdd2`.
+  * These results will be collected and filter using `metapred` for partitions with potential matches.
```
jamesmcclain: Small typo: "... collected and filter**ed** using ..."

spark/src/main/scala/org/apache/spark/rdd/FilteredCartesianRDD.scala
```diff
@@ -28,13 +28,33 @@ import scala.reflect.ClassTag
...
+  * Partition pairings with possible matches will be checked using `pred` in a refinement step.
+  * No shuffle will be performed for partition pairings not flagged by filter step.
```
jamesmcclain: At the end of this sentence, I would just like to add 

> ... by the filter step, but of the metardds produced using the summary functions **will** be shuffled (as they must be).  The metardds contain one item per partition (a "bounding box" around the corresponding partition in the parent rdd), so it is assumed that this shuffle will be much less costly than shuffling the parent rdds.

or something like that.

spark/src/main/scala/org/apache/spark/rdd/FilteredCartesianRDD.scala
```diff
@@ -62,9 +82,20 @@ sealed class FilteredCartesianRDD[T: ClassTag, U: ClassTag, V: ClassTag](
     val part2 = currSplit.s2 // partition for (|meta)rdd2
     require(metardd1.iterator(part1, context).hasNext, "Meta-RDD Failure")
     require(metardd2.iterator(part2, context).hasNext, "Meta-RDD Failure")
+
+    /* A call to RDD.iterator invokes the block manager to compute and fetch
+     * required partition of metardd1 and metardd2. Because subsequent calls
+     * to RDD.iterator are happening within the same TaskContext the partitions
+     * of metardds will only be computed once per task execution.
+     * This is essentially a partition-wise .collect it may cause network traffic.
+     */
```
jamesmcclain: I think that you already addressed something that I wrote above with this comment.

jamesmcclain:
I made a few suggestions, but none of them are super-critical.

echeipesh:
Thank you for suggestions, integrated them and rebased.

PR #2050, opened 2017-03-10
HadoopGeoTiffRDD and S3GeoTiffRDD add overloads to use file names in keys

### LINKED ISSUE #1958: Keeping filepath in HadoopGeoTiffRDD
Hello,

It's really a pity that the filepath is dropped when reading geotiffs using:
https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopGeoTiffRDD.scala#L108

Is there any chance to update the function so we have a possibility to keep track of the files that we are processing in the rdd.  ?

Can you also add a similar function to read local files to an RDD ?

Thanks,
Loic

pomadchin:
Fixes #1958 without a significant API changes. This PR introduces a `keyTransform: (URI, I) => K` function to have ability to read necessary metadata from files paths, and to modify `Input key`.

- [x] Tests

### DIFF
--- a/s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffRDD.scala
+++ b/s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffRDD.scala
@@ -37,8 +37,6 @@ import java.nio.ByteBuffer
  * The S3GeoTiffRDD object allows for the creation of whole or windowed RDD[(K, V)]s from files on S3.
  */
 object S3GeoTiffRDD {
-  implicit def stringToUri(str: String): URI = new URI(str)
-
   final val GEOTIFF_TIME_TAG_DEFAULT = "TIFFTAG_DATETIME"
   final val GEOTIFF_TIME_FORMAT_DEFAULT = "yyyy:MM:dd HH:mm:ss"
 
@@ -124,7 +122,7 @@ object S3GeoTiffRDD {
         ).mapPartitions(
           _.map { case (key, bytes) =>
             val (k, v) = rr.readFully(ByteBuffer.wrap(bytes), options)
-            keyTransform(key, k) -> v
+            keyTransform(new URI(key), k) -> v
           },
           preservesPartitioning = true
         )
@@ -175,7 +173,7 @@ object S3GeoTiffRDD {
 
       val (k, v) = rr.readWindow(reader, pixelWindow, options)
 
-      keyTransform(objectRequest.getKey, k) -> v
+      keyTransform(new URI(objectRequest.getKey), k) -> v
     }
   }
 
--- a/s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffRDD.scala
+++ b/s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffRDD.scala
@@ -37,8 +37,6 @@ import java.nio.ByteBuffer
  * The S3GeoTiffRDD object allows for the creation of whole or windowed RDD[(K, V)]s from files on S3.
  */
 object S3GeoTiffRDD {
-  import geotrellis.spark.io.hadoop.HadoopGeoTiffRDD.keyTransformId
-
   implicit def stringToUri(str: String): URI = new URI(str)
 
   final val GEOTIFF_TIME_TAG_DEFAULT = "TIFFTAG_DATETIME"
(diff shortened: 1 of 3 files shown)


lossyrob:
This breaks API - I'd want to think through this in a way that won't break the API if possible... Is this the only way to accomplish this?

lossyrob:
Maybe if we just create the necessary overloads of public methods to maintain the signatures that existed before, this will work.

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopGeoTiffRDD.scala
```diff
@@ -74,18 +75,26 @@ object HadoopGeoTiffRDD {
...
     val conf = sc.hadoopConfiguration.withInputDirectory(path, options.tiffExtensions)
     conf
   }
 
   /**
+    * Identity key transform function
+    * @tparam K
+    * @return
+    */
+  def keyTransformId[K] = (_: URI, key: K) => key
```
lossyrob: Unclear that this is necessary, if the overloads are properly aligned.

pomadchin:
@lossyrob sure: first approach to create smth like spark did: ~ `NewHadoopGeoTiffRDD` and to deprecate the old one, or I can create functions with keyTransform different and still to keep them. P.S. all public signatures are here, except old apply.

lossyrob:
We can just overload apply to pass through the same sort of default keyTransform as the other methods.

pomadchin:
Usage example in tests and [here](https://github.com/pomadchin/geotrellis-keras-segmentation/blob/master/ingest/src/main/scala/com/azavea/ingest/Ingest.scala#L29).

lossyrob:
Is there a good spot in docs for noting this?

pomadchin:
@lossyrob eh we don't have a section for it, @fosskers can you navigate me where it would be better to write this information? Looks like it should be done as a separate PR, as it would include the whole new section with this part of API description.

fosskers:
What needs to be noted, exactly? If there was a deprecation, it should be explained in a `@deprecated` annotation on the old, overloaded method. The note will appear at compile time for the user, so that they'll know they have to switch eventually.

s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffRDD.scala
```diff
@@ -156,7 +171,9 @@ object S3GeoTiffRDD {
           StreamingByteReader(S3RangeReader(objectRequest, options.getS3Client()))
       }
 
-      rr.readWindow(reader, pixelWindow, options)
+      val (k, v) = rr.readWindow(reader, pixelWindow, options)
+
+      keyTransform(new URI(objectRequest.getKey), k) -> v
```
echeipesh: `getKey` will only give you the path without the `s3://bucket/` prefix

s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffRDD.scala
```diff
@@ -166,6 +183,48 @@ object S3GeoTiffRDD {
     *
     * @param bucket Name of the bucket on S3 where the files are kept.
     * @param prefix Prefix of all of the keys on S3 that are to be read in.
+    * @param keyTransform function to transform input key basing on the URI information.
+    */
+  def singleband[I, K](bucket: String, prefix: String, keyTransform: (URI, I) => K, options: Options)(implicit sc: SparkContext, rr: RasterReader[Options, (I, Tile)]): RDD[(K, Tile)] =
+    apply[I, K, Tile](bucket, prefix, keyTransform, options)
```
echeipesh: `keyTransform` is a little overloaded with the way we use it for SFC. This should be renamed to `uriToKey` or something similar.

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopGeoTiffRDD.scala
```diff
@@ -83,9 +84,10 @@ object HadoopGeoTiffRDD {
     * Creates a RDD[(K, V)] whose K and V depends on the type of the GeoTiff that is going to be read in.
     *
     * @param path     Hdfs GeoTiff path.
+    * @param keyTransform function to transform input key basing on the URI information.
     * @param options  An instance of [[Options]] that contains any user defined or default settings.
     */
-  def apply[K, V](path: Path, options: Options = Options.DEFAULT)(implicit sc: SparkContext, rr: RasterReader[Options, (K, V)]): RDD[(K, V)] = {
+  def apply[I, K, V](path: Path, keyTransform: (URI, I) => K, options: Options)(implicit sc: SparkContext, rr: RasterReader[Options, (I, V)]): RDD[(K, V)] = {
```
echeipesh: @lossyrob Not certain if use of `URI` here provides opportunity to use a more general function or just introduces an extra type here and should be replaced with hadoop `Path`.

PR #2055, opened 2017-03-13
Docs: Add shading information to FAQ

pomadchin:
A very popular question about Guava versions incompatibilities, this PR clarifies a bit the situation and provides a solution.

### DIFF
--- a/docs/guide/faq.rst
+++ b/docs/guide/faq.rst
@@ -158,3 +158,41 @@ This does.
 
 TL;DR: Make sure you're importing the appropriate implicits. They define
 methods that extend GeoTrellis types.
+
+How do I resolve a Guava / other library dependency incompatibility issue(s)?
+===================================
+
+Full possible exception message:
+
+.. code::
+
+    Caused by: java.lang.IllegalStateException: Detected Guava issue #1635 which indicates that a version of Guava less
+    than 16.01 is in use.  This introduces codec resolution issues and potentially other incompatibility issues in the driver.
+    Please upgrade to Guava 16.01 or later.
+
+
+GeoTrellis depends on a huge number of complex dependencies that may cause a dependency hell. One of such dependency
+is a guava library. ``GeoTrellis ETL`` and ``GeoTrellis Cassandra`` depend on ``Guava 16.01``, but Hadoop depends on ``Guava 11.0.2``
+that causes strong libs incompatibility runtime issues. When two different versions of the same library both available in a
+Spark classpath and in a fat assembly jar, Spark would use library version from its classpath.
+
+There're two possible solutions:
+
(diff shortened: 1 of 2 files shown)


fosskers:
Do you think this would be better placed in our FAQ, as opposed to the Setup? I'm concerned that users would perceive it as "GeoTrellis is hard to even get started".

pomadchin:
@fosskers I agree with you, let's move to FAQ

fosskers:
How do Guava conflicts usually manifest? I know they've happened to me before, but I feel like they don't come straight out and say "java.lang.Exception: Your guava situation is crazy, bro".

pomadchin:
@fosskers :D they happened to you more times than you even remember.

fosskers:
I know :( Runtime error heaven over here.

docs/guide/faq.rst
```diff
@@ -158,3 +158,41 @@ This does.
...
+Full possible exception message:
+
+.. code::
+
+    Caused by: java.lang.IllegalStateException: Detected Guava issue #1635 which indicates that a version of Guava less
+    than 16.01 is in use.  This introduces codec resolution issues and potentially other incompatibility issues in the driver.
+    Please upgrade to Guava 16.01 or later.
+
+
+GeoTrellis depends on a huge number of complex dependencies that may cause a dependency hell. One of such dependency
```
fosskers: `a dependency hell` -> `dependency hell`

docs/guide/faq.rst
```diff
@@ -158,3 +158,41 @@ This does.
...
+is a guava library. ``GeoTrellis ETL`` and ``GeoTrellis Cassandra`` depend on ``Guava 16.01``, but Hadoop depends on ``Guava 11.0.2``
```
fosskers: `is a guava library` -> `is the Guava library`

docs/guide/faq.rst
```diff
@@ -158,3 +158,41 @@ This does.
...
+that causes strong libs incompatibility runtime issues. When two different versions of the same library both available in a
```
fosskers: `that causes strong libs...` -> `which causes runtime issues due to library incompatibility`

`the same library both available` -> `the same library are both available`

docs/guide/faq.rst
```diff
@@ -158,3 +158,41 @@ This does.
...
+Spark classpath and in a fat assembly jar, Spark would use library version from its classpath.
```
fosskers: `in a Spark classpath` -> `in the Spark classpath`

`Spark would use` -> `Spark will use`

docs/guide/faq.rst
```diff
@@ -158,3 +158,41 @@ This does.
...
+
+There're two possible solutions:
```
fosskers: `There're` -> `There are`

PR #2056, opened 2017-03-13
Docs: Wrangling more wild markup

fosskers:
I saw these over the weekend, here's the fix.

### DIFF
--- a/docs/guide/etl.rst
+++ b/docs/guide/etl.rst
@@ -564,9 +564,10 @@ Extension
 ---------
 
 In order to provide your own input or output modules you must extend
-```InputPlugin`` <src/main/scala/geotrellis/spark/etl/InputPlugin>`__
-and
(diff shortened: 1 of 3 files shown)


docs/guide/etl.rst
```diff
@@ -564,9 +564,10 @@ Extension
 ---------
```
echeipesh: @fosskers what is going on here? Is this github flagging this section as invalid RST ?

fosskers: I've only seen that once before a year or two ago, and I'm not sure what it is. It's valid RST, because the time I saw it before it was flagging JSON. Maybe it's character encoding? Lines ending with `\n\r`?

echeipesh: I think its the JSON code block above is not closed property, with missing '}' ?

fosskers: Good catch.

fosskers: It could be a general "hey bro your markup is wrong" warning. Either way, it's fixed now.

PR #2065, opened 2017-03-14
Docs: Added 1.1 changelog

lossyrob:


lossyrob:
@fosskers why the title rename, did we start a standard for docs PRs?

fosskers:
We haven't officially, but we could if you're interested. Since the RTD move I've been labelling them like that, since changes to docs are of a different "world" than changes to code. I often search through GH repos for keywords, and when you do that with the top bar you get a listing of Issues and PRs. Nicely named ones are a resource that help us narrow down what we're looking for, while ones like "fixed stuff" are the opposite. 

Your PR here is pretty clear in what it does from the title. My adding `Docs:` here probably wasn't necessary to clarify that, but we can imagine that users less familiar with the project wouldn't know we keep our changelog in our docs.

lossyrob:
Ok, I'm fine with making it standard - you should probably talk about it somewhere (mailing list?) and then make a comment to change PRs that don't follow suit.

fosskers:
> talk about it somewhere

Cool, will do.

PR #2073, opened 2017-03-15
Added library of land use classification color maps.

metasim:


metasim:
Travis error seems to have nothing to do with the actual build. Any way to re-trigger it?

raster/src/main/scala/geotrellis/raster/render/ColorMaps.scala
```diff
@@ -0,0 +1,97 @@
+/*
+ * Copyright 2016 Azavea
```
lossyrob: Since you/your company is submitting this file, this should be copyright to you or your company.

raster/src/main/scala/geotrellis/raster/render/ColorMaps.scala
```diff
@@ -0,0 +1,97 @@
...
+    84 -> 0xA7AB9FFF,     // Fallow
+    85 -> 0xE68A2AFF,     // Urban/Recreational Grasses
+    91 -> 0xB6D8F5FF,     // Woody Wetlands
+    92 -> 0xB6D8F5FF      // Emergent Herbaceous Wetlands
+    // @formatter:on
+  ),
+    defaultOptions
+  )
+
+  private val defaultOptions = ColorMap.Options(
```
lossyrob: Why not just use ColorMap.Options.DEFAULT?

metasim: Since the `noDataColor` was the same as the `DEFAULT`, I'm removing that. However, I think the `classBoundaryType` should be `EXACT`, because operationally, you want to know if you have a classification assignment that's not part of your class set (sign of an error, or something else).

lossyrob: Fair point

PR #2075, opened 2017-03-16
Fix usage of IntCachedColorMap in Indexed PNG encoding

echeipesh:
`IntCachedColorMap` implementation was not covered by unit tests and turned out to be incorrect.
This PR adds unit tests for its usage and changes the behavior to something that works.

It looks like this class fell through some cracks in the refactors. In reality it leans on implementation of `FastMapHistogram` to convert pixels directly to colors, which uses a hash table lookup.

In effect we get `eC` color encoding instead of `Log` which is provided by `IntColorMap`

### DIFF
--- a/raster-test/src/test/scala/geotrellis/raster/render/ColorMapSpec.scala
+++ b/raster-test/src/test/scala/geotrellis/raster/render/ColorMapSpec.scala
@@ -113,6 +113,40 @@ class ColorMapSpec extends FunSpec with Matchers
       }
     }
 
+    it("should correctly map values to colors using IntCacheColorMap") {
+      val limits = Array(25,50,80,100)
+      val colors = Array(100,110,120,130)
+
+      val colorMap1 = ColorMap(limits, colors)
+      val arr = (0 until 90 by 5).toArray
+      val r = createTile(arr)
+      val h = r.histogram
+      val colorMap = colorMap1.withNoDataColor(0).withBoundaryType(LessThanOrEqualTo).asInstanceOf[IntColorMap].cache(h)
+
+      val color: IndexedPngEncoding =
+        PngColorEncoding(colorMap.colors, colorMap.options.noDataColor, colorMap.options.fallbackColor) match {
+          case i @ IndexedPngEncoding(_, _) => i
+          case _ =>
+            withClue(s"Color should be Indexed") { sys.error("") }
+        }
+
+      // check that PNG will correctly convert raster to their color index values
+      val pngMap = color.convertColorMap(colorMap)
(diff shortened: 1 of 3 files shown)


echeipesh:
@metasim This is the bug you ran into. 
Does the test cover the way you were trying to use this feature?

fosskers:
Right, I recall this not being subject to the `BreakMap` refactor.

metasim:
@echeipesh I believe so. Glad there was actually something behind it.

raster-test/src/test/scala/geotrellis/raster/render/ColorMapSpec.scala
```diff
@@ -113,6 +113,42 @@ class ColorMapSpec extends FunSpec with Matchers
...
+    it("should correctly map values to colors using IntCacheColorMap") {
+      val limits = Array(25,50,80,100)
+      val colors = Array(100,110,120,130)
+
+      val colorMap1 = ColorMap(limits, colors)
+      val arr = (0 until 90 by 5).toArray
+      val r = createTile(arr)
+      val h = r.histogram
+      val colorMap = colorMap1.withNoDataColor(0).withBoundaryType(LessThanOrEqualTo).asInstanceOf[IntColorMap].cache(h)
+
```
pomadchin: Can we avoid somehow this runtime cast?

raster/src/main/scala/geotrellis/raster/render/ColorMap.scala
```diff
@@ -228,8 +230,15 @@ class IntColorMap(breaksToColors: Map[Int, Int], val options: Options = Options.
 
   def cache(h: Histogram[Int]): ColorMap = {
     val ch = h.mutable
-    h.foreachValue(z => ch.setItem(z, map(z)))
-    new IntCachedColorMap(colors, ch, options)
+    val cachedColors = h.values()
+    cfor(0)( _ < cachedColors.length, _ + 1) { i =>
```
pomadchin: Is `h.totalCount` would be values number?

echeipesh: `h.totalCount` would give me the total number of times any value was seen bytes histogram across all buckets. Not sure if thats useful here. What are you thinking?

pomadchin: Eh, nothing smart, that's why i approved it. I tried to avoid somehow calculating length for the second time (as it's already known in fact) and trying to avoid distinct function usage. But no; it's fine as it is.

PR #2078, opened 2017-03-17
Improve HalfEdgeTable memory usage

jpolchlo:
The HalfEdgeTable was naïve in how it managed its memory.  After half edges were created and then discarded, the now-useless entry in memory was lost.  Given the number of edges that are created during Delaunay triangulation only to be deleted later, this created a substantial space leak.  This PR addresses the problem by keeping a queue of deleted edge IDs, and reuses that space as new edges are created.


vector/src/main/scala/geotrellis/vector/triangulation/HalfEdgeTable.scala
```diff
@@ -42,11 +44,16 @@ class HalfEdgeTable(_size: Int) {
...
 
+    val toKill = (getFlip(e), getFlip(opp))
+
     setNext(getPrev(getFlip(e)), getNext(getFlip(opp)))
     setNext(getPrev(getFlip(opp)), getNext(getFlip(e)))
     setFlip(e, opp)
     setFlip(opp, e)
 
+    killEdge(toKill._1)
+    killEdge(toKill._2)
```
lossyrob: Tuple = object allocation, no reason to pay that overhead here I don't think

lossyrob:
Any timing comparisons to see how this has an effect on performance?

jpolchlo:
Not to be cheeky, but it's an infinite time difference, since without this fix some code won't complete without blowing the heap. ;)  I can't at this point say if there's a significant time difference for executing two tasks that would complete.  Let me get back to you on that.

jpolchlo:
Some timings:

# Points   Time before    Time after PR
404161     10.86              10.2
672513      7.04                7.19
510490      6.09                5.85
476529      9.10                6.19
516750      5.65                10.67

So, it seems like, modulo the vagaries of system performance, it's as fast or faster after this PR.

lossyrob:
Thanks for the timings.

lossyrob:
Travis reports two test failures:

0m[[31merror[0m] [0m	geotrellis.pointcloud.spark.triangulation.BoundaryDelaunaySpec[0m
[0m[[31merror[0m] [0m	geotrellis.pointcloud.spark.triangulation.StitchedDelaunaySpec[0m

jpolchlo:
Some kind of weirdness with Travis.  Kick the build again?

vector/src/main/scala/geotrellis/vector/triangulation/DelaunayTriangulation.scala
```diff
@@ -779,8 +789,8 @@ case class DelaunayTriangulation(pointSet: DelaunayPointSet, halfEdgeTable: Half
 }
 
 object DelaunayTriangulation {
-  def apply(pointSet: DelaunayPointSet, debug: Boolean = false) = {
+  def apply(pointSet: DelaunayPointSet, tolerance: Double = 1e-8, debug: Boolean = false) = {
```
pomadchin: This change caused `geotrellis.pointcloud.spark.triangulation.BoundaryDelaunaySpec.scala` spec fail. 

`BoundaryDelaunaySpec.scala:68` change to `DelaunayTriangulation(pts, debug = false)`

pomadchin: This change caused `geotrellis.pointcloud.spark.triangulation.BoundaryDelaunaySpec.scala` 
and `geotrellis.vector.triangulation.DelaunayTriangulationSpec.scala` specs fail. 

Change `DelaunayTriangulation(pts, false)` to `DelaunayTriangulation(pts, debug = false)`

jpolchlo: Fixed by `a1f1f24`

PR #2090, opened 2017-03-24
Fix to window reads for multiband pixel interleave GeoTiffs

lossyrob:
This PR fixes windowed reads of pixel-interleaved geotiffs (as opposed to band-interleaved).

The problem was in the `SegmentTransform` (formerly called GridIndexTransform in the parent trait) : It would not take into account the band information when dealing with pixel interleave segments, which caused it to report incorrect indexes.

I included test coverage that was missing (and caused us to miss this case). I also made the segment transform types private to the geotiff package.


jbouffard:
:+1:

PR #2102, opened 2017-03-27
Rasters write support to HDFS / S3

### LINKED ISSUE #2080: Add rasters write support to HDFS / S3
Currently it's possible to save raster only on local fs.

pomadchin:
Hadoop based PNG / JPG / GeoTiff images write. Fixes #2080

- [x] tests


echeipesh:
This looks good and comment from @lossyrob are addressed.

PR #2106, opened 2017-03-29
Fix errors in StitchedDelaunay when used with BoundaryDelaunay extracts

jpolchlo:
In order to enable distributed Delaunay triangulations, we opted to allow triangulations defined over non-overlapping extents to be "stitched" together.  This requires, for each extent, the eight neighboring triangulations to be made available to the process.  Given that triangulations are actually rather heavy in terms of memory consumption, this can lead to large volumes of shuffle traffic in practice.  For example, when dealing with 3GB of pointcloud input (.LAS files), the stitch observes on the order of a terabyte of shuffle reads/writes.

This is unnecessary.  During the stitch, only a thin layer of triangles around the boundary are actually used.  We have had an implementation of the stitching algorithm using BoundaryDelaunay objects in the codebase for some time; however, the implementation of this boundary-stitching algorithm has been plagued with problems.  This PR aims to put this issue to bed.

BoundaryDelaunay construction has been affected by numerical problems which prevented correct extraction of the relevant triangles---notably involving very thin triangles on the boundary failing to be recognized as a part of the set of extracted triangles.  This patch is predominantly numerical in nature, with the main contribution being a reimplementation of the circleCenter function in the geotrellis.vector.RobustPredicates object.  I've also loosened the definition of the triangles that can be included in the boundary construction to hopefully skirt around numerically ill-conditioned cases.

### DIFF
--- a/pointcloud/src/main/scala/geotrellis/pointcloud/spark/triangulation/BoundaryDelaunay.scala
+++ b/pointcloud/src/main/scala/geotrellis/pointcloud/spark/triangulation/BoundaryDelaunay.scala
@@ -72,8 +72,8 @@ object BoundaryDelaunay {
           val predicates = new TriangulationPredicates(DelaunayPointSet(verts.toMap), halfEdgeTable)
           import predicates._
           import halfEdgeTable._
-          val (_, center) = circleCenter(getDest(e), getDest(getNext(e)), getDest(getNext(getNext(e))))
-          println(s"Circle center: $center")
+          val (_, center, valid) = circleCenter(getDest(e), getDest(getNext(e)), getDest(getNext(getNext(e))))
+          println(s"Circle center: $center " + {if (valid) "[valid]" else "[invalid]"})
           e
         })),
         't' -> (("export triangles to triangles.wkt", { e => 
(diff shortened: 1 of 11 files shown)


jpolchlo:
Currently testing the new implementation using the `geotrellis/geotrellis-pointcloud-demo`.  Will report on results when they come in.  (The included test suite currently lacks test cases of sufficient complexity to fish out the problems observed in practice, and it's unclear that extending the test suite will be all that useful.)

jpolchlo:
Still some lingering issues in the pointcloud demo.  Continuing to debug.

pointcloud/src/main/scala/geotrellis/pointcloud/spark/triangulation/BoundaryDelaunay.scala
```diff
@@ -47,16 +47,87 @@ object BoundaryDelaunay {
...
       import dt.predicates._
 
-      val center = circleCenter(getDest(tri), getDest(getNext(tri)), getDest(getNext(getNext(tri))))
-      val radius = center.distance(new Coordinate(dt.pointSet.getX(getDest(tri)), dt.pointSet.getY(getDest(tri))))
+      val (radius, center, valid) = circleCenter(getDest(tri), getDest(getNext(tri)), getDest(getNext(getNext(tri))))
       val ppd = new PointPairDistance
 
       DistanceToPoint.computeDistance(extent.toPolygon.jtsGeom, center, ppd)
-      ppd.getDistance < radius
+      !valid || ppd.getDistance < radius
```
lossyrob: Should bail on the invalid case before doing unnecessary-at-that-point computation, something like

```scala
!valid || {
  val ppd = new PointPairDistance
  DistanceToPoint.computeDistance(extent.toPolygon.jtsGeom, center, ppd)
  ppd.getDistance < radius
}
```

pointcloud/src/main/scala/geotrellis/pointcloud/spark/triangulation/BoundaryDelaunay.scala
```diff
@@ -383,17 +454,27 @@ object BoundaryDelaunay {
...
+        //navigateThis(ne)
         e = getNext(e)
         ne = halfEdgeTable.getNext(ne)
       } while (e != dt.boundary)
 
+      writeWKT("bounds.wkt")
       //isMeshValid(triangles, halfEdgeTable)
 
       fillInnerLoop
+      writeWKT("filled.wkt")
```
pomadchin: should these `writeWKT` calls be here? O: or we need to cleanup that code (as there are lots of commented out other lines), or that would be a sep issue?

jpolchlo: They shouldn't be there, no.  And there are many comments and unclean sections of code that need to be touched up.  However, we can do this when we break out the Delaunay stuff from milestone/pointcloud into its own PR against master.  That code can be taken to a finished state prior to merging.

PR #2115, opened 2017-04-03
Streaming band subsetting

lossyrob:
This PR is aimed at allowing subsets of bands to be pulled from GeoTiffs in a way that avoids reading undesired bands in the case that the GeoTiffs is stored with band interleave.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/io/geotiff/LazySegmentBytes.scala
+++ b/raster/src/main/scala/geotrellis/raster/io/geotiff/LazySegmentBytes.scala
@@ -36,21 +36,10 @@ import geotrellis.raster.io.geotiff.util._
 class LazySegmentBytes(
   byteReader: ByteReader,
   tiffTags: TiffTags,
-  maxChunkSize: Int,
-  maxOffsetBetweenChunks: Int
+  maxChunkSize: Int = 32 * 1024 * 1024,
+  maxOffsetBetweenChunks: Int = 1024
 ) extends SegmentBytes with LazyLogging {
   import LazySegmentBytes.Segment
-
-  def this(
-    byteReader: ByteReader,
-    tiffTags: TiffTags,
-    maxChunkSize: Int
-  ) = this(byteReader, tiffTags, maxChunkSize, 1024)
-
-  def this(
-    byteReader: ByteReader,
-    tiffTags: TiffTags
-  ) = this(byteReader, tiffTags, 32 * 1024 * 1024, 1024)
 
   def length: Int = tiffTags.segmentCount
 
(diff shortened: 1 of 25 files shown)


raster/src/main/scala/geotrellis/raster/io/geotiff/LazySegmentBytes.scala
```diff
@@ -29,6 +29,11 @@ class LazySegmentBytes(
 ) extends SegmentBytes with LazyLogging {
   import LazySegmentBytes.Segment
 
+  // TODO: make it a parameter in GeoTrellis 2.0
+  // max distance between two segments in a group
+  // used in a chunkSegments function
+  private val minChunkSize: Int = 1024
```
lossyrob: Is there an issue for this, or some way to track this TODO? Is it possible to use deprecation or overloads to solve this?

raster/src/main/scala/geotrellis/raster/io/geotiff/LazySegmentBytes.scala
```diff
@@ -65,7 +70,15 @@ class LazySegmentBytes(
...
       .sortBy(_.startOffset) // sort segments such that we inspect them in disk order
       .foldLeft((0l, List(List.empty[Segment]))) { case ((chunkSize, headChunk :: commitedChunks), seg) =>
-      if (chunkSize + seg.size <= maxChunkSize)
+      // difference of offsets should be <= minChunkSize
+      // otherwise everything between these offsets would be read by reader
+      // and the intention is to group segments by location and to limit groups by size
+      val notChunkByOffset =
+        headChunk.headOption.map { c =>
+          seg.startOffset - c.endOffset <= minChunkSize
+        }.getOrElse(true)
```
lossyrob: Confusing variable name. Perhaps `isSegmentNearChunk` or something else more descriptive?

raster/src/main/scala/geotrellis/raster/io/geotiff/LazySegmentBytes.scala
pomadchin: I'll try some variants, mb it's possible smth, I agree that we can forget about it

raster/src/main/scala/geotrellis/raster/io/geotiff/LazySegmentBytes.scala
pomadchin: Yea, your variant looks much better

pomadchin:
@lossyrob @echeipesh I tried to use constructor overloads https://github.com/locationtech/geotrellis/pull/2115/commits/e4f4ca511afb8cf7aeba1a14fc280e3845434317 but it still looks like an API change. The main problem is that we can't overload functions with default values, i'd be glad if we could use just the following thing: 

```scala
class LazySegmentBytes(
  byteReader: ByteReader,
  tiffTags: TiffTags,
  maxChunkSize: Int = 32 * 1024 * 1024,
  maxOffsetBetweenChunks: Int = 1024
) extends SegmentBytes with LazyLogging { ... }
```

pomadchin:
I think, as we changed the milestone, i can remove constructor overloads and use default methods.

PR #2123, opened 2017-04-04
Get absolute path for attributePath

### LINKED ISSUE #2113: ETL: Layer exists returns false for layerIds(0)
I'm not entirely clear on the intended behavior here, but based on [these tests](https://github.com/locationtech/geotrellis/blob/ac1a460e7a2c00d1dd9fc12492c1ee18000b6a6d/spark/src/test/scala/geotrellis/spark/io/AttributeStoreSpec.scala#L44), I expected that if an attribute store has an id in its list of id's, then they should exist when queried via "layerExists": 

File structure: 

```
etl
├── backend-profiles.json
├── geotrellis-spark-etl-assembly-1.1.0-SNAPSHOT.jar
├── input.json
├── output.json
└── tiles
    ├── _attributes
    │   └── mobile___8___metadata.json
    └── mobile
        └── 8
            ├── _SUCCESS
            └── part-r-00000-49914
                ├── data
                └── index
```

Code: 

```
val store = HadoopAttributeStore("./etl/tiles")
store.layerExists(store.layerIds(0))
```

Results:

```
res32: Boolean = false
```

It seems to have a layerId: 

```
store.layerIds(0)
res33: geotrellis.spark.LayerId = Layer(name = "mobile", zoom = 8)
```

This also causes the reader to throw, naturally. 

The problem is that HdfsUtils.listFiles will return an absolute path, but when following the ETL tutorial, it was not clear to me that we must first create an absolute path. Not sure if that's what [this comment](https://github.com/locationtech/geotrellis/blob/147c6a230e92f88917a54af130764f9fb2da5bc2/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopAttributeStore.scala#L115) is referring to?

lossyrob:
Fixes #2113 

What was happening is, because the attributePath was being stored as a local path, the `attributePath` method returned the relative path as well; when being compared to listed paths in `layerExists`, it would always fail the match. This didn't surface itself beforehand because we would always use absolute paths with HadoopLayer types; this prevents a problem if a user decides to user a relative path when the default filesystem is local.

This was a bit tough to capture in a unit test, but I did test this manually against geotrellis-landsat-tutorial.

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopAttributeStore.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopAttributeStore.scala
@@ -34,8 +34,12 @@ import scala.util.matching.Regex
 class HadoopAttributeStore(val rootPath: Path, val hadoopConfiguration: Configuration) extends BlobLayerAttributeStore {
   import HadoopAttributeStore._
 
-  val attributePath = new Path(rootPath, "_attributes")
-  val fs = attributePath.getFileSystem(hadoopConfiguration)
+  val (fs, attributePath) = {
+    val ap = new Path(rootPath, "_attributes")
+    val fs = ap.getFileSystem(hadoopConfiguration)
+    // Get the absolute path to attributes
+    (fs, fs.getFileStatus(ap).getPath)
+  }
 
   // Create directory if it doesn't exist
   if(!fs.exists(attributePath)) {
--- a/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopAttributeStore.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopAttributeStore.scala
@@ -18,17 +18,14 @@ package geotrellis.spark.io.hadoop
 
 import geotrellis.spark._
 import geotrellis.spark.io._
-import geotrellis.spark.io.index.KeyIndex
-import org.apache.avro.Schema
 
 import spray.json._
 import DefaultJsonProtocol._
 import org.apache.hadoop.fs.Path
 import org.apache.spark._
-import java.io.PrintWriter
-
 import org.apache.hadoop.conf.Configuration
 
+import java.io.PrintWriter
 import scala.util.matching.Regex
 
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopAttributeStore.scala
```diff
@@ -34,8 +34,12 @@ import scala.util.matching.Regex
 class HadoopAttributeStore(val rootPath: Path, val hadoopConfiguration: Configuration) extends BlobLayerAttributeStore {
   import HadoopAttributeStore._
 
-  val attributePath = new Path(rootPath, "_attributes")
-  val fs = attributePath.getFileSystem(hadoopConfiguration)
+  val (fs, attributePath) = {
+    val ap = new Path(rootPath, "_attributes")
+    val fs = ap.getFileSystem(hadoopConfiguration)
+    // Get the absolute path to attributes
+    (fs, fs.getFileStatus(ap).getPath)
```
pomadchin: `getFileStatus` throws `FileNotFoundException` when the path does not exist

pomadchin: A straightforward solution would be to use try catch, and to create dirs on exception. 
Another solution would be to `getFileStatus` of a root path and to create only `_attributes` folder, but that would require manual catalog creation, and still would throw runtime exceptions.

```scala
val (fs, attributePath) = {
    val ap = new Path(rootPath, "_attributes")
    val fs = ap.getFileSystem(hadoopConfiguration)
    // Get the absolute path to attributes
    try {
      fs.getFileStatus(ap)
    } catch {
      case _: FileNotFoundException => fs.mkdirs(ap)
    }

    (fs, fs.getFileStatus(ap).getPath)
  }
```

pomadchin: Oh, try catch way would be 100% good, here is `exists` method implementation: 

```java
public boolean exists(Path f) throws IOException {
  try {
    return getFileStatus(f) != null;
  } catch (FileNotFoundException e) {
    return false;
  }
}
```

PR #2131, opened 2017-04-07
Produce Voronoi diagrams from DelaunayTriagulations

jpolchlo:
After reintroducing the custom Delaunay triangulator, it is necessary to allow for Voronoi diagrams to be constructed from these triangulations.  Herein, we provide the necessary classes to support this operation.


raster/src/main/scala/geotrellis/raster/distance/EuclideanDistanceTileMethods.scala
```diff
@@ -27,3 +29,11 @@ trait EuclideanDistanceTileArrayMethods extends MethodExtensions[Array[Point]] {
 trait EuclideanDistanceTileMethods extends MethodExtensions[Traversable[Point]] {
   def euclideanDistanceTile(rasterExtent: RasterExtent): Tile = { EuclideanDistanceTile(self.toArray, rasterExtent) }
 }
+
+trait EuclideanDistanceTileCoordinateArrayMethods extends MethodExtensions[Array[Coordinate]] {
+  def euclideanDistanceTile(rasterExtent: RasterExtent): Tile = { EuclideanDistanceTile(self, rasterExtent) }
+}
```
echeipesh: This method extension isn't needed. The `.toArray` call in the `EuclideanDistanceTileCoordinateMethods` would take care of this case just as well.

PR #2134, opened 2017-04-11
Added back CONTRIBUTING.md

lossyrob:
We need a contributing doc in the root folder as per the Eclipse guidelines; looks like it was removed and placed into the docs; so this readds it.


fosskers:
How about a symlink instead, to avoid divergence of the two files?

Issue #2129, opened 2017-04-05
Incorrect logic during BufferedReproject with set zoom level


lossyrob:
This line:
https://github.com/locationtech/geotrellis/blob/5cb43cb8ccec72e841537e765ae9db6f3f069620/spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala#L166

Incorrectly gets the metadata for the target CRS, when in a buffered reproject, the first tiling happens on the source CRS and a floating layout scheme; _followed_ by a reproject and tiling to the dest layout scheme and CRS.

This might needs some additional API on reproject to the case where you want to use a ZoomedLayoutScheme for a specific zoom level in the post-reproject tiling process.

PR #2135, opened 2017-04-11
Repartition in ETL when re-tiling increases layer resolution

echeipesh:
During ETL process there are two ways in which the job specification can cause increase in resolution:

- When `maxZoom` parameter forces a higher than necessary level using `ZoomedLayoutScheme`
- When `LayoutDefinition` is provided that has higher resolution than source imagery

This is possible both in per-tile reprojected and buffered reproject methods.

If the increase in resolution is significant it will dramatically increase the size of the partitions, which were originally mapped to source imagery. Past certain point the partitions become too large to be processed.

For per-tile reproject this PR handles the logic in `geotrellis.spark.etl.Etl` class by inspecting the difference in resolutions of pre-tiled metadata and post-tiled metadata.

However in buffered reproject the check must happen during after the reprojection took place. The resolutions in different CRS are not comparable so we must check the tiles covered by the `KeyBounds`, assuming that the tiles themselves remain mostly constant in size.

### Incidental Fixes
- Render module was not callable due to incorrect name specification
- Logging added to CutTiles for when a single resample may OOM
- `Etl` must return `maxZoom` as zoom level when its specified

Resolves: https://github.com/locationtech/geotrellis/issues/2129

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/tiling/CutTiles.scala
+++ b/spark/src/main/scala/geotrellis/spark/tiling/CutTiles.scala
@@ -16,19 +16,20 @@
 
 package geotrellis.spark.tiling
 
-import geotrellis.vector.Extent
 import geotrellis.raster._
 import geotrellis.raster.merge._
 import geotrellis.raster.prototype._
 import geotrellis.raster.resample._
 import geotrellis.spark._
+import geotrellis.util.LazyLogging
 
-import org.apache.spark._
 import org.apache.spark.rdd._
 
 import scala.reflect.ClassTag
 
 object CutTiles {
+  @transient private lazy val logger = LazyLogging(this)
+
   def apply[
     K1: (? => TilerKeyMethods[K1, K2]),
     K2: SpatialComponent: ClassTag,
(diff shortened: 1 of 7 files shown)


echeipesh:
## Buffered reproject ingest
buffered-ingest

spark/src/main/scala/geotrellis/spark/tiling/CutTiles.scala
```diff
@@ -16,19 +16,20 @@
...
 import geotrellis.spark._
+import geotrellis.util.LazyLogging
 
-import org.apache.spark._
 import org.apache.spark.rdd._
 
 import scala.reflect.ClassTag
 
 object CutTiles {
+  @transient private lazy val logger = LazyLogging(this)
```
pomadchin: Why you don't want to extend `LazyLogging`? API compatibility? Looks ugly and like we can forget about it ):

util/src/main/scala/geotrellis/util/LazyLogging.scala
```diff
@@ -29,3 +29,9 @@ trait LazyLogging {
   @transient protected lazy val logger: Logger =
     Logger(LoggerFactory.getLogger(getClass.getName))
```
pomadchin: It you are sure in this approach, `LazyLogging(this)`

spark/src/main/scala/geotrellis/spark/tiling/CutTiles.scala
echeipesh: Yeah, that was the intent is to avoid adding types to classes when all we want is this private field. I think this should be the prefered approach.

PR #2139, opened 2017-04-12
Update EuclideanDistance operation to new triangulator with spark capability

jpolchlo:
It has been possible to produce a tile representing the Euclidean distance to the closest point in a set of points for some time.  In light of some improvements to the Delaunay triangulator—notably, the ability to perform distributed triangulations—it is time to fold these changes into Euclidean distance and enable distributed generation of those tiles. (Note this PR is on top of #2131.)

- [x] Basic implementation
- [x] Optimization
- [x] Spark tests
- [x] Cleanup


jpolchlo:
The basic implementation is complete.  Consider that the following Euclidean distance tile was generated from a small excerpt of LAS data (please ignore the minor color mapping glitch in the red area):

spark_distance

Note that the influence of points outside the current extent are clearly visible along the edges of the tile.

jpolchlo:
Runtime performance of this implementation is shockingly bad.  The above test requires about 25 seconds.  The baseline test which triangulates all the points in a 3x3 neighborhood of the original layout (which we triangulate individually and stitch in the distributed version), computes a Voronoi diagram, and rasterizes the Euclidean distance tile runs in a second or two.  The problem is in StitchedDelaunay, which takes an absolutely inordinate amount of time to complete.

jpolchlo:
Runtimes are down to <2 seconds for the Euclidean distance test.  Replacing function calls with array accesses makes the difference as usual.

jpolchlo:
Now with more Spark!  The following image, produced by one of the unit tests which generates randomized data of a given shape, and calculates the Euclidean distance raster over a 4x4 tiling of the domain.  This is the stitched result:

stitched

lossyrob:
A view with the vector data:

screen shot 2017-04-14 at 2 01 08 pm

PR #2142, opened 2017-04-18
Reorganized handling of `CellType.name`

metasim:
to make it the authoritative representation of cell type, over `toString`.
Moved `name` resolution to companion object to make it symmetric with `fromName` implementation.

In the previous implementation, inheritance was used as the means of having different values of `name` per type. That caused problems when trying to come up with a general solution based on calling `super` to generate derivatives, since self-types were used over direct inheritance. Having the to/from name handling in the companion also keeps coupled implementations closer to each other.

cc: @moradology 

Signed-off-by: Simeon H.K. fitch <fitch@astraea.io>

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/CellType.scala
+++ b/raster/src/main/scala/geotrellis/raster/CellType.scala
@@ -18,7 +18,6 @@ package geotrellis.raster
 
 import scala.util.matching.Regex
 import java.awt.image.DataBuffer
-import java.lang.IllegalArgumentException
 
 // Note: CellType defined in package object as
 // `type CellType = DataType with NoDataHandling`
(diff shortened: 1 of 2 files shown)


raster/src/main/scala/geotrellis/raster/CellType.scala
```diff
@@ -398,36 +386,29 @@ case class DoubleUserDefinedNoDataCellType(noDataValue: Double)
...
-    case DataBuffer.TYPE_FLOAT => FloatConstantNoDataCellType
-    case DataBuffer.TYPE_DOUBLE => DoubleConstantNoDataCellType
-    case _ => throw new IllegalArgumentException(s"AWT type $awtType is not supported")
-  }
+   * Translate a string representing a cell type into a [[CellType]].
+   *
+   * @param name A string representing a cell type, as reported by [[DataType.name]] e.g. "uint32"
+   * @return The CellType corresponding to `name`
+   */
+  @deprecated("Use `fromName` instead", "1.2.0")
```
lossyrob: Can you change to 1.1.0?

Issue #2160, opened 2017-04-21
Documentation for geotrellis.spark.KeyBounds


metasim:
The semantics of `include` vs. `combine` and `includes` vs. `contains` are not entirely clear from the function signatures, nor even from reading the code on first pass. The purpose of `get` is also confusing. API docs in this area would be helpful.

metasim:
Furthermore, the implementations of `Bounds.map` and `Bounds.flatMap` don't really follow monadic conventions. One would expect something more along the lines of `map[B](f: A => B): Bounds[B]` and `flatMap[B](f: A => Bounds[B]): Bounds[B]`. Deviation from expected structures should be supported by documentation.

PR #2161, opened 2017-04-21
Documentation for KeyBounds

echeipesh:
Resolves: https://github.com/locationtech/geotrellis/issues/2160

@metasim This should provide enough context to discuss validity further. Looking through the code it looks like the usage from `map` and `flatMap` is gone, so there may be a discussion if they have a useful shape at all. 

The original reasoning was that `Bounds` captures the idea of optionality where our version of `None`, which is `EmptyBounds` can still be used as valid arguments for `include`, `combine` and `intersect` removing the need to do these case checks at call sites since they all follow the same pattern.

Given that `map` and `flatMap` should expose the contents of `Some` case, which is really a tuple of `minKey` and `maxKey` providing them as `KeyBounds` instead of `Tuple2` is more descriptive and allows for manipulations which are not possible with `.map(f: A => B): Bounds[B]`. Because in the latter case you can't discriminate on min vs max key.

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/KeyBounds.scala
+++ b/spark/src/main/scala/geotrellis/spark/KeyBounds.scala
@@ -21,45 +21,79 @@ import geotrellis.util._
 
 import org.apache.spark.rdd.RDD
 
+/** Represents a region of discrete space, bounding it by minimum and maximum points.
+ * The bounds maybe [[EmptyBounds]] as result of intersection operation.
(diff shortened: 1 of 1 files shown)


metasim:
@echeipesh Thanks much. I'm sure this will help others out a lot too.

echeipesh:
Appreciate you calling this out. I've been guilty of letting it slide.

PR #2163, opened 2017-04-26
Table scan on layer read in AccumuloLayerReader

echeipesh:
This optimization is to avoid the current edge case of generating too many index ranges when performing table scan. Given the implementation in this PR the single unbounded Accumulo `Range` will be split across the index and will generate sub-range per existing tablet.

### DIFF
--- a/accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerReader.scala
+++ b/accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerReader.scala
@@ -48,10 +48,14 @@ class AccumuloLayerReader(val attributeStore: AttributeStore)(implicit sc: Spark
 
     val queryKeyBounds = tileQuery(metadata)
 
-    val decompose = (bounds: KeyBounds[K]) =>
-      keyIndex.indexRanges(bounds).map { case (min, max) =>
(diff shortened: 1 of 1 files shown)


accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerReader.scala
```diff
@@ -48,10 +48,14 @@ class AccumuloLayerReader(val attributeStore: AttributeStore)(implicit sc: Spark
 
     val queryKeyBounds = tileQuery(metadata)
 
-    val decompose = (bounds: KeyBounds[K]) =>
-      keyIndex.indexRanges(bounds).map { case (min, max) =>
-        new AccumuloRange(new Text(AccumuloKeyEncoder.long2Bytes(min)), new Text(AccumuloKeyEncoder.long2Bytes(max)))
-      }
+    val decompose = { (bounds: KeyBounds[K]) =>
+      if  (bounds == metadata.getComponent[Bounds[K]].get)
```
pomadchin: Kinda very unsafe `.get` call here.

echeipesh: I don't think we're able to write a layer with `EmptyKeyBounds`, so this seemed safe to me.

PR #2174, opened 2017-05-02
CellSize values should not be truncated to integer when parsing from Json.

mteldridge:


raster/src/main/scala/geotrellis/raster/io/json/Implicits.scala
```diff
@@ -46,7 +46,7 @@ trait Implicits extends HistogramJsonFormats {
     )
     def read(value: JsValue): CellSize =
       value.asJsObject.getFields("width", "height") match {
-        case Seq(JsNumber(width), JsNumber(height)) => CellSize(width.toInt, height.toInt)
+        case Seq(JsNumber(width), JsNumber(height)) => CellSize(width.toDouble, height.toDouble)
```
fosskers: A reminder to anyone checking this out that `JsNumber` contains a `BigDecimal`, so the explicit cast to `Double` is necessary.

PR #2175, opened 2017-05-02
Expose `isEmpty` for Geometries

fosskers:
Patchy patch!

Calling down to JTS's `isEmpty` is faster than doing, say:

```scala
val mp: MultiPoint = ...

mp.points.isEmpty
```

### DIFF
--- a/vector/src/main/scala/geotrellis/vector/Geometry.scala
+++ b/vector/src/main/scala/geotrellis/vector/Geometry.scala
@@ -34,6 +34,18 @@ trait Geometry {
   def isValid: Boolean =
     jtsGeom.isValid
 
+  /** Is this Geometry empty? This is faster than checking manually like:
+    * {{{
+    * val mp: MultiPoint = ...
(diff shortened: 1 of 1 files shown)


vector/src/main/scala/geotrellis/vector/Geometry.scala
```diff
@@ -34,6 +34,18 @@ trait Geometry {
...
+  /** Is this Geometry empty? This is faster than checking manually like:
+    * {{{
+    * val mp: MultiPoint = ...
+    * val ps: Array[Point] = mp.points  // `.points` is a lazy val with processing overhead
+    *
+    * ps.isEmpty  // possible, but mp.isEmpty is faster
+    * }}}
+    * It would be similar for [[MultiLine]] or [[MultiPolygon]].
+    */
+  def isEmpty: Boolean =
```
echeipesh: Calling out that use of `isEmpty` without parens is consistent with scala collections API even though in general its against the suggested style.

PR #2176, opened 2017-05-03
Added `MinimalMetadata`

fosskers:
This is the smallest class that fits Metadata constraints.


fosskers:
This is going to get moved into the VT uber PR.

PR #2180, opened 2017-05-07
Fixes to GeoTiff writing with original LZW compression.

### LINKED ISSUE #2177: Savings a streaming MultibandGeoTiff creates a corrupted geotiff
Fails on Geotrellis Version: 1.1.0-RC4 and RC3

Issue cannot be reproduced in 1.1.0-RC2 but still does not work due to another issue where we just never return.

The following code produces the error above:

```
package com.agrian.analytics.ingest

import geotrellis.raster.io.geotiff.MultibandGeoTiff
import geotrellis.spark.io.http.util.HttpRangeReader
import geotrellis.util.StreamingByteReader

import scalaj.http.Http
import com.typesafe.scalalogging.LazyLogging

object TestApp extends LazyLogging {
  def main(args: Array[String]): Unit = {

    val s3Link = "https://s3-us-west-2.amazonaws.com/vtpkmsdybsdflk/1056026_2017-03-02_RE2_3A.tif"
    val streamer = HttpRangeReader.withoutHeadRequest(s3Link)
    val reader = StreamingByteReader(streamer)

    MultibandGeoTiff.streaming(reader).write("streamed_basetile.tiff")
  }
}

```

If you run `gdalinfo basetile.tiff -hist -stats` you get the resulting info:

```
Warning 1: LZWDecode:LZWDecode: Strip 515 not terminated with EOI code
ERROR 1: LZWDecode:Not enough data at scanline 515 (short 1179 bytes)
ERROR 1: TIFFReadEncodedStrip() failed.
ERROR 1: basetile.tiff, band 1: IReadBlock failed at X offset 0, Y offset 515
```

Some additional notes, if you change the `MultibandGeoTiff.streaming(reader).write("streamed_basetile.tiff")` line to: 
```
    val geotiff = MultibandGeoTiff.streaming(reader)
    val extent@Extent(xmin, ymin, xmax, ymax) = geotiff.extent

    geotiff
      .crop(Extent(xmin + (extent.width * 0.1), ymin + (extent.height * 0.1), xmax, ymax))
      .write("streamed_basetile.tiff")
```

this will actually work on RC3 and RC4.

lossyrob:
This PR fixes several issues with GeoTiff writing, in the case where you write a GeoTiff tile that has LZW compression with a predictor. We don't have support for writing LZW from uncompressed values, but we allow LZW to be used if the original compressed GeoTiff segments exist. This case had several issues:

- Decompression with a predictor that had little endian byte order did not write the correct byte order.
- The predictor GeoTiff tag was not being written

Writing bytes from a streaming source (using LazySegmentBytes) was also broken, because changes were made to not guarantee the ordering of the sequence of segments, which the writer relied on. A docstring mentions not to assume that the segments are iterated over in segment index order, and the writer code changed to gather segment byte counts in a way that provides the correct order.

Fixes #2177


raster/src/main/scala/geotrellis/raster/io/geotiff/writer/TiffTagFieldValue.scala
```diff
@@ -187,7 +188,15 @@ object TiffTagFieldValue {
...
 
-    val segmentByteCounts = imageData.segmentBytes.map { _.length }.toArray
+    val segmentByteCounts = {
+      val len = imageData.segmentBytes.length
+      val arr = Array.ofDim[Int](len)
+      cfor(0)(_ < len, _ + 1) { i =>
+        arr(i) = imageData.segmentBytes.getSegmentByteCount(i)
+      }
+      arr
+    }
```
lossyrob: @echeipesh @jbouffard this is the offending line for the streaming problem. Notice the doc string I made for SegmentBytes. The fixed line's reliance on the ordering of the segments must have been missed when SegmentBytes moved from always being segment order (which was clear since it used to only be `Array[Array[Byte]]` to an arbitrary order based on the `getSegments(0 until length)` call.

jbouffard:
:+1:

kthompson:
Thanks! :D

lossyrob:
@kthompson no problem. Can you test it out just to make sure it works on your end as well?

PR #2182, opened 2017-05-08
Script to collect dependencies between versions

lossyrob:


echeipesh:
Is it still in progress? The notebook loos like it may use a cleanup but otherwise this was obviously used to great effect.

PR #2197, opened 2017-05-11
Distributed EuclideanDistance based on DelaunayTriangulation

echeipesh:
Supersedes https://github.com/locationtech/geotrellis/pull/2131
Supersedes https://github.com/locationtech/geotrellis/pull/2139
Supersedes https://github.com/locationtech/geotrellis/pull/2141


spark/src/main/scala/geotrellis/spark/buffer/Direction.scala
```diff
@@ -28,4 +28,18 @@ object Direction {
...
+      case TopLeft => geotrellis.util.Direction.TopLeft
+      case Top => geotrellis.util.Direction.Top
+      case TopRight => geotrellis.util.Direction.TopRight
+      case Left => geotrellis.util.Direction.Left
+      case Center => geotrellis.util.Direction.Center
+      case Right => geotrellis.util.Direction.Right
+      case BottomLeft => geotrellis.util.Direction.BottomLeft
+      case Bottom => geotrellis.util.Direction.Bottom
+      case BottomRight => geotrellis.util.Direction.BottomRight
+    }
```
echeipesh: @lossyrob This was the unfortunate side effect of having two `Direction`s. All of the local Delaunay code uses the `geotrellis.util.Direction` while the spark still uses its own because of reliance on `collectNeighbors`. Doesn't seem like there is a way around this until next revision.

PR #2198, opened 2017-05-12
Extend Delaunay/Voronoi/EuclideanDistance interfaces to work with other types

jpolchlo:
Method extensions existed in the Delaunay interface for working with `Coordinates`.  This isn't always the most useful interface.  This PR adds other modes of interaction including `Double` pairs and triples and `MultiPoint`s.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/distance/EuclideanDistanceTile.scala
+++ b/raster/src/main/scala/geotrellis/raster/distance/EuclideanDistanceTile.scala
@@ -58,4 +58,20 @@ object EuclideanDistanceTile {
     tile
   }
 
+  def apply(pts: Array[(Double, Double)], rasterExtent: RasterExtent): Tile = {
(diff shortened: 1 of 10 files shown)


jpolchlo:
Based on feedback from @echeipesh, because Double pairs and triples are not a very "natural format" (data will most likely need to be coerced into that form from some input format), the method extensions corresponding to tuples have been dropped.

raster/src/main/scala/geotrellis/raster/distance/EuclideanDistanceTile.scala
```diff
@@ -58,4 +58,20 @@ object EuclideanDistanceTile {
     tile
   }
 
+  def apply(pts: Array[(Double, Double)], rasterExtent: RasterExtent): Tile = {
```
echeipesh: These should just map the `pts` and delegate to the `Coordinate` apply

PR #2216, opened 2017-05-28
Fix ETL upsampling bug

### LINKED ISSUE #2170: Buffered up-sampling reproject in ETL
Up-sampling reproject using GDAL:
gdal-rep

Up-sampling reproject using ETL buffered reproject:
etl-rep

Call site: https://github.com/locationtech/geotrellis/blob/master/spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala#L199-L201

This appears like a legitimate way to call the reproject, if not parameter validation may be required.

pomadchin:
Target sellSize should be passed in all the reproject functions

- [x] Test (on bees)

Closes #2170

### DIFF
--- a/spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala
+++ b/spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala
@@ -16,12 +16,13 @@
 
 package geotrellis.spark.etl
 
-import geotrellis.raster.CellGrid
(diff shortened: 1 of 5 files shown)


lossyrob:
There's some performance improvements that I want to make on top of this, but this functionality is being waited on so I'm going to merge and make an issue with the performance stuff

PR #2217, opened 2017-05-29
Add `MultibandTile.safeBand`

fosskers:
`MultibandTile.band` should not be used unless you're absolutely sure that the band index you're calling for is valid.

### DIFF
--- a/raster-test/src/test/scala/geotrellis/raster/io/geotiff/ArrayMultibandTileSpec.scala
+++ b/raster-test/src/test/scala/geotrellis/raster/io/geotiff/ArrayMultibandTileSpec.scala
@@ -20,7 +20,7 @@ import geotrellis.raster.testkit._
 
 import org.scalatest._
 
-class ArrayMultibandTileSpec extends FunSpec {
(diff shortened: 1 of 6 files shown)


fosskers:
Unclear why that failed.

echeipesh:
`getBand` would be consistent with `Map` interface, which this is closest to in my mind: `Map.get(k: K): Option[V]`.

@lossyrob @fosskers There was a naming discussion but I want to throw my 2-cents in on this one.

lossyrob:
I like @echeipesh's suggestion

fosskers:
I was motivated by the name of the current function, `band`, and by the trend to add `safe` to function names when it's the `Option`-yielding variant.

I hear the point about compatibility with `Map`, except that we wouldn't be matching it with `apply`:
```scala
Map[K, V]

def apply(k: K): V
```
`get` seems to imply safety in the case of `Map`, but that isn't the case with `Option` as we well know. In general, just thinking about the names, `band` and `getBand` don't portray the different in their intentions. `bandSafe` (or the original `safeBand`) does.

lossyrob:
I buy it.

PR #2222, opened 2017-05-31
Move to a different Json Schema validator

pomadchin:
A lightweight https://github.com/networknt/json-schema-validator is used instead of https://github.com/java-json-tools/json-schema-validator.

Were applied some changes into the input schema definition (it was incorrect). 
Some changes into printing validation error messages were added (logger used, error messages are red)

A source with tests for different json schema validators in different languages:
https://github.com/json-schema-org/JSON-Schema-Test-Suite

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -48,4 +48,6 @@ object Dependencies {
   val slickPG      = "com.github.tminglei" %% "slick-pg" % "0.14.6"
 
   val parserCombinators = "org.scala-lang.modules" %% "scala-parser-combinators" % "1.0.5"
+
+  val jsonSchemaValidator = "com.networknt" % "json-schema-validator" % "0.1.7"
 }
--- a/spark-etl/build.sbt
+++ b/spark-etl/build.sbt
@@ -2,7 +2,7 @@ import Dependencies._
 
 name := "geotrellis-spark-etl"
 libraryDependencies ++= Seq(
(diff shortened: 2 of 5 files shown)


lossyrob:
## Dependency Additions

json-schema-validator-dependency-graph

### CQs required for

#### Json Schema Validator

- com.networknt:json-schema-validator:0.1.7

#### Jackson

- com.fasterxml.jackson.core:jackson-annotations:2.8.0
- com.fasterxml.jackson.core:jackson-core:2.8.7
- com.fasterxml.jackson.core:jackson-databind:2.8.7

__Question__: Why is jackson-annotations at 2.8.0 for the 2.8.7 versions? Can we submit these three as a single CQ (do they share the same repository?)

Note there are approved CQs for jackson 1.5.0

#### Apache Commons Lang
- org.apache.commons:commons-lang3:3.5

3.5 should be submitted, Apache commons-lang3 version 3.4 is already approved.

pomadchin:
I think these `jackson-annotations:2.8.0` is from some different dep. I decided to draw the whole project graph using 

```bash
./sbt spark-etl/dependencyGraphMl
```

image

This information makes sense, as it corresponds to `com.networknt:json-schema-validator:0.1.7` [pom.xml](https://github.com/networknt/json-schema-validator/blob/0.1.7/pom.xml#L68)

lossyrob:
CQ's Approved. +1

PR #2223, opened 2017-06-01
AttributeStoreProvider for AttributeStore  SPI

echeipesh:
This allows `AttributeStore` companion object to implement `apply(uri: URI): AttributeStore` based on java SPI.

Examples:

```rst
=============  ============
Backend        URI
=============  ============
``Hadoop``     ``hdfs://path/to/catalog``
``S3``.        ``s3://bucket/catalog-key``
``File``       ``file://tmp/local-catalog``
``Accumulo``   ``accumulo://[user[:password]@]zookeeper/instance-name[?attributes=table1[&layers=table2]]``
``Cassandra``  ``cassandra://[user:password@]zookeeper[:port][/keyspace][?attributes=table1[&layers=table2]]``
``HBase``      ``hbase://zookeeper[:port][?master=host][?attributes=table1[&layers=table2]]``
=============  ============
```

TODO:
- [x] Agree on URI structure
- [x] Add SoftReference cache to AttributeStore URI apply
- [x] Update tests for URI builders
- [x] Update docs

CQs:

- [x] Caffeine 2.5.2 [CQ 13804](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=13804)
- [x] Scaffeine 2.2.0 [CQ 13805](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=13805)

Signed-off-by: Eugene Cheipesh <echeipesh@gmail.com>


metasim:
This is awesome! Can't wait.... what's the target milestone?

lossyrob:
Is this API breaking at all? What's the merge plan for this?

metasim:
Exciting!

echeipesh:
After creating the LayerReader and LayerWriter providers I found that the URI structure needed a little more consideration. 
`LayerReader("accumulo://root:secret@zookeeper.local/gis")`  call needs to have both ways to specify the metadata table and the values table. The values table only needs to be present when writing a layer because the attribute store will have table name for layer being read.

There are two options for this:
Option 1: Put everything into query 
```
# Loads attribute store
accumulo://root:secret@zookeeper.local/gis?attributes=metadata
# Loads layer writer
accumulo://root:secret@zookeeper.local/gis?layers=tiles
# Loads both
accumulo://root:secret@zookeeper.local/gis?attributes=metadata&layers=tiles
```

Option 2: Make use of the URI (fragment)[https://en.wikipedia.org/wiki/Fragment_identifier]

```
accumulo://root:secret@zookeeper.local/gis?attributes=metadata
accumulo://root:secret@zookeeper.local/gis#tiles
accumulo://root:secret@zookeeper.local/gis#tiles?attributes=metadata
```

spark/src/main/scala/geotrellis/spark/io/LayerReader.scala
```diff
@@ -59,6 +62,33 @@ trait LayerReader[ID] {
 }
 
 object LayerReader {
+  /**
+   * Produce LayerReader instance based on URI description.
+   * Find instances of [[LayerReaderProvider]] through Java SPI.
+   */
+  def apply(attributeStore: AttributeStore, layerReaderUri: URI)(implicit sc: SparkContext): LayerReader[LayerId] = {
```
echeipesh: @lossyrob Calling that that this `apply` is placed here even through it is producing `FilteringLayerReader` to keep the user facing API symmetrical `LayerReader`/`LayerWriter`

metasim:
Ooof, that's a tough one. Just so I'm clear, are you wanting to account for the case where the metadata and values can be stored with different schemes? My original though of using the old SVN way of doing it with an aggregate scheme (e.g. `accumulo+postgres://`) doesn't work because you can't store separate credentials. Between option 1 and option 2 , I like 1 a bit more, but not sure why. But still seems like there should be a better way.

As a reminder, why not have the ability to have a separate URI for metadata vs values (with fallback to single URI for both)?

echeipesh:
I don't think its worth while to account for that in URI design, I think that case is covered by accepting to URIs. The question is basically: In those cases when you are instantiating an AttributeStore and LayerWriter from the same URI, same scheme, and you need to specify explicitly the attributes table as well as the values table, where do those two things go in the URI structure. It feels like that case should require only a single URI and that choice should be pretty consistent across the backends.

metasim:
Food for thought?:
http://www.h2database.com/html/features.html#database_url

echeipesh:
That's a pretty good point actually. Looks like CacheBuilder is since 10.0 and I can't imagine guava not being available in some form. Why never-mind? 

The only problem is that I would still need wrap it to avoid new-ing up `Callable` all over the place. But where to place the wrapper? I suppose we can give `util` compile time dependency and depend on guava always lurking in the class path?

metasim:
FWIW, it's often the *only* class in the library I use when depending on it. Correct caching is hard, but that class tends to get it right. 

I think keeping SoftCache is ok for the purposes of wrapping the guava implementation. I personally wouldn't trust my own implementation in a spark or multithreaded context without lots of tests. But I'm pessimistic that way. 

I'd bump the guava dep to the latest version and see if the consequential eviction causes problems in other code (it's a stable lib, so unlikely).

accumulo/build.sbt
```diff
@@ -7,6 +7,7 @@ libraryDependencies ++= Seq(
     exclude("org.apache.hadoop", "hadoop-client"),
   sparkCore % "provided",
   spire,
+  scaffeine,
```
pomadchin: Making it only spark subproject dep is not enough?

accumulo/src/test/scala/geotrellis/spark/io/accumulo/AccumuloAttributeStoreSpec.scala
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Azavea
+ * Copyright 2017 Azavea
```
pomadchin: Sounds like [here](https://github.com/locationtech/geotrellis/blob/master/build.sbt#L71-L72) changes should be done too (mb in a sep PR though)

metasim: `sbt-header` FTW! :-)

pomadchin:
Added CQs in the PR description.

pomadchin:
CQs passed!

accumulo/build.sbt
pomadchin: ^ bump this question up, as all backends depend on spark, should it be only a `spark` project dep?

accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloInstance.scala
```diff
@@ -50,6 +51,18 @@ object AccumuloInstance {
     val tokenClass = token.getClass.getCanonicalName
     BaseAccumuloInstance(instanceName, zookeeper, user, (tokenClass, tokenBytes))
   }
+
+  def apply(uri: URI): AccumuloInstance = {
+    import geotrellis.util.UriUtils._
+
+    val zookeeper = uri.getHost
+    val instance = uri.getPath.drop(1)
```
pomadchin: `uri.getPath.tail` looks nicer :D

accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloInstance.scala
```diff
@@ -50,6 +51,18 @@ object AccumuloInstance {
...
+    val (user, pass) = getUserInfo(uri)
+    AccumuloInstance(
+      instance, zookeeper,
+      user.getOrElse("root"),
+      new PasswordToken(pass.getOrElse("")))
```
pomadchin: Can Accumulo has an empty string as a pwd? May be an exception can take place here?

spark/src/main/scala/geotrellis/spark/io/file/FileLayerProvider.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  def canProcess(uri: URI): Boolean = uri.getScheme.toLowerCase == "file"
+
+  def attributeStore(uri: URI): AttributeStore = {
+    val file = new File(uri)
+    FileLayerProvider.cache.get(file.getCanonicalPath,
+      canonicalPath => new FileAttributeStore(canonicalPath))
+  }
+
+  def layerReader(uri: URI, store: AttributeStore, sc: SparkContext): FilteringLayerReader[LayerId] = {
```
pomadchin: Why `SparkContext` can't be implicit param here?

lossyrob:
Something I want to raise issue with is the caching. It's a smell to me that every provider has a caching mechanism, e.g.

```scala

object HBaseLayerProvider {
  private val cache: Cache[(String, String), AttributeStore] = Scaffeine().softValues().build()
 }
```
which the end user has no control over. This type of built-in caching seems like a application-level issue, and not a library issue. If client code wanted to cache attribute stores, they should do so themselves. There's currently a lot of discussion over on Raster Foundry about what the ideal caching strategy for the attribute stores are. Why should we bake in this application level caching concern into the library code? It seems like hidden unmodifiable functionality that might end up being a headache down the road.

metasim:
@lossyrob Interesting points. I don't know the operational arguments for caching that led Eugene to put it in there, but here's a thought exercise: Would you feel any different if you thought about the `XXXProvider`s being akin to JDBC drivers? If the driver implementer does performance testing and determines that query caching results in a significant performance benefit, would they be better off implementing it behind the scenes, or telling the users of the driver that they should perform caching themselves to get the best performance? I wouldn't be surprised if JDBC drivers cache some of the database metadata for that reason, and the `AttributeStore` feels a bit like database metadata.

On the threading concerns, part of that is addressed by using Scaffene. However, I'll go review the code that actually gets/adds items to/from the cache and see if anything jumps out at me. In google Guice's cache you provide a callback that handles creation of values, thereby allowing the library to take care of all the locking needs.

metasim:
@lossyrob Based on [the `get` method docs](https://static.javadoc.io/com.github.blemale/scaffeine_2.11/2.2.0/index.html#com.github.blemale.scaffeine.Cache@get(key:K,mappingFunction:K=>V):V) I'd personally be comfortable that the operations isn't of the "cross-yer-fingers" kind.

pomadchin:
In RF case these caching makes sense (and even current implementation, which is in master branch). It totaly makes sense in S3AttributeStore (to cache downloaded metadata blob). However making possible to override cache implementation (jsr107 api) or make it possible to disable default caching makes a lot of sense.

lossyrob:
Here's a situation which would show some headache:
```scala
val catalogUri: URI = ??? // Some valid uri
def grabALayer(layerId: LayerId, polygon: Polygon): TileLayerRDD[SpatialKey] =
 LayerReader(catalogUri)
    .query[SpatialKey, Tile, TileLayerMetadata[SpatialKey]](layerId)
    .where(Intersects(polygon))
    .result

val layerId: LayerId = ???
val layer = grabALayer(layerId, somePolygon)

// We get the tiles that intersect with the polygon.
// Fine, things move on. 
// But later that layer was updated elsewhere, in some other ingest process.
// More tiles were added, and so the extent and keybounds have changed.
// We haven't quit the application in that time.
// Perhaps we are a server of sorts.

val newLayer = grabALayer(layerId, polygon)
// Well, our polygon intersects some of the new tiles. But we don't see our new tiles!
```

I scratch my head..._But I called `LayerReader` to get a new layer reader, why aren't my new tiles here?_
After digging through GeoTrellis code for a while, I find https://github.com/locationtech/geotrellis/pull/2223/files#diff-80bed8ae5a88c047a7c1c87e6ff70d12R82, which is where a 'new' AttributeStore is grabbed from the URI, but because of caching that is in library code that I was unaware of, it doesn't actually refresh the AttributeStore. The attribute of the metadata for my layer is the old metadata, which has the old keybounds, which means my layer only loads based on the old keybounds and therefore misses my new tiles.

lossyrob:
@metasim 

> Based on [the `get` method docs](https://static.javadoc.io/com.github.blemale/scaffeine_2.11/2.2.0/index.html#com.github.blemale.scaffeine.Cache@get(key:K,mappingFunction:K=%3EV):V) I'd personally be comfortable that the operations isn't of the "cross-yer-fingers" kind.

I'm not trying to dig on scaffine or say it doesn't do what it says it does. But if for some reason we found out that a weird edge case caused the caching to be bad in whatever library you would use, there wouldn't be an opportunity to mess with it.

echeipesh:
## Original Considerations

The `AttributeStore` instances contain state, the layer metadata cache. The reason this cache exists is because in many `AttributeStore` sub-classes reading metadata is expensive. Additionally a common access pattern is to check for layer existence, read the layer metadata, switch based on K,V,M type and read the layer. Without some kind of cache this would result in 2-3 re-reads of the same metadata record.

The downside is that cache implies need for cache invalidation, on which we totally punted. Whether the cache was a good idea or not is perhaps up for debate but its one of the things that keeps responsiveness of GeoTrellis/spark operations up.

Given that the providers provide an instance of `AttributeStore` from a universal identifier it is helpful for it to continue sharing the state of the cache as long as possible. In fact as long as the layer manipulation originate from the same JVM than this caching mechanism ensures that the view of the layer remains transparently consistent. This mostly lines up with "One driver controls a cluster" thing we have used so far.

Client code attempting to cache instances of `AttributeStore` would have a bad time. It would need to have a way identify which part of `URI` are relevant for identity for each provider it might use. We already lean on query parameters for configuration that does not impact identity. A number of those values and providers do not even share same `URI` structure (by necessity). So if there is caching based on `URI` it makes sense that the provider instances do it, because they are the last part of the interface that connects `URI` to a specific `AttributeStore` subclass.

## Cache Invalidation Blues

So the use case @lossyrob lays out is legitimate problem with cache invalidation. If the layer updates are not sourced from the same JVM as the query, we're serving stale results. I can point out that `AttributeStore` has `clearCache()` and `clearCache(layerId)` methods but you wouldn't really think to look there.

One option we discussed is to decouple SPI for reader/writers from attribute store and introduce caching flag in the latter:

```scala
def grabALayer(layerId: LayerId, polygon: Polygon): TileLayerRDD[SpatialKey] = {
  val store = AttributeStore(catalog,Uri, cache=false)
  LayerReader(catalogUri, store)
    .query[SpatialKey, Tile, TileLayerMetadata[SpatialKey]](layerId)
    .where(Intersects(polygon))
    .result
}
```

Follow up question to this if `LayerReader(uri)` and `LayerWriter(uri)` should be removed to force independent generation of the `AttributeStore` where the `cache` flag is exposed? I would argue to keep those overloads since removing them impacts the utility of this feature a little bit but could be a compromise in least surprise. Scaladocs and docs and dispense appropriate warnings.

It is unfortunate that this is a really crude switch:

-   if `cache=true` you have better responsiveness but cache invalidation problems.
-   if `cache=false` your cache is only good per single request.

metasim:
@echeipesh Would `cache=...` not work as a URI parameter?

metasim:
I'd also note that Caffeine supports caching with expiration, so  if "eventual consistency" between ingest and query is OK, you could have the cached attributes be automatically purged after some period of time.

metasim:
@lossyrob If there's a concern about being tied to a single cache implementation, [ScalaCache](https://github.com/cb372/scalacache) supports swappable back-ends, including Caffeine and some distributed cache implementations. Not sure if ScalaCache is in the CQ list... I don't seem to have permissions to search that bugzilla.

pomadchin:
@metasim a small note about Scala Cache - we had a bad experience with it, that's why it is not in this PR. Probably @echeipesh and @moradology can tell you some insights related to Scala Cache.

metasim:
@pomadchin Good to know. 👍

metasim:
As I re-read this thread, there seem to be three issues: 1) should caching be done in the `AttributeStore`; 2) if so, should caching always be done automatically, or be controllable by the user; 3) should we engineer a solution to cache invalidation? Am I missing anything?

accumulo/build.sbt
echeipesh: Nope, I expected that as well, but it wasn't the case. I had to add it explicitly to each project.

accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloInstance.scala
echeipesh: Eh, 'drop(1)` is more explicit for what its worth.

lossyrob:
I think the attribute store caching is something we've had in place for a while, and I'm not keen on changing it's behavior, even if it is not the most ideal. I will advocate against a second layer of caching on top of that caching, which caches the caching attribute stores.

If we make it clear that an AttributeStore can "go stale" (perhaps we already do in the docs? if not then that is a doc task), then the user can understand when he or she needs to new up a new one in order to prevent staleness, based on their application requirements. The problem I describe above is a situation where it seems like a new AttributeStore/Reader was created, which is an action of the user to prevent staleness, but it results in similar staleness. Which is why I don't think the secondary caching that is being introduced in this PR should be included.

metasim:
@lossyrob My apologies; I misunderstood what was going on from the start. I thought it was the actual stored attribute values that were being cached. Please disregard my opinion then; not sure I know enough of what I'm talking about here to have an opinion.

lossyrob:
@metasim it's a good point about the confusion that two layers of caching introduces. Also probably the way I'm talking about it isn't helping 🥇 

The AttributeStore caching caches the individual attribute values per layer ID, so for instance if I were to request the metadata from a specific layer twice from the same S3AttributeStore, it will only read the metadata from S3 once.

This PR introduces another layer of caching (see [here](https://github.com/locationtech/geotrellis/pull/2223/files#diff-c10056bcab6e28d0a50f320a8dc11cfeR41)), which caches AttributeStores (not attribute store values, but actual AttributeStore instances) per URI. This is the one I'm raising flags about.

lossyrob:
__zemblanity:__
> zemblanity to mean somewhat the opposite of serendipity: "making unhappy, unlucky and unexpected discoveries occurring by design". A zemblanity is, effectively, an "unpleasant unsurprise".

A real world anecdote that happened today/last night that is related to the discussion. I'm working with an iterative ingest, one that accumulates sums of previous column values over a nationwide RDD. I wrote the ingesting algorithm iteratively to fetch the previously written row from the last iteration, and join them to the source row of the current iteration.

The shape of the input data looks like this (reprojected to EPSG:4326 from EPSG:5070, hence the tilt):

screen shot 2017-07-12 at 6 26 34 pm

I saw result tiles resetting their accumulated values every 8 tiles (I was writing 128x128 tiles off of 1024x1024 input tiles). Banged my head against it for a while (longer than I want to say). And it turns out, it was a....wait for it....caching issue!

I wasn't smart enough to realize that when I declared

```scala
 val layerReader = Ingest.getLayerReader(isLocal)
```

at the beginning of the method, and used that to fetch previously written rows, that the AttributeStore inside of LayerReader would have an opinion about the KeyBounds of the result layer after the first time I tried to fetch the first row, and that it's opinion would be that 

```scala
    def getRow(layerId: LayerId, row: Int): TileLayerRDD[SpatialKey] = {
      val p1 = mapTransform(SpatialKey(minKey.col, row)).center
      val p2 = mapTransform(SpatialKey(maxKey.col, row)).center
      layerReader
        .query[SpatialKey, Tile, TileLayerMetadata[SpatialKey]](layerId, numPartitions = if(isLocal) 100 else 1000)
        .where(Intersects(Line(p1, p2)))
        .result
    }
```

would return no values, based on the LayerQuery logic. Once I changed the declaration to:

```scala
    def layerReader = Ingest.getLayerReader(isLocal)
```

I got past this problem. I should have known better, but I didn't. Speaks to the first level of caching causing a real problem. Now if I had written `Ingest.getLayerReader` using the URI instead of a direct creation, with additional caching, the `val` -> `def` fix wouldn't have worked in this case, and I might be further confused.

metasim:
Good example of a "universal access principle" nuance!

Maybe this is standard practice, but if this were a public API I'd declare it as `def layerReader() = ...` (with parens) to emphasize that there's some construction going on every time its called.

echeipesh:
@lossyrob To be cheeky I'll point out that the current iteration would return the same of `AttributeStore` on all calls your problem would not have happened as `LayerWriter` and `LayerReader` would share the same store and call to `write` would update the cache.

But in all reality the caching behavior clearly a pitfall. I think it should be removed from the `AttributeStoreProvider` interfaces.

Furthermore there should be a follow-up PR that replaces the `HashMap` based caching in the `AttributeStore` instances (which is not thread safe) with scaffeine based cache that uses soft keys and has expiration time. The expiration should be configuration through `library.conf`

metasim:
🎉

ben-manes:
For caches, please be aware that soft references are not advised. Their simplicity is attractive, but experience has shown it to be a poor approach. The reason is that it increases GC pressure and has unpredictable evictions. Both cause hard to debug slowdowns only observable in production / heavy loads. I can provide more details if interested.

Preferably use a size bounding, allow custom configurations, and expose the statistics. You'll thank yourself if when on-call.

lossyrob:
@ben-manes thank you for the advice!

PR #2233, opened 2017-06-09
Example for creating a MultiBandTile RDD from Landsat GeoTiffs.

### LINKED ISSUE #2231: S3InputFormat does not filter according to extensions when partitionCount is used
This logic correctly filters out keys by extensions: https://github.com/locationtech/geotrellis/blob/41f695c569af6fe62cb1741377ad427dd8a14a3e/s3/src/main/scala/geotrellis/spark/io/s3/S3InputFormat.scala#L129-L136

That only executes `if (null == partitionCountConf)`. If there is a partition count, that logic is missing:

https://github.com/locationtech/geotrellis/blob/41f695c569af6fe62cb1741377ad427dd8a14a3e/s3/src/main/scala/geotrellis/spark/io/s3/S3InputFormat.scala#L162-L163

### LINKED ISSUE #2232: In S3GeoTiffReader, partitionBytes has no effect if maxTileSize is set
partitionBytes only has an effect in S3InputFormat, and is not taken into consideration when reading windows via maxTileSize. We should modify the maxTileSize path to try and respect this option.

lossyrob:
Also:
- Fixes #2231 
- Fixes #2232 

See: https://stackoverflow.com/questions/44421705/whats-the-easiest-way-to-load-single-band-landsat-8-rasters-from-aws-and-combin

### DIFF
--- a/doc-examples/build.sbt
+++ b/doc-examples/build.sbt
@@ -2,6 +2,6 @@ import Dependencies._
 
 name := "geotrellis-doc-examples"
 libraryDependencies ++= Seq(
-  sparkCore % "provided",
+  sparkCore,
   logging,
   scalatest % "test")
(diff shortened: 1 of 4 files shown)


lossyrob:
Commits 086cb16 and 	7a43bf6 should go into 1.1

PR #2243, opened 2017-06-21
Fixed problematic polygons in Voronoi diagram generation

jpolchlo:
Voronoi cells are constructed by creating a boundary representation in terms of half planes which are then used to clip the extent polygon.  We were previously including half planes for degenerate cell boundaries which were introducing duplicate vertices to the clipped polygons that were causing problems for the rasterizer and negatively affecting EuclideanDistanceTile.  

This PR adjusts numerical tolerances to make sure that degenerate edges are not included.  We use a more conservative value for the tolerance that may interpret valid edges as degenerate.  This might result in losing some edges, but the failure mode will be to have corners where the desired result should have been a very slightly truncated corner.  A mild problem that should only appear very rarely.

### DIFF
--- a/vector/src/main/scala/geotrellis/vector/voronoi/VoronoiDiagram.scala
+++ b/vector/src/main/scala/geotrellis/vector/voronoi/VoronoiDiagram.scala
@@ -170,23 +170,6 @@ object VoronoiDiagram {
     } else {
       clippedCorners += clippedCorners.head
       val poly = Polygon(clippedCorners.map(Point.jtsCoord2Point(_)))
-      // if (!poly.isValid) {
-      //   import het._
-
-      //   println("Invalid polygon generated:")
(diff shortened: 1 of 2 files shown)


vector/src/main/scala/geotrellis/vector/voronoi/VoronoiDiagram.scala
```diff
@@ -175,13 +169,31 @@ object VoronoiDiagram {
       None
     } else {
       clippedCorners += clippedCorners.head
-      Some(Polygon(clippedCorners.map(Point.jtsCoord2Point(_))))
+      val poly = Polygon(clippedCorners.map(Point.jtsCoord2Point(_)))
+      // if (!poly.isValid) {
```
echeipesh: Any reason to keep the commented code?

PR #2246, opened 2017-06-21
Fix writing multiband GeoTiff with compression

lossyrob:
Ass reported by @stevekuo on Gitter:

> I'm seeing GeoTiffOptions(DeflateCompression) being correctly observed with SinglebandGeoTiff, but ignored with MultibandGeoTiff.
> Roughly my code is

```scala
val geotiff = SinglebandGeoTiff(tile, extent, crs, Tags.empty, GeoTiffOptions(DeflateCompression))
// OR
val geotiff = MultibandGeoTiff(tile, extent, crs, Tags.empty, GeoTiffOptions(DeflateCompression))

val geotiffArray = GeoTiffWriter.write(geotiff)
val stream = new ByteArrayInputStream(geotiffArray)
```

> I then write out to a file. Using gdalinfo, the single band is COMPRESSION=DEFLATE, where as the multi band is not compressed.

This fixes the oversight that was causing the GeoTiff to be written as uncompressed.

### DIFF
--- a/raster-test/src/test/scala/geotrellis/raster/io/geotiff/writer/GeoTiffWriterSpec.scala
+++ b/raster-test/src/test/scala/geotrellis/raster/io/geotiff/writer/GeoTiffWriterSpec.scala
@@ -19,6 +19,7 @@ package geotrellis.raster.io.geotiff.writer
 import geotrellis.proj4.{CRS, LatLng}
 import geotrellis.raster._
 import geotrellis.raster.io.geotiff._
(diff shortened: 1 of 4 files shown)


stevekuo:
`tile` is a `MultibandTile`. Workaround is to convert `MultibandTile` to `MultibandGeoTiff` and specify the `GeoTiffOptions`.

```
val geotiffTile = MultibandGeoTiff(tile, options)
MultibandGeoTiff(geotiffTile, extent, metadata.crs, Tags.empty, options)
```

PR #2247, opened 2017-06-22
Fixed issue with BigTiff vs non-BigTiff offset value packing

### LINKED ISSUE #2241: TiffTagsReader fails to read bigtiff format
The TiffTagsReader is unable to read files in bigtiff format.  I have attached a bigtiff image which can be opened in **qgis** and can also be read by **gdalinfo**.

Repro:
download [foo.zip](https://github.com/locationtech/geotrellis/files/1089577/foo.zip)
`jar -xf foo.zip` 
`echo 'scalaVersion  := "2.11.8"' > build.sbt`
`echo 'libraryDependencies += "org.locationtech.geotrellis" %%  "geotrellis-raster" % "1.1.0-RC3"' >> build.sbt`
`sbt console`
`geotrellis.raster.io.geotiff.reader.TiffTagsReader.read("foo.tif")`

Result:
```
java.lang.IllegalArgumentException
  at java.nio.Buffer.position(Buffer.java:244)
  at geotrellis.util.ByteReader$$anon$1.position(ByteReader.scala:51)
  at geotrellis.raster.io.geotiff.util.ByteReaderExtensions$ByteReaderUtilities.getString(ByteReaderExtensions.scala:148)
  at geotrellis.raster.io.geotiff.reader.TiffTagsReader$ByteReaderTagReaderWrapper$.readAsciisTag$extension(TiffTagsReader.scala:253)
  at geotrellis.raster.io.geotiff.reader.TiffTagsReader$.readTag(TiffTagsReader.scala:124)
  at geotrellis.raster.io.geotiff.reader.TiffTagsReader$.read(TiffTagsReader.scala:102)
  at geotrellis.raster.io.geotiff.reader.TiffTagsReader$.read(TiffTagsReader.scala:57)
  at geotrellis.raster.io.geotiff.reader.TiffTagsReader$.read(TiffTagsReader.scala:33)
  ... 42 elided
```

It is possible to rewrite the attached sample image with 
`gdal_translate foo.tif bar.tif` which will rewrite it as a regular geotiff file.  However, for images which exceed the 4GB limit, this is not possible.  I have verified that the above repro steps occur on file over 4GB.

lossyrob:
Fixes #2241 

The TIFF Spec has a condition that is an optimization, but has caused a lot of issues.

A TIFF-tag is structured like `[code, type, length, offset]`. The `code` determines what tag it is, the `type` is the datatype such as Byte or Double, the `length` is the number of elements in the array of values, and the `offset` is the offset in the file where the values live.

If the value of the data is a small size, the value can be packed into the space that stores the offset. In this way, the tiff tag reads as `[code, type, length, value]` instead. This only happens if the byte size of `type`, multiplied by the `length` of the data, is less than the byte size of the `offset` value. In the regular TIFF case, the offset is a 4-byte wide `Int`, so any value that can be stored in 4 or less bytes is stored in the offset field, instead of some other location in the file which is pointed to by the value in the `offset` field.

For BigTiff, the offset field is an 8-byte wide `Long` value. We accounted for this when we first implemented bigtiffs. However, we didn't account for the changes in the packing mechanism: now if the value of the tag is less than or equal to 8 bytes, that value can be packed into the offset field of the tiff tag. The problem with not catching this packing is that, if you don't know the offset actually refers to a packed value, then you read that value as an offset into the file. For many cases this will be a very large number that will point to some huge offset in the file that doesn't actually exist; this is where the strange errors around BigTiff tag reading were coming from.

This PR fixes this. The architecture of how we were utilizing implicit classes to add method extensions to ByteReader for reading arrays is clever but a bit unmalleable (I might have designed it, or at least didn't change it in the past, so that's on me). The type class I introduced was the least invasive way to implement this that I could think of; perhaps not perfect but it gets the job done. It changes technically `public` API, but in reality the API is internal enough that I can't imagine the change effecting any users (though I will note the API change in the CHANGELOG)

### DIFF
--- a/raster-test/src/test/scala/geotrellis/raster/io/geotiff/BigTiffSpec.scala
+++ b/raster-test/src/test/scala/geotrellis/raster/io/geotiff/BigTiffSpec.scala
@@ -103,5 +103,11 @@ class BigTiffSpec extends FunSpec with RasterMatchers with GeoTiffTestUtils {
 
       assertEqual(actual.tile, expected.tile)
     }
+
+    it("should read a previously problematic big tiff") {
+      val tags = TiffTagsReader.read(geoTiffPath("bigtiff-marcuswr.tif"))
+      val e = tags.extent
+      e should be (Extent(-105.06398320198056, 40.743636546229, -105.05724549293515, 40.751667086819424))
+    }
   }
 }
--- a/raster/src/main/scala/geotrellis/raster/io/geotiff/reader/GeoTiffReader.scala
+++ b/raster/src/main/scala/geotrellis/raster/io/geotiff/reader/GeoTiffReader.scala
@@ -20,6 +20,7 @@ import geotrellis.raster._
 import geotrellis.raster.io.geotiff._
 import geotrellis.raster.io.geotiff.compression._
 import geotrellis.raster.io.geotiff.tags._
+import geotrellis.raster.io.geotiff.util._
 import geotrellis.vector.Extent
 import geotrellis.proj4.CRS
 import geotrellis.util.{ByteReader, Filesystem}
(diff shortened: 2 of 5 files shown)


jbouffard:
:+1:

PR #2144, opened 2017-04-19
ETL ADR

pomadchin:
- [ ] another version with a single json file
- [ ] split adrs into sep files
- [ ] fix urls

ADR v2 (just not to create tons of PR's and to keep history): https://github.com/pomadchin/geotrellis/blob/adr/etl-v2/docs/architecture/004-etl-pipeline.rst

Related to #1767

PR #2264, opened 2017-06-28
Spark streaming and Pipeline ADRs

pomadchin:
This PR includes ADRs refactor (adr per file) and includes following prs: #2144 and #1936 (ADR part)


echeipesh:
Thank you. The pipeline ADR is good work. Some things may change as we start to implement it but it states the problem well and looks like its going to be the route we're heading down. Any changes to ADR will come in with the PRs for the code.

PR #2275, opened 2017-07-07
DeflateCompression is Now Serializable

jbouffard:
This PR makes `DeflateCompression` serializable. This is needed as trying to use this form of compression in GeoPySpark fails.

```
Caused by: java.io.NotSerializableException: geotrellis.raster.io.geotiff.compression.DeflateCompression$
E                   Serialization stack:
E                   	- object not serializable (class: geotrellis.raster.io.geotiff.compression.DeflateCompression$, value: geotrellis.raster.io.geotiff.compression.DeflateCompression$@5a2b1895)
E                   	- field (class: geotrellis.raster.io.geotiff.GeoTiffOptions, name: compression, type: interface geotrellis.raster.io.geotiff.compression.Compression)
E                   	- object (class geotrellis.raster.io.geotiff.GeoTiffOptions, GeoTiffOptions(Tiled(256,256),geotrellis.raster.io.geotiff.compression.DeflateCompression$@5a2b1895,0,None))
E                   	- field (class: geopyspark.geotrellis.ProjectedRasterRDD$$anonfun$7, name: geotiffOptions$1, type: class geotrellis.raster.io.geotiff.GeoTiffOptions)
E                   	- object (class geopyspark.geotrellis.ProjectedRasterRDD$$anonfun$7, <function1>)
E                   	at org.apache.spark.serializer.SerializationDebugger$.improveException(SerializationDebugger.scala:40)
E                   	at org.apache.spark.serializer.JavaSerializationStream.writeObject(JavaSerializer.scala:46)
E                   	at org.apache.spark.serializer.JavaSerializerInstance.serialize(JavaSerializer.scala:100)
E                   	at org.apache.spark.util.ClosureCleaner$.ensureSerializable(ClosureCleaner.scala:295)
E                   	... 22 more
```

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/io/geotiff/compression/Compression.scala
+++ b/raster/src/main/scala/geotrellis/raster/io/geotiff/compression/Compression.scala
@@ -16,6 +16,6 @@
 
 package geotrellis.raster.io.geotiff.compression
 
-trait Compression {
+trait Compression extends Serializable {
   def createCompressor(segmentCount: Int): Compressor
 }
--- a/raster/src/main/scala/geotrellis/raster/io/geotiff/compression/DeflateCompression.scala
+++ b/raster/src/main/scala/geotrellis/raster/io/geotiff/compression/DeflateCompression.scala
@@ -27,7 +27,7 @@ import scala.collection.mutable
 import spire.syntax.cfor._
 
 object DeflateCompression extends Compression {
-  def createCompressor(segmentCount: Int): Compressor = 
+  def createCompressor(segmentCount: Int): Compressor =
     new Compressor {
       private val segmentSizes = Array.ofDim[Int](segmentCount)
       def compress(segment: Array[Byte], segmentIndex: Int): Array[Byte] = {


pomadchin:
My it makes sense to make Compression trait itself Serializable?

jbouffard:
Ah, yeah that makes more sense. Though, it is a bit weird that the only other compression type is serializable because it extends both `Compressor` and `Compression`. https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/io/geotiff/compression/NoCompression.scala#L21 In fact, `DelflateCompression` is the only de/compression strategy that isn't serializable.

jbouffard:
@pomadchin @echeipesh Could either one of you guys please review this PR when you get the chance? This needs to be merged so that I can finish https://github.com/locationtech-labs/geopyspark/pull/328

pomadchin:
Ok, in process...

jbouffard:
@pomadchin Thanks for reviewing this PR. I'm not using `Predictor`, and I can't think of reason we'd need it to be serializable now. Though, it might be good to make it serializable since everything else in that package will be.

echeipesh:
The `Predictor` seems to be just a bunch of primitive flags, so it should already be serializable. Also it doesn't look like its references through `GeoTiffOptions` which is what is dragging in the `Compressor`.

Issue #2281, opened 2017-07-14
Cannot create ValueReaders from URI


jpolchlo:
In the recent PR #2223, a means to create LayerReader instances from URIs was provided, but no matching methods exist for the creation of ValueReader.

PR #2286, opened 2017-07-17
ValueReaderProvier

echeipesh:
Closes: https://github.com/locationtech/geotrellis/issues/2281

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopLayerProvider.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopLayerProvider.scala
@@ -31,7 +31,7 @@ import java.net.URI
  * That support is provided by [[S3Attributestore]]
  */
 class HadoopLayerProvider extends AttributeStoreProvider
-    with LayerReaderProvider with LayerWriterProvider {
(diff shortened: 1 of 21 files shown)


spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopLayerProvider.scala
```diff
@@ -59,4 +59,13 @@ class HadoopLayerProvider extends AttributeStoreProvider
     val interval = params.getOrElse("interval", "4").toInt
     new HadoopLayerWriter(path, store, interval)
   }
+
+  def valueReader(uri: URI, store: AttributeStore): ValueReader[LayerId] = {
+    val _uri = trim(uri)
+    val path = new Path(_uri)
+    val params = UriUtils.getParams(_uri)
```
pomadchin: For `valueReader`, `layerWriter`, `attribuetStore` 

```scala
val _uri = trim(uri)
val path = new Path(_uri)
val params = UriUtils.getParams(_uri)
```

Mb that can be sugared?

echeipesh: Not really user functionality, so I wanted to leave it as bare bones as possible

Issue #2287, opened 2017-07-18
Time-series Polygonal Summary operation over spatiotemporal raster layers


lossyrob:
Currently we have a way to take a `TileLayerRDD[SpatialKey]`, do a polygonal summary over it e.g. `layer.polygonalMax(polygon)`, and get a single answer.

This feature would enable a `TileLayerRDD[SpaceTimeKey]` to do a similar polygonal summary _per time period_, and return a `Map[ZonedDateTime, R]` where `R` is the result type.

For example,
```scala
val layer: TileLayerRDD[SpaceTimeKey] = ???
val polygon: Polygon = ??? // or multipolygon, anything the current summaries support
val result: Map[ZonedDateTime, Histogram[Double]] = layer.polygonalHistogram(polygon) // or any other currently supported polygonal summary operation
```

jamesmcclain:
I will start on this.  I can be handed off as appropriate if situation changes.

PR #2288, opened 2017-07-19
Time Series

jamesmcclain:
screenshot from 2017-07-20 17 09 38

- [x] Unit Tests

Fixes locationtech/geotrellis#2287


jamesmcclain:
@CloudNiner and @fungjj92 asked to be tagged on this pull request.  See also locationtech-labs/geopyspark#377.

spark/src/main/scala/geotrellis/spark/timeseries/RDDTimeSeriesMethods.scala
```diff
@@ -0,0 +1,185 @@
...
+  ): Map[ZonedDateTime, Double] =
+    sumSeries(List(polygon), options)
+
+  def sumSeries(
+    polygon: Geometry
+  ): Map[ZonedDateTime, Double] =
+    sumSeries(List(polygon), Options.DEFAULT)
+
+  def sumSeries(
+    polygons: Traversable[Geometry],
```
lossyrob: Having this be Geometries makes the code less typesafe. If we pass in other geometries, we get a match error that is potentially not very descriptive. We should change this to only accept a Polygon or MultiPolygon, both versions (the user can create the MultiPolygon themselves if needed).

jamesmcclain:
Okay, I have modified it.

spark/src/main/scala/geotrellis/spark/timeseries/RDDTimeSeriesMethods.scala
```diff
@@ -0,0 +1,194 @@
...
+  ): Map[ZonedDateTime, Double] =
+    meanSeries(List(polygon), options)
+
+  def meanSeries(
+    polygon: MultiPolygon
+  ): Map[ZonedDateTime, Double] =
+    meanSeries(polygon, Options.DEFAULT)
+
+  def meanSeries(
+    polygons: Traversable[MultiPolygon]
```
lossyrob: Sorry, I should have been more clear. The overloads that should exist are `Polygon` and `MultiPolygon`, not `MultiPolygon` and `Traversable[MultiPolygon]`.

If a user has a single polygon, which seems like a pretty regular case, then they can call the method with just the polygon.

If a user has a multipolygon, same thing.

If the user has a sequence of polygons or sequence of multipolygons, then they can wrap/union them into a multipolygon before calling this method.

jamesmcclain: Okay, there are only methods for `Polygon` and `MultiPolygon`.

echeipesh:
My only question here is about the interface, should this be returning `Map` or `RDD`, `.collect.toMap` could be the client calls and returning an `RDD[ZonedDateTime, V]` would be the first case of its usage. Hard to say right now, this does what it needs to do at the moment.

I think the method extensions need to be `@Experimental` so we have the freedom to change them as we use them more but otherwise it looks good.

PR #2289, opened 2017-07-19
Estimate partitions number basing on a desired partition size

pomadchin:
This PR also introduces a new default `S3GeoTiffRDD` behaviour.
The question is: should we change the default reader behaviour? Or it's still should be an optional feature.


echeipesh:
To answer the question: The default behavior on principle should "just work" as much as possible. 

That is not the case currently, we pretty much require partition and window size tuning before we can expect good performance. So in such a case the defaults can be changed to provide the correct behavior.

pomadchin:
Fixed in #2296

PR #2293, opened 2017-07-20
Use getDouble instead of get in DoubleHistogramSummary

aaron-santos:
Fixes polygonal summarization of double-valued tiles


lossyrob:
@aaron-santos seems like you did this in the github editor, which puts the email as noreply@github.com.

Can you make the commit from your local clone and do the signoff `-s` on the commit? That way the Eclipse IP Lords remain pleased.

pomadchin:
@aaron-santos `git commit --amend -s && git push origin -f patch-1`

PR #2296, opened 2017-07-24
Estimate partitions number basing on GeoTiff segments

pomadchin:
- [x] Poor performance fix
- [x] Tests (new & fix old)
- [x] Better repartition in border cases (size / windows size in stripped / tiles cases)
- [x] Rethink multiband `fromSegments` function, see https://github.com/locationtech/geotrellis/pull/2296/commits/d95266521da63559c0be4f33f01a402ababe9368
- [x] Rethink a singleband `fromSegments` function

The aim of two last points is to reduce memory usage


raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
```diff
@@ -318,12 +336,114 @@ abstract class GeoTiffMultibandTile(
     tiles
   }
 
+  private def _subsetBandsFromSegments(
+    bandSequence: Seq[Int],
+    segmentIds: Traversable[Int],
+    deinterleaveBitSegment: (GeoTiffSegment, Int, Int, Int, Traversable[Int]) => Array[Array[Byte]],
```
echeipesh: Thats a rough signature. Is there any way to sugar it up? Its used a couple of times, maybe type alias where we can hang a comment?

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
```diff
@@ -219,6 +219,14 @@ abstract class GeoTiffMultibandTile(
       (bytes, bandCount, bytesPerSample, _) => GeoTiffSegment.deinterleave(bytes, bandCount, bytesPerSample)
     ).toVector
 
+  def bandsFromSegments(segmentIds: Traversable[Int]): Vector[Tile] =
```
echeipesh: Should be `private` as not to add to API ?

spark/src/main/scala/geotrellis/spark/io/RasterReader.scala
```diff
@@ -123,8 +136,12 @@ object RasterReader {
       val crs = options.crs.getOrElse(geotiff.crs)
       (TemporalProjectedExtent(raster.extent, crs, time), raster.tile)
     }
-  }
 
+    def readSegments(ids: Traversable[Int], info: GeoTiffReader.GeoTiffInfo, options: Options) = {
```
echeipesh: Needs return type

s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffInfoReader.scala
```diff
@@ -0,0 +1,190 @@
...
+import geotrellis.raster.io.geotiff.reader.GeoTiffReader.GeoTiffInfo
+import geotrellis.util.LazyLogging
+
+import com.amazonaws.services.s3.model.ListObjectsRequest
+import org.apache.spark.SparkContext
+import org.apache.spark.rdd.RDD
+
+import scala.collection.mutable
+
+case class S3GeoTiffInfoReader(
```
echeipesh: By the way this is used it looks like it could be decomposed into a companion object with some methods on it, that would probably be safer than a case class that potentially carries client around.

spark/src/main/scala/geotrellis/spark/io/RasterReader.scala
```diff
@@ -40,6 +41,11 @@ import spire.syntax.cfor._
 trait RasterReader[-O, R] extends Serializable {
   def readFully(byteReader: ByteReader, options: O): R
   def readWindow(byteReader: StreamingByteReader, pixelWindow: GridBounds, options: O): R
+
+  /** Read segments iterator, each segment into a single R */
+  def readSegmentsIterator(gbs: Array[GridBounds], info: GeoTiffReader.GeoTiffInfo, options: O): Iterator[R]
+  /** Reads all segments into one R */
+  def readSegments(gbs: Array[GridBounds], info: GeoTiffReader.GeoTiffInfo, options: O): Option[R]
```
pomadchin: Two strategies to read segments, which is better for our use cases?

spark/src/main/scala/geotrellis/spark/io/GeoTiffInfoReader.scala
```diff
@@ -0,0 +1,162 @@
...
+
+  /**
+    * Function calculates a split of segments, to minimize segments reads.
+    *
+    * Returns RDD of pairs: ((String, GeoTiffInfo), Array[GridBounds])
+    * where GridBounds are gird bounds of a particular segment,
+    * each segment can only be in a single partition.
+    * */
+  def segmentsByPartitionBytes(partitionBytes: Long = Long.MaxValue, maxTileSize: Option[Int] = None)
+                              (implicit sc: SparkContext): RDD[((String, GeoTiffInfo), Array[GridBounds])] = {
```
pomadchin: We can still operate with segment refs

s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffRDD.scala
```diff
@@ -50,24 +52,30 @@ object S3GeoTiffRDD extends LazyLogging {
...
+    *                       1024 by defaut.
     * @param numPartitions  How many partitions Spark should create when it repartitions the data.
     * @param partitionBytes Desired partition size in bytes, at least one item per partition will be assigned.
                             This option is incompatible with the maxTileSize option.
+    *                       128 Mb by default.
     * @param chunkSize      How many bytes should be read in at a time.
     * @param delimiter      Delimiter to use for S3 objet listings. See
     * @param getS3Client    A function to instantiate an S3Client. Must be serializable.
+    * @param persistLevel   A spark persist sotrage level, MEMORY_ONLY by default (similar to RDD.cache())
+    * @param bySegments     Minimize segments reads, read input data by segments.
```
echeipesh: Semver doesn't give a lot of room here, can't change the `Options` object.

Lets assume that `persistLevel` is `MEMORY_ONLY`, if there is not enough memory to read the tiff headers, its probably a lost cause anyway. Also re-reading them twice if memory cache is purged is good soft failover.

Lets keep the `maxWindowSize=None` as default, that will be the flag for using the segment wise read. Setting it to some value will be the flag for reading exactly those window chunks, as before.

This means that we will never read full GeoTiff again unless its smaller than the `maxWindowSize` value. It seems that that GeoTiffs small enough to be appropriate for RDD record basically don't get written, so this is good default

s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffRDD.scala
```diff
@@ -105,9 +113,32 @@ object S3GeoTiffRDD extends LazyLogging {
...
+        val segmentsCount = segments.count.toInt
+
+        logger.info(s"repartition into ${segmentsCount} partitions.")
+
+        val repartition =
+          if(segmentsCount > segments.partitions.length) segments.repartition(segmentsCount)
+          else segments
+
+        val result = repartition.flatMap { case ((key, md), segmentIndices) =>
+          rr.readWindows(segmentIndices, md, options).map { case (k, v) =>
```
echeipesh: `segmentIndices` is stale variable name, I guess this is `segmentBounds` now ?

spark/src/main/scala/geotrellis/spark/io/GeoTiffInfoReader.scala
```diff
@@ -0,0 +1,162 @@
...
+import geotrellis.raster.io.geotiff.reader.GeoTiffReader.GeoTiffInfo
+import geotrellis.util.LazyLogging
+import geotrellis.raster.GridBounds
+
+import org.apache.spark.SparkContext
+import org.apache.spark.rdd.RDD
+
+import scala.collection.mutable
+
+trait GeoTiffInfoReader extends LazyLogging {
```
echeipesh: Why not just make the the `private [geotrellis] trait` ?

s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffRDD.scala
pomadchin: So you think for by segment reads we don't need that window option by default? (at least in 1.2); But that can be a good feature to try to pack segments into windows we want.

echeipesh: It seems like the actual size of the window is not important there because the partition size will control it. If there are too few windows in partitions, you are packing more. You never splice segments together, so the effect of the window size is just to provide logic to keep the segments spatially related. So I would assume that it can just be hard coded to 1024 for segment reads.

Issue #2294, opened 2017-07-21
Remove `.collectMetadata` from TileRDDReproject


lossyrob:
See the `collectMetadata` calls here:

https://github.com/locationtech/geotrellis/blob/a0280ebd51e3bec1a765c623138684e8a0fcd966/spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala#L123-L138

Ideally we'd like to derive the new, reprojected metadata from the old metadata, instead of having to we create a `TileLayerMetadata[K] => TileLayerMetadata[K]` function that takes in a target CRS and produces the new metadata directly without having to perform a Spark action. This function would probably belong on TileLayerMetadata, or be an implicit method that hangs off of any `M` that has the necessary components (LayoutDefinition, Bounds, Extent).

The source RDD metadata exists here: https://github.com/locationtech/geotrellis/blob/a0280ebd51e3bec1a765c623138684e8a0fcd966/spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala#L60, so we have access to it.

We also have the ability to reproject GridExtents (like the LayoutDefinition of the source metadata): https://github.com/locationtech/geotrellis/blob/a0280ebd51e3bec1a765c623138684e8a0fcd966/raster/src/main/scala/geotrellis/raster/reproject/ReprojectRasterExtent.scala#L55

The one concern here is the `cellSize`: If we reproject the grid extent, we get a cell size that is uniform over the entire bounds of the GridExtent. In the `collectMetadata` approach, we are iterating through each individual reprojected RasterExtent for each tile, which is locally reprojected and therefore can have varying cellSizes. The `collectMetadata` will choose the smallest cell size for the new metadata, which is fed into the eventual `tileToLayout` call, so that if there is a portion of the GridExtent that has significantly smaller resolution on reproject than would be calculated for the overall grid extent, that resolution is not degraded by taking some overall cellSize.

lossyrob:
@jamesmcclain

lossyrob:
@echeipesh should we close this?

echeipesh:
Yup, resolved by https://github.com/locationtech/geotrellis/pull/2425

PR #2300, opened 2017-07-26
Remove `.collectMetadata` From TileRDDReproject

jamesmcclain:
Fixes locationtech/geotrellis#2294

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
+++ b/spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
@@ -81,46 +81,56 @@ object TileRDDReproject {
           }
       }
 
+    val (sampleKey, BufferedTile(sampleTile, sampleGridBounds)) = bufferedTiles.first() // XXX
+    val tileCols = sampleTile.cols // XXX metadata.layout.tileCols
+    val tileRows = sampleTile.rows // XXX metadata.layout.tileRows
+    val KeyBounds(SpatialKey(keyColMin, keyRowMin), SpatialKey(keyColMax, keyRowMax)) = metadata.bounds
+    val lb = scala.collection.mutable.ListBuffer.empty[K]
+
+    }
+
     val layerInfo: (Extent, CellSize, KeyBounds[K]) =
       bufferedTiles
         .mapPartitions({ partition =>
           val transform = Transform(crs, destCrs)
           val inverseTransform = Transform(destCrs, crs)
 
-          partition.map { case (key, BufferedTile(tile, gridBounds)) =>
+          partition.map { case (key, BufferedTile(_, _)) =>
             val innerExtent = mapTransform(key)
-            val innerRasterExtent = RasterExtent(innerExtent, gridBounds.width, gridBounds.height)
+            val innerRasterExtent = RasterExtent(innerExtent, sampleGridBounds.width, sampleGridBounds.height)
             val outerGridBounds =
               GridBounds(
-                -gridBounds.colMin,
-                -gridBounds.rowMin,
-                tile.cols - gridBounds.colMin - 1,
-                tile.rows - gridBounds.rowMin - 1
+                -sampleGridBounds.colMin,
+                -sampleGridBounds.rowMin,
+                tileCols - sampleGridBounds.colMin - 1,
+                tileRows - sampleGridBounds.rowMin - 1
               )
             val outerExtent = innerRasterExtent.extentFor(outerGridBounds, clamp = false)
 
             val window =
               if(options.matchLayerExtent) {
-                gridBounds
+                sampleGridBounds
               } else {
                 // Reproject extra cells that are half the buffer size, as to avoid
                 // any missed cells between tiles.
                 GridBounds(
-                  gridBounds.colMin / 2,
-                  gridBounds.rowMin / 2,
-                  (tile.cols + gridBounds.colMax - 1) / 2,
-                  (tile.rows + gridBounds.rowMax - 1) / 2
+                  sampleGridBounds.colMin / 2,
+                  sampleGridBounds.rowMin / 2,
+                  (tileCols + sampleGridBounds.colMax - 1) / 2,
+                  (tileRows + sampleGridBounds.rowMax - 1) / 2
                 )
               }
 
-            val Raster(newTile, newExtent) =
-              tile.reproject(outerExtent, window, transform, inverseTransform, rasterReprojectOptions)
+            val rasterExtent = RasterExtent(outerExtent, tileCols, tileRows)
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
```diff
@@ -81,6 +82,76 @@ object TileRDDReproject {
           }
       }
 
+    val (sampleKey, BufferedTile(sampleTile, sampleGridBounds)) = bufferedTiles.first() // XXX
```
jamesmcclain: Need `sampleKey` because it is not possible (as far as I know) to construct an object of type K, and I need such an object in order to be sample to build the list of keys (see line 94).

Need `sampleTile` because in the unit tests, the metadata reports that tiles are 50x50, but the actual tiles are of variable size and always larger than 50x50.  The fixed tile size taken from the first tile works for the unit tests, but I do not know if this is a universally viable strategy.  A "good enough" tile size is needed in order for the cellSize and tile-contents checks in the unit tests to pass.

Need `sampleGridBounds` for much the same reason as `sampleTile`.

spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
```diff
@@ -60,7 +60,8 @@ object TileRDDReproject {
     metadata: TileLayerMetadata[K],
     destCrs: CRS,
     targetLayout: Either[LayoutScheme, LayoutDefinition],
-    options: Options
+    options: Options,
+    sc: Option[SparkContext] = None
```
echeipesh: Can't add parameter in order to keep the API. But you can grab `SparkContext` from `bufferedTiles.context`.

spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
echeipesh: I think I see what you're saying. You should be able to avoid this by just using `SpatialKey` and performing the reproject on that. You'll get the `cellSize`, `Extent`, and `KeyBounds[SpatialKey]`. The last can be set on the `layer.metadata` using [`KeyBounds.setSpatialComponent`](https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/KeyBounds.scala#L140). That makes sense, since reprojection is spatial only operation. 

Tile dimensions are in the `metadata.layout` has the tile size, like you're using it. Does that satisfy the question about tile size?

spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
```diff
@@ -81,6 +82,76 @@ object TileRDDReproject {
...
+            val windowExtent = rasterExtent.extentFor(window)
+            val windowRasterExtent = RasterExtent(windowExtent, window.width, window.height)
+            val targetRasterExtent = ReprojectRasterExtent(windowRasterExtent, transform)
+
+            (key, targetRasterExtent.extent, targetRasterExtent.cols, targetRasterExtent.rows)
+          }
+        })
+        .map({ case (key, extent, cols, rows) =>
+          (extent, CellSize(extent, cols, rows), KeyBounds(key, key))
+        })
```
echeipesh: Not sure how much this will save but this can easily just be what the previous step returns and avoid the map. CPU cycle saves is CPU cycle earned, I suppose.

jamesmcclain:
> Tile dimensions are in the metadata.layout has the tile size, like you're using it. Does that satisfy the question about tile size?

In the `TileRDDReproject` tests, `metadata.layer.tile(Cols|Rows)` reports 50x50 which is not close enough to the actual sizes for those tests to pass.

spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
jamesmcclain: Removed

spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
jamesmcclain: In the `TileRDDReproject` tests, `metadata.layer.tile(Cols|Rows)` reports 50x50 which is not close enough to the actual sizes for those tests to pass.

At present, I still need `sampleTile` and `sampleGridBounds`.

spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
jamesmcclain: Me neither.

jamesmcclain:
> Tile dimensions are in the `metadata.layout` has the tile size, like you're using it. Does that satisfy the question about tile size?

In the `TileRDDReproject` tests, `metadata.layer.tile(Cols|Rows)` reports 50x50 which is not close enough to the actual sizes for those tests to pass.

```
[info] TileRDDReprojectSpec:
[info] TileRDDReproject
[info] - should reproject a raster split into tiles the same as the raster itself: constant border and Bilinear *** FAILED ***
[info]   -78.78159268682867 was not -78.7746204948163 plus or minus 1.0E-5 (TileRDDReprojectSpec.scala:82)
[info] - should reproject a raster split into tiles the same as the raster itself: dynamic border and Bilinear *** FAILED ***
[info]   -78.78159268682867 was not -78.7746204948163 plus or minus 1.0E-5 (TileRDDReprojectSpec.scala:82)
[info] - should reproject a raster split into tiles the same as the raster itself: constant border and NearestNeighbor *** FAILED ***
[info]   -78.78159268682867 was not -78.7746204948163 plus or minus 1.0E-5 (TileRDDReprojectSpec.scala:82)
[info] - should reproject a raster split into tiles the same as the raster itself: dynamic border and NearestNeighbor *** FAILED ***
[info]   -78.78159268682867 was not -78.7746204948163 plus or minus 1.0E-5 (TileRDDReprojectSpec.scala:82)
[info] Reprojected with the same scheme and CRS
[info] - should tile with minimum number of tiles
[info] Run completed in 13 seconds, 974 milliseconds.
[info] Total number of tests run: 5
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 1, failed 4, canceled 0, ignored 0, pending 0
[info] *** 4 TESTS FAILED ***
```

echeipesh:
Ah, I see. But would not the original code also collect reprojected extents from 50x50 source tiles?

jamesmcclain:
No, it did not.

jamesmcclain:
Please do not close this PR.  If what is here does not pass muster, I will change this to remove the extra collectMetadata call.

PR #2303, opened 2017-08-01
Add Terraform EMR deployment

pomadchin:
Since [Terraform PR with spot instances support](https://github.com/terraform-providers/terraform-provider-aws/pull/1071) is still not merged in probably we don't want to merge it.

- [x] Test Spark UI
- [x] Terraform release with EMR spot instances support

### Env variables

* KEY_PAIR_FILE
* TF_VAR_pem_path

Depends on #2312


pomadchin:
Without ingest example: 

```
import geotrellis.raster._
import geotrellis.vector._
import geotrellis.vector.io._
import geotrellis.raster.io.geotiff._
import geotrellis.raster.render._
import geotrellis.raster.resample._
import geotrellis.raster.reproject._
import geotrellis.proj4._
import geotrellis.spark._
import geotrellis.spark.ingest.MultibandIngest
import geotrellis.spark.io._
import geotrellis.spark.io.file._
import geotrellis.spark.io.hadoop._
import geotrellis.spark.io.s3._
import geotrellis.spark.io.index._
import geotrellis.spark.pyramid._
import geotrellis.spark.reproject._
import geotrellis.spark.render._
import org.apache.spark.SparkContext

val bucket = "geotrellis-test"
val prefix = "daunnc/LC81750172014163LGN00_LOW4.TIF"

implicit val _sc = sc

val inputRdd: RDD[(ProjectedExtent, MultibandTile)] = S3GeoTiffRDD.spatialMultiband(bucket, prefix, S3GeoTiffRDD.Options(maxTileSize = Some(1024)))

MultibandIngest[ProjectedExtent, SpatialKey](
    sourceTiles = inputRdd,
    destCRS = WebMercator,
    layoutScheme = ZoomedLayoutScheme(WebMercator, tileSize = 256),
    pyramid = true
  ) { case (rdd, zoomHere) =>  println(zoomHere) }
```

fosskers:
Both the examples posted above work as expected in Zeppelin.

jamesmcclain:
https://github.com/terraform-providers/terraform-provider-aws/pull/1071 has been merged

scripts/emr/terraform/emr-spark.tf
```diff
@@ -0,0 +1,69 @@
+# Marks AWS as a resource provider.
+provider "aws" {
+  access_key = "${var.access_key}"
+  secret_key = "${var.secret_key}"
+  region     = "${var.region}"
+}
+
+# `aws_emr_cluster` is built-in to Terraform. We name ours `emrSparkCluster`.
+resource "aws_emr_cluster" "emrSparkCluster" {
```
jamesmcclain: Minor style point: Camel case seems to be less common than separating words with hyphens and use all lower case (e.g. `"emr-spark-cluster"`).

fosskers: That's the case, yeah.

fosskers:
> I don't have a problem with merging this, but if it is going to become the basis for many/most/all deployments, perhaps in the future it should be moved into its own repository then depended on as a terraform module from each place where it is used.

Especially since the config itself will go mostly unchanged between projects. Rather than cargo-cult it around, we could just use the model system, yeah.

pomadchin:
Complete test with pyramid ingest, read, stitch and writing everything into s3 bucket.

```scala
%spark

import geotrellis.raster._
import geotrellis.vector._
import geotrellis.vector.io._
import geotrellis.raster.io.geotiff._
import geotrellis.raster.render._
import geotrellis.raster.resample._
import geotrellis.raster.reproject._
import geotrellis.proj4._
import geotrellis.spark._
import geotrellis.spark.ingest.MultibandIngest
import geotrellis.spark.io._
import geotrellis.spark.io.file._
import geotrellis.spark.io.hadoop._
import geotrellis.spark.io.s3._
import geotrellis.spark.io.index._
import geotrellis.spark.pyramid._
import geotrellis.spark.reproject._
import geotrellis.spark.render._
import org.apache.spark.SparkContext
import org.apache.hadoop.fs.Path

val bucket = "geotrellis-test"
val prefix = "daunnc/LC81750172014163LGN00_LOW4.TIF"

implicit val _sc = sc

val inputRdd: RDD[(ProjectedExtent, MultibandTile)] = S3GeoTiffRDD.spatialMultiband(bucket, prefix)

val attributeStore = S3AttributeStore("geotrellis-test", "daunnc/chatta/catalog")
val writer = S3LayerWriter(attributeStore)

MultibandIngest[ProjectedExtent, SpatialKey](
  sourceTiles = inputRdd,
  destCRS = WebMercator,
  layoutScheme = ZoomedLayoutScheme(WebMercator, tileSize = 256),
  pyramid = true
) { case (rdd, zoom) =>
  val layerId = LayerId("LC81750172014163LGN00_LOW4", zoom)
  println(s"zoom: $zoom")

  if(attributeStore.layerExists(layerId))
  new S3LayerManager(attributeStore).delete(layerId)

  writer.write(layerId, rdd, ZCurveKeyIndexMethod)
}

val reader = S3LayerReader(attributeStore)
val layerId = LayerId("LC81750172014163LGN00_LOW4", 10)
val rdd = reader.read[SpatialKey, MultibandTile, TileLayerMetadata[SpatialKey]](layerId)
val raster = rdd.stitch
val tiff = GeoTiff(raster, rdd.metadata.crs)

tiff.write(new Path("s3://geotrellis-test/daunnc/chatta/test-avro.tif"))
```

pomadchin:
@jpolchlo would this fix be fine or you want to have a different API?

jpolchlo:
Man, it would be amazing if it would just pull the necessary information from `Version.scala` and `Environment.scala`.  That might be a big ask.  It may be that environment variables are the best we can do?

pomadchin:
@jpolchlo i made them to behave similar to Environment.scala, we just need to keep it updated;

```
GEOTRELLIS_VERSION ?= 1.2.0
GEOTRELLIS_VERSION_SUFFIX ?= -SNAPSHOT
```

pomadchin:
I noticed that they have a [new release](https://github.com/terraform-providers/terraform-provider-aws/releases/tag/v1.1.0), @jpolchlo if you still work with it can you double check it, and probably to update this PR with necessary changes, as probably you used it more than anybody else.

fosskers:
This is the release we've been waiting for.

jpolchlo:
I'll be happy to check this when I get a minute.

scripts/emr/terraform/variables.tf.json
```diff
@@ -0,0 +1,33 @@
+{
```
jamesmcclain: Why is this a json file instead of an actual terraform file (`.tf`)?

scripts/emr/terraform/variables.tf.json
```diff
@@ -0,0 +1,33 @@
+{ 
+  "variable": {
+    "access_key": { 
+        "description": "From your `~/.aws/credentials"
+    },
```
jamesmcclain: Why are the access key and secret key needed?  Shouldn't those be picked up from the environment?

scripts/emr/terraform/variables.tf.json
```diff
@@ -0,0 +1,33 @@
...
+    "secret_key": {
+        "description": "From your `~/.aws/credentials"
+    },
+    "pem_path": {
```
jamesmcclain: I think that it would be more convenient to specify `key_name` rather than the path to the key on the filesystem.

scripts/emr/terraform/variables.tf.json
pomadchin: To parse `json` we can use `jq`, but to use `.tf` we need to parse it somehow ugly.

scripts/emr/terraform/variables.tf.json
pomadchin: Look into two last @jpolchlo commits, he already introduced this feature. I prefer to configure env using configuration files though.

scripts/emr/terraform/variables.tf.json
pomadchin: by introducing a `key_name` you want to introduce a `key_path`? Or to hardcode path to aws key?

jamesmcclain:
I can conform that binary download from the Terraform website (`0.10.7` at time of writing) does have support for spot prices on EMR.  On a fresh virtual machine with that version of Terraform installed, the aws plugin was downloaded and provisioning (mostly) succeeded.

The AWS console showed that the cluster was brought up successfully, but the `terraform` command exited with an error after displaying this message in a continuous loop for a while:
```
aws_emr_cluster.emr-spark-cluster: Still creating... (9m50s elapsed)
aws_emr_cluster.emr-spark-cluster (remote-exec): Connecting to remote host via SSH...
aws_emr_cluster.emr-spark-cluster (remote-exec):   Host: ec2-54-160-235-90.compute-1.amazonaws.com
aws_emr_cluster.emr-spark-cluster (remote-exec):   User: hadoop
aws_emr_cluster.emr-spark-cluster (remote-exec):   Password: false
aws_emr_cluster.emr-spark-cluster (remote-exec):   Private key: true
aws_emr_cluster.emr-spark-cluster (remote-exec):   SSH Agent: false
```

[This issue](https://github.com/hashicorp/terraform/issues/15219) might be related.

I am not sure that we actually need to use `remote-exec` at all.  If `remote-exec` is the motivation for asking for the location of the ssh key on the filesystem, then perhaps two birds can be had with one stone.

jamesmcclain:
Conclusion: `0.10.7` has EMR spot price support, but the `remote-exec` logic may or may not be compatible (I experienced a problem, but I do not know the cause).

scripts/emr/terraform/variables.tf.json
jamesmcclain: Why do we need to parse a file?  What is that in service of?

scripts/emr/terraform/variables.tf.json
jamesmcclain: Looking from a strictly behavioral point of view, all I know is that I was prompted for those things when I tried to run `terraform apply`.  Providing them through environment variables is possible (I did it), but why are they needed at all?  The AWS credentials can be picked up from the environment (please see the terraform configuration [here](https://github.com/geodocker/geodocker-jupyter-geopyspark)).

scripts/emr/terraform/variables.tf.json
jamesmcclain: `key_name` can be brought in through a variable.  It would replace `key_path`.  Please see the terraform configuration [here](https://github.com/geodocker/geodocker-jupyter-geopyspark).

scripts/emr/terraform/variables.tf.json
pomadchin: To parse `KEY_PAIR_FILE` from file, if it's necessary. `pem` file is required to have a convient way to access master node.

scripts/emr/terraform/variables.tf.json
pomadchin: @jamesmcclain your links in both last comments are broken, can you fix them?

scripts/emr/terraform/variables.tf.json
pomadchin: @jamesmcclain sry your link is broken

pomadchin: Btw I don't use global variables, would it pick up my `.aws/credentials`?

jamesmcclain: I was able to click on it successfully.  https://github.com/geodocker/geodocker-jupyter-geopyspark

scripts/emr/terraform/variables.tf.json
jamesmcclain: https://github.com/geodocker/geodocker-jupyter-geopyspark

scripts/emr/terraform/variables.tf.json
pomadchin: Ah i thought it was a link on their docs.

scripts/emr/terraform/variables.tf.json
pomadchin: Ok, to answer more complete: there is set of useful commands in [this Makefile](https://github.com/pomadchin/geotrellis/blob/c2b3443fd68fe5c5fcb70f296df00dc51d3d5dcb/scripts/emr/Makefile#L29-L53), that's why all these things are required.

scripts/emr/terraform/variables.tf.json
jamesmcclain: I am referring in particular to [this](https://github.com/geodocker/geodocker-jupyter-geopyspark/tree/799bc76957c8370d728d6344ed5d8db940744c92/emr-nodocker/terraform) and [this](https://github.com/geodocker/geodocker-jupyter-geopyspark/tree/799bc76957c8370d728d6344ed5d8db940744c92/terraform).  (Be advised that I have a PR up right now will change the directory structure.)

scripts/emr/terraform/variables.tf.json
jamesmcclain: Okay

spark-etl/build.sbt
```diff
@@ -4,17 +4,29 @@ name := "geotrellis-spark-etl"
...
-  case "application.conf" => MergeStrategy.concat
-  case "META-INF/MANIFEST.MF" => MergeStrategy.discard
-  case "META-INF\\MANIFEST.MF" => MergeStrategy.discard
-  case "META-INF/ECLIPSEF.RSA" => MergeStrategy.discard
-  case "META-INF/ECLIPSEF.SF" => MergeStrategy.discard
+  case "reference.conf" | "application.conf"  => MergeStrategy.concat
+  case "META-INF/MANIFEST.MF" | "META-INF\\MANIFEST.MF" => MergeStrategy.discard
+  case "META-INF/ECLIPSEF.RSA" | "META-INF/ECLIPSEF.SF" => MergeStrategy.discard
   case _ => MergeStrategy.first
 }
```
metasim: FWIW, we used to have an assembly strategy like this for our Zeppelin integration, but it needed revising as some config files were getting dropped. Here's what we have now. Can't remember all the details, but it's closer to what the defaults are in the most recent `sbt-assembly` settings:

```scala
    assemblyMergeStrategy in assembly := {
      case "logback.xml" ⇒ MergeStrategy.singleOrError
      case x if Assembly.isConfigFile(x) ⇒ MergeStrategy.concat
      case PathList(ps @ _*) if Assembly.isReadme(ps.last) || Assembly.isLicenseFile(ps.last) ⇒
        MergeStrategy.rename
      case PathList("META-INF", xs @ _*) ⇒
        (xs map { _.toLowerCase }) match {
          case ("manifest.mf" :: Nil) | ("index.list" :: Nil) | ("dependencies" :: Nil) ⇒
            MergeStrategy.discard
          case ps @ (x :: _) if ps.last.endsWith(".sf") || ps.last.endsWith(".dsa") ⇒
            MergeStrategy.discard
          case "plexus" :: _ ⇒
            MergeStrategy.discard
          case "services" :: _ ⇒
            MergeStrategy.filterDistinctLines
          case ("spring.schemas" :: Nil) | ("spring.handlers" :: Nil) ⇒
            MergeStrategy.filterDistinctLines
          case ("maven" :: rest ) if rest.lastOption.exists(_.startsWith("pom")) ⇒
            MergeStrategy.discard
          case _ ⇒ MergeStrategy.deduplicate
        }

      case _ ⇒ MergeStrategy.deduplicate
    }
```

The first time you try it you're likely to get a lot of `deduplicate` errors... but this is good, as in our case found additional bugs relating to conflicting versions or identifying the need for additional shading rules.

metasim: PS: One of the primary things that would get trashed with the `MergeStrategy.first` default was `META-INF/services` files. The version above makes sure they get merged.

pomadchin: @metasim can you cllarify a bit where do you merge `META-INF/services` in your assembly reules? What problems did you have with the simplier (and not perfect) version?

I believe you had smth more than just an sbt-assembly build sbt file, it makes sense; and it's definitely a good point to add this generic assembly strategy.

metasim: The relevant line is 
```
    case "services" :: _ ⇒
            MergeStrategy.filterDistinctLines
```

In our case we had to live *months* without an "Executors" page in the Spark UI associated with the Zeppelin driver. Finally got fed up and hunted it down to a particular Servlet endpoint not getting registered because of some SPI (I forget which one) registration getting smashed by some other one. Only happened when we pulled our über jar into Zeppelin; all was fine in `spark-submit`, etc.

pomadchin:
@metasim awesome, thank you!

spark-etl/build.sbt
metasim: After the hours and hours we lost, I now think `MergeStrategy.first` or `MergeStrategy.discard` as wildcard defaults are evil. Use `MergeStrategy.deduplicate` to really understand what's getting dropped, and if you're OK with it then add a rule, filter the jar, or shade it.

metasim:
Despite what the `sbt-assembly` does say, I think it's better to use `assemblyExcludedJars in assembly` than to `MergeStrategy.first`

spark-etl/build.sbt
pomadchin: Well, it depends, we use merge by default for services meta-inf; it didn't work for you?

pomadchin:
@metasim thx for feedback; we definitely need to add your more generic rules

PR #2332, opened 2017-08-16
Added New Section to README

jbouffard:
This PR adds the "Using GeoTrellis in Other Languages" section to the `README`. The purpose of this new section is to inform users that an alternative to GeoTrellis, GeoPySpark, exists if they are unable to use Scala.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,13 +22,15 @@ fast batch processing of large raster data sets.
 Please visit the **[project site](http://geotrellis.io)**
 for more information as well as some interactive demos.
 
-#### Using GeoTrellis in Other Languages
+#### GeoTrellis with Python
 
-If you are interested in GeoTrellis but are unable to work in Scala, then
-there is an alternative available,
-[GeoPySpark](http://github.com/locationtech-labs/geopyspark). GeoPySpark
-is a Python bindings library for GeoTrellis and can do many (but not
-all) of the operations present in GeoTrellis.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,6 +22,16 @@ fast batch processing of large raster data sets.
...
 for more information as well as some interactive demos.
 
+#### Using GeoTrellis in Other Languages
+
+If you are interested in GeoTrellis but are unable to work in Scala, then
+there is an alternative available,
+[GeoPySpark](http://github.com/locationtech-labs/geopyspark). GeoPySpark
+is a Python bindings library for GeoTrellis and can do many (but not
+all) of the operations present in GeoTrellis. While not as feature rich,
+GeoPySpark can be integrated with other tools, such as Jupyter, which
```
rossbernet: "GeoPySpark can be integrated with other tools in the Python ecosystem, such as NumPy, SciPy, and Jupyter notebooks. Many GeoPySpark tutorials have been developed that leverage the visualization capability of [GeoNotebook](https://github.com/OpenGeoscience/geonotebook), an open-source Jupyter extension with interactive map displays.

README.md
```diff
@@ -22,6 +22,16 @@ fast batch processing of large raster data sets.
...
+GeoTrellis is unable to do.
```
echeipesh: GeoTrellis could integrate with Jupyter through [Toree](https://toree.apache.org/) and we've prototyped it. More attractive option though is clearly for python workbook and not Scala workbook. The last clause is a little misleading but full explanation is out of place here, so I would suggest removing it.

PR #2334, opened 2017-08-17
MapKeyTransform method aliases

fosskers:
### TODO 

- [x] Named aliases for each `.apply` variant
- [x] Provide these transformations directly in the types they affect (where possible)
- [x] Deprecations?

### Motivation

`MapKeyTransform` is a "worker class", it does not itself represent data. It provides mechanisms for transformations involving `SpatialKey`s, namely:

- discovering the key corresponding to some Geometry
- discovering the `Extent` that corresponds to some key
- discovering the set of keys (`GridBounds`) that corresponds to some Geometry envelope (an `Extent`)

Until now, each of these transformations occurred through a very overloaded `.apply` method, making `MapKeyTransform` somewhat of a mysterious black box. This PR adds sensibly named aliases for each of these transformations, like `keyToExtent: SpatialKey => Extent`.

#### Duplicate Deprecations

If it's true that reducing API surface area increases discoverability and improves usability, is it necessary to have:

- both `apply: Point => SpatialKey` and `apply: (Double, Double) => SpatialKey`
- all of `apply: (Int, Int) => Extent` and `apply: SpatialKey => Extent` and `apply[K: SpatialComponent]: K => Extent`

given that the extra forms are simple pure transformations that could occur in application code?
For instance, as currently exists in the code base:

```scala
 def apply[K: SpatialComponent](key: K): Extent = apply(key.getComponent[SpatialKey])
```

My recommendation is to keep `SpatialKey => Extent` and `Point => SpatialKey` and deprecate the duplicates.

#### `.apply` Deprecations

Echoing the "surface area" argument, would it be prudent to deprecate the `.apply` methods as well, given that they were the original source of confusion? Replacing their usage with the named aliases within GT couldn't be more than a 15 minute find-replace job.

#### `geotrellis-core`

Each of these transformations can be modelled by the form `A.someAToB: LayoutDefinition => B`, where the data of the `A` is used implicitly inside the method. Here, the transformation is a method on `A`, not a function through `MapKeyTransform`. For the case of `SpatialKey` and `Extent`, it's: `SpatialKey.toExtent: LayoutDefinition => Extent`. This PR provides this method as well, so that users don't need to go through a `MapKeyTransform`.

Unfortunately, due to the way the project is structured, we can't provide type-local methods for the other types without incurring circular dependencies:

- `Point.toKey: LayoutDefinition => SpatialKey` 
  - `LayoutDefinition` lives in `geotrellis.spark.tiling`
  - `SpatialKey` lives in `geotrellis.spark`
  - `Point` lives in `geotrellis.vector`
- `GridBounds.toExtent: LayoutDefinition => Extent`
  - `GridBounds` lives in `geotrellis.raster`
  - `Extent` lives in `geotrellis.vector`

and so on. So `MapKeyTransform` who lives in `geotrellis.spark.tiling` has to act as a sort of
"type broker", a place for all these necessary transformations to live. This (and other issues) would be solved by a central `geotrellis-core` package, which would contain types useful to any conceivable GeoTrellis application (i.e. one shouldn't need to depend on `spark` just to get access to `SpatialKey`). This way, we could get rid of `MapKeyTransform`, and each transformation could occur directly on the class in question.

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/SpatialKey.scala
+++ b/spark/src/main/scala/geotrellis/spark/SpatialKey.scala
@@ -16,12 +16,17 @@
 
 package geotrellis.spark
 
+import geotrellis.spark.tiling.LayoutDefinition
+import geotrellis.vector.Extent
 import org.apache.spark.rdd.RDD
 
 /** A SpatialKey designates the spatial positioning of a layer's tile. */
 case class SpatialKey(col: Int, row: Int) extends Product2[Int, Int] {
   def _1 = col
   def _2 = row
+
+  /** Retrieve the [[Extent]] that corresponds to this key, given a layout. */
+  def toExtent(layout: LayoutDefinition): Extent = layout.mapTransform(this)
 }
 
 object SpatialKey {
--- a/spark/src/main/scala/geotrellis/spark/SpatialKey.scala
+++ b/spark/src/main/scala/geotrellis/spark/SpatialKey.scala
@@ -26,7 +26,7 @@ case class SpatialKey(col: Int, row: Int) extends Product2[Int, Int] {
   def _2 = row
 
   /** Retrieve the [[Extent]] that corresponds to this key, given a layout. */
-  def tileExtent(layout: LayoutDefinition): Extent = layout.mapTransform(this)
+  def extent(layout: LayoutDefinition): Extent = layout.mapTransform.keyToExtent(this)
 }
 
 object SpatialKey {
--- a/spark/src/main/scala/geotrellis/spark/SpatialKey.scala
+++ b/spark/src/main/scala/geotrellis/spark/SpatialKey.scala
@@ -26,7 +26,7 @@ case class SpatialKey(col: Int, row: Int) extends Product2[Int, Int] {
   def _2 = row
 
   /** Retrieve the [[Extent]] that corresponds to this key, given a layout. */
-  def toExtent(layout: LayoutDefinition): Extent = layout.mapTransform(this)
+  def tileExtent(layout: LayoutDefinition): Extent = layout.mapTransform(this)
 }
 
 object SpatialKey {
--- a/spark/src/main/scala/geotrellis/spark/tiling/MapKeyTransform.scala
+++ b/spark/src/main/scala/geotrellis/spark/tiling/MapKeyTransform.scala
@@ -107,12 +107,13 @@ class MapKeyTransform(val extent: Extent, val layoutCols: Int, val layoutRows: I
   @deprecated("Use this instead: MapKeyTransform.keyToExtent(key.getComponent[SpatialKey])", "1.2")
   def apply[K: SpatialComponent](key: K): Extent = apply(key.getComponent[SpatialKey])
 
-  def keyToExtent(key: SpatialKey): Extent = apply(key)
+  /** Get the [[Extent]] corresponding to a [[SpatialKey]] in some zoom level. */
+  def tileExtent(key: SpatialKey): Extent = apply(key)
 
   def apply(key: SpatialKey): Extent = apply(key.col, key.row)
 
   /** 'col' and 'row' correspond to a [[SpatialKey]] column and row in some grid. */
-  def coordsToExtent(col: Int, row: Int): Extent = apply(col, row)
+  def tileExtent(col: Int, row: Int): Extent = apply(col, row)
 
   def apply(col: Int, row: Int): Extent =
     Extent(
--- a/spark/src/main/scala/geotrellis/spark/tiling/MapKeyTransform.scala
+++ b/spark/src/main/scala/geotrellis/spark/tiling/MapKeyTransform.scala
@@ -86,12 +86,13 @@ class MapKeyTransform(val extent: Extent, val layoutCols: Int, val layoutRows: I
     e1.expandToInclude(e2)
   }
 
+  /** Fetch the [[SpatialKey]] that corresponds to some coordinates in some CRS on the Earth. */
   def pointToKey(p: Point): SpatialKey = apply(p)
 
   def apply(p: Point): SpatialKey = apply(p.x, p.y)
 
   /** Fetch the [[SpatialKey]] that corresponds to some coordinates in some CRS on the Earth. */
-  def pointCoordsToKey(x: Double, y: Double): SpatialKey = apply(x, y)
+  def pointToKey(x: Double, y: Double): SpatialKey = apply(x, y)
 
   def apply(x: Double, y: Double): SpatialKey = {
     val tcol =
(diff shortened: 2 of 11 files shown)


spark/src/main/scala/geotrellis/spark/tiling/MapKeyTransform.scala
```diff
@@ -76,14 +78,20 @@ class MapKeyTransform(val extent: Extent, val layoutCols: Int, val layoutRows: I
...
   }
 
-  def apply(p: Point): SpatialKey =
-    apply(p.x, p.y)
+  def pointToKey(p: Point): SpatialKey = apply(p)
+
+  def apply(p: Point): SpatialKey = apply(p.x, p.y)
+
+  /** Fetch the [[SpatialKey]] that corresponds to some coordinates in some CRS on the Earth. */
+  def pointCoordsToKey(x: Double, y: Double): SpatialKey = apply(x, y)
```
echeipesh: This feels like a place where overloading is helpful to overload between `pointToKey(Point(3,5))` and `pointToKey(x=3,y=5)` are both unsurprising and more discoverable.

spark/src/main/scala/geotrellis/spark/tiling/MapKeyTransform.scala
```diff
@@ -95,12 +103,16 @@ class MapKeyTransform(val extent: Extent, val layoutCols: Int, val layoutRows: I
     (tcol.floor.toInt, trow.floor.toInt)
   }
 
-  def apply[K: SpatialComponent](key: K): Extent = {
-    apply(key.getComponent[SpatialKey])
-  }
+  def keyLikeToExtent[K: SpatialComponent](key: K): Extent = apply(key)
```
echeipesh: Hmm, this could actually take care of the SpatialKey case just fine. Do you think its more clear to have an explicit function not based on `SpatialComponent`?

fosskers: Yeah, I think `SpatialKey => Extent` signals its intent better. If the user knows what the purpose of `SpatialComponent` is, then should be able to make the `getComponent` call themselves (if say their key type was `SpaceTimeKey`).

spark/src/main/scala/geotrellis/spark/tiling/MapKeyTransform.scala
fosskers: This I've added.

spark/src/main/scala/geotrellis/spark/tiling/MapKeyTransform.scala
echeipesh: I think a buy that, what are some of the impacted lines in our codebase for reference ?

spark/src/main/scala/geotrellis/spark/tiling/MapKeyTransform.scala
```diff
@@ -95,12 +104,15 @@ class MapKeyTransform(val extent: Extent, val layoutCols: Int, val layoutRows: I
...
+  def apply[K: SpatialComponent](key: K): Extent = apply(key.getComponent[SpatialKey])
+
+  def keyToExtent(key: SpatialKey): Extent = apply(key)
+
+  def apply(key: SpatialKey): Extent = apply(key.col, key.row)
 
-  def apply(key: SpatialKey): Extent =
-    apply(key.col, key.row)
+  /** 'col' and 'row' correspond to a [[SpatialKey]] column and row in some grid. */
+  def coordsToExtent(col: Int, row: Int): Extent = apply(col, row)
```
echeipesh: Can this be `keyToExtent` overload? `coords` is a little overloaded, is it pixel cords, tile cords, map cords, a bit of ambiguity.

fosskers: I don't think those are the same thing. See the `apply` that it calls out to right below it.

fosskers: Oh wait, I might be dumb.

spark/src/main/scala/geotrellis/spark/SpatialKey.scala
```diff
@@ -16,12 +16,17 @@
...
+import geotrellis.vector.Extent
 import org.apache.spark.rdd.RDD
 
 /** A SpatialKey designates the spatial positioning of a layer's tile. */
 case class SpatialKey(col: Int, row: Int) extends Product2[Int, Int] {
   def _1 = col
   def _2 = row
+
+  /** Retrieve the [[Extent]] that corresponds to this key, given a layout. */
+  def tileExtent(layout: LayoutDefinition): Extent = layout.mapTransform(this)
```
lossyrob: The reference to `tile` here makes this key, which is agnostic of Value type, seem like it's specific to `Tile`. We've been avoiding placing specific language on things that could be generally useful, e.g. to point clouds that are stored based on spatial keys.

spark/src/main/scala/geotrellis/spark/tiling/MapKeyTransform.scala
```diff
@@ -95,12 +104,16 @@ class MapKeyTransform(val extent: Extent, val layoutCols: Int, val layoutRows: I
...
   }
 
-  def apply[K: SpatialComponent](key: K): Extent = {
-    apply(key.getComponent[SpatialKey])
-  }
+  @deprecated("Use this instead: MapKeyTransform.keyToExtent(key.getComponent[SpatialKey])", "1.2")
+  def apply[K: SpatialComponent](key: K): Extent = apply(key.getComponent[SpatialKey])
+
+  /** Get the [[Extent]] corresponding to a [[SpatialKey]] in some zoom level. */
+  def tileExtent(key: SpatialKey): Extent = apply(key)
```
lossyrob: As mentioned above, I think we should keep away from naming things with `tile`.

Suggestion: `keyExtent`

fosskers: It was previously `toExtent`. Which do you think is better?

spark/src/main/scala/geotrellis/spark/SpatialKey.scala
echeipesh: Hmm, thats tough. `SpatialKey` covers a rectangular, tile-like area on the map based on `LayoutDefinition`.

spark/src/main/scala/geotrellis/spark/tiling/MapKeyTransform.scala
fosskers: After a talk with @echeipesh , the current vote is `SpatialKey.extent` (maximum minimalism, the word "key" doesn't need to be repeated since that's implied by the parent class), and `keyToExtent` for the `MapKeyTransform` methods so that they match the naming of their sibling methods. Thoughts?

lossyrob: :+1:

PR #2336, opened 2017-08-18
4-Connected Line Drawing

jamesmcclain:
This adds 4-connected line drawing capability.

The API is "broken" in the sense that the line drawing subroutine now takes and pays attention to a potential `options` parameter whereas it did not before.  `Options.DEFAULT` preserves the original (8-connected) behavior, whereas 4-connected drawing is done when `options.sampleType == PixelIsArea`.

In terms of the API, it seemed as though the choice were
   - This
   - Awkwardly add something along side the existing API
   - Extend the `Options` case class (in terms of invasiveness, seems >= the approach taken here)

This PR should probably either be closed or given the "2.0" label.

Also, while typing this PR I realized that it does not solve the [vector rasterization problem](https://github.com/locationtech/geotrellis/pull/2266) for which it was intended because the line drawing code snaps line endpoints to the centers of pixels before drawing; because of that, the reported pixels are not guaranteed to cover the line.  The picture below has an example.

example

The blue line is the original with the tiles that should be reported.  The red is the snapped-to-centers line with the tiles that will be reported.  Since there are some tiles in the blue set that are not in the red set, this cannot be used.  (If the set of tiles is 3x3 but the set of pixels is 3nx3n, then there will be some pixels which should be drawn but which are not because there is no tile to contain them.)

I did not change endpoint-snapping behavior because seems like an even larger change than the one presented here, so the three-part list above applies.  I am planning to put in a [solution to the original](https://github.com/echeipesh/geotrellis/pull/11) problem that does not involve the line rasterizer.

Note: I said above that it appears from the code that the line endpoints are being snapped to the centers of the pixels.  Even if that is not correct -- if they are being snapped to a corner or the middle of an edge -- the fact that the endpoints are being converted to integers means that such an example can probably still be constructed.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/rasterize/Rasterizer.scala
+++ b/raster/src/main/scala/geotrellis/raster/rasterize/Rasterizer.scala
@@ -226,16 +226,14 @@ object Rasterizer {
     re: RasterExtent,
     options: Options
   )(f: (Int, Int) => Unit) {
-    val cells = (for(coord <- line.jtsGeom.getCoordinates()) yield {
-      (re.mapXToGrid(coord.x), re.mapYToGrid(coord.y))
-    }).toList
-
-    for(i <- 1 until cells.size) {
-      foreachCellInGridLine(
-        cells(i - 1)._1, cells(i - 1)._2,
-        cells(i)._1, cells(i)._2,
-        re, i != cells.size - 1,
-        options)(f)
+    val coords = line.jtsGeom.getCoordinates()
+    var i = 1; while (i < coords.size) {
+      val x1 = re.mapXToGrid(coords(i-1).x)
+      val y1 = re.mapYToGrid(coords(i-1).y)
+      val x2 = re.mapXToGrid(coords(i+0).x)
+      val y2 = re.mapYToGrid(coords(i+0).y)
+      foreachCellInGridLine(x1, y1, x2, y2, re, i != coords.size - 1, options)(f)
+      i += 1
     }
   }
 
@@ -315,7 +313,13 @@ object Rasterizer {
          0 <= y && y < re.rows) { f(x, y); }
       e2 = err
       if (e2 > -dx) { err -= dy; x += sx; }
-      if (e2 < dy) { err += dx; y += sy; }
+      if (e2 < dy) {
+        if (options.sampleType == PixelIsArea &&
+            e2 > -dx &&
+            0 <= x && x < re.cols &&
+            0 <= y && y < re.rows) f(x, y)
+        err += dx; y += sy;
+      }
     }
     if(!skipLast &&
        0 <= x && x < re.cols &&
--- a/raster/src/main/scala/geotrellis/raster/rasterize/Rasterizer.scala
+++ b/raster/src/main/scala/geotrellis/raster/rasterize/Rasterizer.scala
@@ -111,8 +111,8 @@ object Rasterizer {
     geom match {
       case geom: Point         => foreachCellByPoint(geom, re)(f)
       case geom: MultiPoint    => foreachCellByMultiPoint(geom, re)(f)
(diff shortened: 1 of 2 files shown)


fosskers:
Existing user-facing functions don't appear to have had their api changed, so this shouldn't incur a major version increment.

jamesmcclain:
> so this shouldn't incur a major version increment.

From your lips to Ahuramazda's ears.

echeipesh:
Overloading the `sampleType` really strains the imagination here unless there is a conceptual explanation I'm missing. The shape of the solution is good though except the new overloads for `foreachCellInGridLine`, `foreachCellByLineString`, and `foreachCellByMultiLineString ` should accept [Connectivity](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/Connectivity.scala) that will not break the API.

Changes to `foreachCellByGeometry` are no go, they should just use the default 8-connected algorithm. Until 2.0 the user interested in 4-connected variant will have to match on the geometry themselves and pass in the appropriate `Connectivity`

raster/src/main/scala/geotrellis/raster/rasterize/Rasterizer.scala
```diff
@@ -205,6 +221,29 @@ object Rasterizer {
...
     * first point to the last point.
     */
+  def foreachCellByLineString(
+    line: Line,
+    re: RasterExtent,
+    options: Options
+  )(f: (Int, Int) => Unit) {
+    val cells = (for(coord <- line.jtsGeom.getCoordinates()) yield {
+      (re.mapXToGrid(coord.x), re.mapYToGrid(coord.y))
+    }).toList
```
echeipesh: Rasterizer is a performance sensitive part of code in my experience. `getCoordinates` already returns an `Array[Coordinate]` this loop and loop below can be rolled into one. That avoids an extra collection allocation and the takes care of indexing cost comment below.

jamesmcclain: Unless I misunderstand what you are saying, [this is a copy of the existing code](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/rasterize/Rasterizer.scala#L208-L219).  Does it need to be changed too?

raster/src/main/scala/geotrellis/raster/rasterize/Rasterizer.scala
```diff
@@ -205,6 +221,29 @@ object Rasterizer {
...
+
+    for(i <- 1 until cells.size) {
+      foreachCellInGridLine(
+        cells(i - 1)._1, cells(i - 1)._2,
```
jamesmcclain: I'll confess that I did not look very closely at this code, I just copied what was there.

jamesmcclain:
> Changes to foreachCellByGeometry are no go, they should just use the default 8-connected algorithm. Until 2.0 the user interested in 4-connected variant will have to match on the geometry themselves and pass in the appropriate Connectivity

Okay, I will rework the interface.

jamesmcclain:
I made [an issue](https://github.com/locationtech/geotrellis/issues/2343) for the rasterizer issues.

raster/src/main/scala/geotrellis/raster/rasterize/Rasterizer.scala
echeipesh: Wow, that's been there a long time... well, thank you for bringing it to the light.

raster/src/main/scala/geotrellis/raster/rasterize/Rasterizer.scala
echeipesh: Yes, please. I suppose many eyes finally have their day :)

jamesmcclain:
Changes made

PR #2337, opened 2017-08-19
Added requirement for `CRS` implementations to provide a readable `toString` representation.

metasim:
Added implementations for anonymous CRS implementations and singletons.


metasim:
I could see an argument for *not* requiring a `show` implementation (burden on possible third-party implementations). Let me know and I can change the implementation to just specific overrides of `toString` within GeoTrellis itself.

PR #2338, opened 2017-08-21
New GridBounds Methods: buffer and offset

jbouffard:
This PR adds two new methods to `GridBounds`: `buffer` and `offset`. These will allow the user to create a new `GridBounds` by buffering around a preexisting one and to transpose a `GridBounds` to a new location, respectively.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/GridBounds.scala
+++ b/raster/src/main/scala/geotrellis/raster/GridBounds.scala
@@ -106,6 +106,64 @@ case class GridBounds(colMin: Int, rowMin: Int, colMax: Int, rowMax: Int) {
     !(rowMax < other.rowMin || other.rowMax < rowMin)
 
   /**
+   * Creates a new [[GridBounds]] using a buffer around this
+   * GridBounds.
+   *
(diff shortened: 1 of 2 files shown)


raster/src/main/scala/geotrellis/raster/GridBounds.scala
```diff
@@ -106,6 +106,67 @@ case class GridBounds(colMin: Int, rowMin: Int, colMax: Int, rowMax: Int) {
...
+      rowMax + rowBuffer
+    )
+
+  /**
+   * Offsets this [[GridBounds]] to a new location relative to its current
+   * position. If this is offsetted past 0, then None is returned.
+   *
+   * @param boundsOffset The amount the GridBounds should be shifted.
+   */
+  def offset(boundsOffset: Int): Option[GridBounds] =
```
echeipesh: Not sure `Option` makes sense here. We defiantly use negative indexed `GridBounds` and in fact `buffer` may produce them if `clamp=false` above.

jbouffard: Yeah, you're right. It doesn't make that much sense. I'll change it now.

PR #2352, opened 2017-08-31
Improve S3 makePath function

pomadchin:
Fixes writing to an "empty named" folder, if s3 URI is with a trailing slash.


fosskers:
This was discovered during some Pipeline testing.

Issue #2327, opened 2017-08-10
LayerUpdate feature that does not try to merge tiles


lossyrob:
If the user knows there are no tiles to merge, the LayerUpdater shouldn't need to read the old tiles. This speeds up things a lot. For instance, I did this in a recent ingest:

```scala
            val lid = LayerId(layerName, z)
            if(as.layerExists(lid)) {
              // Some manual updating to speed things up
              // Taken from S3LayerUpdater
              val LayerAttributes(header, metadata, keyIndex, writerSchema) =
                as.readLayerAttributes[S3LayerHeader, TileLayerMetadata[SpaceTimeKey], SpaceTimeKey](lid)
              val prefix = header.key
              val bucket = header.bucket
              val maxWidth = Index.digits(keyIndex.toIndex(keyIndex.keyBounds.maxKey))
              val keyPath = (key: SpaceTimeKey) => makePath(prefix, Index.encode(keyIndex.toIndex(key), maxWidth))
              val updatedMetadata = metadata.merge(layer.metadata)
              val codec  = KeyValueRecordCodec[SpaceTimeKey, MultibandTile]
              val schema = codec.schema
              as.writeLayerAttributes(lid, header, updatedMetadata, keyIndex, schema)
              S3RDDWriter.write(layer, bucket, keyPath)
            } else {
              writer.write(lid, layer, keyIndex)
            }

```

echeipesh:
Overall problem LayerUpdater does not size partitions when reading tiles from already existing layer:
https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/io/file/FileLayerUpdater.scala#L75

When number of tiles returned are too large to be fit into default number of partitions (2-8 typically) this is will fail. A good way to reproduce this problem is to use a layer updater in https://github.com/geotrellis/geotrellis-landsat-tutorial
example: https://github.com/echeipesh/geotrellis-landsat-tutorial/tree/layer-updater

### Proposed Solution
Overall the strategy of using LayerReader to fetch the tiles is suboptimal because it requires both a stage of query preparation and then RDD level shuffle to line up the tiles to be updated.


What we need is a `_RDDUpdater.scala` that similar to the writer groups the records by their z-index, reads the existing block, applies the merge function to resolve the duplicates and writes the updated record.

Possible parameters here:
- `(Vector[V], Vector[V] => Vector[V])` merge function
- alwaysOverwrite flag

echeipesh:
[...]

PR #2357, opened 2017-09-06
Add Layer Overwriting Support

jamesmcclain:
In addition to the `update` method, an `overwrite` method has been added to allow portions of a layer to be over-written without looking up what is already there (fast path when no merging is needed and meta tiles are not in use).

The `update` and `overwrite` methods exist primarily in `*LayerWriter` now, `*LayerUpdater` makes calls into the former.

Connects #2327

### DIFF
--- a/accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerWriter.scala
+++ b/accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerWriter.scala
@@ -66,7 +66,7 @@ class AccumuloLayerWriter(
     _update(sc, id, rdd, keyBounds, Some(mergeFunc))
   }
 
-  def _update[
+  private def _update[
     K: AvroRecordCodec: Boundable: JsonFormat: ClassTag,
     V: AvroRecordCodec: ClassTag,
     M: JsonFormat: GetComponent[?, Bounds[K]]: Mergable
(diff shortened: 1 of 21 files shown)


spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopLayerWriter.scala
```diff
@@ -37,8 +40,98 @@ class HadoopLayerWriter(
...
+  ): Unit = {
+    _update(sc, id, rdd, keyBounds, Some(mergeFunc))
+  }
+
+  private def _update[
+    K: AvroRecordCodec: Boundable: JsonFormat: ClassTag,
+    V: AvroRecordCodec: ClassTag,
+    M: JsonFormat: GetComponent[?, Bounds[K]]: Mergable
+  ](
+    sc: SparkContext,
```
pomadchin: Mb to make it implicit param?

accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerWriter.scala
```diff
@@ -33,8 +36,107 @@ class AccumuloLayerWriter(
...
+  ): Unit = {
+    _update(sc, id, rdd, keyBounds, Some(mergeFunc))
+  }
+
+  private def _update[
+    K: AvroRecordCodec: Boundable: JsonFormat: ClassTag,
+    V: AvroRecordCodec: ClassTag,
+    M: JsonFormat: GetComponent[?, Bounds[K]]: Mergable
+  ](
+    sc: SparkContext,
```
pomadchin: Why spark context is not an implicit param?

pomadchin:
Ah I'll readdress everything to @echeipesh in https://github.com/locationtech/geotrellis/pull/2396 :D

Issue #2330, opened 2017-08-14
Negative grid bounds returned from RasterExtent.gridBoundsFor() method


mteldridge:
Even though `clamp` is set to `true`, `RasterExtent.gridBoundsFor` returns negative grid bounds under certain conditions.  In particular if the `subExtent` parameter has bounds that are within `RasterExtent.epsilon`, negative bounds can be returned.  Downstream effects include the inability to rasterize over valid vector geometries due to thrown exceptions.  

I have reproduced this in a set of tests in my fork of geotrellis:  [grid-bounds-bug](https://github.com/locationtech/geotrellis/compare/master...mteldridge:grid-bounds-bug)

metasim:
@rossbernet Could this be bumped to "Priority"?

@mteldridge Was there an easy fix for this (with perhaps not-so-easy testing required)?

rossbernet:
@metasim We have 2 two-weeks sprints to work on these issues. So to better reflect what I meant by those columns on the project, I changed the `priority` column to indicate we are hoping to complete it by September 21.

It's definitely a priority for the 1.2 release, and thanks for tracking it!

mteldridge:
Proposed change is here: 
https://github.com/mteldridge/geotrellis/commit/1324d25ac51c0ff7799d9151adec6ef5e31d8cc3

Basically, just fully clip the grid bounds to the raster boundary.  Passes the unit tests.  Should I submit this as a PR?

echeipesh:
Resolved by https://github.com/locationtech/geotrellis/pull/2364

PR #2364, opened 2017-09-08
Grid bounds bug

mteldridge:
Addresses issue #2330


echeipesh:
Thank you for both issue and the PR, ❤️

Issue #2348, opened 2017-08-24
JMH for Benchmarking


fosskers:
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
"http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" lang="en" xml:lang="en">
<head>
</head>
<body>

*This is a formalization of a conversation had earlier in the week.*

<div id="content">
<div id="table-of-contents">
<h2>Table of Contents</h2>
<div id="text-table-of-contents">
<ul>
<li><a href="#orgd7732fa">1. Scaliper</a></li>
<li><a href="#org5a40b18">2. JMH w/ SBT Plugin</a>
<ul>
<li><a href="#orgd7cdd08">2.1. Defining Benchmarks</a></li>
<li><a href="#org96eabbc">2.2. Running Benchmarks</a></li>
<li><a href="#org69c02bd">2.3. Project Pulses</a></li>
</ul>
</li>
<li><a href="#orgbad0f43">3. Proposal</a></li>
</ul>
</div>
</div>
<div id="outline-container-orgd7732fa" class="outline-2">
<h2 id="orgd7732fa"><span class="section-number-2">1</span> Scaliper</h2>
<div class="outline-text-2" id="text-1">
<p>
We're a data science library, so benchmarking is obviously very important for us.
Up until now we're been using Rob's <a href="https://github.com/azavea/scaliper">scaliper</a> library for this. It was originally
written to fill a void in Scala benchmarking, since Rob had had bad experiences
with ScalaMetre in the past. Scaliper is a great tool that gives us consistent results.
However, what bars it from continued use is the fact that it has no maintainance cycle.
Given that we are already stretched in many directions as a team, this may be one
aspect of our work that we can "off-shore".
</p>
</div>
</div>

<div id="outline-container-org5a40b18" class="outline-2">
<h2 id="org5a40b18"><span class="section-number-2">2</span> JMH w/ SBT Plugin</h2>
<div class="outline-text-2" id="text-2">
<p>
<a href="http://openjdk.java.net/projects/code-tools/jmh/">JMH</a> is a benchmarking tool from the OpenJDK people. It features:
</p>

<ul class="org-ul">
<li>Good control over JVM forking / thread counts</li>
<li>Warm-up iterations</li>
[...]

Issue #2358, opened 2017-09-07
MergeQueue is slow for some query range sets


lossyrob:
From @mteldridge on Gitter:

> On a different subject, I think I have found an geotrellis issue/bug. I am trying to query a geotrellis layer (hadoop) with a multipolygon made up of about 1600 polys. It looks like MergeQueue isn't up to the task. The layer is indexed with Z3SpaceTimeKey and the query ends up with about 600K index ranges. MergeQueue ends up attempting to insert into the middle of an array 600K times. When it returns, it usually takes >10mins. Often it doesn't return. Perhaps having MergeQueue backed by an array isn't ideal as the query gets large. Any suggestions?
For what its worth, the problem seems to be fixed if the sequence of index ranges is presorted before going into the MergeQueue.
> Offending line is here: https://github.com/locationtech/geotrellis/blob/5c5ebed09f66cde92c77d9f21ce22e65a44573c4/spark/src/main/scala/geotrellis/spark/io/hadoop/formats/FilterMapFileInputFormat.scala#L99

I've experienced similar slowdowns when including time large time ranges against our azavea-datahub Landsat8 2016 layer (used in the FOSS4G 2017 workshop).

If presorting ranges speeds things up, that would be an easy optimization (e.g. sorting here https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/io/index/MergeQueue.scala#L24). This should be benchmarked and put in if it helps.

metasim:
Do we need #2348 addressed first (or at least the tech stack decided on)? 

Per [@mteldridge's comments](https://gitter.im/geotrellis/geotrellis?at=59b16defee5c9a4c5f42e578), changing the line above to

```scala
val compressedRanges = MergeQueue(r.sortBy(_._1)).toArray
```
significantly improves performance in our use case.

mteldridge:
As @lossyrob mentioned in Gitter, the appropriate fix is to sort by start value in the MergeQueue.apply method prior to inserting the index ranges.  The fix will then get picked up by all the modules that use the MergeQueue.  Pre-sorting should allow one to do the merging in a single pass through the array.   I would also consider replacing the array that backs MergeQueue with something that at least provides constant time insertion/deletion.  This should make the complexity O(nlogn) instead of what it is now - O(n^2) or perhaps worse.

mteldridge:
[...]

PR #2365, opened 2017-09-10
JMH Prototype 

metasim:
In response to #2348, this is an integration of the `sbt-jmh` plugin as a custom build plugin with some additional configuration keys for using and configuring benchmarking in whatever sub-project wishes to use it. Also created a `bench` sub-project to serve as an example of how to use it, and as a start for adding global benchmarking code (such as porting [legacy benchmarks](https://github.com/geotrellis/geotrellis-benchmark/tree/master/geotrellis/src/test/scala/geotrellis)).

Also committed the results from a run based on MergeQueue in an `archive` directory, providing an example of keeping historical results around.


metasim:
Plot of MergeQueue vs. alternate implementation by @mteldridge in several multi-axis test configurations. See `geotrellis.spark.io.index.MergeQueueBench` in the `bench` subproject for details.

screen shot 2017-09-09 at 8 58 34 pm

lossyrob:
This is looking great.

If I'm reading the graph correctly, that means @mteldridge's #2358 is a winner? Can we include a benchmark that tests the presort with the current MergeQueue as well?

metasim:
@lossyrob The "shuffle" param indicates whether the data was randomized (true) or already sorted (false). "span" is how wide the interval is, and "skip" is the distance between the starting points of subsequent intervals. Varying these triggers overlapping vs. non-overlapping cases.

 That said, I welcome a review of my methodology. This is my first time really trying to measure this kind of thing.

JMH is pretty darn cool. The documentation sucks, but the examples are great and the features are impressive.

metasim:
PS: I've ported a couple of the old benchmarks. So far it's pretty straightforward.

fosskers:
I can review this today.

bench/build.sbt
```diff
@@ -0,0 +1,7 @@
+enablePlugins(GTBenchmarkPlugin)
+
+libraryDependencies += "org.slf4j" % "slf4j-nop" % "1.7.25"
+
+jmhIterations := Some(5)
+jmhTimeUnit := Some("ms")
+jmhExtraOptions := Some("-jvmArgsAppend -Xmx8G")
```
fosskers: By hard coding these, we're assuming we'd never want to run the benches via say `jmh:run -t 1 -f 1 -wi 10 -i 10 .*Bench.*`? The `jmh:run` approach is definitely more "mystical".

bench/src/main/scala/geotrellis/raster/GenericRasterBench.scala
```diff
@@ -0,0 +1,62 @@
...
+import geotrellis.bench.init
+import org.openjdk.jmh.annotations._
+import org.openjdk.jmh.infra.BenchmarkParams
+
+import scala.util.Random
+
+
+@BenchmarkMode(Array(Mode.AverageTime))
+@State(Scope.Thread)
+@Threads(4)
```
fosskers: Always 4?

bench/src/main/scala/geotrellis/raster/GenericRasterBench.scala
```diff
@@ -0,0 +1,62 @@
...
+      case "genericRasterMap" ⇒
+        genericRaster = new GRaster(init(len)(Random.nextInt))
+      case "rasterMap" ⇒
+        tile = ArrayTile(init(len)(Random.nextInt), size, size)
+      case _ ⇒ throw new MatchError("Have a new benchmark without initialization?")
+    }
+  }
+
+  @Benchmark
+  def genericRasterMap = {
```
fosskers: Return type here and elsewhere for the other functions.

bench/src/main/scala/geotrellis/raster/render/RenderBench.scala
```diff
@@ -0,0 +1,46 @@
...
+      }).toMap
+    cmap =
+      ColorMap(map)
+
+    colorEncoding =
+      PngColorEncoding(cmap.colors, cmap.options.noDataColor, cmap.options.fallbackColor)
+  }
+
+  @Benchmark
+  def rendering = cmap.render(tile)
```
fosskers: Return type here too.

bench/build.sbt
metasim: You can still do `jmh:run -t 1 -f 1 -wi 10 -i 10 .*Bench.*` if you want to; the plugin just allows some of these things to be set and managed individually so you could do:

```
sbt> set jmhIterations := Some(5)
sbt> bench
```

bench/src/main/scala/geotrellis/raster/GenericRasterBench.scala
metasim: Can't remember why I did that. I'll either remove or comment why the reason.

bench/build.sbt
metasim: PS: I think it's important to have a certain set of defaults version controlled to ensure you get somewhat consistent runs between analysis sessions (although the machine you run probably influences things heavily, so maybe it's false assumption of control).

metasim:
@fosskers This is my first time using the review feature, so let me know if there's something I'm supposed to do when I'm done with the updates.

bench/build.sbt
fosskers: I think it's good to have sensible defaults.

fosskers:
@metasim Nope, it's just a matter of me looking over the changes again and hitting "Approve". Looks great!

fosskers:
How much of `geotrellis-benchmark` got ported here? Can we deprecate that repo yet?

metasim:
I only ported a handful to work out the patterns involved. I think we should create a new ticket for porting the rest (or an identified subset) over.

PR #2375, opened 2017-09-13
sizeLong Bug Fix and Replacement of Deprecated GridBound Methods

### LINKED ISSUE #2314: Move code from using GridBounds.size to GridBounds.sizeLong
E.g. https://github.com/locationtech/geotrellis/blob/a0280ebd51e3bec1a765c623138684e8a0fcd966/spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala#L145

jbouffard:
This PR replaces `size` to `sizeLong` and `coords` to coordsIter` in the GeoTrellis code base. In addition, a bug has been fixed in `sizeLong` so that it will now produce the correct result.

This PR resolves #2314

### DIFF
--- a/raster-test/src/test/scala/geotrellis/raster/GridBoundsSpec.scala
+++ b/raster-test/src/test/scala/geotrellis/raster/GridBoundsSpec.scala
@@ -96,8 +96,7 @@ class GridBoundsSpec extends FunSpec with Matchers{
           GridBounds(0, 0, 75, 75),
           GridBounds(25, 25, 100, 100)
         )
-      println(GridBounds.distinct(gridBounds))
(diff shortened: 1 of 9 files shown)


raster-test/src/test/scala/geotrellis/raster/GridBoundsSpec.scala
```diff
@@ -106,7 +105,7 @@ class GridBoundsSpec extends FunSpec with Matchers{
     it("should match the output of coordsIter") {
       val gbs = GridBounds(0, 0, 10, 10)
 
-      gbs.coordsIter.toSeq shouldBe gbs.coords.toSeq
+      gbs.coordsIter.toSeq shouldBe gbs.coordsIter.toSeq
```
echeipesh: This breaks the test, we should keep the way it was before until the deprecated method is gone

PR #2380, opened 2017-09-14
Allow specification of JTS precision model in application.conf

### LINKED ISSUE #2219: Specify precision model for GeometryFactory based on configuration

jpolchlo:
Fixes #2219 

Currently, `geotrellis.vector.GeomFactory` defaults to use JTS's floating `PrecisionModel`.  This means that all geometry will naturally be represented with double-precision coordinates.  This can lead to TopologyExceptions.  We want to be able to specify the precision model for numerically robust geometric code.  This PR uses the `com.typesafe.config` mechanisms to allow for an application.conf which specifies the precision model.

The config key `geotrellis.jts.precision.type` can take on one of three string values: "floating", "floating_single", and "fixed".  See the [JTS docs](https://locationtech.github.io/jts/javadoc/org/locationtech/jts/geom/PrecisionModel.html) for details.  In the event of a fixed precision model, the user may also give the scale parameter as `geotrellis.jts.precision.scale`; this defaults to 1e12, or 12 decimal places of precision.  [Note that this PR fixes a bug where the scale parameter was given as simply 12, which is a misreading of what the scale parameter means.]

### DIFF
--- a/vector/src/main/scala/geotrellis/vector/GeomFactory.scala
+++ b/vector/src/main/scala/geotrellis/vector/GeomFactory.scala
@@ -16,6 +16,8 @@
 
 package geotrellis.vector
 
+import geotrellis.util.LazyLogging
+
 import com.typesafe.config.ConfigFactory
 import com.vividsolutions.jts.geom
 import com.vividsolutions.jts.geom.PrecisionModel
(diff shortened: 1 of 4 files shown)


vector/src/main/scala/geotrellis/vector/GeomFactory.scala
```diff
@@ -16,15 +16,29 @@
...
 private[vector] object GeomFactory {
 
-  val factory = new geom.GeometryFactory()
+  val precisionType = Try(ConfigFactory.load().getString("geotrellis.jts.precision.type").toLowerCase).getOrElse("floating")
+
+  val precisionModel = precisionType match {
+    case "floating" => new PrecisionModel()
+    case "floating_single" => new PrecisionModel(PrecisionModel.FLOATING_SINGLE)
+    case "fixed" => 
+      val scale = new PrecisionModel(Try(ConfigFactory.load().getDouble("geotrellis.jts.precision.scale")).getOrElse(1e12))
```
pomadchin: Mb to add a warn here, that the default value would be used.

pomadchin: Also, probably it makes sense to add `precision.scale` and `precision.type` default values into vector project configuration file, that can be a sort of a usage example.

echeipesh: Good point, we should provide ‘reference.conf’ with default values. Those will be overwritten by values defined in ‘application.conf’.

PR #2388, opened 2017-09-19
Allow rasterizer to store z value at double precision

moradology:
This PR allows rasterization to store values at double precision (which is immediately useful for elevation rasters) and allows the z values to be stored with a provided `CellType`

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/rasterize/CellValue.scala
+++ b/raster/src/main/scala/geotrellis/raster/rasterize/CellValue.scala
@@ -3,5 +3,5 @@ package geotrellis.raster.rasterize
 import geotrellis.raster.CellType
 
 
-/** Cell value with its zindex and celltype to be used by the rasterizer. */
-case class CellValue(value: Double, zindex: Double, celltype: CellType)
+/** Cell value with its zindex to be used by the rasterizer. */
+case class CellValue(value: Double, zindex: Double)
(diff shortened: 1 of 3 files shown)


jamesmcclain:
I think that this was originally the case, but these values were narrowed for economic reasons.

echeipesh:
Yep, this PR parameterizes it so economy and precision can be served.

raster/src/main/scala/geotrellis/raster/rasterize/CellValue.scala
```diff
@@ -1,5 +1,7 @@
 package geotrellis.raster.rasterize
 
+import geotrellis.raster.CellType
 
-/** Cell value with its zindex used by the rasterizer. */
-case class CellValue(value: Double, zindex: Short)
+
+/** Cell value with its zindex and celltype to be used by the rasterizer. */
+case class CellValue(value: Double, zindex: Double, celltype: CellType)
```
echeipesh: I'm realizing now that the issue was misstated, exposing `cellType` for zindex per feature actually leads ambiguous behavior: the order in which tiles are merged is undefined and if two priority tiles do not share the same zindex type it is unclear which will be updated and propagated.

So this parameter should be at function invocation level as `zindexCellType`. Probably `ByteConstantNoDataCellType` is a nice conservative default value for it.

moradology:
@echeipesh I believe I've addressed the issue you raised - let me know if I'm missing something

Issue #2058, opened 2017-03-13
Find mechanism to move `-test` projects back into parent projects


metasim:
Possibly enabled by #2036.

metasim:
Possibly stalled by sbt/sbt/#3023.

fosskers:
If the testkits were brought back into their parent projects, would their bytecode be included in the assembly jars when published? If so, it would be better to keep them separate.

pomadchin:
@fosskers won't be

fosskers:
The `testkit` subprojects are published, actually.

pomadchin:
@fosskers ah, we are talking about `-test` projects not `-testkit`

metasim:
An update on this: I was [able to move](https://github.com/metasim/geotrellis/commit/40477459e1ea5d13feac8cab86ac8c678c0582ba) `vector-test` into `vector`, but ran into all sorts of problems with both `raster-test` and `s3-test`, including the sbt crash above. Basically, sbt doesn't like circular dependencies. The best plan, long term will be to draw out the dependency graph and rethink some of the organization.

PR #2389, opened 2017-09-20
Merge test projects

echeipesh:
Resolves https://github.com/locationtech/geotrellis/issues/2058

`scalacOptions ++= Seq("-optimize")` in `raster/build.sbt` appears to be the culprit for NPE seen in https://github.com/sbt/sbt/issues/3023


echeipesh:
Did some benchmarks using https://github.com/locationtech/geotrellis/pull/2365

`optimized` benchmark is benchmark of PR #2365 and `non-optimized` is benchmark of this PR with #2365 merged into it.

https://gist.github.com/echeipesh/b555da8d12c79b65a5f58c98aed0be69

The files can be visualized using: http://jmh.morethan.io

But here is the interesting result: (`non-optimized` -> `optimized`)
screen shot 2017-09-29 at 11 52 12 am
screen shot 2017-09-29 at 11 52 28 am

Googling around reveals that `-optimize` is suspect in scalac `2.11`

> However, there are some known bugs in the 2.11.8 optimizer that we're only planning to fix with the rewrite in 2.12.0.

fosskers:
Files changed: 2,045. :beer:

metasim:
@echeipesh This is awesome! Glad to know about that visualization tool! Should we change the default output format to json?

echeipesh:
Yes, I think that makes sense. Being able to drop them in and get a visual report is way more useful than being able to parse the results. There doesn't seem to be a pre-made way to convert between the formats either.

metasim:
Changed the  default JSON:

https://github.com/locationtech/geotrellis/pull/2365/commits/c1b4d6e0b07816b8749570d4d97f9dbe6835a0b3

metasim:
🎉

PR #2395, opened 2017-09-26
Remove duplicate SCM Info

aheyne:
Remove duplicate SCM Info that breaks Maven POMs

Signed-off-by: Austin Heyne <aheyne@ccri.com>


aheyne:
https://github.com/locationtech/geotrellis/blob/master/build.sbt#L12-L14
Adds this part automatically, the duplicated tags cause the issue.

aheyne:
https://github.com/locationtech/geotrellis/blob/master/build.sbt#L12-L14
Adds the SCM Info to the POM automatically. Duplicating the tags causes Maven to not validate the POM.

PR #2396, opened 2017-09-27
Incremental Layer Updater

echeipesh:
Supersedes: https://github.com/locationtech/geotrellis/pull/2357 https://github.com/locationtech/geotrellis/pull/2369 https://github.com/locationtech/geotrellis/pull/2370 
https://github.com/locationtech/geotrellis/pull/2371 https://github.com/locationtech/geotrellis/pull/2378 https://github.com/locationtech/geotrellis/pull/2379 https://github.com/locationtech/geotrellis/pull/2384 

- Add layer updating functionality to `LayerWriter` interface
- Deprecates `LayerUpdater` trait
- Update no longer supports changes in Avro schema
- Updates RDDs with empty key bounds are skipped with WARN rather than throwing
- `AccumuloLayerWriter.update` does not support HDFS write strategy

### DIFF
--- a/accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerUpdater.scala
+++ b/accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerUpdater.scala
@@ -46,7 +46,7 @@ class AccumuloLayerUpdater(
     val LayerAttributes(header, metadata, keyIndex, writerSchema) = try {
       attributeStore.readLayerAttributes[AccumuloLayerHeader, M, K](id)
     } catch {
-      case e: AttributeNotFoundError => throw new LayerReadError(id).initCause(e)
+      case e: AttributeNotFoundError => throw new LayerWriteError(id).initCause(e)
     }
     val layerWriter = new AccumuloLayerWriter(attributeStore, instance, header.tileTable, options)
     layerWriter.update(id, rdd, mergeFunc)
@@ -60,7 +60,7 @@ class AccumuloLayerUpdater(
     val LayerAttributes(header, metadata, keyIndex, writerSchema) = try {
       attributeStore.readLayerAttributes[AccumuloLayerHeader, M, K](id)
     } catch {
-      case e: AttributeNotFoundError => throw new LayerReadError(id).initCause(e)
+      case e: AttributeNotFoundError => throw new LayerWriteError(id).initCause(e)
     }
     val layerWriter = new AccumuloLayerWriter(attributeStore, instance, header.tileTable, options)
     layerWriter.overwrite(id, rdd)
--- a/accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerUpdater.scala
+++ b/accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerUpdater.scala
@@ -38,8 +38,6 @@ class AccumuloLayerUpdater(
   options: Options
 ) extends LayerUpdater[LayerId] with LazyLogging {
 
-  implicit private val sc: SparkContext = layerReader.sparkContext
-
   def update[
     K: AvroRecordCodec: Boundable: JsonFormat: ClassTag,
     V: AvroRecordCodec: ClassTag,
@@ -51,6 +49,7 @@ class AccumuloLayerUpdater(
       case e: AttributeNotFoundError => throw new LayerReadError(id).initCause(e)
     }
     val layerWriter = new AccumuloLayerWriter(attributeStore, instance, header.tileTable, options)
+    implicit val sc: SparkContext = rdd.sparkContext
     layerWriter.update(id, rdd, mergeFunc)
   }
 
@@ -65,6 +64,7 @@ class AccumuloLayerUpdater(
       case e: AttributeNotFoundError => throw new LayerReadError(id).initCause(e)
     }
     val layerWriter = new AccumuloLayerWriter(attributeStore, instance, header.tileTable, options)
+    implicit val sc: SparkContext = rdd.sparkContext
     layerWriter.update(id, rdd)
   }
 
@@ -79,6 +79,7 @@ class AccumuloLayerUpdater(
       case e: AttributeNotFoundError => throw new LayerReadError(id).initCause(e)
     }
     val layerWriter = new AccumuloLayerWriter(attributeStore, instance, header.tileTable, options)
+    implicit val sc: SparkContext = rdd.sparkContext
     layerWriter.overwrite(id, rdd)
   }
 }
--- a/accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerUpdater.scala
+++ b/accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerUpdater.scala
@@ -34,99 +34,52 @@ import AccumuloLayerWriter.Options
 class AccumuloLayerUpdater(
   val instance: AccumuloInstance,
   val attributeStore: AttributeStore,
-  layerReader: AccumuloLayerReader,
+  layerReader: AccumuloLayerReader, // XXX unused
   options: Options
 ) extends LayerUpdater[LayerId] with LazyLogging {
 
-  protected def _overwrite[
+  implicit val sc: SparkContext = layerReader.sparkContext
+
+  def update[
     K: AvroRecordCodec: Boundable: JsonFormat: ClassTag,
     V: AvroRecordCodec: ClassTag,
     M: JsonFormat: GetComponent[?, Bounds[K]]: Mergable
-  ](
-    id: LayerId,
-    rdd: RDD[(K, V)] with Metadata[M],
-    keyBounds: KeyBounds[K]
-  ): Unit = {
-    _update(id, rdd, keyBounds, None)
-  }
-
-  protected def _update[
-    K: AvroRecordCodec: Boundable: JsonFormat: ClassTag,
-    V: AvroRecordCodec: ClassTag,
-    M: JsonFormat: GetComponent[?, Bounds[K]]: Mergable
-  ](
-    id: LayerId,
-    rdd: RDD[(K, V)] with Metadata[M],
-    keyBounds: KeyBounds[K],
-    mergeFunc: (V, V) => V
-  ): Unit = {
-    _update(id, rdd, keyBounds, Some(mergeFunc))
-  }
-
-  def _update[
-    K: AvroRecordCodec: Boundable: JsonFormat: ClassTag,
-    V: AvroRecordCodec: ClassTag,
-    M: JsonFormat: GetComponent[?, Bounds[K]]: Mergable
-  ](
-    id: LayerId,
-    rdd: RDD[(K, V)] with Metadata[M],
-    keyBounds: KeyBounds[K],
-    mergeFunc: Option[(V, V) => V]
-  ) = {
-    if (!attributeStore.layerExists(id)) throw new LayerNotFoundError(id)
-
+  ](id: LayerId, rdd: RDD[(K, V)] with Metadata[M], mergeFunc: (V, V) => V): Unit = {
     val LayerAttributes(header, metadata, keyIndex, writerSchema) = try {
       attributeStore.readLayerAttributes[AccumuloLayerHeader, M, K](id)
     } catch {
-      case e: AttributeNotFoundError => throw new LayerUpdateError(id).initCause(e)
+      case e: AttributeNotFoundError => throw new LayerReadError(id).initCause(e)
     }
+    val layerWriter = new AccumuloLayerWriter(attributeStore, instance, header.tileTable, options)
+    layerWriter.update(id, rdd, mergeFunc)
+  }
 
-    val table = header.tileTable
+  def update[
+    K: AvroRecordCodec: Boundable: JsonFormat: ClassTag,
+    V: AvroRecordCodec: ClassTag,
+    M: JsonFormat: GetComponent[?, Bounds[K]]: Mergable
+  ](id: LayerId, rdd: RDD[(K, V)] with Metadata[M]): Unit = {
+    val LayerAttributes(header, metadata, keyIndex, writerSchema) = try {
+      attributeStore.readLayerAttributes[AccumuloLayerHeader, M, K](id)
+    } catch {
+      case e: AttributeNotFoundError => throw new LayerReadError(id).initCause(e)
+    }
+    val layerWriter = new AccumuloLayerWriter(attributeStore, instance, header.tileTable, options)
+    layerWriter.update(id, rdd)
+  }
 
-    if (!(keyIndex.keyBounds contains keyBounds))
-      throw new LayerOutOfKeyBoundsError(id, keyIndex.keyBounds)
-
-    val encodeKey = (key: K) => AccumuloKeyEncoder.encode(id, key, keyIndex.toIndex(key))
-
-    logger.info(s"Saving updated RDD for layer ${id} to table $table")
-    val existingTiles =
-      if(schemaHasChanged[K, V](writerSchema)) {
-        logger.warn(s"RDD schema has changed, this requires rewriting the entire layer.")
-        layerReader
-          .read[K, V, M](id)
-
-      } else {
-        val query =
-          new LayerQuery[K, M]
-            .where(Intersects(rdd.metadata.getComponent[Bounds[K]].get))
-
-        layerReader.read[K, V, M](id, query, layerReader.defaultNumPartitions, filterIndexOnly = true)
-      }
-
-    val updatedMetadata: M =
-      metadata.merge(rdd.metadata)
-
-    val updatedRdd: RDD[(K, V)] =
-      mergeFunc match {
-        case Some(mergeFunc) =>
-          existingTiles
-            .fullOuterJoin(rdd)
-            .flatMapValues {
-            case (Some(layerTile), Some(updateTile)) => Some(mergeFunc(layerTile, updateTile))
-            case (Some(layerTile), _) => Some(layerTile)
-            case (_, Some(updateTile)) => Some(updateTile)
-            case _ => None
-          }
-        case None => rdd
-      }
-
-    val codec  = KeyValueRecordCodec[K, V]
-    val schema = codec.schema
-
-    // Write updated metadata, and the possibly updated schema
-    // Only really need to write the metadata and schema
-    attributeStore.writeLayerAttributes(id, header, updatedMetadata, keyIndex, schema)
-    AccumuloRDDWriter.write(updatedRdd, instance, encodeKey, options.writeStrategy, table)
+  def overwrite[
+    K: AvroRecordCodec: Boundable: JsonFormat: ClassTag,
+    V: AvroRecordCodec: ClassTag,
+    M: JsonFormat: GetComponent[?, Bounds[K]]: Mergable
+  ](id: LayerId, rdd: RDD[(K, V)] with Metadata[M]): Unit = {
+    val LayerAttributes(header, metadata, keyIndex, writerSchema) = try {
+      attributeStore.readLayerAttributes[AccumuloLayerHeader, M, K](id)
+    } catch {
+      case e: AttributeNotFoundError => throw new LayerReadError(id).initCause(e)
+    }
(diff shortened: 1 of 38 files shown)


accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloRDDWriter.scala
```diff
@@ -44,14 +55,41 @@ object AccumuloRDDWriter {
...
         // groupBy will reuse the partitioner on the parent RDD if it is set, which could be typed
         // on a key type that may no longer by valid for the key type of the resulting RDD.
         .groupBy({ row => encodeKey(row._1) }, numPartitions = raster.partitions.length)
-        .map { case (key, pairs) =>
-          (key, new Value(AvroEncoder.toBinary(pairs.toVector)(codec)))
+        .map { case (key, _kvs1) =>
+          val kvs1: Vector[(K,V)] = _kvs1.toVector
+          val kvs2: Vector[(K,V)] =
+            if (mergeFunc != None) {
+              val scanner = instance.connector.createScanner(table, new Authorizations())
```
metasim: I'm not much on the details of Accumulo, but is `new Authorizations()` the same as [`Authorizations.EMPTY`](https://accumulo.apache.org/1.8/apidocs/org/apache/accumulo/core/security/Authorizations.html#EMPTY)? If so I'd use the latter to make it clear/explicit `Authorizations` aren't being used. At any rate, it caught my eye.

accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloRDDWriter.scala
```diff
@@ -44,14 +55,41 @@ object AccumuloRDDWriter {
...
         // Call groupBy with numPartitions; if called without that argument or a partitioner,
         // groupBy will reuse the partitioner on the parent RDD if it is set, which could be typed
         // on a key type that may no longer by valid for the key type of the resulting RDD.
         .groupBy({ row => encodeKey(row._1) }, numPartitions = raster.partitions.length)
-        .map { case (key, pairs) =>
-          (key, new Value(AvroEncoder.toBinary(pairs.toVector)(codec)))
+        .map { case (key, _kvs1) =>
+          val kvs1: Vector[(K,V)] = _kvs1.toVector
+          val kvs2: Vector[(K,V)] =
+            if (mergeFunc != None) {
```
metasim: Minor nitpick, so ignore if you don't care, but below you do pattern matching on `Option[(V,V) => V]` and here using identity boolean check. Consistency might help readability. I prefer pattern matching or `mergeFunc.nonEmpty` myself.

cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraLayerWriter.scala
```diff
@@ -33,8 +36,57 @@ class CassandraLayerWriter(
...
+    V: AvroRecordCodec: ClassTag,
+    M: JsonFormat: GetComponent[?, Bounds[K]]: Mergable
+  ](
+    id: LayerId,
+    rdd: RDD[(K, V)] with Metadata[M]
+  ): Unit = {
+    update(id, rdd, None)
+  }
+
+  def update[
```
metasim: These `LayerWriter`s seem to have a number of common delegate/convenience methods. Any way they can be abstracted over all/most of the cases?

cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraRDDWriter.scala
```diff
@@ -35,15 +41,28 @@ import java.util.concurrent.Executors
...
 object CassandraRDDWriter {
 
   def write[K: AvroRecordCodec, V: AvroRecordCodec](
+    rdd: RDD[(K, V)],
+    instance: CassandraInstance,
+    layerId: LayerId,
+    decomposeKey: K => Long,
+    keyspace: String,
+    table: String,
+    threads: Int = ConfigFactory.load().getThreads("geotrellis.cassandra.threads.rdd.write")
```
metasim: I suggest pulling the default number of threads into a final (constant) member of the object.

s3/src/main/scala/geotrellis/spark/io/s3/S3RDDWriter.scala
```diff
@@ -56,42 +75,80 @@ trait S3RDDWriter {
...
-            }
-          }
+              val record = iter.next()
+              val key = record._1
+              val kvs = record._2.toVector
+              Some(((key, kvs), iter))
+            } else None
+          })
+
+        def elaborateRow(row: (String, Vector[(K,V)])): Process[Task, (String, Vector[(K,V)])] = {
```
metasim: Looks like a similar pattern to Cassandra case. Opportunity for refactoring?

scripts/cassandraTestDB.sh
```diff
@@ -4,7 +4,8 @@ docker pull cassandra:latest
 
 docker run \
   --rm \
-  --net=host \
+  -p 9160:9160 \
+  -p 9042:9042 \
```
metasim: 👍

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopRDDWriter.scala
```diff
@@ -72,15 +94,129 @@ object HadoopRDDWriter extends LazyLogging {
       if (null != writer) writer.close()
   }
 
+  private[hadoop] def update[K: AvroRecordCodec: ClassTag, V: AvroRecordCodec: ClassTag](
+    rdd: RDD[(K, V)],
+    layerPath: Path,
+    id: LayerId,
+    as: AttributeStore,
+    mergeFunc: Option[(V,V) => V],
+    indexInterval: Int = 4
```
metasim: Curious to know what's magic about 4... move to object constant with comment?

spark/src/test/scala/geotrellis/spark/io/LayerUpdateSpaceTimeTileSpec.scala
```diff
@@ -65,12 +65,21 @@ trait LayerUpdateSpaceTimeTileSpec
         updater.update(layerId, sample)
       }
 
+      it("should overwrite a layer") {
+        updater.overwrite(layerId, sample)
```
metasim: Does this actually confirm that the update happened?

spark/src/test/scala/geotrellis/spark/io/LayerUpdateSpaceTimeTileSpec.scala
```diff
@@ -65,12 +65,21 @@ trait LayerUpdateSpaceTimeTileSpec
...
+      }
+
       it("should not update a layer (empty set)") {
         intercept[EmptyBoundsError] {
           updater.update(layerId, new ContextRDD[SpaceTimeKey, Tile, TileLayerMetadata[SpaceTimeKey]](sc.emptyRDD[(SpaceTimeKey, Tile)], emptyTileLayerMetadata))
         }
       }
 
+      it("should silently not overwrite a layer (empty set)") {
+        updater.overwrite(layerId, new ContextRDD[SpaceTimeKey, Tile, TileLayerMetadata[SpaceTimeKey]](sc.emptyRDD[(SpaceTimeKey, Tile)], emptyTileLayerMetadata))
```
metasim: ditto...

cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraLayerWriter.scala
echeipesh: We could expose the overload with `mergeFunc: Option[(V, V) => V]` but its a little iffy because having that as `None` turns a function into overwrite mode, which is drastically more dangerous. I hesitated to that that, do you think it would be helpful though?

spark/src/test/scala/geotrellis/spark/io/LayerUpdateSpaceTimeTileSpec.scala
echeipesh: It only confirms it didn't through, its possible that that an update was empty, either in bounds or in records in which case it would be NOOP. 

Overall I agree it would be much more helpful if this returned number of records updated/written but that would require changing interface of `_RDDWriter.write`... or (I'm realizing as I'm typing this) to add a new counting method and wrap it to preserve the interface.

I would push this change off to `2.0` though, what do you think?

hbase/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,13 @@
+# Settings to quiet third party logs that are too verbose
```
pomadchin: It would be interesting to uncomment logs to be sure that the comment above about reducing connections number worked.

s3/src/main/scala/geotrellis/spark/io/s3/S3LayerUpdater.scala
```diff
@@ -37,71 +38,40 @@ class S3LayerUpdater(
 ) extends LayerUpdater[LayerId] with LazyLogging {
 
   def rddWriter: S3RDDWriter = S3RDDWriter
+  def _rddWriter(): S3RDDWriter = rddWriter
+
+  class InnerS3LayerWriter(
```
pomadchin: Mb to make this class private? Or add some overloads to make construction of `S3LayerWriter` with a custom `rddWriter` more convenient?

s3/src/main/scala/geotrellis/spark/io/s3/S3LayerUpdater.scala
```diff
@@ -37,71 +38,40 @@ class S3LayerUpdater(
...
+    attributeStore: AttributeStore,
+    bucket: String,
+    keyPrefix: String
+  ) extends S3LayerWriter(attributeStore, bucket, keyPrefix) {
+    override def rddWriter() = _rddWriter
+  }
+
+  val as = attributeStore.asInstanceOf[S3AttributeStore]
```
pomadchin: This cast looks very dirty and makes `LayerUpdater` incompatible with other backends, it's a normal use case when our `AttributeStore` and the `catalog` itself are in different backends. What do you think about it? Or mb to make `AttributeStore` types more restrictive in `2.0`?

s3/src/main/scala/geotrellis/spark/io/s3/S3RDDWriter.scala
```diff
@@ -56,42 +77,80 @@ trait S3RDDWriter {
...
-          }(pool).retryEBO {
+          })(pool).retryEBO {
             case e: AmazonS3Exception if e.getStatusCode == 503 => true
             case _ => false
           }
         }
 
-        val results = nondeterminism.njoin(maxOpen = threads, maxQueued = threads) { requests map write }(Strategy.Executor(pool))
+        val results = nondeterminism.njoin(maxOpen = threads, maxQueued = threads) {
+          rows flatMap elaborateRow flatMap rowToRequest map retire
```
pomadchin: Can't we abstract over it a bit more? Cassandra uses the same thing; or it's hardly possible?

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopRDDWriter.scala
```diff
@@ -17,21 +17,47 @@
...
+
 
 object HadoopRDDWriter extends LazyLogging {
+  /** Index innterval at which map files should store an offset into sequence file.
+    * This value is picked as a compromize between in-memory footprint and IO cost of retreiving a single record.
+    */
+  final val DefaultIndexInterval = 4
+
+  // From https://github.com/apache/spark/blob/3b049abf102908ca72674139367e3b8d9ffcc283/core/src/main/scala/org/apache/spark/util/SerializableConfiguration.scala
+  private class SerializableConfiguration(@transient var value: Configuration) extends Serializable {
```
pomadchin: Can we put it into `util` package? It could be reused by RF and in my COGs layer support.

* [RF](https://github.com/raster-foundry/raster-foundry/blob/develop/app-backend/batch/src/main/scala/com/azavea/rf/batch/util/HadoopConfiguration.scala) // checked by time!
* [COGs](https://github.com/pomadchin/geotrellis/blob/feature/cog-layer/spark/src/main/scala/geotrellis/spark/util/HadoopConfiguration.scala)

I suggest to use [this](https://github.com/pomadchin/geotrellis/blob/feature/cog-layer/spark/src/main/scala/geotrellis/spark/util/HadoopConfiguration.scala) though not sure about necessity in `@transient`

util/src/main/scala/geotrellis/util/Filesystem.scala
```diff
@@ -24,6 +24,11 @@ import java.nio.file._
 
 
 object Filesystem {
+
+  def exists(path : String): Boolean = {
+    Files.exists(Paths.get(path))
```
pomadchin: :+1:

cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraLayerUpdater.scala
```diff
@@ -39,57 +39,20 @@ class CassandraLayerUpdater(
...
-    // Write updated metadata, and the possibly updated schema
-    // Only really need to write the metadata and schema
-    attributeStore.writeLayerAttributes(id, header, updatedMetadata, keyIndex, schema)
-    CassandraRDDWriter.write(updatedRdd, instance, id, encodeKey, keyspace, table)
+  def overwrite[
+    K: AvroRecordCodec: Boundable: JsonFormat: ClassTag,
+    V: AvroRecordCodec: ClassTag,
+    M: JsonFormat: GetComponent[?, Bounds[K]]: Mergable
+  ](id: LayerId, rdd: RDD[(K, V)] with Metadata[M]): Unit = {
+    val CassandraLayerHeader(_, _, keyspace, table) = attributeStore.readHeader[CassandraLayerHeader](id)
```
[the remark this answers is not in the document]
echeipesh: You'll get `AttributeReadError` with `readHeader` in the stack trace. I think thats pretty descriptive.

cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraLayerWriter.scala
echeipesh: Not sure which way the thumb is going is it:
- "No, way too dangerous!"
- "Yes, Lets expose the common API"

accumulo/src/main/scala/geotrellis/spark/io/accumulo/AccumuloLayerUpdater.scala
```diff
@@ -42,57 +42,28 @@ class AccumuloLayerUpdater(
...
     M: JsonFormat: GetComponent[?, Bounds[K]]: Mergable
-  ](id: LayerId, rdd: RDD[(K, V)] with Metadata[M], keyBounds: KeyBounds[K], mergeFunc: (V, V) => V) = {
-    if (!attributeStore.layerExists(id)) throw new LayerNotFoundError(id)
-
+  ](id: LayerId, rdd: RDD[(K, V)] with Metadata[M], keyBounds: KeyBounds[K], mergeFunc: (V, V) => V): Unit = {
     val LayerAttributes(header, metadata, keyIndex, writerSchema) = try {
       attributeStore.readLayerAttributes[AccumuloLayerHeader, M, K](id)
     } catch {
-      case e: AttributeNotFoundError => throw new LayerUpdateError(id).initCause(e)
+      case e: AttributeNotFoundError => throw new LayerReadError(id).initCause(e)
```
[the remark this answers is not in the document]
echeipesh: good catch.

cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraRDDWriter.scala
```diff
@@ -70,39 +99,72 @@ object CassandraRDDWriter {
...
+                        val bytes = oldRow.one().getBytes("value").array()
+                        AvroEncoder.fromBinary(kwWriterSchema.value.getOrElse(_recordCodec.schema), bytes)(_recordCodec)
+                      } else Vector.empty
+                    } else Vector.empty
+                  val kvs = mergeFunc match {
+                    case Some(fn) =>
+                      (kvs2 ++ kvs1)
+                        .groupBy({ case (k,v) => k })
+                        .map({ case (k, kvs) =>
+                          val vs = kvs.map({ case (k,v) => v }).toSeq
```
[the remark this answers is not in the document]
echeipesh: Doesn't appear to be, but also it'll do basically nothing except waste a stack frame.

cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraRDDWriter.scala
```diff
@@ -70,39 +99,72 @@ object CassandraRDDWriter {
...
+              def retire(row: (java.lang.Long, ByteBuffer)): Process[Task, ResultSet] = {
+                val (id, value) = row
+                Process eval Task({
+                  session.execute(writeStatement.bind(id, value))
+                })(pool)
               }
 
               val results = nondeterminism.njoin(maxOpen = threads, maxQueued = threads) {
-                queries map write
+                rows flatMap elaborateRow flatMap rowToBytes map retire
```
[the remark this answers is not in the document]
metasim: I think you meant:
`rows flatMap elaborateRow flatMap rowToBytes map buyBoat map retire`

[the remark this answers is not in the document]
echeipesh: `rows flatMap boat flatMap upstream flatMap map lifeTodream`

s3/src/main/scala/geotrellis/spark/io/s3/S3LayerUpdater.scala
echeipesh: Yes, the interface for `S3LayerUpdater` is not great. I'm going to make these patch fields private and call it a day. The updaters should be removed in 2.0 and the way they're constructed right now should never trigger the error case.

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopRDDWriter.scala
echeipesh: I wonder if a lack of `Serializable` trait has ever caused so much grief as in `Configuration`

echeipesh: Can't put it into `geotrellis.util` on account of the missing hadoop dependency but it has a home at `geotrellis.spark.util`

pomadchin: yes, that's what I meant; thanks!

hbase/src/main/scala/geotrellis/spark/io/hbase/HBaseRDDWriter.scala
```diff
@@ -51,25 +68,59 @@ object HBaseRDDWriter {
...
     // Call groupBy with numPartitions; if called without that argument or a partitioner,
     // groupBy will reuse the partitioner on the parent RDD if it is set, which could be typed
     // on a key type that may no longer by valid for the key type of the resulting RDD.
-
     raster.groupBy({ row => decomposeKey(row._1) }, numPartitions = raster.partitions.length)
       .foreachPartition { partition: Iterator[(Long, Iterable[(K, V)])] =>
         if(partition.nonEmpty) {
           instance.withConnectionDo { connection =>
             val mutator = connection.getBufferedMutator(table)
+            val _table = instance.getConnection.getTable(table)
```
pomadchin: Ah, sry I missed it %) we already have a `connection`

Should be:
```scala
val _table = connection.getTable(table)
```

`withConnectionDo` is a magic function which closes it after the `block` of code is done: 

```
def withConnectionDo[T](block: Connection => T): T = {
    val connection = getConnection
    try block(connection) finally connection.close()
  }
```

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopRDDWriter.scala
metasim: > I wonder if a lack of `Serializable` trait has ever caused so much grief as in `Configuration`

`CRS` not being `Product` has a similar feel.

Issue #2173, opened 2017-04-28
Partition GeoTiff reading by GeoTiff segment layout


echeipesh:
Currently in `S3GeoTiffRDD` and `HadoopGeoTiffRDD` when performing a windowed read the windows will be generated from (0,0) pixel offset to [cover the whole geotiff](https://github.com/locationtech/geotrellis/blob/master/s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffRDD.scala#L159) without the regard for its [layout](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffSegmentLayout.scala): 

This will likely result in multiple read requests for the same segments of GeoTiff which is suboptimal. Instead given that we have the GeoTiff header available we should consider the segment layout and plan reads such that windows will be no larger than `maxTileSize` but intersect only single segment.

~Likely this can be carried further to [`GeoTiffTile`](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffTile.scala#L537) where `.crop(windows: Seq[GridBounds])` would do similar thing.~

Edit: Improvements to crop are happening here https://github.com/locationtech/geotrellis/pull/2115 and should not be considered part of this issue.

- [x] S3GeoTiffRDD: https://github.com/locationtech/geotrellis/pull/2296
- [x] HadoopGeoTiffRDD https://github.com/locationtech/geotrellis/pull/2402

jamesmcclain:
> ... [cover the whole geotiff](https://github.com/locationtech/geotrellis/blob/e50d43d4c90771d2154bb3834ad04be1e11c9e0f/s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffRDD.scala#L159) ...

PR #2402, opened 2017-10-02
Generate Windows That Conform To GeoTiff Segments

jamesmcclain:
- [x] Ensure that S3 window size is being properly interpreted (is `windowsSize` a length or an area in this context?)
- [x] Ensure that handling of striped segments is viable (are tiles of one pixel in height okay?)

Connects https://github.com/locationtech/geotrellis/issues/2173


s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffRDD.scala
```diff
@@ -132,28 +132,30 @@ object S3GeoTiffRDD extends LazyLogging {
     lazy val sourceGeoTiffInfo = S3GeoTiffInfoReader(bucket, prefix, options)
 
     (options.maxTileSize, options.partitionBytes) match {
-      case (None, Some(partitionBytes)) =>
-        val segments: RDD[(String, Array[GridBounds])] =
-          sourceGeoTiffInfo.segmentsByPartitionBytes(partitionBytes, windowSize)
+      case (_, Some(partitionBytes)) => {
+        val maxSize = math.min(options.maxTileSize.getOrElse(1<<10), windowSize.getOrElse(1<<10)) // XXX is windowSize a length or an area?
```
echeipesh: `windowSize` is length, implicitly for square windows.

echeipesh:
> Ensure that handling of striped segments is viable (are tiles of one pixel in height okay?)

It seems strange but the thinking went is that if you don't have an opinion about the shape of tiles you want you're probably using this RDD as input to tiler. From the perspective of optimizing for IO reading the segments as skinny tiles is fine, from the perspective of tiler chunking a skinny tile into multiple tiles is not problematic either.

jamesmcclain:
> It seems strange but the thinking went is that if you don't have an opinion about the shape of tiles you want you're probably using this RDD as input to tiler. From the perspective of optimizing for IO reading the segments as skinny tiles is fine, from the perspective of tiler chunking a skinny tile into multiple tiles is not problematic either.

Sounds capital.

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopGeoTiffRDD.scala
```diff
@@ -87,22 +92,54 @@ object HadoopGeoTiffRDD {
...
-    val conf = configuration(path, options)
-    options.maxTileSize match {
-      case Some(tileSize) =>
+  def apply[I, K, V](
+    path: Path,
+    uriToKey: (URI, I) => K,
+    options: Options
+  )(implicit sc: SparkContext, rr: RasterReader[Options, (I, V)]): RDD[(K, V)] = {
+
+    val conf = new SerializableConfiguration(configuration(path, options))
```
pomadchin: Why `conf` should be wrapped here? Looks like it's only used for input formats arguments => can be used a common configuration without extra wrapper.

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopGeoTiffRDD.scala
```diff
@@ -87,22 +92,54 @@ object HadoopGeoTiffRDD {
...
+    val path2 = path.toString
```
pomadchin: Well it's always a bit confusing when you see in code such names. In fact you can write just 

```scala
HadoopGeoTiffInfoReader(path.toString, conf, options.tiffExtensions)
```

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopGeoTiffRDD.scala
```diff
@@ -129,18 +166,25 @@ object HadoopGeoTiffRDD {
...
-
     val windows: RDD[(Path, GridBounds)] =
       pathsToDimensions
         .flatMap { case (objectRequest, (cols, rows)) =>
-          RasterReader.listWindows(cols, rows, options.maxTileSize).map((objectRequest, _))
+          val info = HadoopGeoTiffInfoReader(objectRequest.toString, conf, options.tiffExtensions)
+          val layout = info.getGeoTiffInfo(objectRequest.toString).segmentLayout.tileLayout
+
+          RasterReader
+            .listWindows(cols, rows, options.maxTileSize.getOrElse(1<<10), layout.tileCols, layout.tileRows)
```
pomadchin: `maxTileSize` comment here too

s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffRDD.scala
```diff
@@ -205,7 +207,13 @@ object S3GeoTiffRDD extends LazyLogging {
     val windows =
       objectRequestsToDimensions
         .flatMap { case (objectRequest, (cols, rows)) =>
-          RasterReader.listWindows(cols, rows, options.maxTileSize).map((objectRequest, _))
+          val bucket = objectRequest.getBucketName
+          val key = objectRequest.getKey
+          val layout = sourceGeoTiffInfo.getGeoTiffInfo(s"s3://$bucket/$key").segmentLayout.tileLayout
+
+          RasterReader
+            .listWindows(cols, rows, options.maxTileSize.getOrElse(1<<10), layout.tileCols, layout.tileRows)
```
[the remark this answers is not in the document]
jamesmcclain: As long as changing the default `maxTileSize` is not construed to constitute an API change, I prefer that.

spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopGeoTiffRDD.scala
jamesmcclain: That variable makes its way into a `sc.parallelize` inside of `HadoopGeoTiffInfoReader`.

pomadchin: Good point, my bad.

jamesmcclain:
All comments have been addressed, I believe.

jamesmcclain:
:100:

PR #2408, opened 2017-10-05
Haversine formula fix

### LINKED ISSUE #2383: Haversine formula has bugs
The Haversine formula in ZoomedLayoutScheme seems to have a bug in it; I was using that logic to compute road distance metrics on OSM ways and found odd results. When I changed the implementation to something that [matched this](https://github.com/njj/haversine/blob/master/haversine.js), I started getting sensible results. 

We should move to something like:

```scala
    val R = EARTH_RADIUS
    val (start, end) = (ll1, ll2)
    val dLat = math.toRadians(end.y - start.y)
    val dLon = math.toRadians(end.x - start.x)
    val lat1 = math.toRadians(start.y)
    val lat2 = math.toRadians(end.y)

    val a = math.sin(dLat/2) * math.sin(dLat/2) +
            math.sin(dLon/2) * math.sin(dLon/2) * math.cos(lat1) * math.cos(lat2)
    val c = 2 * math.atan2(math.sqrt(a), math.sqrt(1-a))

    R * c
```

This only would effect area measurements for points that do not have valid UTM zones, so probably doesn't have a huge effect.

We should put the distance formula in geotrellis.utils or geotrellis.vector, so that it can be used outside of the ZoomedLayoutScheme logic.

pomadchin:
Fixes #2383

### DIFF
--- a/util/src/main/scala/geotrellis/util/Haversine.scala
+++ b/util/src/main/scala/geotrellis/util/Haversine.scala
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2017 Azavea
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package geotrellis.util
+
+object Haversine {
+  val EARTH_RADIUS = 6378137d // Use what gdal2tiles uses.
+
+  // (x: Double, y: Double) points
+  def apply(start: (Double, Double), end: (Double, Double), R: Double = EARTH_RADIUS): Double = {
+    val dLat = math.toRadians(end._2 - start._2)
+    val dLon = math.toRadians(end._1 - start._1)
+    val lat1 = math.toRadians(start._2)
+    val lat2 = math.toRadians(end._2)
+
+    val a =
+      math.sin(dLat / 2) * math.sin(dLat / 2) +
+      math.sin(dLon / 2) * math.sin(dLon / 2) * math.cos(lat1) * math.cos(lat2)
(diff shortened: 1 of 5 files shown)


util/src/main/scala/geotrellis/util/Haversine.scala
```diff
@@ -0,0 +1,36 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package geotrellis.util
+
+object Haversine {
+  val EARTH_RADIUS = 6378137d // Use what gdal2tiles uses.
```
metasim: I'd suggest adding a doc comment declaring the units used.

util/src/main/scala/geotrellis/util/Haversine.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  // (x: Double, y: Double) points
+  def apply(start: (Double, Double), end: (Double, Double), R: Double = EARTH_RADIUS): Double = {
```
metasim: For future GeoTrellisites, I'd suggest a doc comment indicating what the components of `start` and `end` are. e.g. (Latitude, Longitude), or (Longitude, Latitude) or (radius, angle), etc.

vector/src/main/scala/geotrellis/vector/Point.scala
```diff
@@ -31,6 +31,9 @@ object Point {
 
   implicit def jtsCoord2Point(coord: jts.Coordinate): Point =
     Point(factory.createPoint(coord))
+
+  implicit def pointToTuple2(point: Point): (Double, Double) =
```
metasim: While this is a nice convenience for people familiar with the particular code that translates between `Point` and `Tuple2`, I've been burned enough by unexpected or unintended conversions in the past to avoid this sort of thing personally, especially when the conversion is to a more general type. If you were converting between `Point` and something like `Position(lat, lng)` I'd be a bit less concerned because the translation semantics of `x` -> `lat` and `y` -> `lng` are clear, but `_1` and `_2` have no inherent semantics and therefore information is lost in the implicit conversion.

pomadchin: Yes, i decided to remove this implicit at all as it is indeed ambiguous.

pomadchin:
@metasim do you have some additional comments?

Issue #2328, opened 2017-08-12
Add filter capability to GeoTiffRDDs


lossyrob:
We should be able to filter the GeoTiff(s) by a polygon. Because we can do this from TiffTag metadata that can be read quickly, this would allow narrow ingests of large datasets, e.g. if I only wanted to pull a state-level amount of data from a single large GeoTiff like the Crop Data Layer.

jamesmcclain:
How does this relate to prospective COG layers?  That is, would COG layers automatically provide this functionality?

fosskers:
Solved by #2409

PR #2409, opened 2017-10-05
Filter GeoTiffRDDs by Geometry

jamesmcclain:
- [x] Requires https://github.com/locationtech/geotrellis/pull/2402

Connects #2328

### DIFF
--- a/s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffRDD.scala
+++ b/s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffRDD.scala
@@ -151,7 +151,7 @@ object S3GeoTiffRDD extends LazyLogging {
     bucket: String, prefix: String,
     uriToKey: (URI, I) => K,
     options: Options,
-    geometry: Option[Geometry] = None
+    geometry: Option[Geometry]
   )(implicit sc: SparkContext, rr: RasterReader[Options, (I, V)]): RDD[(K, V)] = {
 
     val conf = configuration(bucket, prefix, options)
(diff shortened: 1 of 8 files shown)


s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffRDD.scala
```diff
@@ -120,40 +120,51 @@ object S3GeoTiffRDD extends LazyLogging {
...
+    * @param  options   An instance of [[Options]] that contains any user defined or default settings.
+    * @param  geometry  An optional geometry to filter by.  If this is provided, it is assumed that all GeoTiffs are in the same CRS, and that this geometry is in that CRS.
     */
-  def apply[I, K, V](bucket: String, prefix: String, uriToKey: (URI, I) => K, options: Options)
-    (implicit sc: SparkContext, rr: RasterReader[Options, (I, V)]): RDD[(K, V)] = {
+  def apply[I, K, V](
+    bucket: String, prefix: String,
+    uriToKey: (URI, I) => K,
+    options: Options,
+    geometry: Option[Geometry] = None
```
pomadchin: Hm, mb to add some helpful overloads? Though it's smth to discuss, as the best otpion would be to move it into `Options` and it's impossible due to compatibility reasons. ):

pomadchin:
Btw a thought appeared after the review: mb it makes sense to add some extra overloads to use this filter by geometry functionality more convenient.

jamesmcclain:
> Btw a thought appeared after the review: mb it makes sense to add some extra overloads to use this filter by geometry functionality more convenient.

Yes, perhaps we can do that in a subsequent PR.

s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffRDD.scala
echeipesh: There also needs to be an `apply` with the old signature to avoid removing this method from the API.

PR #2414, opened 2017-10-06
Mark Various Vector Classes Serializable

### LINKED ISSUE #2381: Mark all vector types as Serializable
Right now it doesn't look like the Geometries extend Serializable, which is a big oversight.

Let's make sure all of our types implement Serializable so that we can shuffle them around in RDDs, even if Kryo isn't set.

jamesmcclain:
Stray whitespace has also been removed.

Resolves #2381

### DIFF
--- a/vector/src/main/scala/geotrellis/vector/Feature.scala
+++ b/vector/src/main/scala/geotrellis/vector/Feature.scala
@@ -23,7 +23,7 @@ package geotrellis.vector
   * @param geom An instance of G
   * @param data An instance of D
   */
-case class Feature[+G <: Geometry, +D](geom: G, data: D) extends Serializable {
+case class Feature[+G <: Geometry, +D](geom: G, data: D) {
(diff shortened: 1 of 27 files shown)


vector/src/main/scala/geotrellis/vector/Feature.scala
```diff
@@ -23,7 +23,7 @@ package geotrellis.vector
   * @param geom An instance of G
   * @param data An instance of D
   */
-case class Feature[+G <: Geometry, +D](geom: G, data: D) {
+case class Feature[+G <: Geometry, +D](geom: G, data: D) extends Serializable {
```
echeipesh: `case classes` extend `Serializable`

```scala
scala> case class Box(x: Int)
defined class Box

scala> Box(3).asInstanceOf[Serializable]
res0: Serializable = Box(3)
```

Issue #2368, opened 2017-09-12
Provide version of `ToSpatial` that blindly drops the temporal component.


metasim:
The current version of `ToSpatial.apply` requires an `instant` parameter to specify which tiles in the temporal dimension should be selected before converting to a spatieal-only key.

https://github.com/locationtech/geotrellis/blob/f4fa8860cd96acbe5068330d7e4a6673c69cec68/spark/src/main/scala/geotrellis/spark/filter/ToSpatial.scala#L57

There are cases where the temporal component has either been filtered already, or one doesn't care about overlaps, and the the temporal component should be blindly dropped. Without this capability you might have to do a RDD scan to find the valid temporal values and then select one.

Related: #2367

mteldridge:
I find myself following this pattern often:

```scala
val predicate = geotrellis.raster.mapalgebra.local.Mean
val spatial = rdd.map { case (stk, tile) => (stk.spatialKey, tile)  }
val merged = ContextRDD(spatial.groupByKey().mapValues(predicate(_)), 
                                             rdd.metadata.map[SpatialKey](_.spatialKey))
```
Where I map the key to `SpatialKey` and then apply a predicate function (e.g. `Mean`) to the resulting RDD and then stitch it up with (re-mapped) metadata into a new `ContextRDD`.  It would be nice if there were a

```scala
toSpatial(predicate: Traversable[Tile] => Tile) : RDD[K,V] with Metadata[M]
```
function that did all this transparently.

PR #2417, opened 2017-10-09
Time-Oblivious toSpatial Methods

jamesmcclain:
I gave the methods the names `toSpatialEnsureUnique` and `toSpatialNoEnsureUnique` because they had to be put into a different implicit class than the `toSpatial` methods.  The reason that they had to be put into a different implicit class is because `reduceByKey` (needed to ensure unique keys) requires that `K` and `V` are members of the `ClassTag` class.  Adding that requirement to the existing implicit class would have changed the API.

Connects https://github.com/locationtech/geotrellis/issues/2368

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,6 @@ index.js
 nohup.out
 
 site/
+
+derby.log
+metastore_db/
--- a/spark/src/main/scala/geotrellis/spark/filter/Implicits.scala
+++ b/spark/src/main/scala/geotrellis/spark/filter/Implicits.scala
@@ -21,6 +21,9 @@ import geotrellis.util._
 
 import org.apache.spark.rdd._
 
+import scala.reflect.ClassTag
+
+
 object Implicits extends Implicits
 
 trait Implicits {
@@ -36,4 +39,10 @@ trait Implicits {
     V,
     M[_]
   ](val self: RDD[(K, V)] with Metadata[M[K]]) extends SpaceTimeToSpatialMethods[K, V, M]
+
+  implicit class withSpaceTimeToSpatialMethods2[
+    K: ClassTag: SpatialComponent: TemporalComponent: λ[α => M[α] => Functor[M, α]]: λ[α => Component[M[α], Bounds[α]]],
+    V: ClassTag,
+    M[_]
+  ](val self: RDD[(K, V)] with Metadata[M[K]]) extends SpaceTimeToSpatialMethods2[K, V, M]
 }
--- a/spark/src/main/scala/geotrellis/spark/filter/Implicits.scala
+++ b/spark/src/main/scala/geotrellis/spark/filter/Implicits.scala
@@ -40,9 +40,9 @@ trait Implicits {
     M[_]
   ](val self: RDD[(K, V)] with Metadata[M[K]]) extends SpaceTimeToSpatialMethods[K, V, M]
 
-  implicit class withSpaceTimeToSpatialMethodsOblivious[
+  implicit class withSpaceTimeToSpatialReduceMethods[
     K: ClassTag: SpatialComponent: TemporalComponent: λ[α => M[α] => Functor[M, α]]: λ[α => Component[M[α], Bounds[α]]],
     V: ClassTag,
     M[_]
-  ](val self: RDD[(K, V)] with Metadata[M[K]]) extends SpaceTimeToSpatialObliviousMethods[K, V, M]
+  ](val self: RDD[(K, V)] with Metadata[M[K]]) extends SpaceTimeToSpatialReduceMethods[K, V, M]
 }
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -42,3 +42,6 @@ index.js
 nohup.out
 
 site/
+
+derby.log
+metastore_db/
```
pomadchin: What are these files?

spark/src/main/scala/geotrellis/spark/filter/Implicits.scala
```diff
@@ -36,4 +39,10 @@ trait Implicits {
     V,
     M[_]
   ](val self: RDD[(K, V)] with Metadata[M[K]]) extends SpaceTimeToSpatialMethods[K, V, M]
+
+  implicit class withSpaceTimeToSpatialMethods2[
```
pomadchin: Well, it looks like it's not the best name for implicit class :D even taking in account  that it will be removed.

.gitignore
jamesmcclain: Those are (I believe) created by Jacoco when you run tests.

jamesmcclain:
:100:

spark/src/main/scala/geotrellis/spark/filter/ToSpatial.scala
```diff
@@ -90,4 +93,22 @@ object ToSpatial {
...
+    K: ClassTag: SpatialComponent: TemporalComponent: λ[α => M[α] => Functor[M, α]]: λ[α => Component[M[α], Bounds[α]]],
+    V: ClassTag,
+    M[_]
+  ](rdd: RDD[(K, V)] with Metadata[M[K]], ensureUnique: Boolean): RDD[(SpatialKey, V)] with Metadata[M[SpatialKey]] = {
+    val metadata = rdd.metadata.map(_.getComponent[SpatialKey])
+    val rdd2 = ensureUnique match {
+      case true =>
+        rdd
+          .map({ case (k, v) => (k.getComponent[SpatialKey], v) })
+          .map(identity).reduceByKey({ case (left, right) => left })
```
echeipesh: This is only one of the possible arguments to `reduceByKey`, other more common cases would be to reduce by key using `Mean` or `Max` tile functions. To minimize API surprise I would suggest to use the method signatures to disambiguate the cases where merge is not required:


```scala
abstract class SpaceTimeToSpatialMethods[...](...) {
  def toSpatial(): RDD[(SpatialKey, V)] with Metadata[M[SpatialKey]]
  def toSpatial(instant: Long): RDD[(SpatialKey, V)] with Metadata[M[SpatialKey]]
  def toSpatial(dateTime: ZonedDateTime): RDD[(SpatialKey, V)] with Metadata[M[SpatialKey]]
}

abstract class SpaceTimeToSpatialMergeMethods[...](...) {
  def toSpatial(merge: (V, V) => V): RDD[(SpatialKey, V)] with Metadata[M[SpatialKey]]
  def toSpatial(instant: Long, merge: (V, V) => V): RDD[(SpatialKey, V)] with Metadata[M[SpatialKey]]
  def toSpatial(dateTime: ZonedDateTime, merge: (V, V) => V): RDD[(SpatialKey, V)] with Metadata[M[SpatialKey]]
}
```

Also if we're going to delegate calls to `reduceByKey` to be responsible we should expose at least the overload with partitioner parameter: https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/rdd/PairRDDFunctions.scala#L306

jamesmcclain: I was not able to fully understand everything written above, but I have added the capability to specify a merge function and/or a partitioner.  The methods have to have a name different from `toSpatial` because if they are given that name, they hide the methods given in the other class.  A second class had to be created to avoid changing the API.

jamesmcclain:
I retested an observed the same behavior as before w.r.t. method names.  Others are welcome to replicate in case I did something incorrectly.

echeipesh:
Check as well method names are not an option here. Scala mechanics REPL set some false expectations here:

```scala
Welcome to Scala 2.12.2 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_131).
Type in expressions for evaluation. Or try :help.

scala> case class Box(x: Int)
defined class Box

scala> implicit class withPlus(box: Box) { def plus(x: Int) = Box(box.x + x) }
defined class withPlus

scala> implicit class withPlus2(box: Box) { def plus(x: Int, y: Int) = Box(box.x + x + y) }
defined class withPlus2

scala> val box = Box(4)
box: Box = Box(4)

scala> box.plus(4)
res0: Box = Box(8)

scala> box.plus(5,6)
res1: Box = Box(15)
```

jamesmcclain:
> Check as well method names are not an option here. Scala mechanics REPL set some false expectations here:

I see and understand the code block, but I am afraid that I do not understand what is being asked here.

echeipesh:
> I see and understand the code block, but I am afraid that I do not understand what is being asked here.

No ask there, just an illustration of interesting behavior in REPL that would fail as a stand-alone program.

jamesmcclain:
I see.

PR #2420, opened 2017-10-10
Optimized toArrayTile for GeoTiffMultibandTile

### LINKED ISSUE #2087: Optimize GeoTiffMultibandTile.toArrayTile
Currently this method simply maps over the bands:

```scala
    ArrayMultibandTile(bands.map(_.toArrayTile): _*)
```

This is suboptimal for pixel interleave, because geotiff segments will have to be reloaded for each band.

jbouffard:
This PR optimizes `GeoTiffMultibandTile.toArrayTile` for pixel interleave so that the segments will no longer be reloaded.

This PR resolves #2087

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
+++ b/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
@@ -322,7 +322,7 @@ abstract class GeoTiffMultibandTile(
     * Converts the GeoTiffMultibandTile to an
     * [[ArrayMultibandTile]] */
   def toArrayTile(): ArrayMultibandTile =
-    ArrayMultibandTile(bands.map(_.toArrayTile): _*)
+    crop(this.gridBounds)
 
   /**
    * Performs a crop on itself. The returned MultibandGeoTiffTile will


raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
```diff
@@ -321,8 +324,145 @@ abstract class GeoTiffMultibandTile(
...
+            val segmentTransform = segmentLayout.getSegmentTransform(segmentIndex, 1)
+            val bytes = GeoTiffSegment.deinterleaveBitSegment(segment, segCols, segRows, bandCount, 0 until bandCount)
+
+            cfor(0)(_ < bandCount, _ + 1) { bandIndex =>
+              val segmentBytes = bytes(bandIndex)
+              tileSetInt(new BitGeoTiffSegment(segmentBytes, segCols, segRows), segmentTransform, bandIndex)
+            }
+          }
+        case _ =>
+          val segmentProducer =
```
pomadchin: I'm wondering can we abstract over it? I mean to avoid somehow strict types

pomadchin: Btw, if there is no way to do it, probably we can move that segmentProducer function into `GeoTiffSegment` object.

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
```diff
@@ -321,8 +324,145 @@ abstract class GeoTiffMultibandTile(
...
+              case (_, DoubleUserDefinedNoDataCellType(nd)) =>
+                (bytes: Array[Byte]) => new Float64UserDefinedNoDataGeoTiffSegment(bytes, nd)
+            }
+
+          val bytesPerSample = bandType.bytesPerSample
+          val setter = if (cellType.isFloatingPoint) tileSetDouble else tileSetInt
+
+          getSegments(0 until segmentCount).foreach { case (segmentIndex, geoTiffSegment) =>
+            val segmentTransform = segmentLayout.getSegmentTransform(segmentIndex, bandCount)
+            val bytes = GeoTiffSegment.deinterleave(geoTiffSegment.bytes, bandCount, bytesPerSample, 0 until bandCount)
```
lossyrob: I don't think it's necessary to de-interleave here, and causes more array iterations than necessary.

I think a good approach would be to iterate over the segments as-is, and write to the bands that each value should be. i.e. there's no reason why the band count has to be the outer most loop of the iteration.

Is there a reason why we couldn't call `getSegments` and foreach over the segments as the BandInterleave case, and then write the values to the appropriate place in the result bands?

pomadchin: Would not just `crop(this.gridBounds)` be a desired optimisation?

jbouffard: Yeah, simply doing `crop(this.gridBounds)` would work and is implemented in the desired way.

lossyrob:
Nice catch @pomadchin. :+1:

Issue #2116, opened 2017-04-03
Spatially partition RDDs of vector data


lossyrob:
We have a [VectorJoin](https://github.com/locationtech/geotrellis/blob/d9ca7d3953a62ef437d1842e6d009d73c33d39e7/spark/src/main/scala/geotrellis/spark/join/VectorJoin.scala#L30) which can efficiently join RDDs of vectors, __with the caveat that the RDDs must be spatially partitioned__. Currently, we don't have a very good way to do that. This issue is solved when we have a good spark `Partitioner` that can spatially partition RDDs of Geometries or Feature data, such that VectorJoin is as efficient as possible.

lossyrob:
To clarify my thinking on this:

If partitions of geometries are not spatially partitioned, then the extent of any one partition is likely to be a large percentage of the extent of the dataset as a whole. The [line here](https://github.com/locationtech/geotrellis/blob/d9ca7d3953a62ef437d1842e6d009d73c33d39e7/spark/src/main/scala/org/apache/spark/rdd/FilteredCartesianRDD.scala#L103) will tend to always be true as the partition extents tend to overlap each other, for the `metapred` that [the VectorJoin uses, defined here.](https://github.com/locationtech/geotrellis/blob/d9ca7d3953a62ef437d1842e6d009d73c33d39e7/spark/src/main/scala/geotrellis/spark/join/VectorJoin.scala#L72). This will cause the number of partitions shuffled to tends toward a complete shuffle (like a full cross join), i.e. shuffling each partition to each node in the cluster. So for an N node cluster, we send at most, and tending towards, N-1 copies of the data across the network of the cluster for this join.

A spatial partition of the RDD should result in shuffling each piece of data to at most 1 node. There should not be multiple copies of data elements to multiple nodes; either the geometries would stay on the node they happen to reside prior to the spatial partitioning, or they would be sent to a new partition that resides on another node. After that single shuffle, the join should be much more selective in what partitions it shuffles.

If we have two RDDs that are partitioned with the same partitioner, then we can guarantee that the join will shuffle _no_ partitions. So for each partition step, we would be shuffling at most 1 copy of the data, and the join would shuffle nothing - this is as opposed to shuffling potentially N-1 copies of the data across the wire for on of the RDDs during the join.

[...]

PR #2424, opened 2017-10-12
ClipToGrid funtionality

lossyrob:
This PR adds `ClipToGrid`, which takes an RDD of Features and clips them to `SpatialKey`s based on `LayoutDefinition`.

This functionality is generally useful, but was moved from a version in [VectorPipe](https://github.com/geotrellis/vectorpipe) to here in service of a specific class of vector join problems, in service of #2116. I [added and example](https://github.com/locationtech/geotrellis/pull/2424/files#diff-d80a3e82172d43a81e629e86465decbf) in the `doc-examples` project that shows a polygonal summary of `PolygonFeature[UUID]`s that were in an RDD, based on a set of GeoTiffs on S3.

A full solution to #2116 will take some more thought, which I hear is in line with some work the GeoMesa folk are doing, and we can collaborate on.

Some other things to note:
- I broke out some of the logic in the LayerFilters for MultiPolygon and MultiLine so I could reuse that logic. It seemed like the most appropriate place for it was MapKeyTransform.
- I fixed up a deprecation warning message along the way.

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/tiling/MapKeyTransform.scala
+++ b/spark/src/main/scala/geotrellis/spark/tiling/MapKeyTransform.scala
@@ -18,9 +18,13 @@ package geotrellis.spark.tiling
 
 import geotrellis.spark._
 import geotrellis.raster._
+import geotrellis.raster.rasterize.{Rasterizer, Callback}
 import geotrellis.vector._
 import geotrellis.proj4._
 import geotrellis.util._
+
+import java.util.concurrent.ConcurrentHashMap
+import scala.collection.JavaConverters._
 
 object MapKeyTransform {
   def apply(crs: CRS, level: LayoutLevel): MapKeyTransform =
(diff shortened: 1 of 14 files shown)


spark/src/main/scala/geotrellis/spark/tiling/MapKeyTransform.scala
```diff
@@ -122,4 +126,55 @@ class MapKeyTransform(val extent: Extent, val layoutCols: Int, val layoutRows: I
...
+     * array of  tuples which  is then  mapped-over to  produce an
+     * array of KeyBounds.
+     */
+    val tiles = new ConcurrentHashMap[(Int,Int), Unit]
+    val fn = new Callback {
+      def apply(col : Int, row : Int): Unit = {
+        val tile : (Int, Int) = (bounds.colMin + col, bounds.rowMin + row)
+        tiles.put(tile, Unit)
+      }
+    }
```
echeipesh: Some dragons been found around these parts, better to use this method: https://github.com/locationtech/geotrellis/blob/5c5ebed09f66cde92c77d9f21ce22e65a44573c4/spark/src/main/scala/geotrellis/spark/tiling/LayoutDefinition.scala#L38

fosskers: Whoa, that's a sexy way to quote the line.

lossyrob: @echeipesh I'm using the logic from our LayerFilters for MultiPolygon and MultiLine (which the Line and Polygon filters also use). If this is wrong, then they are also (and have been) wrong. I thought we sorted out all the dragons there - are there some specifics about the dragons?

lossyrob:
Need to investigate and benchmark a way to have an efficient and correct line rasterization implementation for spatial keys. That bumps this PR past 1.2. :opposite of tada:

Issue #2315, opened 2017-08-06
Rethink how we are increasing partitions based on resampling in reproject


lossyrob:
https://github.com/locationtech/geotrellis/commit/a0280ebd51e3bec1a765c623138684e8a0fcd966

When I ran an ingest on a reasonable raster, not intentionally upsampling, I got a very large number of partitions and the ingest failed. When I backed out the above changes it succeeded.

<img width="1664" alt="screen shot 2017-08-06 at 4 07 53 am" src="https://user-images.githubusercontent.com/2320142/29001915-fcda198e-7a63-11e7-9cac-4529fcf7a35f.png">

pomadchin:
I think it can be related to RF ingests, where on a small raster ingest can take more than an hour instead of 2 minutes and shuffle 1TB of data.

lossyrob:
Just to note, I had to delete the following lines to make this ingest work:

```diff
-    // Layout change may imply upsampling, creating more tiles. If so compensate by creating more partitions
-    val part: Option[Partitioner] =
-      for {
-        sourceBounds <- metadata.bounds.toOption
-        targetBounds <- newMetadata.bounds.toOption
-        sizeRatio = targetBounds.toGridBounds.size.toDouble / sourceBounds.toGridBounds.size.toDouble
-        if sizeRatio > 1.5
-      } yield {
-        val newPartitionCount = (bufferedTiles.partitions.length * sizeRatio).toInt
-        logger.info(s"Layout change grows potential number of tiles by $sizeRatio times, resizing to $newPartitionCount partitions.")
-        new HashPartitioner(partitions = newPartitionCount )
-      }
-
-    val tiled = reprojectedTiles.tileToLayout(newMetadata,
-      Tiler.Options(resampleMethod = tilerResampleMethod,
-        partitioner = if (part.isDefined) part else bufferedTiles.partitioner ))
```

lossyrob:
I believe this is the Ingest I was using when hitting this issue: https://github.com/lossyrob/foss4g-2017-geopyspark-workshop/blob/master/ingest/src/main/scala/IngestCDL.scala

PR #2425, opened 2017-10-12
Optimize TileRDDReproject

echeipesh:
Connects https://github.com/locationtech/geotrellis/issues/2294
Connects https://github.com/locationtech/geotrellis/issues/2315
Supersedes  https://github.com/locationtech/geotrellis/pull/2300

In the RDD reproject step instead of projecting tiles to their raster extent, re-projecting and collecting the metadata from the results, use the key bounds and the source layout to perform the the reproject on raster extent as a side compute operation. Because every pixel covered by the `RasterExtent` is in the data source we do not have to consider buffer sizes.

Tracking the changes in the pixel count on per-tile basis allows to get more accurate track of possible increase in raster layer memory footprint. The final tileToLayout operation is going to use the most resolute cellSize for the whole layer so we must perform a rolling adjustment on accumulated pixel counts of reprojected raster extents. (Tiles in different parts of the layer will undergo be warped in different manners, depending on the projections involved) 

Notes:
Tracking the pixel count through the reprojected metadata and then adjusting it for possible changes in layout appears to be giving better results wrt #2294 but still needs verification.

Using spark to launch a job for reprojected metadata collect wrt #2315 appears solid and preferable to the side-collect.

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
+++ b/spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
@@ -60,7 +60,8 @@ object TileRDDReproject {
     metadata: TileLayerMetadata[K],
     destCrs: CRS,
     targetLayout: Either[LayoutScheme, LayoutDefinition],
-    options: Options
+    options: Options,
+    sc: Option[SparkContext] = None
   ): (Int, RDD[(K, V)] with Metadata[TileLayerMetadata[K]]) = {
     val crs: CRS = metadata.crs
     val layout = metadata.layout
--- a/spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
+++ b/spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
@@ -44,7 +44,7 @@ object TileRDDReproject {
     * @tparam           K           Key type; requires spatial component.
     * @tparam           V           Tile type; requires the ability to stitch, crop, reproject, merge, and create.
     *
-    * @param            bufferedTiles                An RDD of buffered tiles, created using the BufferTiles operation.
+    * @param            bufferedTiles      An RDD of buffered tiles, created using the BufferTiles operation.
     * @param            metadata           The raster metadata for this keyed tile set.
     * @param            destCrs            The CRS to reproject to.
     * @param            targetLayout       Either the layout scheme or layout definition to use when re-keying the reprojected layers.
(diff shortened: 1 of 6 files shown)


spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
```diff
@@ -119,34 +120,47 @@ object TileRDDReproject {
           }
         }
 
+    val reprojectSummary = matchReprojectRasterExtent(
+      metadata.crs, destCrs,
+      metadata.layout,
+      Some(metadata.bounds.asInstanceOf[KeyBounds[SpatialKey]])
```
pomadchin: Here is a cast issue, can be resolved some how ~
```scala
metadata.bounds match {
        case KeyBounds(s, e) => Some(KeyBounds(s.getComponent[SpatialKey], e.getComponent[SpatialKey]))
        case _ => None
      }
```

spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
```diff
@@ -269,4 +283,88 @@ object TileRDDReproject {
...
+        import scala.concurrent._
+        import scala.concurrent.duration._
+        import ExecutionContext.Implicits.global
+
+        val splitWork: Iterator[Future[ReprojectSummary]] =
+          boundsChunk.split(128,128) // large enough to be worth a Future
+            .map { subChunk =>
+              Future {
+                // Proj4Transform is not thread safe
+                val transform = Proj4Transform(src, dst)
```
pomadchin: Should we add here a `synchronized` call?

pomadchin: Btw i know that it's still WIP, but probably it makes sense to make execution context somehow configurable.

echeipesh: My thought was that `synchronized` is just going to introduce resource contention and defeat the point of forking a `Future` to do the reprojection.

What is your concern about the choice of executor? This is CPU bound computation and the default executor is supposed well suited for those.

spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
echeipesh: Thank you, good catch.

spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
pomadchin: Hm and what did you mean by thread safety of `Proj4Transform`?

In terms of limited sources it makes sense to have at least configurable pool, to control parallelizm level. But I'm fine with the default pool.

PR #2432, opened 2017-10-16
[WIP] Tobler walking speed focal op

jpolchlo:
This PR adds methods to generate a Tobler walking speed layer from an elevation raster.

**There are issues inherent to this PR that must be resolved prior to merge!**

Notably, the use of this Tobler calculation is direction-agnostic, so all slopes are assumed to be uphill.  (This should be the Cosby distance—his father, after all, did walk to school uphill both ways.) Changes to the way we calculate slope are essential to improve this contribution.


echeipesh:
Going to wait until we have vector slope to merge tobler

PR #2438, opened 2017-10-17
Exact line to grid cell rasterization

jpolchlo:
As described in #2336, rasterizing lines snaps points to their pixel centers.  This can cause problems in certain applications when it is expected that the exact line geometry should be used.  This PR intends to fix this problem by introducing the `Rasterizer.foreachCellInGridLineExact` method.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/rasterize/Rasterizer.scala
+++ b/raster/src/main/scala/geotrellis/raster/rasterize/Rasterizer.scala
@@ -328,6 +328,19 @@ object Rasterizer {
   }
 
   /**
+    * Iterates over the cells determined by the segments of a
+    * LineString.  The iteration happens in the direction from the
+    * first point to the last point.
+    */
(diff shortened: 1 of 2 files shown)


lossyrob:
Is there a reason to not replace our current Rasterizer with this logic (besides risk mitigation)?

jpolchlo:
I was a bit hesitant to mess with the existing stuff before (and so close to) the 1.2 release.  Also seemed like this would result in a change in behavior, which seemed like a definite no-no on a minor point release.  If that's not the case, then we can update the main rasterizer, but it might take a bit of time given that there are probably a few tests that will need modification.  Probably this can be exchanged for the 4-connected rasterizer.  We can quietly tuck in this rasterizing function now without much effect on the API (especially if it's tagged experimental) and not derail the release timing.  Then we can properly stitch it in and make broader API changes for the 2.0 release?  If I'm being too conservative, I'm happy to make more radical changes now.

Issue #2437, opened 2017-10-17
Extent of Layer Not as Expected When Reading RDDs in With HadoopGeoTiffRDD and S3GeoTiffRDD


jbouffard:
This was encountered in GeoPySpark, but this issue most likely exists in GeoTrellis as well. When collecting the `Metadata` for an `RDD`, it was found that the returned `Extent` for the whole layer is different than what was expected.

Below shows the expected and actual returned `Extent`:

```
expected extent: Extent(xmin=141.7066666666667, ymin=-17.73333333333334, xmax=141.92000000000004, ymax=-17.52000000000001)
actual extent:   Extent(xmin=141.7066666666667, ymin=-18.373333333333342, xmax=142.56000000000003, ymax=-17.52000000000001)
```

jamesmcclain:
This was caused by the fact that a tile was of unexpected size and the "expected" extent was being taken from the projected extent of the first entry of a RDD.

PR #2439, opened 2017-10-17
Fix Tile Sizes for maxTileSize < Segment Dimension

jamesmcclain:
The code was producing tiles of the second-largest possible size when `maxSize` was a factor of the `segment` size.  Now they are of the largest possible size.

@jbouffard please test.

Connects https://github.com/locationtech/geotrellis/issues/2437

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/io/RasterReader.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/RasterReader.scala
@@ -61,16 +61,16 @@ object RasterReader {
     }
   }
 
-  private def best(m: Int, n: Int): Int = {
+  private def best(maxSize: Int, segment: Int): Int = {
     var i: Int = 1
-    var result: Int = m
+    var result: Int = -1
     // Search for the largest factor of n that is > 1 and <= m.  If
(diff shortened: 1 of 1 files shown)


jbouffard:
All of the tests pass for GeoPySpark :+1:

jamesmcclain:
:100:

spark/src/main/scala/geotrellis/spark/io/RasterReader.scala
```diff
@@ -61,16 +61,16 @@ object RasterReader {
...
-      if ((n % i == 0) && ((n/i) <= m)) result = (n/i)
+    var result: Int = -1
+    // Search for the largest factor of segment that is > 1 and <=
+    // maxSize.  If one cannot be found, give up and return maxSize.
+    while (i < math.sqrt(segment) && result == -1) {
+      if ((segment % i == 0) && ((segment/i) <= maxSize)) result = (segment/i)
       i += 1
     }
-    result
+    if (result == -1) maxSize; else result
```
pomadchin: btw, semicolon can be removed

PR #2446, opened 2017-10-20
Fix incomplete match warning in Rasterizer

echeipesh:
Also implicit conversion from to polygon has caused the doc generation to break


echeipesh:
Build failure appears to be coming from corrupted test files, I'm seeing similar failures locally that are fixed by updating the `geotiff-test-files` going to tackle that in separate PR.

PR #2453, opened 2017-10-24
ClipToGrid Optimizations

fosskers:
### TODO

- [x] Port VectorPipe code
- [x] Remove `Predicates`
- [x] Remove `LayoutDefinition.keysForGeometry` in favour of `MapKeyTransform.keysForGeometry` ( @echeipesh )

### Motivation

Apologies to @lossyrob - I should have reviewed the original PR.

~~This PR optimizes certain sections of `ClipToGrid` based on benchmark results [from here](https://github.com/fosskers/scala-benchmarks).~~

This PR is a port of functionality from VectorPipe which was changed to incorporate the two core concepts of the original `ClipToGrid`, namely:

- efficient association of Geometries to SpatialKeys
- avoidance of JTS intersections where ever possible

This PR maintains the API exposed by the original `ClipToGrid` (save the removal of `Predicates`), brings back clipping by buffered Extents, allows for easier definition of new clipping functions, and also is easy to "hijack" (i.e. borrow the same logic to inject an error logging step, which is a requirement of VectorPipe).

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
+++ b/spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
@@ -24,6 +24,7 @@ import geotrellis.vector._
 
 import com.vividsolutions.jts.geom.prep.{PreparedGeometry, PreparedGeometryFactory}
 import org.apache.spark.rdd._
+import scala.util.Try
 
 object ClipToGrid {
   /** Trait which contains methods to be used in determining
@@ -31,38 +32,37 @@ object ClipToGrid {
     * for ClipToGrid methods.
     */
   trait Predicates {
-    /** Returns true if the feature geometry covers the passed in extent */
+    /** True if the feature geometry covers the passed-in [[Extent]]. */
     def covers(e: Extent): Boolean
-    /** Returns true if the feature geometry is covered bythe passed in extent */
+    /** True if the feature geometry is covered by the passed-in [[Extent]]. */
     def coveredBy(e: Extent): Boolean
   }
 
-  /** Clips a feature to the given extent, that uses the given predicates
+  /** Clips a feature to the given [[Extent]], using the given [[Predicates]]
     * to avoid doing intersections where unnecessary.
     */
   def clipFeatureToExtent[G <: Geometry, D](
     e: Extent,
-    feature: Feature[G, D],
+    f: Feature[G, D],
     preds: Predicates
-  ): Option[Feature[Geometry, D]] =
-    if(preds.covers(e)) { Some(Feature(e, feature.data)) }
-    else if(preds.coveredBy(e)) { Some(feature) }
-    else {
-      feature.geom.intersection(e).toGeometry.map { g =>
-        Feature(g, feature.data)
-      }
-    }
+  ): Option[Feature[Geometry, D]] = e match {
+    /* If a Feature covers the Extent, their intersection would be the Extent itself. */
+    case _ if preds.covers(e) => Some(Feature(e, f.data))
+    /* The Feature may be completely contained within the Extent. In that case, no clipping need occur at all. */
+    case _ if preds.coveredBy(e) => Some(f)
+    /* Otherwise, we need to perform a JTS intersection */
+    case _ => Try(f.geom.intersection(e)).toOption.flatMap(_.toGeometry.map(g => Feature(g, f.data)))
+  }
 
   /** Clip each geometry in the RDD to the set of SpatialKeys
     * which intersect it, where the SpatialKeys map to the
     * given [[LayoutDefinition]].
     */
   def apply[G <: Geometry](
-    rdd: RDD[G],
-    layout: LayoutDefinition
+    layout: LayoutDefinition,
+    rdd: RDD[G]
   )(implicit d: DummyImplicit): RDD[(SpatialKey, Geometry)] =
-    apply[G, Unit](rdd.map(Feature(_, ())), layout)
-      .mapValues(_.geom)
+    apply[G, Unit](layout, rdd.map(Feature(_, ()))).mapValues(_.geom)
 
   /** Clip each geometry in the RDD to the set of SpatialKeys
     * which intersect it, where the SpatialKeys map to the
@@ -70,17 +70,14 @@ object ClipToGrid {
     * to clip each geometry to the extent.
     */
   def apply[G <: Geometry](
-    rdd: RDD[G],
+    clipGeom: (Extent, G, Predicates) => Option[Geometry],
     layout: LayoutDefinition,
-    clipGeom: (Extent, G, Predicates) => Option[Geometry]
-  )(implicit d: DummyImplicit): RDD[(SpatialKey, Geometry)] =
-    apply[G, Unit](
-      rdd.map(Feature(_, ())),
-      layout,
-      { (e: Extent, f: Feature[G, Unit], p: Predicates) =>
-        clipGeom(e, f.geom, p).map(Feature(_, ()))
-      }
-    ).mapValues(_.geom)
+    rdd: RDD[G]
+  )(implicit d: DummyImplicit): RDD[(SpatialKey, Geometry)] = {
+    val f = { (e: Extent, f: Feature[G, Unit], p: Predicates) => clipGeom(e, f.geom, p).map(Feature(_, ())) }
+
+    apply[G, Unit](f, layout, rdd.map(Feature(_, ()))).mapValues(_.geom)
+  }
 
   /** Clip each geometry in the RDD to the set of SpatialKeys
     * which intersect it, where the SpatialKeys map to the
@@ -88,10 +85,10 @@ object ClipToGrid {
     * to clip each geometry to the extent.
     */
   def apply[G <: Geometry, D](
-    rdd: RDD[Feature[G, D]],
-    layout: LayoutDefinition
+    layout: LayoutDefinition,
+    rdd: RDD[Feature[G, D]]
   ): RDD[(SpatialKey, Feature[Geometry, D])] =
-    apply[G, D](rdd, layout, clipFeatureToExtent[G,D] _)
+    apply[G, D](clipFeatureToExtent[G,D] _, layout, rdd)
 
   /** Clip each geometry in the RDD to the set of SpatialKeys
     * which intersect it, where the SpatialKeys map to the
@@ -99,20 +96,31 @@ object ClipToGrid {
     * to clip each geometry to the extent.
     */
   def apply[G <: Geometry, D](
-    rdd: RDD[Feature[G, D]],
+    clipFeature: (Extent, Feature[G, D], Predicates) => Option[Feature[Geometry, D]],
     layout: LayoutDefinition,
-    clipFeature: (Extent, Feature[G, D], Predicates) => Option[Feature[Geometry, D]]
+    rdd: RDD[Feature[G, D]]
   ): RDD[(SpatialKey, Feature[Geometry, D])] = {
     val mapTransform: MapKeyTransform = layout.mapTransform
 
-    rdd.flatMap { f => clipGeom(mapTransform, clipFeature, f) }
+    rdd.flatMap { f => clipGeom(clipFeature, mapTransform, f) }
   }
 
+  /** Given a clipping function, clip a Geometry according to some
+    * sensible, pre-defined [[Predicates]].
+    */
   private def clipGeom[G <: Geometry, D](
-    mapTransform: MapKeyTransform,
     clipFeature: (Extent, Feature[G, D], Predicates) => Option[Feature[Geometry, D]],
+    mapTransform: MapKeyTransform,
     feature: Feature[G, D]
   ): Iterator[(SpatialKey, Feature[Geometry, D])] = {
+
+    /* Perform the actual clipping */
+    def clipToKey(k: SpatialKey, preds: Predicates): Option[(SpatialKey, Feature[Geometry, D])] = {
+      val extent: Extent = mapTransform(k)
+
+      clipFeature(extent, feature, preds).map(k -> _)
+    }
+
     val pointPredicates =
       new Predicates {
         def covers(e: Extent) = false
@@ -132,20 +140,11 @@ object ClipToGrid {
         def coveredBy(e: Extent) = pg.coveredBy(e.toPolygon.jtsGeom)
       }
 
-    def gcPredicates =
+    val gcPredicates =
       new Predicates {
         def covers(e: Extent) = feature.geom.jtsGeom.covers(e.toPolygon.jtsGeom)
         def coveredBy(e: Extent) = feature.geom.jtsGeom.coveredBy(e.toPolygon.jtsGeom)
       }
-
-    def clipToKey(
-      k: SpatialKey,
-      preds: Predicates
-    ): Option[(SpatialKey, Feature[Geometry, D])] = {
-      val extent = mapTransform(k)
-
-      clipFeature(extent, feature, preds).map(k -> _)
-    }
 
     val iterator: Iterator[(SpatialKey, Feature[Geometry, D])] =
       feature.geom match {
--- a/spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
+++ b/spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
@@ -148,56 +148,45 @@ object ClipToGrid {
 
     val iterator: Iterator[(SpatialKey, Feature[Geometry, D])] =
       feature.geom match {
-        case p: Point =>
-          val k = mapTransform(p)
-          clipToKey(k, pointPredicates).toSeq.iterator
+        case p: Point => Iterator.single(mapTransform(p) -> feature)
         case mp: MultiPoint =>
           mp.points
             .map(mapTransform(_))
-            .distinct
-            .map(clipToKey(_, mpOrLinePredicates))
-            .flatten
+            .toSet
+            .flatMap { k: SpatialKey => clipToKey(k, mpOrLinePredicates) }
             .iterator
         case l: Line =>
           mapTransform.multiLineToKeys(MultiLine(l))
-            .map(clipToKey(_, mpOrLinePredicates))
-            .flatten
+            .flatMap(clipToKey(_, mpOrLinePredicates))
         case ml: MultiLine =>
           mapTransform.multiLineToKeys(ml)
-            .map(clipToKey(_, mpOrLinePredicates))
-            .flatten
+            .flatMap(clipToKey(_, mpOrLinePredicates))
         case p: Polygon =>
           val pg = PreparedGeometryFactory.prepare(p.jtsGeom)
           val preds = polyPredicates(pg)
 
-          mapTransform
-            .multiPolygonToKeys(MultiPolygon(p))
-            .map(clipToKey(_, preds))
-            .flatten
+          mapTransform.multiPolygonToKeys(MultiPolygon(p))
+            .flatMap(clipToKey(_, preds))
         case mp: MultiPolygon =>
           val pg = PreparedGeometryFactory.prepare(mp.jtsGeom)
           val preds = polyPredicates(pg)
 
           mapTransform.multiPolygonToKeys(mp)
-            .map(clipToKey(_, preds))
-            .flatten
+            .flatMap(clipToKey(_, preds))
         case gc: GeometryCollection =>
           def keysFromGC(g: GeometryCollection): List[SpatialKey] = {
-            var keys: List[SpatialKey] = List()
-            keys = keys ++ gc.points.map(mapTransform.apply)
-            keys = keys ++ gc.multiPoints.flatMap(_.points.map(mapTransform.apply))
-            keys = keys ++ gc.lines.flatMap { l => mapTransform.multiLineToKeys(MultiLine(l)) }
-            keys = keys ++ gc.multiLines.flatMap { ml => mapTransform.multiLineToKeys(ml) }
-            keys = keys ++ gc.polygons.flatMap { p => mapTransform.multiPolygonToKeys(MultiPolygon(p)) }
-            keys = keys ++ gc.multiPolygons.flatMap { mp => mapTransform.multiPolygonToKeys(mp) }
-            keys = keys ++ gc.geometryCollections.flatMap(keysFromGC)
-            keys
+            List(
+              gc.points.map(mapTransform.pointToKey),
+              gc.multiPoints.flatMap(_.points.map(mapTransform.pointToKey)),
+              gc.lines.flatMap { l => mapTransform.multiLineToKeys(MultiLine(l)) },
+              gc.multiLines.flatMap { ml => mapTransform.multiLineToKeys(ml) },
+              gc.polygons.flatMap { p => mapTransform.multiPolygonToKeys(MultiPolygon(p)) },
+              gc.multiPolygons.flatMap { mp => mapTransform.multiPolygonToKeys(mp) },
+              gc.geometryCollections.flatMap(keysFromGC)
+            ).flatten
           }
 
-          keysFromGC(gc)
-            .map(clipToKey(_, gcPredicates))
-            .flatten
-            .iterator
+          keysFromGC(gc).flatMap(clipToKey(_, gcPredicates)).iterator
       }
 
     iterator
--- a/spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
+++ b/spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
@@ -16,179 +16,111 @@
 
 package geotrellis.spark.clip
 
-import geotrellis.raster.GridBounds
 import geotrellis.spark._
 import geotrellis.spark.tiling._
-import geotrellis.util._
 import geotrellis.vector._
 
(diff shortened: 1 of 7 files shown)


spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
```diff
@@ -24,95 +24,103 @@ import geotrellis.vector._
...
-        Feature(g, feature.data)
-      }
-    }
+  ): Option[Feature[Geometry, D]] = e match {
+    /* If a Feature covers the Extent, their intersection would be the Extent itself. */
+    case _ if preds.covers(e) => Some(Feature(e, f.data))
+    /* The Feature may be completely contained within the Extent. In that case, no clipping need occur at all. */
+    case _ if preds.coveredBy(e) => Some(f)
+    /* Otherwise, we need to perform a JTS intersection */
+    case _ => Try(f.geom.intersection(e)).toOption.flatMap(_.toGeometry.map(g => Feature(g, f.data)))
```
fosskers: The intersection can still throw Topology Exceptions in weird cases, so we guard against it here.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
```diff
@@ -24,95 +24,103 @@ import geotrellis.vector._
...
+    mapTransform: MapKeyTransform,
     feature: Feature[G, D]
   ): Iterator[(SpatialKey, Feature[Geometry, D])] = {
+
+    /* Perform the actual clipping */
+    def clipToKey(k: SpatialKey, preds: Predicates): Option[(SpatialKey, Feature[Geometry, D])] = {
+      val extent: Extent = mapTransform(k)
+
+      clipFeature(extent, feature, preds).map(k -> _)
+    }
```
fosskers: Worker function moved to the top.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
```diff
@@ -132,73 +140,53 @@ object ClipToGrid {
         def coveredBy(e: Extent) = pg.coveredBy(e.toPolygon.jtsGeom)
       }
 
-    def gcPredicates =
```
fosskers: Probably a leftover from a previous signature that involved a `PreparedGeometry`?

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
```diff
@@ -132,73 +140,53 @@ object ClipToGrid {
...
-
-      clipFeature(extent, feature, preds).map(k -> _)
-    }
-
     val iterator: Iterator[(SpatialKey, Feature[Geometry, D])] =
       feature.geom match {
-        case p: Point =>
-          val k = mapTransform(p)
-          clipToKey(k, pointPredicates).toSeq.iterator
+        case p: Point => Iterator.single(mapTransform(p) -> feature)
```
fosskers: `Point` will never be outside its own `Extent`, so we can skip all attempts at clipping. The point predicates *do* help with skipping a lot of the work in `clipToKey`, but by avoiding that altogether we perform even fewer operations.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
```diff
@@ -132,73 +140,53 @@ object ClipToGrid {
...
         case mp: MultiPoint =>
           mp.points
             .map(mapTransform(_))
-            .distinct
-            .map(clipToKey(_, mpOrLinePredicates))
-            .flatten
+            .toSet
```
fosskers: `toSet` is always faster than `distinct` for collections smaller than ~100k elements.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
```diff
@@ -132,73 +140,53 @@ object ClipToGrid {
...
+            .flatMap { k: SpatialKey => clipToKey(k, mpOrLinePredicates) }
```
fosskers: Here and elsewhere, `flatMap := flatten ∘ map`, except `flatMap` is faster.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
```diff
@@ -132,73 +140,53 @@ object ClipToGrid {
...
-            keys = keys ++ gc.multiPoints.flatMap(_.points.map(mapTransform.apply))
-            keys = keys ++ gc.lines.flatMap { l => mapTransform.multiLineToKeys(MultiLine(l)) }
-            keys = keys ++ gc.multiLines.flatMap { ml => mapTransform.multiLineToKeys(ml) }
-            keys = keys ++ gc.polygons.flatMap { p => mapTransform.multiPolygonToKeys(MultiPolygon(p)) }
-            keys = keys ++ gc.multiPolygons.flatMap { mp => mapTransform.multiPolygonToKeys(mp) }
-            keys = keys ++ gc.geometryCollections.flatMap(keysFromGC)
-            keys
+            List(
+              gc.points.map(mapTransform.pointToKey),
+              gc.multiPoints.flatMap(_.points.map(mapTransform.pointToKey)),
```
fosskers: New aliases.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
lossyrob: This assumes that we don't want to change the metadata for the geometry in the clip function. That's an OK assumption.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
fosskers: Haha we could debate this one to the moon. I agree in general that swallowing errors isn't good. Throwing exceptions in `IO` code is sometimes correct. Throwing exceptions in pure code is never correct.

Now with soapboxing out of the way... I refactored this bit first before I realized how it was used further down. The analogue to this in VectorPipe also takes a logging function, which logs in the case of `None` returned by any clippers (like this function). What's our path forward?

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
```diff
@@ -24,95 +24,103 @@ import geotrellis.vector._
...
+    apply[G, Unit](layout, rdd.map(Feature(_, ()))).mapValues(_.geom)
 
   /** Clip each geometry in the RDD to the set of SpatialKeys
     * which intersect it, where the SpatialKeys map to the
     * given [[LayoutDefinition]], using the given method
     * to clip each geometry to the extent.
     */
   def apply[G <: Geometry](
-    rdd: RDD[G],
+    clipGeom: (Extent, G, Predicates) => Option[Geometry],
```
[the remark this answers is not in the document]
fosskers: Ah, gotcha. I was going with the usual FP order which is significant for currying. Is that a conversation you'd like to have? If no, which is totally fine, I can revert them back.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
fosskers: Ah I didn't think of that. Want me to switch it back and add a comment? Or how about a `lazy val`?

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
fosskers: It also assumes something else: that user-invented clipping functions and `Predicates` will always be well-defined. We can imagine a case where the user is passing in clipping functions that do nonsensical things, such that a `Point` being outside of its Extent could be significant. This sounds like "garbage in, garbage out" though, and will probably never come up.

Which made me think that `Predicates` can probably be hidden from the user entirely. Is the set of `Predicates` you've provided hands-down the best? If so, we might be able to apply them in all cases and hide them from the API.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
fosskers: > If there's issues with the input, the user should know.

Yes, we're totally on the same page about this. Swallowing errors is never good (we can leave the exception debate aside). So the GeoTrellis version of this function only returns in `Option` because of the `.toGeometry` then, yeah? Given that there's no logging function passed in to deal with failed clips right away (like in VP), I'm having trouble seeing the integration path with VP. I'll prototype some more on this PR and hopefully we can reach something nice.

fosskers: Logging while being robust and skipping failed clips was a specific requirement of VectorPipe, so that one naughty geom doesn't crash the whole ingest. The only way to do that would be to catch Topology Exceptions directly within an `RDD.map`, so the injected `.clipToGrid` methods here wouldn't be usable by VP as-is.

Still pondering, I'll have an answer some time this morning.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
```diff
@@ -16,191 +16,111 @@
...
     * which intersect it, where the SpatialKeys map to the
     * given [[LayoutDefinition]].
     */
-  def apply[G <: Geometry](
-    rdd: RDD[G],
-    layout: LayoutDefinition
-  )(implicit d: DummyImplicit): RDD[(SpatialKey, Geometry)] =
-    apply[G, Unit](rdd.map(Feature(_, ())), layout)
-      .mapValues(_.geom)
+  def geometries[G <: Geometry](layout: LayoutDefinition, rdd: RDD[G]): RDD[(SpatialKey, Geometry)] =
```
fosskers: I'm completely willing to reinstate overloaded `apply`s. I originally switched the names because the overloads were interfering with type resolution.

These names (`.geometries`, and `.features`) are also in the spirit of the new, more descriptive `MapKeyTransform` method aliases. They're a suggestion for a way forward.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
```diff
@@ -16,191 +16,111 @@
...
   ): RDD[(SpatialKey, Feature[Geometry, D])] = {
     val mapTransform: MapKeyTransform = layout.mapTransform
 
-    rdd.flatMap { f => clipGeom(mapTransform, clipFeature, f) }
+    /* Associate each Feature with its SpatialKeys*/
+    val withKeys: RDD[(Iterator[SpatialKey], Feature[G,D])] =
+      rdd.map(f => (mapTransform.keysForGeometry(f.geom), f))
+
+    /* Clip every Feature */
+    withKeys.flatMap { case (ks, f) => ks.flatMap(k => clipFeature(k.extent(layout), f).map((k, _))) }
```
fosskers: The `Feature <-> Key` association step and clipping step are now separated.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
```diff
@@ -16,191 +16,111 @@
...
+    case g: Line                               => clip(extent, g).map(Feature(_, f.data))
+    case g: MultiLine  if coveredBy(g, extent) => Some(f)
+    case g: MultiLine                          => clip(extent, g).map(Feature(_, f.data))
+    case g =>  /* Polygon and MultiPolygon */
+      val pg = PreparedGeometryFactory.prepare(g.jtsGeom)
+      val ep = extent.toPolygon.jtsGeom
+
+      if (pg.covers(ep)) Some(Feature(extent, f.data))
+      else if (pg.coveredBy(ep)) Some(f)
+      else clip(extent, g).map(Feature(_, f.data))
```
fosskers: The logic contained in `Predicates` was considered the best reasonable approach for the given Geoms, so the same checks are now hard-coded into the canonical clipping function `byExtent`.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
```diff
@@ -16,191 +16,111 @@
...
-      clipFeature(extent, feature, preds).map(k -> _)
+  private def coveredBy[G <: Geometry](g: G, e: Extent): Boolean =
+    g.jtsGeom.coveredBy(e.toPolygon.jtsGeom)
+
+  private def clip[G <: Geometry](e: Extent, g: G): Option[Geometry] = {
+    val exPoly: Polygon = e.toPolygon
+
+    val clipped: Try[Geometry] = g match {
+      case mp: MultiPolygon => Try(MultiPolygon(mp.polygons.flatMap(_.intersection(exPoly).as[Polygon])))
+      case _ => Try(g.intersection(exPoly).toGeometry.get)
```
fosskers: Not throwing an Exception is a requirement of VectorPipe ingest behaviour, so we catch here. Hm, actually the catch might still be able to be done at the VP level, not here, since it will be calling `byExtent` itself in an `RDD.map`, checking for errors anyway, and logging.

spark/src/main/scala/geotrellis/spark/tiling/MapKeyTransform.scala
```diff
@@ -176,4 +176,26 @@ class MapKeyTransform(val extent: Extent, val layoutCols: Int, val layoutRows: I
...
+      List(
+        gc.points.map(pointToKey),
+        gc.multiPoints.flatMap(_.points.map(pointToKey)),
+        gc.lines.flatMap { l => multiLineToKeys(MultiLine(l)) },
+        gc.multiLines.flatMap { ml => multiLineToKeys(ml) },
+        gc.polygons.flatMap { p => multiPolygonToKeys(MultiPolygon(p)) },
+        gc.multiPolygons.flatMap { mp => multiPolygonToKeys(mp) },
+        gc.geometryCollections.flatMap(keysForGeometry)
+      ).flatten.toSet.iterator
+  }
```
fosskers: The new go-to for associating Geometries with their keys generically.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
```diff
@@ -16,191 +16,111 @@
...
-
-      clipFeature(extent, feature, preds).map(k -> _)
+  private def coveredBy[G <: Geometry](g: G, e: Extent): Boolean =
+    g.jtsGeom.coveredBy(e.toPolygon.jtsGeom)
+
+  private def clip[G <: Geometry](e: Extent, g: G): Option[Geometry] = {
+    val exPoly: Polygon = e.toPolygon
+
+    val clipped: Try[Geometry] = g match {
+      case mp: MultiPolygon => Try(MultiPolygon(mp.polygons.flatMap(_.intersection(exPoly).as[Polygon])))
```
fosskers: Not sure if this is the best thing to do. It was a work-around from long ago to help prevent clips on large MultiPolygons frmo going insane.

spark/src/main/scala/geotrellis/spark/tiling/MapKeyTransform.scala
fosskers: Were `LayoutDefinition` and `SpatialKey` in `geotrellis-core`, `Geometry` would have an abstract method `Geometry.keys: LayoutDefinition => Iterator[SpatialKey]` and each subtype would implement directly the functionality seen here in `keysForGeometry`. In large jobs, that approach should save many pattern matches.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
```diff
@@ -16,191 +16,111 @@
...
-            .flatten
-            .iterator
-      }
-
-    iterator
+    clipped.toOption
   }
+
+  /** Clips Features to a 3x3 grid surrounding the current Tile.
+    * This has been found to capture ''most'' Features which stretch
```
lossyrob: Found to capture where? In VectorPipe? We could be using ClipToGrid in a setting where there are tiny tiles and big polygons, so I don't think this comment applies across the board.

byBufferedExtent is a pretty VP specific use case. Not sure this should be elevated into GeoTrellis, or just kept in VP.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
lossyrob: This is a de-optimization.

Building a prepared Geometry takes time. That is why the original code was only doing it once for each of the clips it was going to do on that Geometry.

The predicates also apply across the board no matter what clipping method you use, so I don't think they should be hidden in a single implementation. That's why I elevated them out of the clipping logic and passed into the clip function, so that any clip function can take advantage of the predicates.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
lossyrob: No, we should let the underlying intersection logic of JTS's MultiPolygon do it's thing.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
lossyrob: Flattening out an exception into a None is one step up from swallowing the exception: now the user knows there was something wrong, since they got a None, but have no idea what happens. I'd prefer if the error be transmitted some way, and if that's bubbling the exception up and having VP catch it to do what it needs, then that seems to work.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
lossyrob: We use overloaded applys across the library, especially in this pattern of object methods -> method extensions. I get that you don't like that pattern, but changing my new code to your preferred pattern when it is not established at the project level that we will start moving to not using applys is irksome.

fosskers: No need to be irked, this stuff is never personal. As I said above I'm totally willing to use `apply` if that's deemed best. PRs are always suggestions.

Anyway, my motivation for avoiding `apply` was multifold:

- Team belief: `MapKeyTransform`'s overloaded `apply`s were mysterious to the point of warranting more descriptive aliases (i.e. good names can signal intent better).
- Practical trouble: I was having trouble getting the code to compile with the overloads.
- Team precedent: Since the `MapKeyTransform` stuff, I'm not the first to introduce a lone (i.e. bag-o'-functions only) companion object that included only descriptive names and not overloaded `apply`s.
- Personal belief: an `apply` returning a type other than its parent type is always non-mathematical.

Would you like them switched back to `apply`?

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
fosskers: Good catch. The first thing that comes to mind is switching the signature of clipping functions to:

```scala
(Iterator[Extent], Feature[G, D]) => Iterator[Option[Feature[G, D]]]
```

since the new `MapKeyTransform.keysForGeometry` also returns an `Iterator`. The actual clipping (done in a `Iterator.map`) and the subsequent `foldLeft` to grab out the legal Features would be optimally fast due to the chained, lazy `Iterator` ops. VectorPipe would do the error checking in the `foldLeft`, and vanilla GT can throw as it sees fit.

I had a question for you for when you got back to the office: was the intersection meant to be done on the `PreparedGeometry`, or just the `covers` check? At current both the original `ClipToGrid` and this rewrite only use the `PreparedGeometry` for the `covers` check. Was that an oversight or am I misunderstanding?

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
fosskers: Given [my suggestion above](https://github.com/locationtech/geotrellis/pull/2453#discussion_r147568159), throwing inside the `(Iterator[Extent], Feature[G, D]) => Iterator[Option[Feature[G, D]]]` still wouldn't fit VectorPipe's use-case, as we'd like to document each failed clip separately. 

How about this code here still uses `Try`, casts to `Option`, and the default GT behaviour higher up in the `clipToGrid` overloads is to detect those `Option` and throw? Returning `Try` or `Either` would be the most transparent, but you've said in the past that you prefer Exceptions because it avoids further boxing.

Either way, VectorPipe won't be using the injected `clipToGrid` methods, since it needs fine-grained control over the error reporting process and so will apply the clipping functions itself in its own `RDD.map`.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
fosskers: Cool, I can remove it. It was a fairly old work-around.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
lossyrob: I think we should avoid doing code changes to other people's code if it's a matter of switching from a well-established preference to a personal one, even if the personal one can be elevated to a prefered one in the future but that is not yet happend. Switching back undoes this behavior, so please switch it back.

As for moving forward, I understand your motivations. And if the project wants to set this as a 
standard, we should communicate that clearly. Your bullet points shouldn't go into a PR that changes a recent code addition, but perhaps into a GitHub issue that argues for this change in a 2.0, which we would apply across the board and then keep to moving forward.

It's possible to be irked when things are not personal; I don't think you're making this change out of anything but pure intentions. However spending cycles on debating these types of finer points is something I'd like to avoid. Which is why we should either set the standard or avoid microchanges to other's work.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
lossyrob: There was not an oversight. PreparedGeometry is only an optimization for the predicate, and not the intersection operation.

What's the reason for changing it from what it was? I'm unclear why this part of the code needed to change.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
lossyrob: If you turn a Try into an Option you lose all of the error information.

If VectorPipe uses it's own clip method, it seems like this shouldn't be an issue for VP. If client code wants to deal with clip-specific errors in a certain way, they should be able to do that in the passed in clip method. The current code accounts for that by having the `clipFeature` method coming in from the client return an option, so that if you wanted to handle errors by ignoring them or reporting them with some logging, you can do that and return a None.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
fosskers: Cool, `apply` it is.

And just to be as clear as possible, the primary motivation for the rewrite was that the original form of `ClipToGrid` didn't quite match VectorPipe's use-case. I didn't discover that until checking the new code directly to figure out what the integration path would be. I should've reviewed the original PR - apologies for not having done that.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
fosskers: Ok cool. As for the change, I also mentioned the reasons a bit [here](https://github.com/locationtech/geotrellis/pull/2453#discussion_r147570288). Integrating `ClipToGrid` was my immediate task for VP, but I hit issues. I read through your code a lot and  saw an isomorphic algorithm that didn't involve the `Predicates` concept, so I [prototyped it over on the VP side](https://github.com/geotrellis/vectorpipe/pull/29). I explained everything to the team and was given the go-ahead to port it back.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
fosskers: My plan was to exposes aliases in VP to the clipping functions in this PR, so that the user wouldn't have to think about GT imports and symbols.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
lossyrob: Ok, thanks.

Let's figure out how the original code doesn't fit the use case - I had done some thinking around how it would, and so VectorPipe could rely on the code but also do whatever it wanted, like log errors in it's own specific way. What did I miss that doesn't allow VP to use things? Is it an aesthetic use case (e.g. not wanting to use Predicates in the Clip methods though it solves it) or is VP blocked from using this code for it's use case for some ohter reason?

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
```diff
@@ -16,191 +16,111 @@
...
     * which intersect it, where the SpatialKeys map to the
     * given [[LayoutDefinition]].
     */
-  def apply[G <: Geometry](
-    rdd: RDD[G],
-    layout: LayoutDefinition
-  )(implicit d: DummyImplicit): RDD[(SpatialKey, Geometry)] =
-    apply[G, Unit](rdd.map(Feature(_, ())), layout)
-      .mapValues(_.geom)
+  def apply[G <: Geometry](layout: LayoutDefinition, rdd: RDD[G]): RDD[(SpatialKey, Geometry)] =
```
lossyrob: The argument ordering is intentional and fits with the rest of the method extensions across the library.

the `object Verb` has `apply` methods that take as a first argument the subject. That subject is the `self` in the MethodExtensions.

Issue #2406, opened 2017-10-04
Transitive dependency load of `geotrellis-spark-etl` is significant.


metasim:
While this may be made irrelevant by #1767, I wanted to record the impact and results of a problem we just had to address relating to the transitive dependency load of `geotrellis-spark-etl`. 

We were running into a large number of evictions and assembly/shading problems causing bugs in our Zeppelin integration (e.g. Spark UI missing executors tab). It became clear that shading wasn't enough, and marking `geotrellis-spark-etl` as `intransitive()` caused problems in other code. Part of the problem is our fault, having some ETL dependent stuff mixed in with a bunch of non-ETL features. But after trying several things, the easiest fix was to add the following to our build:

```scala
excludeDependencies ++= Seq(
  "org.slf4j" % "slf4j-log4j12",
  "org.apache.hbase" % "hbase-server",
  "org.apache.hbase" % "hbase-client",
  "org.apache.hbase" % "hbase-common",
  "org.apache.hbase" % "hbase-hadoop-compat",
  "org.apache.accumulo" % "accumulo-core",
  "com.datastax.cassandra" % "cassandra-driver-core"
)
```
As result, our dependency load reduced in half:

**Before** &rarr; **After**

![img_3294](https://user-images.githubusercontent.com/131013/31176333-82901f42-a8e0-11e7-9c7d-aa0fce80b601.png)

`hbase-server` was about a third of that half.

So, if the "pipeline" code ends up being delayed for one reason or another, I suggest we investigate fixing the `geotrellis-spark-etl` architecture to not have such extensive runtime dependencies, either through the recent SPI feature @echeipesh wrote, or hack a solution using reflection that only instantiates drivers if they are available in the classpath (currently the ETL module instantiates entry-points to all the backends at startup). We plan to separate out our own ETL code at some point soon so this problem isn't an issue, but thought bringing it up here was important in case it affects other users.

pomadchin:
Well my opinion is to stop etl project maintaining and either to mark it as deprecated or to do nothing with it, just not to include it into `2.0`.

pomadchin:
It was partially addressed via `pipeline` project, though until we don't have SPI for Writers we can't exclude S3 dep. Closing it for now.

Issue #2107, opened 2017-03-29
Allow selective inclusion of ETL backend libraries


metasim:
To reduce the size of shaded jars and other `spark-submit` payloads, it would be helpfulI one could selectively include the backend libraries as needed. We tried marking `geotrellis-spark-etl` as `intransitive()` and then manually adding in the desired back-end (e.g. `geotrellis-accumulo`) , but the ETL framework eagerly instantiates the modules, causing `ClassNotFoundException`s on things relating to `hbase` and `cassandra`.

To de-couple the ETL from backend providers, I'd suggest using the Java SPI framework. It is very simple to use and implement for plugins conforming to a particular interface. Instead of having hard coded lists of I/O modules, you'd query the `ServiceLoader.load` method for a list of services discovered in the classpath. The official docs from Oracle are [here](https://docs.oracle.com/javase/tutorial/ext/basics/spi.html). Oddly, they make it look more complicated than it is in reality. You basically define an interface/trait for your API, and in your provider libraries add a special file in your resources declaring an implementation of the service.

cc: @mteldridge

pomadchin:
Reasonable comments, thanks!
Related to #1767 
And it's definitely and API break.

metasim:
@pomadchin Yes, we tried basically this:

```
libraryDependencies ++= Seq(
 geotrellis("raster"),
  geotrellis("spark"),
  geotrellis("spark-etl") intransitive(),
  "com.github.fge" % "json-schema-validator" % "2.2.6", // <-- needed for geotrellis-spark-etl
  geotrellis("accumulo")
)
```

but got `ClassNotFoundException` during the Spark job due to hbase, etc. GT classes. It appears that [this construction](https://github.com/locationtech/geotrellis/blob/master/spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala#L42) results in [this being called](https://github.com/locationtech/geotrellis/blob/master/spark-etl/src/main/scala/geotrellis/spark/etl/hbase/HBaseModule.scala#L21), which at some point leads to [this](https://github.com/locationtech/geotrellis/blob/master/spark-etl/src/main/scala/geotrellis/spark/etl/hbase/HBaseModule.scala#L21), which is in the `hbase` module. I'd have to get Matt to re-run to get the exact stack trace, so let me know if it's not clear.

metasim:
Addressed by https://github.com/locationtech/geotrellis/pull/2223

[...]

PR #2458, opened 2017-10-26
Add `shapefile` And `geotools` To ETL Assembly

jamesmcclain:
If the ETL assembly is going to be used as the official (or quasi-official or recommended) überjar, I think that there is an argument to be made for including `shapefile` and `geotools` which are needed for many exploratory workflows.


pomadchin:
Well, taking into account #2107 and #2406 I am not sure that ETL assembly is indeed was designed to be an uberjar, it's more an undesired feature to include as much as it is possible different deps (actually it is a bad design thing).

Probably we can create a separate subproject that will contain all possible gt deps, or transform current ETL into it after making pipeline stable.

PR #2462, opened 2017-10-30
ClipToGrid Optimizations - Round 2

fosskers:
### TODO

- [x] `MapKeyTransform.keysForGeometry`
- [x] Use it in `ClipToGrid`
- [ ] Factor out `coveredBy` ?

@echeipesh , we had decided that this was to be two PRs, but the changes to `MapKeyTransform` (the `Iterator -> Set` change) broke `ClipToGrid`. 

Closes #2453 .

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
+++ b/spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
@@ -30,7 +30,7 @@ object ClipToGrid {
     * the most optimal way to clip geometries and features
     * for ClipToGrid methods.
     */
-  trait Predicates {
+  trait Predicates extends Serializable {
     /** Returns true if the feature geometry covers the passed in extent */
     def covers(e: Extent): Boolean
     /** Returns true if the feature geometry is covered bythe passed in extent */
@@ -113,29 +113,32 @@ object ClipToGrid {
     clipFeature: (Extent, Feature[G, D], Predicates) => Option[Feature[Geometry, D]],
     feature: Feature[G, D]
   ): Iterator[(SpatialKey, Feature[Geometry, D])] = {
-    val pointPredicates =
+
+    /* Unique keys that this Feature touches */
+    val keys: Set[SpatialKey] = mapTransform.keysForGeometry(feature.geom)
+
+    lazy val mpOrLinePredicates =
       new Predicates {
         def covers(e: Extent) = false
-        def coveredBy(e: Extent) = true
+        def coveredBy(e: Extent) = keys.size < 2
       }
 
-    val mpOrLinePredicates =
-      new Predicates {
-        def covers(e: Extent) = false
-        def coveredBy(e: Extent) =
-          feature.geom.jtsGeom.coveredBy(e.toPolygon.jtsGeom)
-      }
-
-    def polyPredicates(pg: PreparedGeometry) =
+    def preparedPredicates(pg: PreparedGeometry) =
       new Predicates {
         def covers(e: Extent) = pg.covers(e.toPolygon.jtsGeom)
-        def coveredBy(e: Extent) = pg.coveredBy(e.toPolygon.jtsGeom)
+        def coveredBy(e: Extent) = keys.size < 2
       }
 
-    def gcPredicates =
+    lazy val polyPredicates =
       new Predicates {
         def covers(e: Extent) = feature.geom.jtsGeom.covers(e.toPolygon.jtsGeom)
-        def coveredBy(e: Extent) = feature.geom.jtsGeom.coveredBy(e.toPolygon.jtsGeom)
+        def coveredBy(e: Extent) = keys.size < 2
+      }
+
+    lazy val gcPredicates =
+      new Predicates {
+        def covers(e: Extent) = feature.geom.jtsGeom.covers(e.toPolygon.jtsGeom)
+        def coveredBy(e: Extent) = keys.size < 2
       }
 
     def clipToKey(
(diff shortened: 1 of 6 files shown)


spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
```diff
@@ -30,7 +30,7 @@ object ClipToGrid {
     * the most optimal way to clip geometries and features
     * for ClipToGrid methods.
     */
-  trait Predicates {
+  trait Predicates extends Serializable {
```
fosskers: ... just in case.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
```diff
@@ -113,29 +113,32 @@ object ClipToGrid {
     clipFeature: (Extent, Feature[G, D], Predicates) => Option[Feature[Geometry, D]],
     feature: Feature[G, D]
   ): Iterator[(SpatialKey, Feature[Geometry, D])] = {
-    val pointPredicates =
+
+    /* Unique keys that this Feature touches */
+    val keys: Set[SpatialKey] = mapTransform.keysForGeometry(feature.geom)
+
+    lazy val mpOrLinePredicates =
```
fosskers: `lazy` now to avoid allocations.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
```diff
@@ -113,29 +113,32 @@ object ClipToGrid {
...
+        def covers(e: Extent) = feature.geom.jtsGeom.covers(e.toPolygon.jtsGeom)
+        def coveredBy(e: Extent) = keys.size < 2
       }
 
-    def gcPredicates =
+    lazy val gcPredicates =
       new Predicates {
         def covers(e: Extent) = feature.geom.jtsGeom.covers(e.toPolygon.jtsGeom)
-        def coveredBy(e: Extent) = feature.geom.jtsGeom.coveredBy(e.toPolygon.jtsGeom)
+        def coveredBy(e: Extent) = keys.size < 2
```
fosskers: The `coveredBy` condition is now the same for all geometry types. It could be factored out with the `keys.size < 2` check happening once below in the `val iterator = ...`. This would avoid billions of pattern matches on the full data set.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
```diff
@@ -147,59 +150,28 @@ object ClipToGrid {
...
-          mapTransform.multiLineToKeys(ml)
-            .map(clipToKey(_, mpOrLinePredicates))
-            .flatten
-        case p: Polygon =>
+        /* Points will always exist inside their Extent */
+        case p:  Point      => keys.iterator.map(k => (k, feature))
+        case mp: MultiPoint => keys.iterator.flatMap(k => clipToKey(k, mpOrLinePredicates))
+        case l:  Line       => keys.iterator.flatMap(k => clipToKey(k, mpOrLinePredicates))
+        case ml: MultiLine  => keys.iterator.flatMap(k => clipToKey(k, mpOrLinePredicates))
+        case p:  Polygon if keys.size > 10 =>
```
fosskers: Pulled 10 out of thin air. Actually now that I think about it, if `keys.size > 8`, then the Polygon will always cover the Extent and we never have to prepare the geom nor do intersections. We only need to *actually* check for covers when the key size is between 8 and 2... in which case would we even both with preparing it?

fosskers: Not true. What if the Polygon stretched outside its Extent (and so had many keys), but had a hole inside the Extent? In that case, returning just the `Extent` as the faux-clipped Geom would be a mistake.

fosskers: Most Polygons in OSM don't have holes, though. It feels like we can still use that information to save tons of operations.

fosskers: https://geotrellis.github.io/scaladocs/latest/index.html#geotrellis.vector.Polygon@hasHoles:Boolean

If no, then the `keys.size > 8` should always be reliable.

fosskers: https://github.com/locationtech/geotrellis/blob/master/vector/src/main/scala/geotrellis/vector/Polygon.scala#L107

I wonder if this is `O(1)`.

fosskers: Doesn't matter, actually. The `covers` check still needs to be performed *per Extent*, since each Extent varies in shape. The number of such Extents doesn't matter.

spark/src/main/scala/geotrellis/spark/tiling/MapKeyTransform.scala
```diff
@@ -140,19 +141,18 @@ class MapKeyTransform(val extent: Extent, val layoutCols: Int, val layoutRows: I
...
       def apply(col: Int, row: Int): Unit = {
-        val tile: (Int, Int) = (bounds.colMin + col, bounds.rowMin + row)
-        tiles.put(tile, Unit)
+        tiles += ((bounds.colMin + col, bounds.rowMin + row))
       }
     }
 
     multiLine.lines.foreach { line => Rasterizer.foreachCellByLineStringDouble(line, rasterExtent)(fn) }
-    tiles.keys.asScala.map { case (col, row) => SpatialKey(col, row) }
+    tiles.map { case (col, row) => SpatialKey(col, row) }.toSet
```
echeipesh: This is a drive by optimization, but it stands out given PR title: `SpatialKey` is just as good as `Tuple2` to be stored in the `MSet` and this map can be just avoided all together.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
echeipesh: Yeah, can't use the number of keys for `covers` because geometry could always be skinnier than the extent no matter the key count.

spark/src/main/scala/geotrellis/spark/tiling/MapKeyTransform.scala
fosskers: My assumption is that @lossyrob had it as `Tuple2` because of some JVM optimizations that occur on a `Tuple2` of primitives. You have a point though, that said gain might be undone by the extra `.map`.

spark/src/main/scala/geotrellis/spark/tiling/MapKeyTransform.scala
```diff
@@ -159,21 +153,15 @@ class MapKeyTransform(val extent: Extent, val layoutCols: Int, val layoutRows: I
...
 
-    /*
-     * Use the Rasterizer to construct  a list of tiles which meet
-     * the  query polygon.   That list  of tiles  is stored  as an
-     * array of  tuples which  is then  mapped-over to  produce an
-     * array of KeyBounds.
-     */
-    val tiles: MSet[(Int, Int)] = MSet.empty
+    /* Use the Rasterizer to construct a list of tiles which meet the query polygon. */
+    val tiles: mutable.Set[SpatialKey] = mutable.Set.empty
```
fosskers: Apparently it's good practice to prefix mutable containers with `mutable`, as opposed to aliasing it like I had done.

fosskers:
Don't merge quite yet - I'm adding some docstrings.

fosskers:
Good to go.

spark/src/main/scala/geotrellis/spark/clip/ClipToGrid.scala
lossyrob: :+1: good catch

PR #2451, opened 2017-10-23
Lightweight BufferTiles operation

jpolchlo:
The current version of `BufferTiles` relies on specifying the `BufferSizes` via an `RDD[(K, BufferSizes)]`.  This method requires a `join` that really hams up the works.  This PR proposes an alternative pathway that takes a function of type `K => BufferSizes`.  Presumably, the former RDD method can be converted to this approach in nearly all cases.

Signed-off-by: jpolchlo <jpolchlopek@azavea.com>

PR #2472, opened 2017-11-07
Avoid using RDD join to buffer sizes in RDD reproject

echeipesh:
Connects: https://github.com/locationtech/geotrellis/pull/2451

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
+++ b/spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
@@ -187,37 +187,29 @@ object TileRDDReproject {
       val crs = rdd.metadata.crs
       val layout = rdd.metadata.layout
       val tileLayout = rdd.metadata.layout.tileLayout
+      // Avoid capturing instantiated Transform in closure because its not Serializable
+      lazy val transform = Transform(crs, destCrs)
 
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
```diff
@@ -187,37 +187,29 @@ object TileRDDReproject {
...
-              right = 1 + (if(gb.colMax >= re1.cols) gb.colMax - (re1.cols - 1) else 0),
-              top = 1 + (if(gb.rowMin < 0) -gb.rowMin else 0),
-              bottom = 1 + (if(gb.rowMax >= re1.rows) gb.rowMax - (re1.rows - 1) else 0)
-            )
-          }
-
-      val bufferedTiles =
-        rdd.bufferTiles(borderSizesPerKey)
+      // Avoid capturing instantiated Transform in closure because its not Serializable
+      lazy val transform = Transform(crs, destCrs)
```
echeipesh: This is a trick worth calling out for smells (I think its inline with normal level of trickery).

Having a `lazy val` means that we're shipping only a `null` reference and a thunk as the value of the `transform`. Then each executor JVM will init its own instance of `transform`, once, and re-use it in the RDD evaluation. Thus this devilry avoids dealing with `TaskNotSerializable` exception and also allows re-use of heavy weight object per partition.

PR #2477, opened 2017-11-09
Filter by extensions in S3GeoTiffInfoReader

pomadchin:
An idea i had during making these changes, probably it makes sense to use in `{Hadoop | S3}GeoTiffInfoReader` `Options` object (from `{Hadoop | S3}GeoTiffRDD`)?


jbouffard:
This looks good to me :+1:

echeipesh:
I don't think it should pull in the `Options` objects, none of the remaining fields in the options could influence the reading the metadata.

PR #2478, opened 2017-11-09
`Line.head` and `Line.last`

fosskers:
This PR adds two efficient functions for quickly grabbing either the first or last `Point` in a `Line`.
This lets us skip having to call `Line.points`, which is expensive.


fosskers:
`head` and `last` are the only methods we can guarantee to be total. Anything else would have to return in `Option`.

pomadchin:
I just want to say, that it's unobvious that Line even has a `getPoint` method.

PR #2479, opened 2017-11-09
Make `Point6D` a top-level type

fosskers:
Hidden inside of `IterativeViewshed`, it wouldn't appear in Scaladocs and was a pain to import.

Ping @echeipesh .


metasim:
Since this isn't a "point" in a geometric sense, is another name more appropriate? It's more like a "camera" or "viewport".

echeipesh:
That's a good ...  point ... maybe `Viewpoint` ?

fosskers:
Good call, I like Viewpoint.

PR #2480, opened 2017-11-09
[WIP] Extend Terraform EMR deployment scripts with Jupyterhub and Jupyter-scala

jpolchlo:
After an initial trial period, it appears that Zeppelin is not the answer to all one's hopes for an accessible, user-friendly, Spark-enabled Scala environment.  This PR takes a different tack to achieve the same end using Jupyter notebook and the `jupyter-scala` kernel.  

The approach offered here is somewhere in the middle of the bare-bones, just-make-it-work-quick-and-dirty and fully-featured-but-boy-are-you-gonna-need-to-do-some-configuration-to-get-this-working ends of the spectrum.  The simplest approach would be to set up JupyterHub and expect the user to log in as `hadoop:hadoop`.  This PR goes through the trouble of setting up GitHub oauth functionality.  This can easily be removed in favor of the quick-and-dirty method.  (The reason to add oauth was in the hope that this authenticator could be used to provide some mirroring of a third-party file share for persistent notebook storage so as not to lose work if the cluster suddenly vanishes.)

In order to test this PR as it stands, one will need to create an OAuth application using their github account.  Initially, the homepage URL and authorization callback URL values are unimportant, so provide dummy values there.  We need the client ID and client secret tokens to start our cluster.

Navigate to the `scripts/emr` directory under the repository root.  Specify the following environment variables:
- `TF_VAR_access_key`: AWS access key
- `TF_VAR_secret_key`: AWS secret key
- `TF_VAR_pem_path`: Path to the AWS account PEM on your local machine
- `TF_VAR_oauth_client_id`: From github OAuth application
- `TF_VAR_oauth_client_secret`: From github OAuth application

Now run `make terraform-init && make create-cluster`.  This will produce a URL for the cluster of the form `ec2-xx-xx-xx-xx.compute-1.amazonaws.com`.  Use that URL to update the GitHub OAuth application settings with `http://ec2-xx-xx-xx-xx.compute-1.amazonaws.com:8000` as the homepage URL, and `http://ec2-xx-xx-xx-xx.compute-1.amazonaws.com:8000/hub/oauth_callback` as the authorization callback URL.  Now visit `http://ec2-xx-xx-xx-xx.compute-1.amazonaws.com:8000`.

Logging in via github will create a new account on the cluster and display the new (empty) home directory in the file browser.  You may create a new scala notebook here or upload a notebook from your local machine.  A good starting point is `https://gist.github.com/jpolchlo/77ae121627a6e28b9b3d2f145a69acb7`.  This notebook shows how to use the `jupyter-scala` kernel to import dependencies (including GT) and use them in a notebook.  I have verified that vanilla Spark operations and GT Spark operations behave normally in this environment.

Things to try before merging this PR include:
- [ ] Using the `hadoop:hadoop` login instead of OAuth
- [ ] Finding a jupyter plugin that allows mirroring the contents of an S3 folder
- [ ] Switching to Google OAuth and mirroring a google drive folder (such a plugin definitely exists)

I'll take recommendations on other things that would be good to try.

### DIFF
--- a/scripts/emr/terraform/emr-spark.tf
+++ b/scripts/emr/terraform/emr-spark.tf
@@ -65,11 +65,25 @@ resource "aws_emr_cluster" "emr-spark-cluster" {
   provisioner "file" {
     source      = "bootstrap.sh"
     destination = "/tmp/bootstrap.sh"
+    connection {
+      type        = "ssh"
+      user        = "hadoop"
+      host        = "${aws_emr_cluster.emr-spark-cluster.master_public_dns}"
+      private_key = "${file("${var.pem_path}")}"
+    }
   }
 
-  bootstrap_action {
-    path = "file:///tmp/bootstrap.sh"
-    name = "Install JupyterHub"
+  provisioner "remote-exec" {
+    inline=[
+      "chmod +x /tmp/bootstrap.sh",
+      "/tmp/bootstrap.sh ${var.oauth_client_id} ${var.oauth_client_secret}"
+    ]
+    connection {
+      type        = "ssh"
+      user        = "hadoop"
+      host        = "${aws_emr_cluster.emr-spark-cluster.master_public_dns}"
+      private_key = "${file("${var.pem_path}")}"
+    }
   }
 }
 
--- a/scripts/emr/terraform/emr-spark.tf
+++ b/scripts/emr/terraform/emr-spark.tf
@@ -7,7 +7,7 @@ provider "aws" {
 
 # `aws_emr_cluster` is built-in to Terraform. We name ours `emr-spark-cluster`.
 resource "aws_emr_cluster" "emr-spark-cluster" {
-  name          = "EMR GeoTrellis Zeppelin"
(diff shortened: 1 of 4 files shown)


jpolchlo:
Note: S3-backed storage could be enabled by `https://github.com/danielfrg/s3contents`

scripts/emr/terraform/emr-spark.tf
```diff
@@ -7,7 +7,7 @@ provider "aws" {
 
 # `aws_emr_cluster` is built-in to Terraform. We name ours `emr-spark-cluster`.
 resource "aws_emr_cluster" "emr-spark-cluster" {
-  name          = "EMR GeoTrellis Zeppelin"
+  name          = "EMR GeoTrellis - ${var.user}"
```
pomadchin: 👍

pomadchin:
Can you add a command to start only Zeppelin cluster too (without jupyther bootstrap)?

jpolchlo:
I'm going to shut down this PR in favor of #2481; your comment should eventually be addressed there.

PR #2481, opened 2017-11-10
Extend Terraform EMR deployment scripts with Jupyterhub, Jupyter-scala, and S3-backed notebook storage

jpolchlo:
Supersedes #2480

After an initial trial period, it appears that Zeppelin is not the answer to all one's hopes for an accessible, user-friendly, Spark-enabled Scala environment. This PR takes a different tack to achieve the same end using Jupyter notebook and the jupyter-scala kernel.

This PR, in contrast to #2480 upon which it is based, dispenses with Oauth, and gives a more direct interface.  The only extravagant feature included here is the ability to have persistent notebook storage on S3.

To use this PR, navigate to the scripts/emr directory under the repository root. Specify the following environment variables:

    TF_VAR_access_key: AWS access key
    TF_VAR_secret_key: AWS secret key
    TF_VAR_pem_path: Path to the AWS account PEM on your local machine
    TF_VAR_s3_notebook_bucket: S3 bucket containing persistent notebook store
    TF_VAR_s3_notebook_prefix: S3 prefix for stored notebooks

Now run make terraform-init && make create-cluster. After an initial startup (between 7 and 8 minutes), this will produce a URL for the cluster of the form ec2-xx-xx-xx-xx.compute-1.amazonaws.com.  Visit http://ec2-xx-xx-xx-xx.compute-1.amazonaws.com:8000 and log in as `user` with the (highly secure) password `password`.

From here, you will be presented with the file browser, which should be populated with the contents of the S3 folder specified above. You may create new scala notebooks here, upload notebooks from your local machine, or simply rely on the files found in the S3 bucket. You may wish to start with the notebook found here: https://gist.github.com/jpolchlo/77ae121627a6e28b9b3d2f145a69acb7. This notebook shows how to use the jupyter-scala kernel to import dependencies (including GT) and use them in a notebook.

This PR is not yet complete.  The following list suggests tasks that should be completed prior to removing the WIP tag. 
- [x] Create a make target for Zeppelin-only clusters.  Cluster creation time is noticeable longer for this version over the previous Zeppelin-based EMR deployment.  It would make sense to offer a way to select between a jupyter-enabled cluster and a standard cluster.
- [x] Learn how to import fat jars/extend the class path in Ammonite and document
- [x] Provide a template so that new Scala notebooks load up with the standard boilerplate code in place for importing Spark/creating a SparkContext/importing GeoTrellis


jpolchlo:
Currently having a bit of trouble with the bootstrap scripts automatically starting up the jupyterhub instance.  For now, for testing purposes, you'll have to `make ssh` after `make create-cluster` and then execute `cd /tmp && launch_hub`.  Following that, the jupyterhub instance should be running on port 8000.

jpolchlo:
This most recent commit should properly start up jupyterhub during bootstrapping.  In the worst case, there might be a slight delay before the service comes on line (I had to use a cron process, which has a minimum granularity of minutes).

jpolchlo:
@echeipesh @pomadchin This is ready for final review.  Pretty well tested it, but give it a shot yourselves.

Issue #2469, opened 2017-11-03
Ingesting large geotiff from s3 doesn't seem to be partitioning data properly


jmelching:
I've been testing out the 1.2.0-RC1 with some existing code that just ingests one year of the USDA's cropland dataset and seeing a strange behavior... It appears that the entire geotiff (13 GB's) might be being read by a single task as it tries to crop the tiff into tiles.
Here's the thread dump of the only running executor:

`
java.net.SocketInputStream.socketRead0(Native Method)
java.net.SocketInputStream.socketRead(SocketInputStream.java:116)
java.net.SocketInputStream.read(SocketInputStream.java:171)
java.net.SocketInputStream.read(SocketInputStream.java:141)
sun.security.ssl.InputRecord.readFully(InputRecord.java:465)
sun.security.ssl.InputRecord.read(InputRecord.java:503)
sun.security.ssl.SSLSocketImpl.readRecord(SSLSocketImpl.java:983) => holding Monitor(java.lang.Object@1820339677})
sun.security.ssl.SSLSocketImpl.readDataRecord(SSLSocketImpl.java:940)
sun.security.ssl.AppInputStream.read(AppInputStream.java:105) => holding Monitor(sun.security.ssl.AppInputStream@1648873021})
org.apache.http.impl.io.SessionInputBufferImpl.streamRead(SessionInputBufferImpl.java:139)
org.apache.http.impl.io.SessionInputBufferImpl.read(SessionInputBufferImpl.java:200)
org.apache.http.impl.io.ContentLengthInputStream.read(ContentLengthInputStream.java:178)
org.apache.http.conn.EofSensorInputStream.read(EofSensorInputStream.java:137)
com.amazonaws.internal.SdkFilterInputStream.read(SdkFilterInputStream.java:82)
com.amazonaws.event.ProgressInputStream.read(ProgressInputStream.java:180)
com.amazonaws.internal.SdkFilterInputStream.read(SdkFilterInputStream.java:82)
com.amazonaws.services.s3.internal.S3AbortableInputStream.read(S3AbortableInputStream.java:125)
com.amazonaws.internal.SdkFilterInputStream.read(SdkFilterInputStream.java:82)
com.amazonaws.internal.SdkFilterInputStream.read(SdkFilterInputStream.java:82)
com.amazonaws.internal.SdkFilterInputStream.read(SdkFilterInputStream.java:82)
com.amazonaws.event.ProgressInputStream.read(ProgressInputStream.java:180)
[...]

PR #2482, opened 2017-11-14
Fix GeoTiffRDD Segment Partitioning

echeipesh:
Resolves: https://github.com/locationtech/geotrellis/issues/2469

Additionally there needs to be some effort to review the added code since 1.1 that enabled these features to make sure it still makes sense given this refactor.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
+++ b/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
@@ -325,31 +325,63 @@ abstract class GeoTiffMultibandTile(
     crop(this.gridBounds)
 
   /**
-   * Performs a crop on itself. The returned MultibandGeoTiffTile will
-   * contain bands that have the same area as the inputte GridBounds.
+   * Crop this tile to given pixel region.
    *
-   * @param  gridBounds  A [[GridBounds]] that contains the area to be cropped.
-   * @return             A [[ArrayMultibandTile]]
+   * @param bounds Pixel bounds specifying the crop area.
    */
-  def crop(gridBounds: GridBounds): ArrayMultibandTile = {
-    val bands = Array.fill(bandCount)(ArrayTile.empty(cellType, gridBounds.width, gridBounds.height))
-    val intersectingSegments = segmentLayout.intersectingSegments(gridBounds)
-    if (hasPixelInterleave) {
-      logger.debug(s"Cropping $gridBounds uses ${intersectingSegments.length} out of ${segmentCount} pixel-interleave segments")
+ def crop(gridBounds: GridBounds): ArrayMultibandTile =
+  crop(List(gridBounds)).next._2
 
-      // de-interlace the pixels from each segment
-      for ((segmentId, segment) <- getSegments(intersectingSegments)) {
-        val segmentBounds = segmentLayout.getGridBounds(segmentId)
-        val segmentTransform = segmentLayout.getSegmentTransform(segmentId, bandCount)
+  /**
+   * Crop this tile to given pixel regions.
+   *
+   * @param windows Pixel bounds specifying the crop areas
+   */
+  def crop(windows: Seq[GridBounds]): Iterator[(GridBounds, ArrayMultibandTile)] = {
+    case class Chip(
+      window: GridBounds,
+      bands: Array[MutableArrayTile],
+      intersectingSegments: Int,
+      var segmentsBurned: Int = 0)
+    val chipsBySegment = mutable.Map.empty[Int, List[Chip]]
+    val intersectingSegments = mutable.SortedSet.empty[Int]
+
+    for (window <- windows) {
+      val segments: Array[Int] =
+        if (hasPixelInterleave) {
+          segmentLayout.intersectingSegments(window)
+        } else {
+          val bandSegmentCount = segmentCount / bandCount
+          for {
+            band <- 0 until bandCount
+            segmentId <- segmentLayout.intersectingSegments(window)
+          } yield segmentId + (band * bandSegmentCount)
+        }.toArray
+
+      val bands = Array.fill(bandCount)(ArrayTile.empty(cellType, window.width, window.height))
+      val chip = Chip(window, bands, segments.length)
+      for (segment <- segments) {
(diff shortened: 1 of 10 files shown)


echeipesh:
Visual confirmation that tiff was tiled/painted. This tiff seems to cover very large area with not a lot of data though.

screen shot 2017-11-14 at 3 23 48 pm

echeipesh:
timings on `S3GeoTiffRDD.spatial("bigtiffs-test", "2469/2013_30m_cdls.tif")(sc).count()`
over 10 `m3.xlarge` seems in line with @pomadchin timings in `1.2.0-M1`
screen shot 2017-11-14 at 3 32 13 pm

s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffRDD.scala
```diff
@@ -75,15 +75,16 @@ object S3GeoTiffRDD extends LazyLogging {
     crs: Option[CRS] = None,
     timeTag: String = GEOTIFF_TIME_TAG_DEFAULT,
     timeFormat: String = GEOTIFF_TIME_FORMAT_DEFAULT,
-    maxTileSize: Option[Int] = None,
+    maxTileSize: Option[Int] = Some(256),
```
pomadchin: probably we don't need `maxWindowSize`, be careful with fixing tests.

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
```diff
@@ -325,31 +325,63 @@ abstract class GeoTiffMultibandTile(
...
-        val segmentBounds = segmentLayout.getGridBounds(segmentId)
-        val segmentTransform = segmentLayout.getSegmentTransform(segmentId, bandCount)
-        val overlap = gridBounds.intersection(segmentBounds).get
+  /**
+   * Crop this tile to given pixel regions.
+   *
+   * @param windows Pixel bounds specifying the crop areas
+   */
+  def crop(windows: Seq[GridBounds]): Iterator[(GridBounds, ArrayMultibandTile)] = {
+    case class Chip(
```
pomadchin: I'm confused that it lives inside a function, mb to put into a package object and to make it private?

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffSegmentLayout.scala
```diff
@@ -122,14 +125,159 @@ case class GeoTiffSegmentLayout(totalCols: Int, totalRows: Int, tileLayout: Tile
...
+    *
+    * @param windows List of pixel windows from this layout
+    * @param maxPartitionSize Maximum pixel count for each partition
+    */
+  def partitionWindowsBySegments(
+    windows: Seq[GridBounds],
+    maxPartitionSize: Long
+  ): Array[Array[GridBounds]] = {
+    val partition = mutable.ArrayBuilder.make[GridBounds]
+    partition.sizeHintBounded(128, windows)
```
pomadchin: Hm, why 128?

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
echeipesh: The singleband case has a slightly different `case class Chip`, it really has no meaning outside of this function. Do you think there is a downside to having it here ?

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffSegmentLayout.scala
echeipesh: Pretty safe guess I guess. If each window is the common case of 256x256 and a double 128 gets you up to 68MB.

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
pomadchin: Nope, just wanted to be sure that it's here because useless anywhere else

s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffRDD.scala
echeipesh: Good point on tests. I'm going to remove the configuration, it doesn't seem to serve its purpose now. One thing to call out is that `maxTileSize` is really that. We don't have any way to generate windows of a fixed size, which this class used to do at some point. I would regard that as unspecified behavior though unless there are strong opinions to the contrary.

pomadchin: https://github.com/locationtech/geotrellis/blob/master/s3/src/test/scala/geotrellis/spark/io/s3/S3GeoTiffRDDSpec.scala#L176

Max tileSize would be 251 in both rdds in this tests, if you'll pass explicit 1024 it would pass

pomadchin: Also remove [this function](https://github.com/locationtech/geotrellis/blob/master/s3/src/test/scala/geotrellis/spark/io/s3/S3GeoTiffRDDSpec.scala#L54) at all :+1:

s3/src/test/scala/geotrellis/spark/io/s3/S3GeoTiffRDDSpec.scala
```diff
@@ -89,11 +80,9 @@ class S3GeoTiffRDDSpec
       mockClient.putObject(bucket, key, geoTiffBytes)
 
       val source1 =
-        S3GeoTiffRDD.spatial(bucket, key, S3GeoTiffRDD.Options(partitionBytes = None, getS3Client = () => new MockS3Client))
-      val source2 = {
-        setWindowSize(128)
-        S3GeoTiffRDD.spatial(bucket, key, S3GeoTiffRDD.Options(getS3Client = () => new MockS3Client))
-      }
+        S3GeoTiffRDD.spatial(bucket, key, S3GeoTiffRDD.Options(maxTileSize = None, partitionBytes = None, getS3Client = () => new MockS3Client))
```
pomadchin: Good catch

PR #2483, opened 2017-11-15
Reprojection Documentation

fosskers:
In order to improve our reprojection API, we first need collated documentation that demonstrates the API's current form. This PR provides that.


fosskers:
### API oddities

First, reprojecting core types like `Line`, `Raster`, and `Extent` follows a pattern:

```scala
val webm: Line = ...
val latlng: Line = webm.reproject(WebMercator, LatLng)
```
implying the typeclass:
```scala
import simulacrum._

@typeclass trait Projectable[A] {
  def reproject(a: A, source: CRS, target: CRS): A
}
```
But the RDD reprojection methods don't follow this pattern. The simplest invocation (of the many) is:
```scala
val webm: TileLayerRDD[SpatialKey] = ...
val layout: LayoutDefinition = ...

val (zoom, latlng): (Int, TileLayerRDD[SpatialKey]) = webm.reproject(LatLng, layout)
```
`TileLayerRDD` knows its `CRS` inside of itself, and also requires the "destination grid".

Second, the RDD-based reproject also returns an `Int` whose meaning is not documented anywhere. 

Third, the validity of the `source` CRS is unchecked. Unlike `TileLayerRDD` who knows who he is, `Line` and friends are implicity in a projection but are unable to report which one.

### Questions

- Is there a single `Projectable` typeclass definition that matches both the primitive and RDD use-case?
- Should GT primitives be aware of their projections, allowing for:
```scala
@typeclass trait Projectable[A] {
  def reproject(a: A, target: CRS): A
}
```
which is at least a bit closer to the `RDD` variant.

docs/guide/rasters.rst
```diff
@@ -443,6 +443,73 @@ each run - often referred to as 'optimized' rendering. By default, a
...
+   val (zoom, ll): (Int, TileLayerRDD[SpatialKey]) = wm.reproject(LatLng, layout)
+
+Let's break down the last line some more:
+
+::
+
+   val (zoom, ll): (Int, TileLayerRDD[SpatialKey]) = wm.reproject(LatLng, layout)
+        [1]                                                       [2]     [3]
+
+- [1]: We are also given back the zoom level corresponding to ??? TODO!
```
fosskers: Help.

docs/guide/rasters.rst
```diff
@@ -443,6 +443,73 @@ each run - often referred to as 'optimized' rendering. By default, a
...
+- [2]: We only need to provide the target ``CRS`` here, since a ``TileLayerRDD``
+  implicitely knows its own projection.
+- [3]: Providing a ``LayoutDefinition`` allows us to blah blah blah TODO
```
fosskers: Help.

docs/guide/rasters.rst
```diff
@@ -443,6 +443,73 @@ each run - often referred to as 'optimized' rendering. By default, a
...
+
+To reproject a ``Tile``:
+
+.. code-block:: scala
+
+   val wm: Tile = ...
+   val extent: Extent = ....  /* Area covered by the Tile */
+   val raster: Raster[Tile] = Raster(wm, extent)  /* A Raster is a "location-aware" Tile */
+
+   val ll: Raster[Tile] = wm.reproject(WebMercator, LatLng)
```
echeipesh: type-o: `wm` is a tile, you meant `raster`

docs/guide/rasters.rst
echeipesh: Zoom level returned has meaning when input is `LayoutScheme`, where the reproject function will determine the appropriate zoom by picking the closest resolution (see example)[https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/tiling/ZoomedLayoutScheme.scala#L67]

In case of `LayoutDefinition` or `FloatingLayoutScheme` there is no possible pyramid that can be referenced with a zoom level, so its always `0`.

docs/guide/rasters.rst
echeipesh: Define the target layout we're retrojecting to. This shouldn't be too obscure, I'm curious where the confusion came from.

docs/guide/rasters.rst
```diff
@@ -443,6 +443,73 @@ each run - often referred to as 'optimized' rendering. By default, a
...
+You may also have a different formulation if your data source is a giant GeoTiff
+on S3, and not a pre-ingested GeoTrellis layer. In that case, you'd have:
+
+.. code-block:: scala
+
+   /* Magically extract windowed Tiles from a GeoTiff */
+   val wm: RDD[(ProjectedExtent, Tile)] = S3GeoTiffRDD.spatial("s3-bucket-name", "your-image.tiff")
+
+   val ll: RDD[(ProjectedExtent, Tile)] = wm.reproject(LatLng)
+
```
echeipesh: The key point here would be that the shape of RDD is different from `TileLayerRDD[SpatialKey]` and they're reprojected on per-tile basis as opposed to per-layer basis. To reproject an RDD as a layer it the tiles need to be buffered by the pixels from their neighbors so the reproject kernel may sample them during transformation.

docs/guide/rasters.rst
```diff
@@ -443,6 +443,73 @@ each run - often referred to as 'optimized' rendering. By default, a
...
+- [2]: We only need to provide the target ``CRS`` here, since a ``TileLayerRDD``
+  implicitely knows its own projection.
+- [3]: Providing a ``LayoutDefinition`` allows us to blah blah blah TODO
+
+You may also have a different formulation if your data source is a giant GeoTiff
+on S3, and not a pre-ingested GeoTrellis layer. In that case, you'd have:
+
+.. code-block:: scala
+
+   /* Magically extract windowed Tiles from a GeoTiff */
```
echeipesh: Why `Magically` and `windowed`?
[`S3GeoTiffRDD`](https://github.com/locationtech/geotrellis/blob/master/s3/src/main/scala/geotrellis/spark/io/s3/S3GeoTiffRDD.scala#L116) gained a little more detail in the docs recently that maybe (?) helpful.

docs/guide/rasters.rst
fosskers: With the `zoom` return val seemingly significant, I was unsure how the layout affected the result. Mostly because `TileLayerRDD` already know their own layout.

docs/guide/rasters.rst
fosskers: > ... the tiles needs to be buffered ...

Meaning the user has to buffer them themselves ahead of time, or the reproject does the buffering behind the scenes?

docs/guide/examples.rst
```diff
@@ -3,6 +3,48 @@ Example Archive
...
+    val ungridded: RDD[(ProjectedExtent, Tile)] =
+      S3GeoTiffRDD.spatial(bucket, key)
+
+    /* Say it was in WebMercator and we want it in ConusAlbers. */
+    val reprojected: RDD[(ProjectedExtent, Tile)] =
+      ungridded.reproject(ConusAlbers)
+
+    /* Our layer will need metadata. Luckily, this can be derived mostly for free. */
+    val (zoom, meta): (Int, TileLayerMetadata[SpatialKey]) =
+      TileLayerMetadata.fromRdd(reprojected, ZoomedLayoutScheme(ConusAlbers))
```
echeipesh: I am not certain this line will actually compile, happy to be proven wrong. IIRC when using `fromRdd` functions you need to provide all the type parameters `K, V, K2`. This is the primary reason to use `.collectMetadata[SpatialKey]` form that infers the params from `reprojected` and just requires target key type, `K2`.

fosskers: This does compile, I wrote the example first in an actual editor.

fosskers: https://github.com/fosskers/playground/blob/master/scala/geotrellis/src/main/scala/playground/S3Geotiff.scala

fosskers:
A thought, having done the example. Conceivably, the following could exist:

```scala
@typeclass trait Layer[F[_]] {
...
  def pyramid[K: Spatial, V](l: F[(K, V)], startZoom: Int): Stream[(Int, F[(K, V)])]
}
```
Or something like that with properly constrained `KVM` type variables. Possibly with the input as `Int Refined Positive` from [refined](https://github.com/fthomas/refined) to prevent a negative zoom being given.

echeipesh:
Another related question that keeps coming up is the need for `Projected[A]` currently there are a bunch of a `Projected_` compositions that tuple a `CRS` and something. I wonder how that would relate to the `Projectable[A]` considering that the `reproject` method would have access the original CRS.

docs/guide/examples.rst
echeipesh: I withdraw my baseless accusation!

echeipesh:
Thank you, this nice addition and should help with common questions.

Issue #2484, opened 2017-11-16
Incorrect Bounds in Reprojected Layer


jbouffard:
When reprojecting a layer in GPS, I'm noticing that the `Bounds` of the layer are way off. These are the `Bounds` that are expected vs. what it actual is:

```
This is the expected Bounds(minKey=SpatialKey(col=1450, row=966), maxKey=SpatialKey(col=1479, row=996))
This is the actual Bounds(minKey=SpatialKey(col=39135189, row=-5244971), maxKey=SpatialKey(col=41807069, row=-2549900))
This is the expected Bounds(minKey=SpatialKey(col=725, row=483), maxKey=SpatialKey(col=739, row=498))
This is the actual Bounds(minKey=SpatialKey(col=19567594, row=-2622486), maxKey=SpatialKey(col=20903534, row=-1274950))
This is the expected Bounds(minKey=SpatialKey(col=362, row=241), maxKey=SpatialKey(col=369, row=249))
This is the actual Bounds(minKey=SpatialKey(col=9783797, row=-1311243), maxKey=SpatialKey(col=10451767, row=-637475))
This is the expected Bounds(minKey=SpatialKey(col=181, row=120), maxKey=SpatialKey(col=184, row=124))
This is the actual Bounds(minKey=SpatialKey(col=4891898, row=-655622), maxKey=SpatialKey(col=5225883, row=-318738))
This is the expected Bounds(minKey=SpatialKey(col=90, row=60), maxKey=SpatialKey(col=92, row=62))
This is the actual Bounds(minKey=SpatialKey(col=2445949, row=-327811), maxKey=SpatialKey(col=2612941, row=-159369))
This is the expected Bounds(minKey=SpatialKey(col=45, row=30), maxKey=SpatialKey(col=46, row=31))
This is the actual Bounds(minKey=SpatialKey(col=1222974, row=-163906), maxKey=SpatialKey(col=1306470, row=-79685))
This is the expected Bounds(minKey=SpatialKey(col=22, row=15), maxKey=SpatialKey(col=23, row=15))
This is the actual Bounds(minKey=SpatialKey(col=611487, row=-81953), maxKey=SpatialKey(col=653235, row=-39843))
This is the expected Bounds(minKey=SpatialKey(col=11, row=7), maxKey=SpatialKey(col=11, row=7))
This is the actual Bounds(minKey=SpatialKey(col=305743, row=-40977), maxKey=SpatialKey(col=326617, row=-19922))
[...]

PR #2485, opened 2017-11-17
Fix TileRDDMetadata generation

echeipesh:
Resolves: https://github.com/locationtech/geotrellis/issues/2484

Type-o caused using the source layout definition rather than target layout definition when generating reprojected metadata


jbouffard:
:+1:

Issue #2486, opened 2017-11-17
RDDs Read in Using Hadoop/S3GeoTiffRDD Have Different Layout Extents Depending on the maxTileSize


jbouffard:
It has been found that when reading in `RDD`s using either `Hadoop` or `S3` `GeoTiffRDD`, that the resulting `layout.extent` for each `RDD` will differ depending on the `maxTileSize` passed in.

This code sample will create a collection of `layout.extent`s from each `RDD` that was read in.

```scala
def ingest(size: Int) = HadoopGeoTiffRDD.spatial(new Path("/tmp/srtm_52_11.tif"), options=optMaxTile(size))

val sizes = 7 to 58 by 9
val floating = FloatingLayoutScheme(256, 256)
val layoutExtents = sizes.map(ingest(_).collectMetadata[SpatialKey](floating)._2.layout.extent)
```

When checking the distinct values of the collection, it was found that there were 6 different `Extent`s, one for each `RDD`. However, there should have only been one, as each `layout.extent` should be the same.

```
scala> layoutExtents.toSet
res68: scala.collection.immutable.Set[geotrellis.vector.Extent] =Set(Extent(74.99958369653905, 4.8804167635160525, 80.11958369653439, 10.000416763514124), Extent(74.99958369653905, 4.880416763514688, 80.11958369653439, 10.000416763514124), Extent(74.99958369653905, 4.880416763514287, 80.11958369653696, 10.000416763514124), Extent(74.99958369653905, 4.880416763514163, 80.11958369653775, 10.000416763514124), Extent(74.99958369653905, 4.8804167635142335, 80.11958369653803, 10.000416763514124), Extent(74.99958369653905, 4.880416763514415, 80.11958369653439, 10.000416763514124))

scala> layoutExtents.toSet.size
res69: Int = 6
```

The cause of this bug can be found in `MapKeyTransform` in this method: https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/tiling/MapKeyTransform.scala#L125-L131 The resulting `Extent` produced will be incorrect.

```
scala> val e = Extent(0, 0, 10, 10)
e: geotrellis.vector.Extent = Extent(0.0, 0.0, 10.0, 10.0)

scala> val cs = CellSize(e, 10, 10)
cs: geotrellis.raster.CellSize = CellSize(1.0,1.0)

scala> val ge = GridExtent(e, cs)
[...]

PR #2487, opened 2017-11-17
Fix mapTransform on GeoTiffSegmentLayout

echeipesh:
MapKeyTransform is intended to provide context for tile layouts, not pixels.
Also replacing its current usage with RasterExtent

Connects: https://github.com/locationtech/geotrellis/issues/2486
(Discovered during investigation)


jbouffard:
:+1:

PR #2491, opened 2017-11-29
Rasterizer-based reprojection

jpolchlo:
The current reprojection code reprojects tiles in their entirety, requiring a CutTiles step followed by a merge step.  This additional cutting stage can cause problems in large-scale ingests.  The present PR aims to produce a new implementation of TileRDDReproject that produces tiles directly using a rasterizer-based approach.  The new operation produces a polygon representation of the post-projection coverage area, and flat-mapping to fill in the new tiles using a rasterization function that back-projects and resamples from the original tile.

This method may be more computationally-intensive, but loses the cutting stage, potentially saving a shuffle step.  It also should allow a more general approach to tiling a collection of rasters, allowing for rasters in different projections/resolutions to be joined in a single step.

- [x] Basic raster implementation
- [x] Reimplementation of TileRDDReproject


spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
```diff
@@ -54,126 +55,163 @@ object TileRDDReproject {
...
-          val m = TileLayerMetadata(metadata.cellType, layoutDefinition, extent, destCrs, kb)
+    val rasterReprojectOptions = options.rasterReprojectOptions.copy(
+      parentGridExtent = Some(targetLayerLayout: GridExtent),
+      targetCellSize = None,
+      targetRasterExtent = None
+    )
+
+    val newMetadata = {
+      metadata.copy(
+        layout = targetLayerLayout,
```
echeipesh: forgot to change the CRS in the target layout

raster/src/main/scala/geotrellis/raster/reproject/MultibandRasterizeReprojectMethods.scala
```diff
@@ -0,0 +1,66 @@
...
+    destRegion: Polygon,
+    destRE: RasterExtent, 
+    destCRS: CRS,
+    resampleMethod: ResampleMethod, 
+    destCellType: CellType
+  ): ProjectedRaster[MultibandTile] = {
+    val trans = Proj4Transform(destCRS, srcCRS)
+    val bands = Array.ofDim[MutableArrayTile](self.tile.bandCount)
+
+    cfor(0)(_ < self.tile.bandCount, _ + 1) { i =>
```
echeipesh: Iteration of the bands can be brought inside the loop to re-use the `val (tx, ty) = trans(x, y)` call for every pixel in the stack.

raster/src/main/scala/geotrellis/raster/reproject/MultibandRasterizeReprojectMethods.scala
```diff
@@ -0,0 +1,66 @@
...
+      bands(i) = self.band(i).prototype(destCellType, destRE.cols, destRE.rows).mutable
+      val resampler = Resample.apply(resampleMethod, 
+                                     bands(i),
```
echeipesh: This `resampler` is resampling from the empty bands that were just allocated, also producing much `NODATA` and gnashing of teeth. 

`self.bands(i)` was what was intended.

Some unit tests are required for the `MultibandTile` version of this reproject.

spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
echeipesh: To be clear, I forgot this.  This is c/p from my prototype:🥇

PR #2496, opened 2017-11-30
Pyramid optimization: Resample on map side

lossyrob:
This change introduces an optimization to pyramiding for raster layers. This requires a new Context Bounds for the `V` of the Pyramid, which represents an additional restriction of the types - however the `V` is already a `CellGrid`, and all implemented `CellGrid`s have `ResampleMethods`, so this doesn't seem like a bad step.

With this change, the same type of interpolation that happens in the merge step after the shuffle happens on the map side. The resample method on the reduce side turns into NearestNeighbor since the grids of the incoming tile and the new time will align.

This is an API change, though only in implicit - there should not be any change necessary for the user.


echeipesh:
Looks good, used this to re-tile NLCD 8-0:
screen shot 2017-12-04 at 1 21 58 pm

Going to test for effects on shuffle on a large job next.

PR #2508, opened 2017-12-08
Added some utilities for dealing with HSV color space

jpolchlo:
We have simple tools for creating RGB(A) colors, but nothing comparable for the HSV color space.  HSV is a convenient space for building color maps for things like aspect, where the current stable of ramps doesn't make total sense.  This PR adds some simple HSV tools that allow for the production of RGB colors from HSV coordinates.

Note that the reverse translation from RGB to HSV is not provided.  If there is user demand, it can be quickly added.


pomadchin:
Hm, can I request a HSL color space support? It’s a good color space to experiment with color corrections too.

jpolchlo:
You can request it...  :þ

(That is, make an issue)

PR #2509, opened 2017-12-09
Updated benchmark for compile

lossyrob:
@metasim Seemed like reasonable updates, but as you're the author can you look over and make sure it makes sense?


metasim:
LGTM

PR #2516, opened 2017-12-13
Fix partitioning in a bit CellType case

pomadchin:
Consider `BitCellType` as bytes when deciding on appropriate number of partitions.
Its expected that this "over-estimate" is acceptable in all cases.

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/io/GeoTiffInfoReader.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/GeoTiffInfoReader.scala
@@ -57,7 +57,7 @@ private [geotrellis] trait GeoTiffInfoReader extends LazyLogging {
           info.segmentLayout.listWindows(maxSize)
       }
 
-    info.segmentLayout.partitionWindowsBySegments(windows, partitionBytes / info.cellType.bytes)
+    info.segmentLayout.partitionWindowsBySegments(windows, partitionBytes / math.max(info.cellType.bytes, 1))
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/geotrellis/spark/io/GeoTiffInfoReader.scala
```diff
@@ -57,7 +57,7 @@ private [geotrellis] trait GeoTiffInfoReader extends LazyLogging {
           info.segmentLayout.listWindows(maxSize)
       }
 
-    info.segmentLayout.partitionWindowsBySegments(windows, partitionBytes / info.cellType.bytes)
+    info.segmentLayout.partitionWindowsBySegments(windows, partitionBytes / math.max(info.cellType.bytes, 1))
```
echeipesh: Good catch.
Should we just do division based on bits or is "over allocation" of partitions in bit case good enough in practice?

pomadchin: Technically `bit` is `byte`, probably i can just change it into `math.max(info.cellType.bytes, 8)`?

Not sure about bits, mb we can experiment with it in 2.0 / 1.3?

echeipesh: `could just use `cellType.bits` and multiply the result by 8, will have to do double division to be safe though ?

PR #2518, opened 2017-12-14
GitHub PR Template

echeipesh:
Before each release we have a scramble to udpate the changelog and double check the documentation. This led to a discussion that it would be good to have a reminder in form of a PR template. 

Overall the intent is:
- Help keep ongoing work well documented and easily discoverable
- Encourage thinking about REPL usage of GeoTrellis API
- Provide more clear information for participation in PR review

Possible drawbacks are increasing the burden of accepting community contributions and longer PR times. Soliciting feedback on this general idea and specifics of the template: @metasim @lossyrob @fosskers @moradology @hectcastro @rossbernet @jpolchlo @jamesmcclain 

--------
## Overview

Brief description of what this PR does, and why it is needed.

### Checklist

### Demo

Optional. Screenshots/REPL

### Notes

Closes #XXX

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,20 @@
+## Overview
+
+Brief description of what this PR does, and why it is needed.
+
+### Checklist
+
+- [ ] `docs/CHANGELOG.rst` updated, if necessary
+- [ ] `docs` guides update, if necessary
+- [ ] New user API has useful Scaladoc strings
+- [ ] Unit tests added for bug-fix or new feature
+
+### Demo
+
+Optional. Screenshots/REPL
+
+### Notes
+
+Optional. Ancillary topics, caveats, alternative strategies that didn't work out, anything else.
+
+Closes #XXX


fosskers:
> New user API has useful Scaladoc strings ...

Potential addition:

> ... and has been assigned a `@group`

metasim:
FWIW, in the RasterFrames documentation, I'm using `sbt-tut` to evaluate the code examples, partly for the teaching benefit, but if I'm more honest, it's so I can make sure API changes get caught and fixed before release. Here's an example:

http://rasterframes.io/apps/geotrellis-ops.html

`tut` has a number of pros and cons, which I can get into if there's interest. There are many things I don't like about it when I have to get into the internals, but as long as I treat it as a black box it works pretty well. 

I'm using it within the Paradox static site generator (which I like because it's written in Scala, and is what the Akka team is using), but it's not a requirement. `tut` is agnostic to the site generation tool you use.

Another option (I've not used it) is `sbt-doctest`. It is more opinionated than `tut`, but in some way more robust than `tut` if you want to go down the "test as documentation" route. It is probably the best option if you want unified ScalaDoc and User Manual roles in one output. It looks to me like a more cleanly designed and extensible system overall, and deserves a look, but I'm not convinced that API Docs and User Manual should be combined in the same output.

metasim:
PS: I realize now that this Issue was around an actual PR template, and not about changing the way docs were kept up to date with the API. So just ignore if it's not of interests.

fosskers:
@echeipesh and I have talked about both `tut` and `sbt-doctest`. The doctests seem like a no-brainer - we should clearly be doing that. Our side-library [VectorPipe](https://geotrellis.github.io/vectorpipe/) uses `tut` in its microsite to great effect.

echeipesh:
Agreed, we should start using and leaning on `sbt-doctest`. I don't think we should include it in PR template until it becomes more common practice but I at least. I haven't really played around with it yet. Same goes for `tut` but integrating that with ReadTheDocs and `.rst` based documentation is tricky.

Either way going to apply this PR template for and change/amend it as we use it.

Issue #1789, opened 2016-11-10
MultibandTile MapAlgebra


echeipesh:
We have `MultibandTile` but all of our operations are expressed against `Tile`. This is somewhat Ok because `MultibandTile` is essentially a `Seq[Tile]`, but not really because using them in map algebra context soon leads to very tangled client code.

Before considering any method of overloads we need to support use of map algebra operations in `MultibandTile.map` and `MultibandTile.combine` methods  (ex: [`Add`](https://github.com/geotrellis/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/mapalgebra/local/Add.scala) needs to support `MultibandTile` arguments.

It appears we can get quite a lot long way to extending its super-type [`LocalTileBinaryOp`](https://github.com/geotrellis/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/mapalgebra/local/LocalTileBinaryOp.scala). 

One question to consider is what is the implication of combing two `MultibandTile` with mismatched band count. Requiring band match would be the easiest option but not the most convenient from the user perspective. 

It would be most useful to not throw in this case but upgrade missing bands to `NODATA` tiles. However, for this to be a useful operation we will need to parametrize `NODATA` semantics. `NODATA + 5` = `NODATA` or `5`.

Related: https://github.com/geotrellis/geotrellis/issues/779

echeipesh:
The overall problem is that local operations, which a defined by a function from a cell value 
to cell value, are exposed with API that is tied to the `Tile` trait.

This lack of generality will also impact tiles like GeoTiffTile, LazyTile, etc.
Basically this is a job for a type class.


## LocalTileBinaryOp

The abstract part of the method are `combine` methods.
The trait relies on `Tile.dualCombine` and `Tile.dualMap` to implement the operation.
The map case handles combining with a constant.

```scala
    trait LocalTileBinaryOp extends Serializable {
      def combine(z1: Int, z2: Int): Int
      def combine(z1: Double, z2: Double): Double
    
      def apply(r1: Tile, r2: Tile): Tile = 
        r1.dualCombine(r2)(combine)(combine)
    
      def apply(r: Tile, c: Double): Tile =
        r.dualMap({z => d2i(combine(i2d(z), c))})(combine(_, c))
    }
```

Then we notice some duplication in that the method extensions for local ops 
just delegate to an instance of `LocalTileBinaryOp.apply`

```scala
    trait AddMethods extends MethodExtensions[Tile] {
[...]

Issue #779, opened 2014-04-11
Add ability to modify how NODATA values are handled in local operations.


lossyrob:
Currently in local operations the default is that NODATA values will combine with other values such that the result is NODATA. For instance, if I were to do the addition operation on 5 rasters, and the values at a certain cell were 1, 3, 5, 4, and NODATA, then the cell value of the resulting raster will be NODATA. There are cases where a user might want the result to instead be 1+3+5+4 = 13, i.e. ignoring all NODATA values.

There should be a mechanism implemented for switching the behavior so that users can choose which way they prefer.

PR #2520, opened 2017-12-15
Map Algebra Typeclasses

fosskers:
### TODO

- [x] ADR outlining philosophy and best practices
- [x] Proposal for `Local` typeclass
- [x] Proposal for `Focal` typeclass
- [x] Implementations of the above
- [x] Benchmarks

### Motivation

This PR wounds if not slays the `Tile[T]` dragon, but maybe not in the way we all had imagined. In general I'll assert that true `Tile[T]` is not possible with Scala 2.x and Java 8, our current platform. But, we can still achieve something mathematically equivalent by shifting our core assumptions.

If:

- all current `Tile` behaviour can be broken out into typeclasses
- it's considered an acceptable tradeoff for `Local` and `Focal` ops to go from `A => A` and not `A => B` like true Functors

Then:

- `Tile` is just data with no inherant behaviour (FP-style rather than OO-style)
- `Tile` is just a wrapper around `Array`
- `Tile` doesn't exist, thus `Tile[T]` _is_ `Array[T]`
- `Local` and `Focal` instances should be written directly for `Array`

The consequences of thinking this way are:

- `while`-speed Local and Focal ops, specialized on `Int` and `Double`
- Map Algebra also available any `A` (Strings, case classes, etc)
- the elimination of manual `CellType` passing
- the elimination of `foo`/`fooDouble` operation pairs
- the elimination of all our custom method injection boilerplate

If it is not acceptable to phase out `Tile`, here are two suggestions:

- keep higher-kinded `Local[F[_]]` and `Focal[F[_]]` and their `Array` instances. Expose some function `Tile.focalSum` which just calls down to the inner Array, then uses `Focal[Array].focalSum(arr, ...)` on that. This would minimize code duplication and boilerplate.
- Provide monomorphic `MonoLocal[A]` and `MonoFocal[A]` which assume the usual `foo`/`fooDouble` pattern. This would maximize code duplication and boilerplate.

I personally recommend exploring the former.

### Related Issues

https://github.com/mpilquist/simulacrum/issues/93
https://github.com/mpilquist/simulacrum/issues/94


jpolchlo:
I am concerned that the term `Matrix` carries a very specific connotation, and users may expect that this class would provide the usual matrix-matrix/matrix-vector math operations and may even be expected to be indexed in the usual matrix (row, col) ordering.  Providing such operations may, in fact, be useful, but if they are not provided, I wonder if we will start to be hit by requests to do so?  That may not be difficult, especially if we design the internal structure to work with the Apache commons math library.  Am I overblowing this?  Maybe we consider a nomenclature that is more closely linked to `Array` than `Matrix`?

fosskers:
I chose `Matrix` because any mention of `Tile` didn't seem right, at least for the candidate operations and laws I have down on paper. `Grid` might be closer, but there's already a `Grid` symbol in GeoTrellis.

fosskers:
Either way, it's still too early to discuss things. Let me get some code down and we can reconvene on Mondayish.

fosskers:
Over the weekend I realized we had made a mistake - trying to force something structural / stateful (i.e. `Tile`) into a typeclass. If typeclasses are supposed to be about *behaviour*, then it's map algebra itself that should be abstracted over, not the "shapes" of the implementing classes.

I've arrived at typeclasses that given an implementation of how to iterate over a type, automatically provide all Local and Focal ops. They assume next to nothing about the "shapes" of potential implementing classes, not even rectangularity.

metasim:
Wow.

fosskers:
Current `Focal` times:
```
[info] Benchmark             Mode  Cnt      Score     Error  Units
[info] FocalBench.manualSum  avgt    5  16515.513 ± 169.742  us/op
[info] FocalBench.original   avgt    5   1181.609 ±  40.420  us/op
```

fosskers:
Updated `Focal` times:

```
[info] Benchmark              Mode  Cnt      Score     Error  Units
[info] FocalBench.cloneArray  avgt    5     23.458 ±   0.289  us/op
[info] FocalBench.focalClass  avgt    5  19577.618 ± 723.807  us/op
[info] FocalBench.fold        avgt    5  18202.044 ± 366.563  us/op
[info] FocalBench.foldClass   avgt    5  18134.416 ± 733.964  us/op
[info] FocalBench.manualSum   avgt    5  13349.337 ± 387.994  us/op
[info] FocalBench.original    avgt    5   1209.252 ±  68.962  us/op
```
Given the time difference between `fold` and `manualSum`, I don't think the slowdown compared to `original` is due to boxing.

metasim:
While this probably blows up the attempt to spire-ify all the things, my gut says that `Tile[CellType, StorageType]` is going to be preferable long-term (given JVM erasure over HKT). That said, perhaps `StorageType` can be handled via subclasses. IMO Tiles are *definitely* not "just" `Array`s, and some future GeoTrellis is going to require non-`Array` storage mechanisms.

metasim:
To flesh out my [comment of Dec 22](https://github.com/locationtech/geotrellis/pull/2520#issuecomment-353663104) above, I've started sketching some ideas out with type parameterized tiles. It attacks this from a very different approach than this PR, completely due to me wanting to define the end state first and work downward to a solution (or maybe motivated by my lack of adeptness with the more abstract algebra approach). I suspect a final solution would combine both approaches.

It's got a loooooooong way to go; no `NoData` support, nor cross cell-type or storage format conversions, but it does show Spire-based numerical algebra on `Int`, and `UByte` tiles in JVM arrays, and `Int` in NIO buffers (other cell types are just a boilerplate exercise):

* [package `über`](https://github.com/metasim/geotrellis/tree/%C3%BCber-tile/raster/src/main/scala/geotrellis/raster/%C3%BCber)

(Of course calling it `über` is me being silly, and not a long term name...)

Here's my test code:

* [`ÜberTileSpec.scala`](https://github.com/metasim/geotrellis/blob/%C3%BCber-tile/raster/src/test/scala/geotrellis/raster/%C3%BCber/%C3%9CberTileSpec.scala)

The TL;DR on the current state is basically Spire-backed vector ops, but done so in a way that [doesn't make assumptions](https://github.com/metasim/geotrellis/blob/%C3%BCber-tile/raster/src/main/scala/geotrellis/raster/%C3%BCber/Characteristics.scala) about how cells are addressed/accessed.

As I said, it doesn't yet support cross cell-type or storage format operations. This is the hard part: how do we allow strict tile algebra that doesn't jump storage or cell-type boundaries, while also allowing other modes where tiles can be coerced between cell representations and moved from one storage format to another. My current thought is that `Tile` could eventually become the more dynamic mode: it contains a reference to the underlying raw representation and more dynamically manages the restrictions/coercions between cell-types and storage mechanisms. The operational semantics of `Tile` would remain close to what they are now, keeping things simple for general operations, but for cases where efficiency needs to be managed carefully (e.g. inhibiting unintended conversions, array copies, controlled memory management, etc.) you can drop down into the more heavily typed implementations.

Currently my motivating example is [InternalRowTile](https://github.com/s22s/raster-frames/blob/develop/src/main/scala/org/apache/spark/sql/gt/InternalRowTile.scala) in RasterFrames. It screams of accidental complexity, and doesn't completely keep you safe from unintended memory copies.

From there I'm looking ahead to ND4J-backed tiles for GPU-enabled operations. In both cases, once you've incurred the expense of moving from JVM heap to unmanaged memory, you want eliminate accidental copies of the tile data back into the JVM.

Then comes figuring out how to abstract the higher level map algebra operations such that they could be implemented in lower level ND4J GPU registered operations. For example, having storage-specific typeclasses for implementations of things like `geotrellis.raster.mapalgebra.local.LocalTileBinaryOp` and `geotrellis.raster.mapalgebra.focal.Convolve` for which GPUs excel.

After all the above is worked out, I was going to tinker with was `NoData` handling that can be specified via an algebra (e.g. does "1 + ND == ND" or "1 + ND == 1"? Is ND more like multiplicative one or zero?). We have use cases where we need both.

cc: @echeipesh  @lossyrob

metasim:
Related, on the `NoData` front:
locationtech/geotrellis#779 locationtech/geotrellis#1789

fosskers:
RIP, we hardly knew ye.

pomadchin:
@fosskers it is not forgotten though; not in terms of a 3.0 release though

Issue #2284, opened 2017-07-16
Feature: GeoTiff Layers and Cloud Optimized GeoTiff Support


lossyrob:
This issue will explain the reasoning behind and initial architecture considerations of
a set of related features that are upcoming in the next release of
GeoTrellis (after 1.1): __GeoTiff Layers__ and __Cloud Optimized GeoTiff Support__

### Issues with current GeoTrellis Layer architecture 

These two features will address a number of issues:

#### The "many small files" issue

A large pain point for using GeoTrellis in AWS S3 is the number of small files a GeoTrellis ingest
produces. If you ingest a layer to a tile layout with 256x256 tiles covering a large area, you'll be
saving off millions of small tiles to S3. S3 is not built for this use case, and is not priced for
it either: each GET and PUT costs, and the high number of them when dealing with GeoTrellis layers
is a burden on our users.

This issue is mostly felt with S3 backend users, but could potentially effect Hadoop and file system
users, as well as any future object store backends we might support like Google Cloud Storage.
This is less of a concern with the Cassandra, Accumulo, or HBase backends.

#### Data format interopability

GeoTrellis saves tiles in an Avro serialization format. Once a GeoTrellis layer is written,
that data can only be read using the GeoTrellis library. This is unlike storing raster
data as GeoTiffs, which can be read by most raster libraries under the sun (and most importantly
GDAL-based software).

#### Duplication of Data

Users can use GeoTrellis to perform spark jobs on GeoTiffs currently, without having to save off
GeoTrellis layers. However, if they want to take full advantage of GeoTrellis, utilizing it's ability
to query space-filling curve indexed tile sets and combine raster data quickly, it is
always advised to ingest data into GeoTrellis layers stored in one of our supported backends.
However, once the data is saved as GeoTrellis layers, 
not only does this make the data readable only by GeoTrellis, as mentioned above, but it also
creates a duplicate copy of the data. For users dealing with very large raster data, this
is a burden; unless they commit to only storing their data as GeoTrellis layers, there
is a requirement to hold onto twice the amount of data than they would like to.

### Cloud Optimized GeoTiffs (COGs)

As [the page from GDAL's wiki on Cloud Optimized GeoTiffs states:](https://trac.osgeo.org/gdal/wiki/CloudOptimizedGeoTIFF)

[...]

PR #2521, opened 2017-12-17
Structured COG Layers

lossyrob:
This PR adds structured COG layers to GeoTrellis. Structured COG layers are a replacement for our SinglebandTile and MultibandTile layers for the S3, File and Hadoop backends. It uses Cloud Optimized GeoTiffs instead of Avro-serialized tiles to encode the raster data, and utilizes GeoTiff overviews to store multiple levels of pyramids into a single file. The mechanics are very different, however the API is very similar to the Avro GeoTrellis Layers, and so should be close to a drop-in replacement.

Work done by @pomadchin, this PR is part of code review/tweaks/merging into master.

Relates to #2284


lossyrob:
Superseded by #2523

PR #2523, opened 2017-12-17
Structured COG Layers

lossyrob:
This PR adds structured COG layers to GeoTrellis. Structured COG layers are a replacement for our SinglebandTile and MultibandTile layers for the S3, File and Hadoop backends. It uses Cloud Optimized GeoTiffs instead of Avro-serialized tiles to encode the raster data, and utilizes GeoTiff overviews to store multiple levels of pyramids into a single file. The mechanics are very different, however the API is very similar to the Avro GeoTrellis Layers, and so should be close to a drop-in replacement.

Work done by @pomadchin, this PR is part of code review/tweaks/merging into master.

Relates to #2284 
Supersedes #2521


s3/src/main/scala/geotrellis/spark/io/s3/AmazonS3Client.scala
```diff
@@ -20,7 +20,7 @@ import com.amazonaws.ClientConfiguration
 import com.amazonaws.auth._
 import com.amazonaws.services.s3.{AmazonS3Client => AWSAmazonS3Client}
 import com.amazonaws.services.s3.model._
-import org.apache.commons.io.IOUtils
+import sun.misc.IOUtils
```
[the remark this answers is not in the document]
metasim: Does this work with both Oracle and OpenJDK versions of the JRE?

Can't help to worry we're [taking a risk](https://blog.dripstat.com/removal-of-sun-misc-unsafe-a-disaster-in-the-making/) using this API. That said, if it provides measurable benefits and if properly documented (and what to do if it's ever removed), I'll stop worrying. ;-)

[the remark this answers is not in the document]
pomadchin: @metasim yes, [works for both](https://github.com/locationtech/geotrellis/blob/master/.travis.yml#L27-L29)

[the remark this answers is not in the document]
pomadchin: I think I'll revert this change back: [the sun.misc package is internal to Sun/Oracle and should not be used as they are not publicly supported, and not portable across all JVMs](http://www.oracle.com/technetwork/java/faq-sun-packages-142232.html).

spark/src/main/scala/geotrellis/spark/io/cog/COGRDDReader.scala
```diff
@@ -0,0 +1,76 @@
...
+
+    val tiffMethods = implicitly[TiffMethods[V]]
+    val kwFormat = KryoWrapper(implicitly[JsonFormat[K]])
+
+    val ranges = if (baseQueryKeyBounds.length > 1)
+      MergeQueue(baseQueryKeyBounds.flatMap(decomposeBounds))
+    else
+      baseQueryKeyBounds.flatMap(decomposeBounds)
+
+    val bins = IndexRanges.bin(ranges, numPartitions.getOrElse(sc.defaultParallelism))
```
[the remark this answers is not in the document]
pomadchin: @echeipesh well, it's hard to do 

```scala
val bins = IndexRanges.bin(ranges, numPartitions.getOrElse(bins.length))
```

as we don't know the real cogs amount.

s3/src/main/scala/geotrellis/spark/io/s3/AmazonS3Client.scala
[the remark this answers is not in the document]
metasim: I think erring on the side of caution is best. If there's a bit performance difference, maybe we could implement something using java.nio? If so, happy to help.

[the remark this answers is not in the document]
pomadchin: well, it's a bit faster, don't think it's too significant at the moment, but it would be good to remember that we're using a not optimal function to convert input stream into a byte array.

metasim:
The work that @echeipesh did to convert the tile layer backends to use the Java SPI framework has been hugely beneficial to us in managing transitive dependencies and deployment payload size. Our assembled jar reduced in size considerably. Have you considered doing something similar with the `geotrellis.spark.io.s3` features? If I'm using HDFS or just local FS, pulling all the S3 API dependencies seems to be heading back in the direction of dependency bloat. Thoughts?

metasim:
PS: The other benefit from the SPI  work was that we had less issues with version conflicts (e.g. Guava) to contend with.

pomadchin:
@metasim sry i misunderstood your question a bit, why do you need to pull down s3 to use file and hdfs backends only in this case?

pomadchin:
But to add: that's a good point to add `COGLayerProviders` to support Java SPI, don't think it's smth super important at the moment, though can be implemented easily (right now i have just incorrect links on common (old) layer providers, which would produce an incorrect design).

metasim:
@pomadchin 

> @metasim sry i misunderstood your question a bit, why do you need to pull down s3 to use file and hdfs backends only in this case?

Sorry, I wasn't clear. I meant that if I *didn't* want S3, I don't want GT to pull in the S3 libraries when I don't need them.

That said, I may be inferring too much from my cursory look at the implementation.

pomadchin:
@metasim ah, cogs are not in a separate sbt project, it's smth very core => all s3 related stuff is only in the s3 project. If you'd decide to use hdfs i don't see any reasons why depending only on a spark project may lead to the s3 deps download.

spark/src/main/scala/geotrellis/spark/io/hadoop/geotiff/GeoTiffLayerReader.scala
```diff
@@ -0,0 +1,82 @@
...
+import geotrellis.spark.tiling.ZoomedLayoutScheme
+import geotrellis.spark.{LayerId, SpatialKey}
+import geotrellis.vector.{Extent, ProjectedExtent}
+import geotrellis.raster.crop.Crop
+import geotrellis.raster.reproject.Reproject.{Options => ReprojectOptions}
+
+import java.net.URI
+
+// global context only for test purposes, should be refactored
+import scala.concurrent.ExecutionContext.Implicits.global
```
echeipesh: Generally not safe, to rely on global thread pool for IO.
Ideally we could just capture a `Monoid` that would either be a `Future[V]`, an `Option[V]` or `Future[Option[V]]` but I'm not certain how that would discover the thread pool though.

pomadchin: Let's make it an implicit context, I agree. 
But in the future that should be an own blocking threadpool

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffSegmentLayout.scala
```diff
@@ -199,6 +199,56 @@ case class GeoTiffSegmentLayout(totalCols: Int, totalRows: Int, tileLayout: Tile
...
+      if(layoutRow == tileLayout.layoutRows - 1) {
+        totalRows - ((tileLayout.layoutRows - 1) * tileLayout.tileRows)
+      } else {
+        tileLayout.tileRows
+      }
+
+    (cols, rows)
+  }
+
+  private [geotrellis] def getGridBounds(segmentIndex: Int, isBit: Boolean = false): GridBounds = {
```
echeipesh: `getSegmentGridBounds`

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffSegmentLayout.scala
```diff
@@ -199,6 +199,56 @@ case class GeoTiffSegmentLayout(totalCols: Int, totalRows: Int, tileLayout: Tile
     }
     result.result
   }
+
+  private def bandSegmentCount: Int =
```
echeipesh: Can be public

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffSegmentLayout.scala
echeipesh: `isBit` is not used?

s3/src/main/scala/geotrellis/spark/io/s3/cog/package.scala
```diff
@@ -0,0 +1,29 @@
...
+
+import geotrellis.spark.io.s3.util.S3RangeReader
+import geotrellis.util.{ByteReader, StreamingByteReader}
+
+import com.amazonaws.services.s3.AmazonS3URI
+
+import java.net.URI
+
+package object cog {
+  def s3PathExists(str: String, s3Client: S3Client): Boolean = {
```
echeipesh: Why not make it hang directly on `S3Client` ?

s3/src/main/scala/geotrellis/spark/io/s3/cog/package.scala
```diff
@@ -0,0 +1,29 @@
...
+    val arr = str.split("/").filterNot(_.isEmpty)
+    val bucket = arr.head
+    val prefix = arr.tail.mkString("/")
+    s3Client.doesObjectExist(bucket, prefix)
+  }
+
+  def byteReader(uri: URI, s3Client: S3Client): ByteReader = {
```
echeipesh: This can live on `S3Client` too

spark/src/main/scala/geotrellis/spark/io/cog/TileLayerMetadataMapAccumulator.scala
```diff
@@ -0,0 +1,52 @@
+package geotrellis.spark.io.cog
+
+import geotrellis.spark.{Boundable, TileLayerMetadata}
+import org.apache.spark.util.AccumulatorV2
+
+import scala.collection.JavaConverters._
+import java.util
+import java.util.function.BiFunction
+
+/** Probably should be removed as no needs in it now. */
```
lossyrob: ...should we remove?

spark/src/main/scala/geotrellis/spark/io/hadoop/geotiff/GeoTiffAttributeStore.scala
```diff
@@ -0,0 +1,26 @@
+package geotrellis.spark.io.hadoop.geotiff
+
+import geotrellis.vector.ProjectedExtent
+
+trait CollectionAttributeStore[T] extends AttributeStore[Seq, T]
+trait IteratorAttributeStore[T] extends AttributeStore[Iterator, T]
+
+/** Layer that works with Metadata Index ?? */
+trait AttributeStore[M[_], T] {
```
lossyrob: Class name doesn't match file name.

spark/src/main/scala/geotrellis/spark/io/hadoop/geotiff/GeoTiffLayerReader.scala
```diff
@@ -0,0 +1,118 @@
...
+import scalaz.concurrent.{Strategy, Task}
+import scalaz.stream.{Process, nondeterminism}
+
+import java.net.URI
+import java.util.concurrent.{ExecutorService, Executors}
+
+import scala.reflect.ClassTag
+
+/** Approach with TiffTags stored in a DB */
+trait GeoTiffLayerReader[M[T] <: Traversable[T]] {
```
lossyrob: Confused between the `geotiff` and `cog` packages. Is the `geotiff` stuff the unstructured layers?

spark/src/main/scala/geotrellis/spark/io/index/HilbertKeyIndexMethod.scala
```diff
@@ -20,6 +20,7 @@ import geotrellis.spark._
 import geotrellis.spark.io.index.hilbert._
 
 import java.time.ZonedDateTime
+import java.lang.Long.numberOfLeadingZeros
```
lossyrob: Is this used?

spark/src/main/scala/geotrellis/spark/io/index/hilbert/HilbertSpatialKeyIndex.scala
```diff
@@ -69,8 +70,10 @@ class HilbertSpatialKeyIndex(val keyBounds: KeyBounds[SpatialKey], val xResoluti
         BitVectorFactories.OPTIMAL.apply(yResolution)
       )
 
-    bitVectors(0).copyFrom(key.col.toLong)
-    bitVectors(1).copyFrom(key.row.toLong)
+    val col = key.col - minKey.col
```
lossyrob: This would break current Hilbert layers! Why is this change in here?

pomadchin: yeah, but it's a bug, otherwise this test would fail with a bit overflow error, in a `.copyFrom` function call: 

```
it("Generates a Correct index given a 4x4 square on a 10th zoom level") {
  val keyBounds = KeyBounds(SpatialKey(177, 409), SpatialKey(178, 410))
  val index = HilbertKeyIndexMethod.createIndex(keyBounds)
  index.toIndex(SpatialKey(178, 410)) should be (2)
}
```

pomadchin: In addition, it's how a curent temporal `HilbertKeyIndex` works: 

https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/io/index/hilbert/HilbertSpaceTimeKeyIndex.scala#L88-L89

s3/src/main/scala/geotrellis/spark/io/s3/geotiff/S3GeoTiffInput.scala
```diff
@@ -0,0 +1,56 @@
...
+    pattern: String,
+    recursive: Boolean = true,
+    getS3Client: () => S3Client = () => S3Client.DEFAULT): List[GeoTiffMetadata] = {
+    val s3Client = getS3Client()
+    val s3Uri = new AmazonS3URI(uri)
+    val regexp = pattern.r
+
+    val objectRequest = (new ListObjectsRequest)
+      .withBucketName(s3Uri.getBucket)
+      .withPrefix(s3Uri.getKey)
```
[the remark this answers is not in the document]
pomadchin: Yes, `s3Uri.getKey` in this case returns `null` and it works fine for the request builder.

spark-testkit/src/main/scala/geotrellis/spark/testkit/io/cog/COGPersistenceSpec.scala
```diff
@@ -0,0 +1,195 @@
...
+  type TestCollectionReader = COGCollectionLayerReader[LayerId]
+
+  def sample: RDD[(K, V)] with Metadata[TileLayerMetadata[K]]
+  def reader: TestReader
+  def writer: TestWriter
+  // def deleter: TestDeleter
+  // def copier: TestCopier
+  // def mover: TestMover
+  // def reindexer: TestReindexer
+  // def updater: TestUpdater
```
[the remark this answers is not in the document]
pomadchin: `TODO`, will mark them as `TODO`, probably there would be `updater` in a new `API`, just `writer`, what do you think?

spark/src/main/scala/geotrellis/spark/io/hadoop/geotiff/GeoTiffLayerReader.scala
pomadchin: Yes, `geotiff` means unstructured. Do you think it would be better to put it into `cog` package and to rename everything or to rename this `geotiff` package into `ucog` or smth like that?

spark/src/main/scala/geotrellis/spark/io/index/hilbert/HilbertSpatialKeyIndex.scala
pomadchin: There is also a workaround (not a real fix but at least to make it not failing): 

```scala
     def apply(temporalResolution: Int): KeyIndexMethod[SpaceTimeKey] =
        new KeyIndexMethod[SpaceTimeKey] {
          def createIndex(keyBounds: KeyBounds[SpaceTimeKey]) = {
            val xResolution = math.max(resolution(keyBounds.maxKey.col, keyBounds.minKey.col), 64 - numberOfLeadingZeros(keyBounds.maxKey.col.toLong))
            val yResolution = math.max(resolution(keyBounds.maxKey.row, keyBounds.minKey.row), 64 - numberOfLeadingZeros(keyBounds.maxKey.row.toLong))

             HilbertSpaceTimeKeyIndex(keyBounds, xResolution, yResolution, temporalResolution)
           }
        }
```

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffTile.scala
```diff
@@ -290,6 +290,60 @@ object GeoTiffTile {
...
+    * for some tile coordinate of the TileLayout, a NoData segment will be constructed.
+    *
+    * @param tiles: The TileLayout coordinates mapped to the Tiles that will make up the segments.
+    * @param tileLayout: The TileLayout that this GeoTiffTile should use for its segments
+    * @param options: GeoTiffOptions for this GeoTiffTile
+    *
+    * @return A GeoTiffTile
+    */
+  def apply(tiles: Map[(Int, Int), Tile], tileLayout: TileLayout, options: GeoTiffOptions): GeoTiffTile = {
+    // TODO: Test
```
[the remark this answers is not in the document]
pomadchin: @echeipesh was the idea of `GeoTiffBuilder[_].makeTile` to remove this function from here?

pomadchin:
Before it would be merged: Implement `persist` for all inmemory attribute storage (it's about unstructured COGs). Currently implemented only for S3. Thx to @notthatbreezy for report.

spark/src/main/scala/geotrellis/spark/io/index/hilbert/HilbertSpatialKeyIndex.scala
lossyrob: I remember this, but should this be a separate PR with it's own discussion on fallout on current layers, and how to mitigate any potential damage to current layers?

spark-testkit/src/main/scala/geotrellis/spark/testkit/io/cog/COGPersistenceSpec.scala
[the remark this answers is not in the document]
lossyrob: Should we just move the updater tests to the writer? If so that's a TODO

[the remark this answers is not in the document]
pomadchin: @lossyrob yea, it's done. I'll remove the confusing line than.

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffTile.scala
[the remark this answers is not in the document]
echeipesh: Yup, that was the idea. This is no longer used so removing it.

PR #2527, opened 2017-12-19
Optimize TileRDDReproject using mutable region reprojection

jpolchlo:
This PR attempts to improve TileRDDReproject's performance by eliminating the merge step.  This is accomplished by using a mutable version of RasterRegionReproject to continually burn reprojected tiles into the same buffers.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/reproject/RasterRegionReproject.scala
+++ b/raster/src/main/scala/geotrellis/raster/reproject/RasterRegionReproject.scala
@@ -27,7 +27,7 @@ trait RasterRegionReproject[T <: CellGrid] extends Serializable {
 }
 
 object RasterRegionReproject {
-  def rowCoords(destRegion: Polygon, destRasterExtent: RasterExtent, toSrcCrs: Transform): Int => (Array[Int], Array[Double], Array[Double]) = {
+  private def rowCoords(destRegion: Polygon, destRasterExtent: RasterExtent, toSrcCrs: Transform): Int => (Array[Int], Array[Double], Array[Double]) = {
     val extent = destRasterExtent.extent
     val rowTransform = RowTransform.approximate(toSrcCrs, 0.125)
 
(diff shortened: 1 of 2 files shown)


raster/src/main/scala/geotrellis/raster/reproject/RasterRegionReproject.scala
```diff
@@ -21,40 +21,162 @@ trait RasterRegionReproject[T <: CellGrid] extends Serializable {
     * @param resampleMethod cell value resample method
     */
   def regionReproject(raster: Raster[T], src: CRS, dest: CRS, rasterExtent: RasterExtent, region: Polygon, resampleMethod: ResampleMethod): Raster[T]
+
+  def mutableRegionReproject(target: T, raster: Raster[T], src: CRS, dest: CRS, rasterExtent: RasterExtent, region: Polygon, resampleMethod: ResampleMethod): Unit
+
 }
 
 object RasterRegionReproject {
+  def rowCoords(destRegion: Polygon, destRasterExtent: RasterExtent, toSrcCrs: Transform): Int => (Array[Int], Array[Double], Array[Double]) = {
```
echeipesh: This should be made `private` so it can be tweaked if desired arrises

spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
```diff
@@ -208,7 +202,21 @@ object TileRDDReproject {
     // val expectedCols = targetDataExtent.width / newMetadata.layout.cellSize.width
     // println(s"Expected width: $expectedCols")
```
echeipesh: debug comments, not part of this PR but can go now.

raster/src/main/scala/geotrellis/raster/reproject/RasterRegionReproject.scala
```diff
@@ -21,40 +21,162 @@ trait RasterRegionReproject[T <: CellGrid] extends Serializable {
...
+          cfor(0)(_ < xs.size, _ + 1){ s =>
+            buffer.set(pxs(s), i, resampler.resample(xs(s), ys(s)))
+          }
+        }
+      }
+    }
+  }
+
+  implicit val multibandInstance = new RasterRegionReproject[MultibandTile] {
+    def regionReproject(raster: Raster[MultibandTile], src: CRS, dest: CRS, rasterExtent: RasterExtent, region: Polygon, resampleMethod: ResampleMethod): Raster[MultibandTile] = {
```
echeipesh: The code duplication between `regionReproject` and `mutableRegionReproejct` should be taken care of. Probably the easiest thing is to delegate from the former to the latter once tiles are allocated.

echeipesh: Any thoughts on naming convention for `mutable` methods now that its used?

jpolchlo:
For the record, test runs on EMR yielded the following timings on a 20 executor cluster.  For the baseline case:

Baseline case

And for the mutable-optimized case:

Optimized case

Slight, but not insignificant performance improvement.

raster/src/main/scala/geotrellis/raster/reproject/RasterRegionReproject.scala
jpolchlo: I don't have significant thoughts on naming convention, sad to say.  This works well enough for me.  Do you have specific concerns about the nomenclature used here?

echeipesh:
Weird that the partition count for base pyramid job doesn't match between two ingests.

PR #2531, opened 2017-12-22
Make GeoKeyDirectory parsing more fault tolerant

notthatbreezy:
This commit makes the reading of GeoKeyDirectory entries more fault tolerant to
cover cases where the `NumberOfKeys` entry in the directory header has an
incorrect number of headers that is actually larger than the number of keys.
This manifested itself when attempting to read a tiff created by a geoprocessing
service from an ArcGIS server. The tiff could be read and parsed by other
tools (GDAL, QGIS) without issue but when attempting to open it with GeoTrellis
a `scala.MatchError` would be thrown in `readGeoKeyEntry`.

The workaround introduced here is that if a geokey is read and fails to match
then it likely means there are no longer any valid geokeys to read. I'm open to
other approaches to the workaround than the one offered here - one option is to
introduce a try/catch and only catch on `java.nio.Exception` which can happen if
the number of keys is large enough that reading too many keys (and ignoring
them) leads to a buffer underflow.

I also added a test case that tests the more fault tolerant parsing of the GeoKeyDirectory header.

Signed-off-by: Chris Brown <cbrown@azavea.com>

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/io/geotiff/tags/GeoKeyReader.scala
+++ b/raster/src/main/scala/geotrellis/raster/io/geotiff/tags/GeoKeyReader.scala
@@ -33,6 +33,10 @@ object GeoKeyReader {
     geoKeyDirectory: GeoKeyDirectory, index: Int = 0
   ): GeoKeyDirectory = {
 
+    /**
+      * Attempt to read GeoKey Entry; if tag does not match then assume there are no
+      * more valid geokeys and the header was written incorrectly so return `None`
+      */
     def readGeoKeyEntry(keyMetadata: GeoKeyMetadata,
       geoKeyDirectory: GeoKeyDirectory): Option[GeoKeyDirectory] = keyMetadata.tiffTagLocation match {
       case 0 => Some(readShort(keyMetadata, geoKeyDirectory))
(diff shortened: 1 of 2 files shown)


notthatbreezy:
I'm not sure how to fix the [test errors](https://travis-ci.org/locationtech/geotrellis/jobs/320359746#L7544) -- they happened locally as well and the only solution was to manually copy the test zip file rather than rely on git.

raster/src/main/scala/geotrellis/raster/io/geotiff/tags/GeoKeyReader.scala
```diff
@@ -34,10 +34,11 @@ object GeoKeyReader {
...
     def readGeoKeyEntry(keyMetadata: GeoKeyMetadata,
-      geoKeyDirectory: GeoKeyDirectory)  = keyMetadata.tiffTagLocation match {
-      case 0 => readShort(keyMetadata, geoKeyDirectory)
-      case DoublesTag => readDoubles(keyMetadata, geoKeyDirectory)
-      case AsciisTag => readAsciis(keyMetadata, geoKeyDirectory)
+      geoKeyDirectory: GeoKeyDirectory): Option[GeoKeyDirectory] = keyMetadata.tiffTagLocation match {
+      case 0 => Some(readShort(keyMetadata, geoKeyDirectory))
+      case DoublesTag => Some(readDoubles(keyMetadata, geoKeyDirectory))
+      case AsciisTag => Some(readAsciis(keyMetadata, geoKeyDirectory))
+      case _ => None
```
lossyrob: Can you put a comment on the `readGeoKeyEntry` method that basically explains the same reasoning as in the PR description? This will save some future version of ourselves time otherwise spent on thinking through why this would be returning an option.

notthatbreezy: Good point, addressed in https://github.com/locationtech/geotrellis/pull/2531/commits/788a03464d05420ebf6f97da9b5b170d223e2072

PR #2533, opened 2018-01-03
Avoid `ListBuffer.prepend` and `ListBuffer.append`

fosskers:
## Overview

I discovered today that `.prepend` and `.append`, when used repeatedly (in a loop, say), are an order of magnitude slower than their operators, `+=:` and `+=` respectively:

```
[info] ListBufferBench.append10000     avgt    5  664.377 ± 10.276  us/op
[info] ListBufferBench.appendOp10000   avgt    5   58.740 ±  1.557  us/op
[info] ListBufferBench.prepend10000    avgt    5  848.419 ± 20.689  us/op
[info] ListBufferBench.prependOp10000  avgt    5   58.248 ±  1.813  us/op
```

Rule of thumb: Don't trust methods that take `a: A*` when you intend to only give it one thing.

This PR removes usage of these two methods for `ListBuffer` and `ArrayBuffer`.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/viewshed/R2Viewshed.scala
+++ b/raster/src/main/scala/geotrellis/raster/viewshed/R2Viewshed.scala
@@ -347,7 +347,7 @@ object R2Viewshed extends Serializable {
       val seg = northSegs(i)
       alpha = thetaToAlpha(from, rays, seg.theta); terminated = false
       Rasterizer.foreachCellInGridLine(seg.x0, seg.y0, seg.x1, seg.y1, null, re, false)(callback)
-      if (!terminated && !seg.isRumpSegment) southRays.append(Ray(seg.theta, alpha))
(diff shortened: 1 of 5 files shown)


raster/src/main/scala/geotrellis/raster/viewshed/R2Viewshed.scala
```diff
@@ -347,7 +347,7 @@ object R2Viewshed extends Serializable {
       val seg = northSegs(i)
       alpha = thetaToAlpha(from, rays, seg.theta); terminated = false
       Rasterizer.foreachCellInGridLine(seg.x0, seg.y0, seg.x1, seg.y1, null, re, false)(callback)
-      if (!terminated && !seg.isRumpSegment) southRays.append(Ray(seg.theta, alpha))
+      if (!terminated && !seg.isRumpSegment) (southRays += Ray(seg.theta, alpha))
       i += 1
```
jamesmcclain: Good catch, thanks.

PR #2538, opened 2018-01-09
Allow new geometries to snap to given PrecisionModel

### LINKED ISSUE #2525: java.lang.ArrayIndexOutOfBoundsException when trying to get value from Tile
If you try and run the above code you will see that it returns `java.lang.ArrayIndexOutOfBoundsException`

```scala
import geotrellis.vector._
import geotrellis.raster._
import geotrellis.proj4.WebMercator
import geotrellis.spark._
import geotrellis.raster.rasterize._
import org.apache.spark.sql.SparkSession
import geotrellis.raster.rasterize.Rasterizer

object TestRasterazerFaliure extends App {

  val _spark = SparkSession
    .builder()
    .master("local[*]")
    .appName("Rasterazer")
    .getOrCreate()

  import _spark.implicits._

  val conf = _spark.sparkContext.getConf
  implicit val sc = _spark.sparkContext

  implicit val _sql = _spark.sqlContext



  val p = Point(17532819.799940586, 5425368.153816739)
  val tileSize = 512
  val zoom = 4

  val layout = ZoomedLayoutScheme(WebMercator, tileSize).levelForZoom(zoom).layout

  val pointRDD = sc.parallelize(Array(p))

  val result = pointRDD.rasterize(1, IntConstantNoDataCellType, layout)

  result.collect()


  println("the end")

}
```

Even thought the collect is succeeded, the x value of the SpatialKey is -1 here 
https://github.com/locationtech/geotrellis/blob/04a60d0ee8734602aaa2635046107a474d5109cd/spark/src/main/scala/geotrellis/spark/rasterize/RasterizeRDD.scala#L71

jpolchlo:
## Overview

When working with geometries, numerical error can arise.  The easiest remedy is to use a fixed-precision model for representing geometries.  This will snap vertices to a grid with a user-defined spacing.  Or at least, that was the idea.  This PR actually makes that happen.  Previously, it was assumed that JTS would perform the snapping of vertices when the fixed-point model was used, but that was not the case.  After this PR, we will explicitly apply the precision model to new geometry; but it is also possible to circumvent the snapping process by explicitly creating JTS geometries and wrapping them in GT vector objects.

### Checklist

- [X] `docs/CHANGELOG.rst` updated, if necessary
- [X] New user API has useful Scaladoc strings

Closes #2525

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -16,6 +16,14 @@ API Changes
   - **Change:** Reprojection has improved performance due to one less shuffle stage and lower memory usage.
   ``TileRDDReproject`` loses dependency on ``TileReprojectMethods`` in favor of ``RasterRegionReproject``
 
+- ``geotrellis.vector``
+
+  - **Fix:** Allow a ``geotrellis.jts.precision.type = floating`` in ``application.conf`` to actually snap vertices of
+    GT geometries to the desired grid.
+
+  - **Change:** Exposed ``GeomFactory.factory`` to allow direct creation of JTS geometries, circumventing the snapping
+    of vertices to a fixed precision model.
+
 1.2.1
 _____
 *2018 Jan 3*
--- a/docs/guide/vectors.rst
+++ b/docs/guide/vectors.rst
@@ -5,33 +5,46 @@ Numerical Precision and Topology Exceptions
 ===========================================
 
 When doing any significant amount of geometric computation, there is the
-potential of encountering JTS TopologyExceptions or other numerical hiccups
-that arise, essentially, from asking too much of floating point numerical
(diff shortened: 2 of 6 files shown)


jpolchlo:
Not 100% sure of the best way to test this.  Willing to take suggestions.

jpolchlo:
After having time to consider this problem, this is a bad solution.  Topology changes can occur by simply snapping points.  The right approach is to use snap rounding from JTS (an implementation exists and an improved implementation is in progress, to be released at some point in the future).  We should offer a method to allow users to bring their geometries in at any precision level, and then use the snap rounder to do the work of bringing geometries into agreement with the chosen precision model.  At that time, we can determine a useful interface for the simplification process.

Automatically forcing coordinates onto a grid is a bad idea, so I'm closing this.

PR #2553, opened 2018-01-17
TileRDDReproject Will Now Preserve the Source RDD's Partitioner

jbouffard:
## Overview

This PR makes it so that the resulting `RDD` of a `reproject` will have the same `Partitioner` as the source `RDD`. This is needed as it can help reduce the number of shuffle steps required, and in turn, improve the performance of a job.

### Checklist

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
+++ b/spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
@@ -72,6 +72,8 @@ object TileRDDReproject {
     val passthroughGridExtent = ReprojectRasterExtent(sourceDataGridExtent, metadata.crs, destCrs)
     val targetDataExtent = passthroughGridExtent.extent
 
+    val sourcePartitioner = bufferedTiles.partitioner
+
     // inspect the change in spatial extent to get the pixel counts
     val reprojectSummary = matchReprojectRasterExtent(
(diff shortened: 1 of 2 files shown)


lossyrob:
IMO this should trigger a 1.2.x bugfix release

echeipesh:
Discovered that this is not an issue in 1.2 while trying to back-port this fix: https://github.com/locationtech/geotrellis/blob/1.2/spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala#L167-L174

However, also found that the partitioner scaling that was removed in this PR was not actually used in 1.2 (notice that `part` goes unused). 

This is an unfortunate situation. I think the best thing to do is to [remove the logic](https://github.com/locationtech/geotrellis/commit/4b71d12f834c30205fea73a422e22c1b92b055cb) that creates the scaled partitioner and produces the log message. The alternative, of actually using the `HashPartitioner` which may be radically different from the source partitioner is too surprising a change for bug-fix release.

The auto-scaling behavior was intended to address cases where the total pixel count (and resolution) of the reprojected raster is significantly larger from source raster. This was result of internal testing and not user report.

I think the best thing to do is leave the 1.2 branch with current functionality and be ready to create 1.3 release if a user encounters a scenario that can be fixed by this auto-scaling. Likely the release of 2.0 (which has different and much improved reprojection logic) will render this issue moot.

PR #2565, opened 2018-02-08
Update LazySegmentBytes to only read required segments

kthompson:
Previously LazySegmentBytes would read all segments between the first and
last segment causing much more data to be downloaded when the
StreamingByteReader is used.

Signed-off-by: Kevin Thompson <kevin.thompson@theautomaters.com>

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/io/geotiff/LazySegmentBytes.scala
+++ b/raster/src/main/scala/geotrellis/raster/io/geotiff/LazySegmentBytes.scala
@@ -92,19 +92,14 @@ class LazySegmentBytes(
     }
   }._2.reverse // get segments back in offset order
 
-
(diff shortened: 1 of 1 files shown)


kthompson:
There may be a good reason this was reading the way it was so let me know if there is some rework you would like to see here.

echeipesh:
The assumption was that that you're likely going to be reading consecutive segments and it would be good to read as many of them at once as possible to avoid additional overhead. Although I don't recall us having done rigorous testing to measure this overhead.

I'm guessing you ran into cases where that wasn't true or helpful?

kthompson:
Yes for streaming geotiffs from S3. Using the existing method typically retrieves way more data than you actually need. From my testing it was around 10x the transfer of using something like vsicurl.

echeipesh:
Thank you for clarification.

PR #2569, opened 2018-02-16
ZoomResmaple for MultibandTileLayerRDDs

jbouffard:
## Overview

This PR makes it so that both `TileLayerRDD`s and `MultibandTileLayerRDD`s can use the `ZoomResample` method.

### Checklist

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/resample/Implicits.scala
+++ b/spark/src/main/scala/geotrellis/spark/resample/Implicits.scala
@@ -16,10 +16,20 @@
 
 package geotrellis.spark.resample
 
+import geotrellis.raster._
+import geotrellis.raster.resample._
 import geotrellis.spark._
+import geotrellis.spark.tiling._
+import geotrellis.util._
+import geotrellis.vector._
+
+import org.apache.spark.rdd._
 
 object Implicits extends Implicits
 
 trait Implicits {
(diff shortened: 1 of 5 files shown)


spark/src/main/scala/geotrellis/spark/resample/Implicits.scala
```diff
@@ -21,5 +21,6 @@ import geotrellis.spark._
 object Implicits extends Implicits
 
 trait Implicits {
-  implicit class withZoomResampleMethods[K: SpatialComponent](self: TileLayerRDD[K]) extends ZoomResampleMethods[K](self)
+  implicit class withTileLayerRDDZoomResampleMethods[K: SpatialComponent](self: TileLayerRDD[K]) extends TileLayerRDDZoomResampleMethods[K](self)
+  implicit class withMultibandTileLayerRDDZoomResampleMethods[K: SpatialComponent](self: MultibandTileLayerRDD[K]) extends MultibandTileLayerRDDZoomResampleMethods[K](self)
```
echeipesh: You shouldn't need to add the specific method extensions of Multiband.
You should be able to add `V` type parameter constrained by `(implicit ev: V => TileResampleMethods[V])` to the `withZoomResampleMethods`/`ZoomResampleMethods` which should allow it to call `ZoomResample` directly.

PR #2570, opened 2018-02-19
Update plugins versions

pomadchin:
## Overview

This PR updates sbt and other plugins. The critical thing here is sbt-scapegoat plugin update as in the current master it doesn't work.


metasim:
@pomadchin Did you test the bench plugin to make sure it works with the new version of JMH?

pomadchin:
@metasim nope, that's why it's not merged yet

pomadchin:
@metasim it looks like it works fine with a new JMH, can you throw here some parameters to check it better?

pomadchin:
This PR requires no CQs as adds no extra deps into geotrellis artifacts.

PR #2573, opened 2018-02-23
Option[Partitioner] Parameter for TileRDDReproject

### LINKED ISSUE #2554: TileRDDReproject should accept Option[Partitioner]
Connects: https://github.com/locationtech/geotrellis/issues/2464

This partitioner should be applied to the result of the reproject, allowing to either increase the partition count or change partitioning method when `RDD.combineByKey` is triggered.

The default behavior should remain the same, keep the partitioner of the input tile RDD if it exists.

jbouffard:
## Overview

This PR adds an `Option[Partitioner]` parameter to `TileRDDReproject`. This will allow users to specify a `Partitioner` for the resulting `RDD` without having to do a `partitionBy`step afterwards . If no `Partitioner` is specified, then the resulting `RDD` will have the same `Partitioner` as its source.

### Checklist

### Demo

```scala

val partitioner = new HashPartitioner(16)

rdd.reproject(ZoomedLayoutScheme(LatLng), Some(partitioner))
```

Closes #2554

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
+++ b/spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
@@ -55,13 +55,14 @@ object TileRDDReproject {
     */
   def apply[
     K: SpatialComponent: Boundable: ClassTag,
-    V <: CellGrid: ClassTag: RasterRegionReproject: Stitcher: (? => TileMergeMethods[V]): (? => TilePrototypeMethods[V])
+    V <: CellGrid: ClassTag: RasterRegionReproject: (? => TileMergeMethods[V]): (? => TilePrototypeMethods[V])
   ](
     bufferedTiles: RDD[(K, BufferedTile[V])],
     metadata: TileLayerMetadata[K],
     destCrs: CRS,
     targetLayout: Either[LayoutScheme, LayoutDefinition],
-    options: Options
+    options: Options,
+    partitioner: Option[Partitioner]
   ): (Int, RDD[(K, V)] with Metadata[TileLayerMetadata[K]]) = {
     val crs: CRS = metadata.crs
     val layout = metadata.layout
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
```diff
@@ -72,7 +73,12 @@ object TileRDDReproject {
     val passthroughGridExtent = ReprojectRasterExtent(sourceDataGridExtent, metadata.crs, destCrs)
     val targetDataExtent = passthroughGridExtent.extent
 
-    val sourcePartitioner = bufferedTiles.partitioner
+    val targetPartitioner: Option[Partitioner] =
```
echeipesh: `Option` has handy [`orElse`](http://www.scala-lang.org/api/2.11.8/index.html#scala.Option@orElse[B>:A](alternative:=>Option[B]):Option[B]) that enables `partitioner.orElse(bufferedTiles.partitioner)`

spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
```diff
@@ -61,7 +61,8 @@ object TileRDDReproject {
     metadata: TileLayerMetadata[K],
     destCrs: CRS,
     targetLayout: Either[LayoutScheme, LayoutDefinition],
-    options: Options
+    options: Options,
+    partitioner: Option[Partitioner]
```
echeipesh: Unrelated to this PR but while reviewing this noticed that `Stitcher` is no longer required constraint here. Might as well remove it in this PR if you don't mind.

jbouffard: :+1:

spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
jbouffard: Didn't know about that method, but it looks pretty handy.

spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
jbouffard: It looks like `Stitcher` is still required due to `BufferTiles` https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/buffer/BufferTiles.scala#L533-L540

echeipesh: Right, but its not needed for the `apply` that already takes RDD of buffered tiles.

PR #2564, opened 2018-02-08
Fix issue where data could be loaded prematurely

kthompson:
When the StreamingByteReader calls `position(x)` it would previously set the buffer position on the chunk which would force loading the data for that chunk.

This change tracks the ByteReader position separately to prevent loading data prematurely.

### Checklist

- [ ] `docs/CHANGELOG.rst` updated, if necessary
- [x] Unit tests added for bug-fix or new feature

PR #2578, opened 2018-02-27
Optimize and Fix StreamingByteReader

echeipesh:
## Overview

This PR fixes the missing logic that is required when tracking position independently from buffer position. 

It also adds some logic to:
 - re-use existing buffer when incrementing to next read to prevent re-reading the same bytes
 - absorb reads that can be satisfied with current buffer
 - allow expanding the buffer on both sides and re-using existing bytes 

### Checklist

~- [ ] `docs/CHANGELOG.rst` updated, if necessary~
~- [ ] Unit tests added for bug-fix or new feature~

Connects: https://github.com/locationtech/geotrellis/pull/2564

### DIFF
--- a/util/src/main/scala/geotrellis/util/StreamingByteReader.scala
+++ b/util/src/main/scala/geotrellis/util/StreamingByteReader.scala
@@ -16,6 +16,9 @@
 
 package geotrellis.util
 
+import spire.math._
+import spire.math.interval.{Overlap, Closed, Open}
+import spire.implicits._
 import java.nio.{ByteOrder, ByteBuffer}
 
 /**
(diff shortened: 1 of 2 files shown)


util/src/main/scala/geotrellis/util/StreamingByteReader.scala
```diff
@@ -155,4 +189,40 @@ object StreamingByteReader {
 
   def apply(rangeReader: RangeReader, chunkSize: Int): StreamingByteReader =
     new StreamingByteReader(rangeReader, chunkSize)
+
+
+  private[util]
+  def intervalBounds(i: Interval[Long]): (Long, Long) = (
+    i.lowerBound match {
+      case Closed(a) => a
+      case Open(a) => a + 1
```
jbouffard: This is more of a genera question than an issue, but when would the `Interval`s have `Open` bounds? In this situation, it seems like they should always be closed.

jbouffard:
Looks good to me.

util/src/main/scala/geotrellis/util/StreamingByteReader.scala
echeipesh: Thats a good question. They can apparently be a result of the difference `\`, taking one interval away from the other other:

```scala
scala> import spire.math._
import spire.math._

scala> import spire.implicits._
import spire.implicits._

scala> Interval(1,10) \ Interval(7,25)
res0: List[spire.math.Interval[Int]] = List([1, 7))
```
Because `7` is on the right, it can't be on the left, so the interval includes things up to `7` but not including `7`. 

This doesn't quite square with the type parameter of `Long`, you kind of know that if its not `7` it must be `6`. I guess this could be useful if you were using integers to model real intervals or more likely encoding the distinction at the type level didn't seem like it was worth while to the library author.

echeipesh:
In the best spirt of thinking after writing it occurs to me now that the real useful implementation here would be to have a block cache, where the underlying file is always read by blocks of fixed size and the results are pieces together from available blocks. 

Making that cache of limited size with LRU would enable keeping most accessed parts of the file (likely the header?) in memory even if the access pattern is alternating between reading from different parts of the file. At this point I'd file this idea under "future improvements" until benchmarking shows that it would be a real benefit.

PR #2582, opened 2018-03-06
Query Results Will Now Have the Correct KeyBounds and Extent

### LINKED ISSUE #2543: Layer query result KeyBounds are of the whole layer
Results of queries should contain `KeyBounds` and `Extent` that are intersections of the query `KeyBounds`, `Extent` and the layer `KeyBounds`. This is more relevant now because `TileRDDReproject` uses layer bounds to calculate target extent.

This behavior change should be called out in the changelog.

jbouffard:
## Overview

This PR makes it so that the resulting `TileLayerMetadata` of a `query` will contain the correct `KeyBounds` and `Extent` of the queried region. This is opposed to the current output which will have the `KeyBounds` and `Extent` for the whole layer.

### Checklist

Closes #2543

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/io/package.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/package.scala
@@ -17,7 +17,9 @@
 package geotrellis.spark
 
 import geotrellis.raster.io.geotiff.reader.GeoTiffReader
-import geotrellis.spark.tiling.MapKeyTransform
+import geotrellis.spark.tiling.{MapKeyTransform, LayoutDefinition}
+import geotrellis.vector.Extent
+import geotrellis.util._
 
 import com.typesafe.config.Config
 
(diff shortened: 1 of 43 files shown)


spark/src/main/scala/geotrellis/spark/io/package.scala
```diff
@@ -50,6 +52,25 @@ package object io
...
+    K: SpatialComponent: Boundable,
+    M: Component[?, Bounds[K]]: Component[?, LayoutDefinition]: Component[?, Extent]
+  ](
+    keyBounds: Seq[KeyBounds[K]],
+    sourceMetadata: M
+  ): M = {
+    val mapTransform = sourceMetadata.getComponent[LayoutDefinition].mapTransform
+
+    if (keyBounds.isEmpty)
+      sourceMetadata.setComponent[Bounds[K]](EmptyBounds).setComponent[Extent](Extent(0, 0, 0, 0))
```
pomadchin: `Extent(0, 0, 0, 0)` would throw a construction exception: https://github.com/locationtech/geotrellis/blob/master/vector/src/main/scala/geotrellis/vector/Extent.scala#L117-L118

I guess it was not covered with tests?

jbouffard: So `Extent.isEmpty` https://github.com/locationtech/geotrellis/blob/master/vector/src/main/scala/geotrellis/vector/Extent.scala#L151 would always return `true`?

pomadchin: @jbouffard ah sry; i was thinking about a different thing. It's good here.

echeipesh:
@pomadchin don't we actually already have that with `Component[Bounds]` [here](https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/TileLayerMetadata.scala#L101-L102) that implements the [update](https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/TileLayerMetadata.scala#L69-L86)

pomadchin:
@echeipesh I'll double check it! that's amazing, how smart we are.

pomadchin:
Replaced by https://github.com/locationtech/geotrellis/pull/2633
@jbouffard can you check #2633 ?

jbouffard:
@pomadchin Why was this PR closed?

PR #2586, opened 2018-03-20
Fix spatial Hilbert to use offset of the minimum range key

echeipesh:
## Overview

Spatial Hilbert index incorrectly did not offset the keys being encoded by the minimum key in the `KeyBounds` of the index. Because Hilbert will attempt to figure out the number of bits required to encode values in the given key range it the following exception can be thrown when `KeyBounds` provided to the [HilbertSpatialKeyIndex constructor](https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/io/index/hilbert/HilbertSpatialKeyIndex.scala#L52) does not contain `minKey = SptialKey(0, 0)`

`java.lang.IllegalArgumentException: value doesn't fit`

However, it is possible that a layer indexed by `HilbertSpatialKeyIndex` with `minKey` > `SpatialKey(0,0)` would not produce this error when writing if the minimum key was large enough to require the same number of bits as the maximum key. This PR will break reading those layers by changing the mapping from stored record index to `SpatialKey` instance.

It seems unlikely that this bug would have gone unreported if this is was a prevailing case, however there is a way to fix such layers without re-ingesting them.

*IF* this PR breaks your layer indexed by `HilbertSpatialKeyIndex` you can fix it by changing the layer metadata `keyIndex.properties.keyBounds.minKey` to have `col = 0` and `row = 0`

Before:
screen shot 2018-03-20 at 4 20 11 pm

After:
screen shot 2018-03-20 at 4 20 25 pm

This will not effect the layer `KeyBounds` as shown in the `TileLayerMetadata[K]` but only adjust the `KeyBounds` used by the `HilbertSpatialKeyIndex` instance, adjusting it to match new reality.


echeipesh:
Because this is basically a glaring bug and evidence is that this is not popular index strategy the intention is to back-port this to `1.2` branch and trigger release of `1.2.2`

PR #2587, opened 2018-03-21
Spark Split Part 1: geotrellis-tiling

jbouffard:
## Overview

This PR is the first of two that seeks to break out types from `geotrellis.spark` that are not dependent on `spark` itself. In this part, the types that deal with tiling were broken out into a new package called, `geotrellis-tiling`. These types include `SpatialKey`, `SLpaceTimeKey`, and `LayoutDefinition` among others.

### Notes

This PR will break any preexisting GeoTrellis code. Before work on the second part of this task, breaking out the `collections` API, this PR should be approved.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -173,14 +173,13 @@ lazy val spark = project
   )
 
 lazy val `spark-testkit` = project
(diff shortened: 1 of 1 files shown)


fosskers:
Sweet.

PR #2592, opened 2018-03-26
COGLayerProviders

jbouffard:
## Overview

This PR adds `LayerProvider` support for `COGLayer`s based on the Java SPI.

### Checklist

### Demo

```scala

val path = "file:///tmp/test-catalog"
val reader = COGFileLayerReader(path)
```

### DIFF
--- a/s3/src/main/scala/geotrellis/spark/io/s3/cog/S3COGCollectionLayerReader.scala
+++ b/s3/src/main/scala/geotrellis/spark/io/s3/cog/S3COGCollectionLayerReader.scala
@@ -39,8 +39,6 @@ import scala.reflect.ClassTag
  */
 class S3COGCollectionLayerReader(
   val attributeStore: AttributeStore,
-  val bucket: String,
-  val prefix: String,
   val getS3Client: () => S3Client = () => S3Client.DEFAULT,
   val defaultThreads: Int = S3COGCollectionLayerReader.defaultThreadCount
 ) extends COGCollectionLayerReader[LayerId] with LazyLogging {
@@ -51,6 +49,17 @@ class S3COGCollectionLayerReader(
     K: SpatialComponent: Boundable: JsonFormat: ClassTag,
     V <: CellGrid: GeoTiffReader: ClassTag
   ](id: LayerId, rasterQuery: LayerQuery[K, TileLayerMetadata[K]]) = {
+    val header =
+      try {
+        attributeStore.read[S3LayerHeader](LayerId(id.name, 0), COGAttributeStore.Fields.header)
+      } catch {
+        // to follow GeoTrellis Layer Readers logic
+        case e: AttributeNotFoundError => throw new LayerNotFoundError(id).initCause(e)
+      }
+
+    val bucket = header.bucket
+    val prefix = header.key
+
     def getKeyPath(zoomRange: ZoomRange, maxWidth: Int): BigInt => String =
       (index: BigInt) =>
         s"$bucket/$prefix/${id.name}/" +
@@ -74,8 +83,9 @@ object S3COGCollectionLayerReader {
   def apply(attributeStore: S3AttributeStore): S3COGCollectionLayerReader =
     new S3COGCollectionLayerReader(
       attributeStore,
-      attributeStore.bucket,
-      attributeStore.prefix,
       () => attributeStore.s3Client
     )
+
+  def apply(bucket: String, prefix: String): S3COGCollectionLayerReader =
+    apply(S3AttributeStore(bucket, prefix))
 }
--- a/s3/src/main/scala/geotrellis/spark/io/s3/cog/S3COGLayerWriter.scala
+++ b/s3/src/main/scala/geotrellis/spark/io/s3/cog/S3COGLayerWriter.scala
@@ -25,7 +25,7 @@ class S3COGLayerWriter(
   keyPrefix: String,
   getS3Client: () => S3Client = () => S3Client.DEFAULT,
   threads: Int = S3RDDWriter.DefaultThreadCount
-) extends COGLayerWriter {
+) extends COGLayerWriter[LayerId] {
 
   def writeCOGLayer[
     K: SpatialComponent: Ordering: JsonFormat: ClassTag,
--- a/s3/src/main/scala/geotrellis/spark/io/s3/cog/S3COGLayerWriter.scala
+++ b/s3/src/main/scala/geotrellis/spark/io/s3/cog/S3COGLayerWriter.scala
@@ -25,7 +25,7 @@ class S3COGLayerWriter(
   keyPrefix: String,
   getS3Client: () => S3Client = () => S3Client.DEFAULT,
   threads: Int = S3RDDWriter.DefaultThreadCount
-) extends COGLayerWriter[LayerId] {
+) extends COGLayerWriter {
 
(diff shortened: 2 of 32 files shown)


spark/src/main/scala/geotrellis/spark/io/cog/COGLayerWriter.scala
```diff
@@ -50,32 +53,30 @@ trait COGLayerWriter extends LazyLogging with Serializable {
...
     tiles: RDD[(K, V)] with Metadata[TileLayerMetadata[K]],
-    tileZoom: Int,
     keyIndexMethod: KeyIndexMethod[K],
     compression: Compression,
     mergeFunc: Option[(GeoTiff[V], GeoTiff[V]) => GeoTiff[V]]
   ): Unit =
     tiles.metadata.bounds match {
       case keyBounds: KeyBounds[K] =>
-        val cogLayer = COGLayer.fromLayerRDD(tiles, tileZoom, compression = compression)
+        val cogLayer = COGLayer.fromLayerRDD(tiles, id.asInstanceOf[LayerId].zoom, compression = compression)
```
pomadchin: Eh, `asInstanceOf[LayerId]` kills the idea of having a generic `ID` in the interface. It looks like in this case we can have only `LayerId` (?)

Would not such API be a bit confusing for the user?

jbouffard: Yeah, I was thinking that it might be better to just use `LayerId` instead of `ID`. Which API do you mean? The one that uses `LayerId`?

pomadchin: Yeah, this move to `LayerId` changes semantics a bit; because COG layer always builds a pyramid right now

jbouffard: Do you think it'd be better to go back to having the `layerName` and `layerZoom` parameters for the `write` methods?

s3/src/main/scala/geotrellis/spark/io/s3/cog/S3COGCollectionLayerReader.scala
```diff
@@ -39,8 +39,6 @@ import scala.reflect.ClassTag
  */
 class S3COGCollectionLayerReader(
   val attributeStore: AttributeStore,
-  val bucket: String,
```
pomadchin: Good catch!

s3/src/main/scala/geotrellis/spark/io/s3/cog/S3COGLayerWriter.scala
```diff
@@ -25,7 +25,7 @@ class S3COGLayerWriter(
   keyPrefix: String,
   getS3Client: () => S3Client = () => S3Client.DEFAULT,
   threads: Int = S3RDDWriter.DefaultThreadCount
-) extends COGLayerWriter {
+) extends COGLayerWriter[LayerId] {
```
pomadchin: Do we really need `[ID]` there?

spark/src/main/scala/geotrellis/spark/io/cog/COGLayerWriter.scala
```diff
@@ -24,16 +24,19 @@ import geotrellis.raster.crop._
...
 import org.apache.spark.rdd.RDD
 import spray.json._
 
+import java.net.URI
+import java.util.ServiceLoader
+
 import scala.reflect._
 
-trait COGLayerWriter extends LazyLogging with Serializable {
+abstract class COGLayerWriter[ID] extends LazyLogging with Serializable {
```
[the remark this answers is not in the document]
jbouffard: The only thing that `ID` is used for is in the `writer` methods https://github.com/jbouffard/geotrellis/blob/8d90d529ada19b2668c47c389772d127a241b6ae/spark/src/main/scala/geotrellis/spark/io/cog/COGLayerWriter.scala#L121-L137

But those could be changed to `LayerId` or `(String, Int)` if either of those would be better.

spark/src/main/scala/geotrellis/spark/io/hadoop/cog/HadoopCOGLayerProvider.scala
```diff
@@ -0,0 +1,80 @@
...
+    val path = new Path(_uri)
+    new HadoopCOGLayerWriter(path.toString, store)
+  }
+
+  def valueReader(uri: URI, store: AttributeStore): COGValueReader[LayerId] = {
+    val _uri = trim(uri)
+    val path = new Path(_uri)
+    val params = UriUtils.getParams(_uri)
+    val conf = new Configuration()
+    val maxOpenFiles = params.getOrElse("maxOpenFiles", "16").toInt
```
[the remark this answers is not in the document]
jbouffard: Whoops, I left those in by accident. I'll remove them.

s3/src/main/scala/geotrellis/spark/io/s3/cog/S3COGCollectionLayerReader.scala
jbouffard: Thanks!

spark/src/main/scala/geotrellis/spark/io/cog/COGLayerWriter.scala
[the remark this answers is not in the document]
jbouffard: Oh, never mind. I thought I had `ID` there haha. We can get rid of the `ID` then

s3/src/main/scala/geotrellis/spark/io/s3/cog/S3COGLayerWriter.scala
jbouffard: Do you mean the `[LayerId]`?

jbouffard: :+1:

spark/src/main/scala/geotrellis/spark/io/hadoop/cog/HadoopCOGLayerProvider.scala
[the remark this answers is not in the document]
jbouffard: Actually, that code is used for `HadoopCOGValue` which does use those values. https://github.com/jbouffard/geotrellis/blob/8d90d529ada19b2668c47c389772d127a241b6ae/spark/src/main/scala/geotrellis/spark/io/hadoop/cog/HadoopCOGLayerProvider.scala#L65 So it might be better to keep them.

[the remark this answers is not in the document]
pomadchin: Nope, HadoopValueReader uses `MapFile.Reader` to read tiffs, but here we don't use it.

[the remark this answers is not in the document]
pomadchin: https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/io/hadoop/cog/HadoopCOGValueReader.scala#L38 

its not passed anywhere, was an oversight by me.

jbouffard:
@pomadchin Is there anything else I need to change in this PR?

PR #2599, opened 2018-03-28
add coursier plugin

dos65:
I tried to compile geotrellis with coursier because @pomadchin claimed that it doesn't work for his project, and it works fine.


pomadchin:
Thank you for double checking it for us, if travis would be happy with this solution we're ready to move on coursier.

metasim:
Just a word of warning: keep a lookout for people logging weird bugs; coursier has burned me in the past with complicated projects. Hopefully they've fixed the various issues, as using it would be awesome. Just keep an eye out.

PR #2614, opened 2018-04-03
Build GeoTiff overviews by resampling the whole base tile

echeipesh:
This approach avoids numerical differences in partial vs while tile resample.
They were particularly apparently with `NearestNeighbor` resample method.

Another change in this PR is the restriction that overviews will always have the same extent as the underlying raster. This change is to match GDAL behavior. It should be noted that it may cause stretching/squishing of the overview pixels where the overview pixel boundaries do not align with base layer pixel boundaries. This will happen when base raster pixel column/row count is not a multiple of the decimation factor.

In structured COG case however the base tile will always have a pixel count that is power of two and the case above will be avoided.

Additionally we trade memory for performance by forcing the base tile to ArrayTile before resampling.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/io/geotiff/MultibandGeoTiff.scala
+++ b/raster/src/main/scala/geotrellis/raster/io/geotiff/MultibandGeoTiff.scala
@@ -84,19 +84,8 @@ case class MultibandGeoTiff(
       .resample(rasterExtent, resampleMethod)
 
   def buildOverview(resampleMethod: ResampleMethod, decimationFactor: Int, blockSize: Int): MultibandGeoTiff = {
-    // pad overview with extra cells to keep 1 source pixel = d overview pixels alignment
-    // this may cause the overview extent to expand to cover the wider pixels as well
-    // val padCols: Int = if (tile.cols % decimationFactor == 0) 0 else decimationFactor - tile.cols % decimationFactor
-    // val padRows: Int = if (tile.rows % decimationFactor == 0) 0 else decimationFactor - tile.rows % decimationFactor
-    val padCols: Int = decimationFactor
-    val padRows: Int = decimationFactor
     val overviewRasterExtent = RasterExtent(
-      Extent(
-        xmin = extent.xmin,
-        ymin = extent.ymin - padRows * cellSize.height,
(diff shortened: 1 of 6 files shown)


raster/src/main/scala/geotrellis/raster/io/geotiff/MultibandGeoTiff.scala
```diff
@@ -136,6 +129,32 @@ case class MultibandGeoTiff(
...
+        case tiffTile: GeoTiffMultibandTile =>
+          tiffTile.toArrayTile() // allow GeoTiff tile to read segments in optimal way
+        case _ =>
+          MultibandTile(tile.bands.map(_.toArrayTile()))
+      }
+      val staged = MultibandGeoTiff(arrayTile, extent, crs, tags, options, Nil)
+      val overviews = overviewDecimations.map { (decimationFactor: Int) =>
+        staged.buildOverview(resampleMethod, decimationFactor, blockSize)
+      }
+      MultibandGeoTiff(tile, extent, crs, tags, options, overviews)
```
echeipesh: @pomadchin What do you think is more appropriate here?
Should we return the base tile, which could be lazy or compressed, or to return the forced `ArrayTile` since we already went through the trouble of fetching it.

pomadchin: @echeipesh from the API perspective, I'd expect to see the result of function as a geotiff. What's your point for the `ArrayTile`?

echeipesh: In order to generate the overviews in a way that match the base tile resample and save on segment thrashing the overview code will pull the base GeoTiff into `ArrayTile` and resample that.

The question that I have: given that we do that, should the returned `GeoTiff` now hold the `ArrayTile` version of the segments or the original geotiff segments, which could be lazy. I was thinking the latter, but wanted a double check.

pomadchin: Yea, `the original geotiff segments` sounds correct :+1:

PR #2618, opened 2018-04-08
Prefer JTS CoordinateSequences to Coordinates

mojodna:
## Overview

This changes internal geometry construction to use `CoordinateSequence`s instead of `Coordinate`s, reducing the number of intermediate `Coordinate` allocations when `PackedCoordinateSequence`s are used (coordinates are stored in a `double[]` and allocated as `Coordinate`s when needed).

It also introduces some new `Polygon` and `Line` constructors to allow `CoordinateSequence`s to be efficiently used by client code (without access to GT's `GeometryFactory`).

`vertices` has been rewritten to avoid intermediate allocations (and added to the `Geometry` trait along with `vertexCount`).

### Notes

Overall, this appears to reduce the number of intermediate object allocations (when used with `PackedCoordinateSequence`s; see #2619) and reduces pressure on the GC.


echeipesh:
The eclipse CLA validation bot seems to be on the fritz.

@mojodna could you please sign the Eclipse CLA: https://www.eclipse.org/legal/ECA.php
and "sign off" on the PR of the by adding a line with your GitHub name and email  as in: https://github.com/locationtech/geotrellis/pull/2571

vector/src/main/scala/geotrellis/vector/Polygon.scala
```diff
@@ -38,34 +37,44 @@ object Polygon {
...
+    apply(factory.createLinearRing(exterior), holes.map(factory.createLinearRing))
+
+  def apply(exterior: Line, holes: GenTraversable[Line]): Polygon = {
     if(!exterior.isClosed) {
       sys.error(s"Cannot create a polygon with unclosed exterior: $exterior")
     }
 
     if(exterior.vertices.length < 4) {
-      sys.error(s"Cannot create a polygon with exterior with less that 4 points: $exterior")
+      sys.error(s"Cannot create a polygon with exterior with fewer than 4 points: $exterior")
```
echeipesh: Using this PR to think about immutability vs performance in `geotrellis.vector` and noticed that ^ in line above where `. vertices.length` is called an `Array[Point]` is created to perform the length call. `.vertexCount` should be safe substitute.

mojodna:
Signed-off-by: Seth Fitzsimmons seth@mojodna.net

pomadchin:
Hm, still `should maintain immutability over points *** FAILED ***`

pomadchin:
The last commit should be considered also in terms of #2622

mojodna:
I've been meaning to weigh back in on what I consider to be the controversial part of this change: the use of `PackedCoordinateSequence`s.

I've spent some more time on the problem that looked like it was benefiting from switching out the `CoordinateSequenceFactory` and managed to get it performant (i.e. it runs to completion now) while keeping the previous behavior (part of the solution involved implementing a `VirtualCoordinateSequence` to avoid recreating arrays of coordinates). I have not had the chance to compare it using `PackedCoordinateSequence`s, so I can't offer a relative comparison, though my instinct is that re-using the existing `Coordinate` objects will work better with GT 1.2's behavior.

tl;dr: I still strongly support the new constructors, use of `CoordinateSequence`s under the hood, and making the `GeometryFactory` and `CoordinateSequenceFactory` public, but I'm less confident about the type of sequence than I was originally.

pomadchin:
@mojodna are you talking about https://github.com/locationtech/geotrellis/pull/2619?

mojodna:
Uhh, yeah. I was thinking this was that. I stand by this PR ;-)

PR #2619, opened 2018-04-08
Expose GeomFactory into the user API and move vector project to use pureconfig

mojodna:
## Overview

When working with JTS geometries directly in client code, it's useful to have access to the same `Geometry` and `CoordinateSequence` factories that GT is using (vs. creating local instances w/ potentially differing settings). All other members remain private to the `geotrellis.vector` package.

More controversially, this switches GT's default `CoordinateSequenceFactory` to `PackedCoordinateSequenceFactory` in order to reduce the number of objects allocated (`double[]` is used rather than `Coordinate[]`) and reduce pressure on the GC.

This shouldn't affect client implementations unless `CoordinateSequence`s are explicitly cast; this shouldn't occur and isn't necessary when the `CoordinateSequences` helper class is used.

### DIFF
--- a/vector/src/main/scala/geotrellis/vector/GeomFactory.scala
+++ b/vector/src/main/scala/geotrellis/vector/GeomFactory.scala
@@ -29,6 +29,5 @@ object GeomFactory extends LazyLogging {
   val precisionModel: PrecisionModel = JtsConfig.precisionModel
   lazy val simplifier: GeometryPrecisionReducer = JtsConfig.simplifier
 
-  val coordinateSequenceFactory: CoordinateSequenceFactory = CoordinateArraySequenceFactory.instance
-  val factory: GeometryFactory = new geom.GeometryFactory(precisionModel, 0, coordinateSequenceFactory)
+  val factory: GeometryFactory = new geom.GeometryFactory(precisionModel)
 }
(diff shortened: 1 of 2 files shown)


pomadchin:
Moving the conversation here, @mojodna do you still think this PR has to be merged? Or you have some ideas on how to modify it to get merged?

mojodna:
@pomadchin changes pushed. The branch name no longer makes sense, but these are the changes I'd like and are ones that should be consistent over time (vs. exposing the `simplifier`, since it sounds like that might change).

vector/src/main/scala/geotrellis/vector/GeomFactory.scala
```diff
@@ -53,10 +53,11 @@ private[vector] object GeomFactory extends LazyLogging {
     case s => throw new IllegalArgumentException(s"""Unrecognized JTS precision model, ${precisionType}; expected "floating", "floating_single", or "fixed" """)
   }
 
-  val factory = new geom.GeometryFactory(precisionModel)
+  val coordinateSequenceFactory: CoordinateSequenceFactory = CoordinateArraySequenceFactory.instance
```
pomadchin: I though that it's a default coordinate sequence factory; the intention is to override it and expose into public API?

pomadchin:
This PR now looks more like a migration on a `pureconfig` :D

Issue #2622, opened 2018-04-11
Scope usage of defensive jtsGeom.clone in geotrellis.vector


echeipesh:
This issue is regarding proper balance of performance and safety by immutability.

Geometry classes in `geotrellis.vector` wrap instances JTS `Geometry` objects in order to provide Scala friendly and type safe interface to work with JTS. One way in which we try to make them Scala friendly is to try to maintain immutability of geometries. JTS `Coordinates` and thus geometries are mutable, `geotrellis.vector` tries to wrap that way by aggressively using `.clone`.

### Problem:
We provide an escape hatch by exposing `jtsGeom` as public field: 
[`case class Point(jtsGeom: jts.Point) extends Geometry`](https://github.com/locationtech/geotrellis/blob/master/vector/src/main/scala/geotrellis/vector/Point.scala#L37)

We kind of have to do this to avoid locking people into more restrictive API because we have not, and probably will not wrap all of JTS API. 

However, this shoots immutability promise in the foot a little because a determined hacker has an easy way to grab the underlying JTS objects and mutate them for whatever reason.

Given that mutability is a step away the benefit of defensive use of `.clone` when returning wrapped component geometries is less clear. A good example is [`MultiLine.lines`](https://github.com/locationtech/geotrellis/blob/master/vector/src/main/scala/geotrellis/vector/MultiLine.scala#L62-L66) method.

In these cases we're just creating a lot of memory pressure without clear intent or win.

### Proposed Solution:

Clarify the semantics in `geotrellis.vector` API to:
- perform `.clone` when wrapping a mutable operation.
   - example: [`Line.normalize`](https://github.com/locationtech/geotrellis/blob/master/vector/src/main/scala/geotrellis/vector/Line.scala#L59-L63)
- avoid `.clone` when returning wrap component geometries.
  - example: [`MultiLine.lines`](https://github.com/locationtech/geotrellis/blob/master/vector/src/main/scala/geotrellis/vector/MultiLine.scala#L62-L66)

Essentially the API contract is:
If you stay within `geotrellis.vector` types we provide type safety and immutability through operations.
If you reach for `jtsGeom` reference, the concern for mutability and its effect is on you.

Essentially I'd suggest declaring [`this`](https://github.com/locationtech/geotrellis/blob/master/vector/src/test/scala/spec/geotrellis/vector/LineSpec.scala#L528-L535) and other tests like it as invalid.

This "refinement" is still change in API behavior so any community comments around this area are very valuable.

PR #2623, opened 2018-04-11
Re-introduce use of defensive .clone in vertices method for Line

echeipesh:
Rolling back the change to get code base to consistent point. I agree with the trade off on performance vs immutability here, but the such usage of `.clone` is prevalent in `geotrellis.vector` and any changes there should be part of a single principled PR.

While not ideal I assume the need currently the need fulfilled by this change can be met by reaching into `line.jtsGeom` and iterating over `getPointN` directly.

Created issue for follow up discussion: https://github.com/locationtech/geotrellis/issues/2622

@mojodna @pomadchin @lossyrob

### DIFF
--- a/vector/src/main/scala/geotrellis/vector/Line.scala
+++ b/vector/src/main/scala/geotrellis/vector/Line.scala
@@ -96,7 +96,7 @@ case class Line(jtsGeom: jts.LineString) extends Geometry
     val size = jtsGeom.getNumPoints
     val arr = Array.ofDim[Point](size)
     cfor(0)(_ < arr.size, _ + 1) { i =>
-      arr(i) = Point(jtsGeom.getPointN(i))
+      arr(i) = Point(jtsGeom.getPointN(i).clone.asInstanceOf[jts.Point])
     }
     arr
   }


mojodna:
FWIW, my original impetus for proposing that change was discovering that JTS already clones the points before they hit GeoTrellis:

https://github.com/locationtech/jts/blob/30aeae1e1530e586f9017a4dd973de8b011dc73f/modules/core/src/main/java/org/locationtech/jts/geom/LineString.java#L123-L125

```java
  public Point getPointN(int n) {
      return getFactory().createPoint(points.getCoordinate(n));
  }
```

lossyrob:
☝️ True, which is why I merged it without worry originally - I don't see how this actually breaks immutability

pomadchin:
Does it really makes a copy? It seems to me that it passes everywhere the same (not cloned) `Coordinate` object. Yes it wraps it into arrays at various places, but it's still the same reference, that's probably why immutability test fails.

```java
public Point createPoint(Coordinate coordinate) {
    return createPoint(coordinate != null ? getCoordinateSequenceFactory().create(new Coordinate[]{coordinate}) : null);
  }
```

mojodna:
Ah, I was looking at `PackedCoordinateSequenceFactory`, where it does effectively copy it into a new `double[]`. Good catch, thanks!

PR #2626, opened 2018-04-16
LayerProviders Can Now Handle a URI That is Missing a Scheme

### LINKED ISSUE #2551: NPE in `FileLayerProvider` when scheme is missing from URI
Happens here:

https://github.com/locationtech/geotrellis/blob/ec10694349c0e06fdf0af789579b966ed2af1f08/spark/src/main/scala/geotrellis/spark/io/file/FileLayerProvider.scala#L33

According to [this](https://docs.oracle.com/javase/7/docs/api/java/net/URI.html#getScheme()), returning `null` is acceptable. 

Should probably check the other `LayerProvider`s for proper null handling. 

I'd argue that in the case of a missing scheme, `FileLayerProvider` should be the fallback selection.

jbouffard:
## Overview

This PR makes it so that a `RuntimeException` will be thrown when a user tries to pass in a `URI` to a `LayerProvider` instead of a `NullPointerException`.

### Checklist

Closes #2551

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/io/file/FileLayerProvider.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/file/FileLayerProvider.scala
@@ -30,7 +30,10 @@ import java.io.File
 class FileLayerProvider extends AttributeStoreProvider
     with LayerReaderProvider with LayerWriterProvider with ValueReaderProvider with CollectionLayerReaderProvider {
 
-  def canProcess(uri: URI): Boolean = uri.getScheme.toLowerCase == "file"
+  def canProcess(uri: URI): Boolean = uri.getScheme match {
+    case str: String => if (str.toLowerCase == "file") true else false
+    case null => true // assume that the user is passing in the path to the catalog
+  }
 
   def attributeStore(uri: URI): AttributeStore = {
     val file = new File(uri)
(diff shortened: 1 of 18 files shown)


spark/src/main/scala/geotrellis/spark/io/file/FileLayerProvider.scala
```diff
@@ -30,7 +30,10 @@ import java.io.File
 class FileLayerProvider extends AttributeStoreProvider
     with LayerReaderProvider with LayerWriterProvider with ValueReaderProvider with CollectionLayerReaderProvider {
 
-  def canProcess(uri: URI): Boolean = uri.getScheme.toLowerCase == "file"
+  def canProcess(uri: URI): Boolean = uri.getScheme match {
+    case str: String => if (str.toLowerCase == "file") true else false
+    case null => false
```
echeipesh: This should be `true` I think? Something like `/Users/eugene/catalog` we can assume implies local file.

jbouffard: I made that `false` so that its behavior is more in line with the other backends; but I can see why we'd want it to be `true`.

jbouffard:
@echeipesh Is there anything else that needs to be done for this PR?

PR #2628, opened 2018-04-17
Specify maxTileSize When Writing a COGLayer

jbouffard:
## Overview

This PR makes it so that a user can specify the `maxTileSize` of a COGLayer that is about to written in the `write` methods themselves. This is done via the newly created `COGLayerWrtier.Options` which can be passed in to the `write` methods.

### Checklist

- [x] `docs/CHANGELOG.rst` updated, if necessary
- [x] `docs` guides update, if necessary
- [x] New user API has useful Scaladoc strings
- [x] Unit tests added for bug-fix or new feature

### Demo

```scala

val options = COGLayerWriter.Options(1024)

writer.write[K, V](layerId.name, tiles, layerId.zoom, keyIndexMethod, options)
```

Closes #2606

Issue #2606, opened 2018-03-29
maxTileSize not able to be set in COGLayerWriter


lossyrob:
Suggesting an API change: we should allow options to be passed into the COGLayerWriter methods. This would avoid users having to learn another type (COGLayer), and pass any necessary options like maxTileSize (which cannot currently be set from a `.write` method) and resample method.

What I want to do:

```scala
    val writer = FileCOGLayerWriter(attributeStore)

    val layerName = "landsat"
  
    val options = 
      COGLayerWriter.Options(
        resampleMethod = Bilinear,
        maxTileSize = 1024,
        compression = Deflate
      )

    writer.write(layerName, reprojected, zoom, options)
```

what I currently have to do:

```scala
    val writer = FileCOGLayerWriter(attributeStore)

    val layerName = "landsat"

    val cogLayer =
      COGLayer.fromLayerRDD(reprojected, zoom, compression = Deflate, maxTileSize = 1024)

    val keyIndexes =
      cogLayer.metadata.zoomRangeInfos.
        map { case (zr, bounds) => zr -> ZCurveKeyIndexMethod.createIndex(bounds) }.
        toMap

    writer.writeCOGLayer(layerName, cogLayer, keyIndexes)
````

jbouffard:
@lossyrob Do we want to encapsulate `maxTileSize`, `resampleMethod`, etc. in `COGLayerWriter.Options`? I thought we were trying to move away from that since it obfuscates the options available to the user.

PR #2630, opened 2018-04-18
Specify a resampleMethod and maxTileSize when writing a COGLayer

### LINKED ISSUE #2605: Implement resample options for COGLayerWriter
We'll need to pass in a resample option for COGLayerWriter methods, so that the pyramid method knows how to properly resample to overviews.

jbouffard:
## Overview

This PR adds the `resampleMethod`, `maxTileSize` and moves `compression` parameter to `COGLayerWriter.Options`.

### Checklist

### Demo

```scala
COGLayer.fromLayerRDD(layerRDD, zoom, options = Bilinear)
```

```scala
val options = COGLayerWriter.Options(1024)

writer.write[K, V](layerId.name, tiles, layerId.zoom, keyIndexMethod, options)
```

### Notes

This PR is based on another, open PR #2628 which should be merged before this one.

Closes #2605 and #2606

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/io/cog/COGLayer.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/cog/COGLayer.scala
@@ -42,9 +42,8 @@ object COGLayer {
     * @param rdd             Layer layer, at highest resolution
     * @param baseZoom        Zoom level of the base layer, assumes [[ZoomedLayoutScheme]]
     * @param compression     Compression method for GeoTiff tiles
-    * @param maxTileSize     The maximum tile size in pixels for any one COG file for this layer.
-    *                        For instance, if 1024, no COG in the layer will have a greater width or height than 1024.
     * @param minZoom         Zoom level at which to stop the pyramiding.
+    * @param options         [[COGLayerWriter.Options]] that contains information on the maxTileSize of the layer.
     */
   def fromLayerRDD[
     K: SpatialComponent: Ordering: JsonFormat: ClassTag,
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/geotrellis/spark/io/cog/COGLayer.scala
```diff
@@ -53,10 +52,11 @@ object COGLayer {
      rdd: RDD[(K, V)] with Metadata[TileLayerMetadata[K]],
      baseZoom: Int,
      compression: Compression = Deflate,
-     maxTileSize: Int = 4096,
-     minZoom: Option[Int] = None
+     minZoom: Option[Int] = None,
+     options: COGLayerWriter.Options
```
pomadchin: Hm, mb it makes sense to put compression into options to? 
And to have options here with a default value.

jbouffard: Yeah, I think it'd make sense to put `compression` into `Options` I only didn't because that was outside the scope of the PR.

jbouffard: But if you think that it should be added in this one then I can do that

pomadchin: Let's do all these `options` things in a single PR (in this one).

jbouffard: Okay, that sounds good.

PR #2633, opened 2018-04-19
Query Results Will Now Have the Correct KeyBounds and Extent

### LINKED ISSUE #2543: Layer query result KeyBounds are of the whole layer

pomadchin:
## Overview

This PR makes it so that the resulting TileLayerMetadata of a query will contain the correct KeyBounds and Extent of the queried region. This is opposed to the current output which will have the KeyBounds and Extent for the whole layer.

This PR is an updated version of https://github.com/locationtech/geotrellis/pull/2582

### Checklist

Closes #2543


echeipesh:
Could use an added checkin `Spec` s

PR #2634, opened 2018-04-20
Tile.slope(cellSize, target = TargetCell.Data) Fix

### LINKED ISSUE #2563: Tile.slope (cellSize, target = TargetCell.Data) problem
If the Tile's first col's value has NoData, then after `Tile.slope` with `TargetCell.Data`, it's entire line will be NoData.
The related code is at  raster\mapalgebra\focal\hillshade\SurfacePointCalculation.scala L268 :
image
I think you need to read the focalValue again in MiddleMiddle.

jbouffard:
## Overview

This PR fixes a bug in the `Tile.slope(cellSize, target = TargetCell.Data)` method call where if the first value of a column was `NoData` then all of the values in that row would be set to `NoData`.

### Checklist

Closes #2563


pomadchin:
LGTM! Merging.

PR #2637, opened 2018-04-24
Migrate from Scalaz streams to FS2 Streams

### LINKED ISSUE #1778: Move to fs2 streams
Since we dropped Scala 2.10 we can move to use [fs2 streams](https://github.com/functional-streams-for-scala/fs2) without scalaz dependency.

pomadchin:
## Overview

This PR removes dependence on Scalaz and introduces dependence on cats.

### Checklist

- [ ] Test parallelism on real ingests (S3 / Accumulo / Hadoop)

Closes #1778


spark/src/main/scala/geotrellis/spark/util/TaskUtils.scala
```diff
@@ -17,32 +17,31 @@
...
-     */
-    def retryEBO(p: (Throwable => Boolean) ): Task[A] = {
-      def help(count: Int): Future[Throwable \/ A] = {
+  /**
+    * Implement non-blocking Exponential Backoff on a Task.
+    * @param  p  returns true for exceptions that trigger a backoff and retry
+    */
+
+  implicit class IOBackoff[A](ioa: IO[A]) {
+    def retryEBO(p: (Throwable => Boolean))(implicit timer: Timer[IO]): IO[A] = {
```
pomadchin: NOTE: Double check how it works

pomadchin:
WIP as it has to be tested

project/Dependencies.scala
```diff
@@ -39,7 +39,8 @@ object Dependencies {
 
   val awsSdkS3            = "com.amazonaws"               % "aws-java-sdk-s3"          % "1.11.143"
 
-  val scalazStream        = "org.scalaz.stream"          %% "scalaz-stream"            % "0.8.6a"
+  val fs2Core             = "co.fs2"                     %% "fs2-core"                 % "0.10.3"
+  val fs2Io               = "co.fs2"                     %% "fs2-io"                   % "0.10.3"
```
fosskers: It looks like the cats dependency is transitive through `fs2`? Make sure that the version of `cats-effect` that it's pulling is at least `0.10` or the performance of the `IO` monad will be garbage. They recently improved it.

Perhaps consider adding an explicit dependency on `cats-effect` here?

pomadchin: `cats-core`: 1.1.0
`cats-effect`: 0.10

pomadchin: Good point to bump up cats versions manually as we plan to depend on it!

pomadchin:
Our travis doesn't work at the moment, but on my repo last commit CI run passed: https://travis-ci.org/pomadchin/geotrellis/builds/371922116 

:tada:

spark/src/main/scala/geotrellis/spark/util/TaskUtils.scala
jpolchlo: Has this comment been rectified?  Did you double-check?

pomadchin: Yea, looks good, and here is the reference: https://typelevel.org/cats-effect/datatypes/io.html#example-retrying-with-exponential-backoff

pomadchin:
Rebased on master, merging after CI would pass.

PR #2639, opened 2018-04-24
Deprecate Slick and move Projected

### LINKED ISSUE #2557: Break `Projected` out of Slick
`Projected` is currently the only way we have to associate a geometry with a projection. This lives in the Slick package and requires us to bring in all those dependencies. We should move `Projected` to vector so that using a library like, say, Doobie doesn't require bringing in Slick dependencies.

As evidence that this is a fine way of representing the relationship, note that JTS and `org.postgis` store projections *on* their geometries...

moradology:
## Overview

Slick is not fun to use. It is nigh-impossible to debug complex queries. We will no longer be suggesting that people use it, though the notion of a Projected Geometry is something we'll need to support for PostGIS usage. This PR deprecates the relevant classes and salvages the useful bits from `geotrellis.slick`

Closes #2557


pomadchin:
Btw, there are also probably useful functions in `PostGisProjectionSupportUtils` and in `PostGisSupportUtils`. Does it make any sense to move them into vector package as well? 
It can be pretty convenient to have `readWktOrWkb` function in `vector` package. But its only a question.

However looks good, we can merge after you'll say that we don't need them as it can be a part of a user API and not a part of a GeoTrellis API.

moradology:
@pomadchin `readWKTorWKB` seems good; useful. The surrounding code to support `Projected[Geometry]` as well as `Geometry` further radicalizes me towards the view that `Projected` isn't a container that has geometries in it but a property of the geometries themselves...

PR #2640, opened 2018-04-25
Improve performance of attribute store queries

### LINKED ISSUE #2529: Overzooming has performance issues for many catalogs with many layers
For Overzooming, https://github.com/locationtech/geotrellis/blob/9881296a667667c8a6a44758c012b4e985b1b574/spark/src/main/scala/geotrellis/spark/io/OverzoomingValueReader.scala#L36 will be very slow for catalogs with many, many layers. We should optimize this.

jpolchlo:
## Overview

As noted in #2529, querying `AttributeStore` for catalogs of many layers can be slow.  This PR introduces some new interface hooks to that class to allow backends to specialize the functions to gain the greatest benefit.  This PR provides an implementation of the new functions into the S3 backend.  These changes are also pushed into `OverzoomingValueReader`.

### Checklist

### Notes

Closes #2529


jpolchlo:
We do, if it makes sense for the particular backend.  I have no experience with the other backends to do the job.  Need to make a separate issue to make sure someone who knows better will do it later.

pomadchin:
Agree with you. I'll take a look and if it makes sense will create a separate issue / will create a separate PR. And this will merge in a while.

pomadchin:
@jbouffard have you ever seen some similar issues with Hadoop backend?

jbouffard:
@pomadchin I've never noticed it being slow, but I'm not sure how many layers would be needed to start seeing performance degradation. It might be worth it to implement this PR for the Hadoop backend.

PR #2643, opened 2018-04-26
Fix issues with simple feature building

### LINKED ISSUE #2575: GeoTools SimpleFeature ID handling
Working with GeoTools is important for all kinds of library interop on the JVM. Unfortunately, we seem to be mishandling the conversion of GT features/geometries to GeoTools' `SimpleFeature`. The argument to `buildFeature` should almost certainly be exposed (rather than simply set to `"SimpleFeature"`) to allow users to set unique identifiers where necessary: https://github.com/locationtech/geotrellis/blob/a00c35b928e96083188d91734252d66574e3b4a7/geotools/src/main/scala/geotrellis/geotools/GeometryToSimpleFeature.scala#L77

As is evident from the relevant section of the GeoTools API, the `buildFeature` is supposed to take a feature ID: http://docs.geotools.org/latest/javadocs/org/geotools/feature/simple/SimpleFeatureBuilder.html#buildFeature-java.lang.String-

jpolchlo:
## Overview

Simple features were being built with a constant ID field.  This PR allows the user to specify that ID or to accept a randomly generated ID.  We were also storing geometry into the nonstandard field "geometry".  This has been updated to store geometry in the standard "the_geom" field.

### Checklist

### Notes

Closes #2575

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -48,6 +48,8 @@ Fixes
 - The regex used to match files for the ``HadoopLayerAttributeStore`` and ``FileLayerAttributeStore`` has been
   expanded to include more characters.
 - ``HadoopAttributeStore.availableAttributes`` has been fixed so that it'll now list all attribute files.
+- Allow for simple features to be generated with a specified or random id with geometry stored in the standard
+  field, "the_geom"
 
 1.2.1
 _____
--- a/geomesa/src/main/scala/geotrellis/geomesa/geotools/GeometryToGeoMesaSimpleFeature.scala
+++ b/geomesa/src/main/scala/geotrellis/geomesa/geotools/GeometryToGeoMesaSimpleFeature.scala
@@ -35,7 +35,7 @@ import com.typesafe.config.ConfigFactory
 object GeometryToGeoMesaSimpleFeature {
 
   val whenField  = "when"
-  val whereField = "where"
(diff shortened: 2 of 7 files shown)


jpolchlo:
There is a remaining open question on this PR as to how to update the `geomesa` project to be consistent.  At the moment, we assign geometry to the `where` field at https://github.com/locationtech/geotrellis/blob/master/geomesa/src/main/scala/geotrellis/geomesa/geotools/GeometryToGeoMesaSimpleFeature.scala#L38.  I have no idea who wrote this code (prior to locationtech changeover), and now I do not know if this is how GeoMesa expects its simple features to be constructed.  Any insight @pomadchin @jnh5y ?

pomadchin:
Hey @jpolchlo these are our internal fields, you can choose a better name for it: required for bidirectional transformation between GT Geometry and GM Geometry: https://github.com/locationtech/geotrellis/blob/master/geomesa/src/test/scala/geotrellis/geomesa/geotools/GeoMesaSimpleFeatureType.scala

pomadchin:
@jpolchlo do you think GeoMesa project should be updated too?

jpolchlo:
What would you like to see it the changeset for the geomesa project?  I've changed the field name for the geometry to be more standardized, and it already does the right thing for the handling of feature ids.  That seems to be the scope of this PR, but I'm happy to entertain other ideas.

PR #2647, opened 2018-04-27
Replace custom Functor class with Cats' Functor class

### LINKED ISSUE #2047: Adopt usage of community Functor in API
We have erroneously written our own `Functor` as illustrated by: https://github.com/locationtech/geotrellis/pull/1948

We should pick either Scalaz or cats Functor implementation and use that instead. Since the implicit type class parameter is threaded all the way to user facing API there is no way to make this change without breaking the binary compatibility of user facing API, placing this change at `2.0.0` version.

Some things we may need to consider when making this decision:
 - API friendliness of Scalaz vs cats we are likely to use
 - If it makes sense to depend on both Scalaz and cats (seems doubtful)
 - If not, should stream based collection reading be based on Scalaz?

jpolchlo:
## Overview

We had implemented a custom Functor interface.  It makes no sense to rely on a custom implementation of a typeclass that is defined by multiple standard libraries.  After deciding to accept Cats as a dependency, we can move to that package to supply those standard typeclasses.  This PR thus moves to the Cats-supplied Functor class.

### Checklist

- [X] `docs/CHANGELOG.rst` updated, if necessary

### Notes

Closes #2047


spark/src/main/scala/geotrellis/spark/filter/ToSpatial.scala
```diff
@@ -34,17 +36,16 @@ object ToSpatial {
     * to define additional constraints on Metadata. M should
     * depend on a K type (M[K]), and two type classes should
     * be provided: [[geotrellis.util.Component]], to extract key bounds
-    * from M[K], and [[geotrellis.util.Functor]] to map
-    * M[K] to M[SpatialKey].
+    * from M[K], and cats.Functor to map M[K] to M[SpatialKey].
     *
     * For those reading the source code directly,
-    * {{{K: λ[α => M[α] => Functor[M, α]]: λ[α => Component[M[α], Bounds[α]]}}}
+    * {{{K: λ[α => M[α] => Functor[M]]: λ[α => Component[M[α], Bounds[α]]}}}
```
pomadchin: Functor is already passed simpler, so this line can be replaced by `{{{K: λ[α => Component[M[α], Bounds[α]]}}}`.
And below you can remove `ev1: M[K] => Functor[M]`. That intention here was to show what is `λ` here in types.

PR #2650, opened 2018-04-30
Fix conflicting GeoTiff reader flags

### LINKED ISSUE #2086: GeoTiff option for "decompress" conflicts with "streaming" in a strange way
In 1.1, if we were to read a GeoTiff in with the "streaming" option, but also setting the "decompress" option, and then crop, the whole geotiff would be read in (since decompress turns the geotiff tile into an ArrayTile):

```scala
val gt = GeoTiffReader.readMultiband("input.tif", decompress = true, streaming = true)
gt.crop(extent) // doesn't do a smart window read, since the tile is already an ArrayTile
```

pomadchin:
## Overview

The thing is that the decompression flag is conflicting with a streaming flag. The solution is to allow "decompression" only in case of not streaming read.

### Checklist

Closes #2086

### DIFF
--- a/raster/src/test/scala/geotrellis/raster/io/geotiff/BitGeoTiffMultibandTileSpec.scala
+++ b/raster/src/test/scala/geotrellis/raster/io/geotiff/BitGeoTiffMultibandTileSpec.scala
@@ -43,50 +43,6 @@ class BitGeoTiffMultibandTileSpec extends FunSpec
 
     it("should combine all bands with pixel interleave, striped") {
       val tile =
-        MultibandGeoTiff.compressed(p("striped", "pixel")).tile
-
-      val actual = tile.combine(_.sum % 2)
-      val expected = BitConstantTile(0, tile.cols, tile.rows)
-
-      assertEqual(actual, expected)
-    }
-
-    it("should combine all bands with pixel interleave, tiled") {
-      val tile =
-        MultibandGeoTiff.compressed(p("tiled", "pixel")).tile
-
-      val actual = tile.combine(_.sum % 2)
-      val expected = BitConstantTile(0, tile.cols, tile.rows)
-
-      assertEqual(actual, expected)
-    }
-
-    it("should combine all bands with band interleave, striped") {
-      val tile =
-        MultibandGeoTiff.compressed(p("striped", "band")).tile
-
-      val actual = tile.combine(_.sum % 2)
-      val expected = BitConstantTile(0, tile.cols, tile.rows)
-
-      assertEqual(actual, expected)
-    }
-
-    it("should combine all bands with band interleave, tiled") {
-      val tile =
-        MultibandGeoTiff.compressed(p("tiled", "band")).tile
-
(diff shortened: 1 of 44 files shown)


echeipesh:
For some reason I thought that we would decompress the segments and keep them as segments. Looks like we never did that, we always just called `.toArrayTile`.  Given that is true and the issue above I think we actually need to go all the way and remove `decompress` as an option everywhere.

The only question if it should become a method on `GeoTiff`.
Right now I guess you would be able to:
```scala
val tiff: SinglebandGeoTiff = ???
tiff.copy(tile = tiff.tile.toArrayTile)
```
That will work on `GeoTiff[T]` as well because it has home grown `copy` method.

My vote would be to not add `decompress` method until there is actually a need for that implementation and avoid growing the API for now.

pomadchin:
@echeipesh i see no sense in this decompress option at the moment too, mb there were some thoughts behind it by @lossyrob ? 

I'd like to have the thought about `tiff.copy(tile = tiff.tile.toArrayTile)` somewhere in order not to forget about it if we'll have this case someday

echeipesh:
I think mentioning it in `CHANGELOG` is good enough. If its really strange for somebody they should see it in RC process. I'm guessing nobody actually relies on that functionality though.

pomadchin:
@echeipesh i had to add an explicit `decompress` function into companion objects as it was very convenient for tests, any comments are appreciated!

raster/src/test/scala/geotrellis/raster/io/geotiff/BitGeoTiffMultibandTileSpec.scala
```diff
@@ -87,7 +87,7 @@ class BitGeoTiffMultibandTileSpec extends FunSpec
 
     it("should combine all bands with pixel interleave, striped") {
       val tile =
-        MultibandGeoTiff(p("striped", "pixel")).tile
+        MultibandGeoTiff.decompressed(p("striped", "pixel")).tile
```
echeipesh: Not trying to be difficult here but do you think this is not suitable substitute
`MultibandGeoTiff(p("striped", "pixel")).tile.toArrayTile`

pomadchin: ah it won't work, `toArrayTile` is in [GeoTiffTile](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffTile.scala#L690)

pomadchin: Ah you're genius and I'm blind!

pomadchin:
@echeipesh looking at how travis tries to work with compressed tiffs by default i think i see some sense in this decompression flag o:

pomadchin:
We'll think about `decompression` function name during docs update.

PR #2651, opened 2018-04-30
Add TMS reader for raster case [WIP] / [Blocked]

### LINKED ISSUE #2143: TmsLayerReader
There are situations where we want to read files from a TMS-style layer (`/{z}/{x}/{y}`), e.g. the [terrain dataset on aws](https://aws.amazon.com/public-datasets/terrain/). We should be able to read a query-able layer reader for layers from that style layout, for anything we can read from binary files.

moradology:
## Overview

This PR adds the ability to read TMS layers to the raster and the spark packages. Compilation won't work until we break tiling out of `Spark` (`ZoomedLayoutScheme` needs to be accessible from within `Raster`)

### Checklist

- [x] Raster case
- [ ] Spark case

Closes #2143

### DIFF
--- a/spark/build.sbt
+++ b/spark/build.sbt
@@ -5,7 +5,7 @@ libraryDependencies ++= Seq(
   sparkCore % Provided,
   hadoopClient % Provided,
   "com.google.uzaygezen" % "uzaygezen-core" % "0.2",
-  "org.scalaj" %% "scalaj-http" % "2.3.0",
+  "org.scalaj" %% "scalaj-http" % "2.4.0",
   avro,
   spire,
   monocleCore, monocleMacro,
--- a/raster/build.sbt
+++ b/raster/build.sbt
@@ -3,6 +3,7 @@ import Dependencies._
 name := "geotrellis-raster"
 
 libraryDependencies ++= Seq(
+  "org.scalaj" %% "scalaj-http" % "2.4.0",
   typesafeConfig,
   jts,
   spire,
--- a/raster/src/main/scala/geotrellis/raster/io/tms/TMSReader.scala
+++ b/raster/src/main/scala/geotrellis/raster/io/tms/TMSReader.scala
@@ -1,0 +1,76 @@
+package geotrellis.raster.io
+
+import geotrellis.raster._
+import geotrellis.vector.{Point, Extent}
+import geotrellis.spark.tiling.ZoomedLayoutScheme
+
+import spire.syntax.cfor._
+import scalaj.http._
+
+import scala.collection.mutable._
+import java.net.{URI, URL}
+import javax.imageio._
+
+class TMSReader[T](
+  uriTemplate: String,
+  f: URI => T = TMSReader.decodeTile _,
+  tileSize: Int = ZoomedLayoutScheme.DEFAULT_TILE_SIZE
+) extends Serializable {
+  def uri(z: Int, x: Int, y: Int): URI = new URI(
+    uriTemplate.replace("{z}", z.toString)
+      .replace("{x}", x.toString)
+      .replace("{y}", y.toString)
+  )
+
+  def read(zoom: Int, extent: Extent): Iterator[((Int, Int), T)] = {
+    val layoutDefinition = ZoomedLayoutScheme.layoutForZoom(zoom, extent, tileSize)
+    val bounds = layoutDefinition.mapTransform.extentToBounds(extent)
+    // construct bounds from zoom + extent
+    for ((x, y) <- bounds.coordsIter)
+    yield (x, y) -> f(uri(zoom, x, y))
+  }
+
+  def read(zoom: Int, x: Int, y: Int): T = f(uri(zoom, x, y))
+}
+
+object TMSReader {
+  import geotrellis.raster.render._
+
+  // ImageIO use here to deal with different decoding behaviors (png vs jpg)
+  def decodeTile(uri: URI): Tile = {
+    val img = ImageIO.read(uri.toURL)
+    val height = img.getHeight()
+    val width = img.getWidth()
+    val rgbArr = img.getRGB(0, 0, width, height, null, 0, width)
+    IntArrayTile(rgbArr, width, height, None)
+  }
+
+  // ImageIO use here to deal with different decoding behaviors (png vs jpg)
+  def decodeMultibandTile(uri: URI): MultibandTile = {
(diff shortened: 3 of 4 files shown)


pomadchin:
Can we put it into spark sub project? Until the break up time.

pomadchin:
@moradology can you rebase it on top of master?

project/Version.scala
```diff
@@ -15,7 +15,7 @@
  */
 
 object Version {
-  val geotrellis  = "2.0.0" + Environment.versionSuffix
+  val geotrellis  = "2.0.0-M2"
```
pomadchin: I believe it's a typo

spark/build.sbt
```diff
@@ -5,7 +5,7 @@ libraryDependencies ++= Seq(
   sparkCore % Provided,
   hadoopClient % Provided,
   "com.google.uzaygezen" % "uzaygezen-core" % "0.2",
-  "org.scalaj" %% "scalaj-http" % "2.3.0",
+  "org.scalaj" %% "scalaj-http" % "2.4.0",
```
pomadchin: Makes  sense to move it into `Dependencies` file? Sounds like a good idea to move all deps there (an issue for a separate PR).

spark/src/main/scala/geotrellis/spark/io/tms/TMSReader.scala
```diff
@@ -0,0 +1,84 @@
...
+}
+
+object TMSReader {
+
+  def apply[T](uriTemplate: String, f: URI => T = TMSReader.decodeTile _): TMSReader[T] =
+    new TMSReader(uriTemplate, f)
+
+  // ImageIO use here to deal with different decoding behaviors (png vs jpg)
+  def decodeTile(uri: URI): Tile = {
+    val img = ImageIO.read(uri.toURL)
```
pomadchin: It's a `png` / `jpg` reader only?

spark/src/main/scala/geotrellis/spark/io/tms/TMSReader.scala
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  // ImageIO use here to deal with different decoding behaviors (png vs jpg)
+  def decodeMultibandTile(uri: URI): MultibandTile = {
+    val img = ImageIO.read(uri.toURL)
+    val height = img.getHeight()
+    val width = img.getWidth()
+    val length = width * height
+
+    val rgbArr = img.getRGB(0, 0, width, height, null, 0, width)
```
pomadchin: It is two nested loops:

```java
for (int y = startY; y < startY+h; y++, yoff+=scansize) {
  off = yoff;
  for (int x = startX; x < startX+w; x++) {
    rgbArray[off++] = colorModel.getRGB(raster.getDataElements(x, y, data));
   }
}
```

Would not it be more efficient instead of doing it just to have two nested loops and to get pixels by col / row directly from memory without this extra copy step?

```scala
cfor(0)(_ < length, _ + 1) { c =>
  cfor(0)(_ < length, _ + 1) { r =>
  val c = img.getRGB(c, r)
  reds(cellIdx) = cell.red.toByte
  // ...
  }
}
```

spark/src/main/scala/geotrellis/spark/io/tms/TMSReader.scala
```diff
@@ -0,0 +1,84 @@
...
+      reds :+ cell.red.toByte
+      greens :+ cell.green.toByte
+      blues :+ cell.blue.toByte
+      alphas :+ cell.alpha.toByte
+    }
+
+    val redTile = ByteArrayTile(reds.toArray, width, height, None)
+    val greenTile = ByteArrayTile(greens.toArray, width, height, None)
+    val blueTile = ByteArrayTile(blues.toArray, width, height, None)
+    val alphaTile = ByteArrayTile(alphas.toArray, width, height, None)
```
pomadchin: Hm, there is this conversion, it's an extra `copyToArray` (actually here it would be just a while loop to iterate over all elements of an `ArrayBuffer`), mb just to use `Array` instead of `ArrayBuffer`?

pomadchin:
Looks like we're not ready to include it into 2.0 release, probably later, mb next `RC`.

project/Version.scala
metasim: Also: you should consider creating a custom sbt key called `versionSuffix`, and then you can just change that on the command line **or** in the sbt console, instead of relying on an environment variable.

spark/build.sbt
metasim: Does this need a QC?

Also: don't we already use akka-http client somewhere?

spark/src/main/scala/geotrellis/spark/io/tms/TMSReader.scala
metasim: Is any other format supported by TMS?
http://mapserver.org/mapcache/formats.html

spark/src/main/scala/geotrellis/spark/io/tms/TMSReader.scala
metasim: I vote for using `cfor` just for consistency with other loops in GT.

spark/src/main/scala/geotrellis/spark/io/tms/TMSReader.scala
metasim: I also suggest rewriting this just to use `Array[Byte]`, e.g.:

```scala
val reds = Array.ofDim[Byte](length)
... 
cfor(0)(_ < length, _ + 1) { cellIdx =>
   val cell = rgbArr(cellIdx)
   reds(cellIdx) = cell.red.toByte
...
}
...
val redTile = ArrayTile(reds, width, height)
```

spark/src/main/scala/geotrellis/spark/io/tms/TMSReader.scala
pomadchin: @metasim that's what i wanted to see actually :D thanks for reference!

spark/build.sbt
pomadchin: CQ: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=16213

Nope, we only used scalaj-http; i believe akka-http was somewhere in tests.

metasim:
@echeipesh A RasterSource version would be 💣 💥. Curious how passing query extents would work.

PR #2653, opened 2018-05-01
Implement persistence functions for Hadoop & File UCOGs

pomadchin:
## Overview

Implement persistence functions for unstructured COGs.

@notthatbreezy if you can make a review of this PR it would be great.

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/io/file/geotiff/FileIMGeoTiffAttributeStore.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/file/geotiff/FileIMGeoTiffAttributeStore.scala
@@ -13,14 +13,12 @@ object FileIMGeoTiffAttributeStore {
   def apply(
     name: String,
     uri: URI
-  ): InMemoryGeoTiffAttributeStore = {
-    def getDataFunction = () => HadoopGeoTiffInput.list(name, uri, new Configuration())
-    new InMemoryGeoTiffAttributeStore(() => GeoTiffMetadataTree.fromGeoTiffMetadataList(getDataFunction())) {
+  ): InMemoryGeoTiffAttributeStore =
+    new InMemoryGeoTiffAttributeStore {
+      lazy val metadataList = HadoopGeoTiffInput.list(name, uri, new Configuration())
       def persist(uri: URI): Unit = {
-        val data = getDataFunction()
-        val str = data.toJson.compactPrint
+        val str = metadataList.toJson.compactPrint
         new PrintWriter(uri.toString) { write(str); close }
       }
     }
-  }
 }
--- a/spark/src/main/scala/geotrellis/spark/io/file/geotiff/FileIMGeoTiffAttributeStore.scala
+++ b/spark/src/main/scala/geotrellis/spark/io/file/geotiff/FileIMGeoTiffAttributeStore.scala
@@ -1,14 +1,26 @@
 package geotrellis.spark.io.file.geotiff
 
 import geotrellis.spark.io.hadoop.geotiff._
-import org.apache.hadoop.conf.Configuration
 
+import org.apache.hadoop.conf.Configuration
+import spray.json._
+import spray.json.DefaultJsonProtocol._
+
+import java.io.PrintWriter
 import java.net.URI
 
 object FileIMGeoTiffAttributeStore {
   def apply(
     name: String,
     uri: URI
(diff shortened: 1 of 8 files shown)


spark/src/main/scala/geotrellis/spark/io/hadoop/geotiff/InMemoryGeoTiffAttributeStore.scala
```diff
@@ -3,7 +3,7 @@ package geotrellis.spark.io.hadoop.geotiff
 import geotrellis.vector.ProjectedExtent
 import java.net.URI
 
-case class InMemoryGeoTiffAttributeStore(getData: () => GeoTiffMetadataTree[GeoTiffMetadata]) extends CollectionAttributeStore[GeoTiffMetadata] {
+abstract class InMemoryGeoTiffAttributeStore(getData: () => GeoTiffMetadataTree[GeoTiffMetadata]) extends CollectionAttributeStore[GeoTiffMetadata] {
```
echeipesh: Making this class `abstract` makes sense, why not move `getData` into abstract method in that case, just like `persist` ?

echeipesh: Its kind of weird that `GeoTiffMetadataTree` is exposed as part of the interface here. Does it need to be. Perhaps it can just be used privately to build an index from incoming list of `GeoTiffMetadata`

spark/src/main/scala/geotrellis/spark/io/hadoop/geotiff/HadoopIMGeoTiffAttributeStore.scala
```diff
@@ -1,13 +1,28 @@
...
 
 object HadoopIMGeoTiffAttributeStore {
   def apply(
     name: String,
     uri: URI,
     conf: Configuration = new Configuration()
-  ): InMemoryGeoTiffAttributeStore =
-    InMemoryGeoTiffAttributeStore(() => GeoTiffMetadataTree.fromGeoTiffMetadataList(HadoopGeoTiffInput.list(name, uri, conf)))
+  ): InMemoryGeoTiffAttributeStore = {
+    def getDataFunction = () => HadoopGeoTiffInput.list(name, uri, conf)
```
echeipesh: Kind of strange that we're mixing in anonymous trait here to define `getDataFunction`. We already have `HadoopIMGeoTiffAttributeStore` name, why not make it a boring class that just extends `HadoopIMGeoTiffAttributeStore`?

spark/src/main/scala/geotrellis/spark/io/hadoop/geotiff/InMemoryGeoTiffAttributeStore.scala
pomadchin: @echeipesh it's more an oversight (`GeoTiffMetadataTree`)

spark/src/main/scala/geotrellis/spark/io/file/geotiff/FileIMGeoTiffAttributeStore.scala
```diff
@@ -1,14 +1,26 @@
...
 object FileIMGeoTiffAttributeStore {
   def apply(
     name: String,
     uri: URI
-  ): InMemoryGeoTiffAttributeStore =
-    InMemoryGeoTiffAttributeStore(() => GeoTiffMetadataTree.fromGeoTiffMetadataList(HadoopGeoTiffInput.list(name, uri, new Configuration())))
+  ): InMemoryGeoTiffAttributeStore = {
+    def getDataFunction = () => HadoopGeoTiffInput.list(name, uri, new Configuration())
+    new InMemoryGeoTiffAttributeStore(() => GeoTiffMetadataTree.fromGeoTiffMetadataList(getDataFunction())) {
+      def persist(uri: URI): Unit = {
```
echeipesh: `GeoTiffMetadataTree.fromGeoTiffMetadataList` asking for `List` is too restrictive, `Seq` or `TraversableOnce` is better, although I can't seem to consistently decide if `TraversableOnce` is preferable.

echeipesh:
@pomadchin What do you think about these comments, do you want to handle them here or split into follow-up PR. The changes here are strictly more-better, so I think they can be merged.

PR #2657, opened 2018-05-02
Remove old deprecations [WIP]

moradology:
This is a work in progress PR in which I'm collecting changes in response to deprecation warnings from the 1.x releases.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/GridBounds.scala
+++ b/raster/src/main/scala/geotrellis/raster/GridBounds.scala
@@ -76,11 +76,7 @@ case class GridBounds(colMin: Int, rowMin: Int, colMax: Int, rowMax: Int) {
   def width = colMax - colMin + 1
   def height = rowMax - rowMin + 1
 
-  @deprecated("This will return a `Long` in 2.0. Until then, sizeLong may be more accurate.", "1.2")
-  def size: Int = width * height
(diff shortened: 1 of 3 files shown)


raster/src/main/scala/geotrellis/raster/GridBounds.scala
```diff
@@ -76,11 +76,7 @@ case class GridBounds(colMin: Int, rowMin: Int, colMax: Int, rowMax: Int) {
   def width = colMax - colMin + 1
   def height = rowMax - rowMin + 1
 
-  @deprecated("This will return a `Long` in 2.0. Until then, sizeLong may be more accurate.", "1.2")
-  def size: Int = width * height
-
-  // TODO Mark for deprecation in 3.0 when 2.0 comes out!
-  def sizeLong: Long = width.toLong * height.toLong
```
fosskers: Is it okay to remove `sizeLong` without deprecating first?

pomadchin:
@moradology want to point here again, that this PR is made against 2.0 branch and not against master branch :)

raster/src/main/scala/geotrellis/raster/GridBounds.scala
moradology: Not at all

PR #2660, opened 2018-05-03
Introduce Pyramid class

### LINKED ISSUE #2654: Updates to Pyramid API
This issue captures a discussion about rehabbing pyramid API in GeoTrellis, which suffers from couple of surprising behaviors.

First notion is to restrict pyramiding to always be "power of two, 2 to 1 reduction" concept. This assumption cuts out the use of the `LayoutScheme` and affords a much simpler way to build pyramids:
```scala
val layer: TileLayerRDD[SpatialKey] = ???

val pyramid: List[(Int, TileLayerRDD[SpatialKey])] = 
    layer.pyramid(max = 15, min = 10, resampleMethod = ???)
// Things I need to know:
// -  We always use LocalLayoutScheme
// Must:
// - The first element in the list is the input layer
// Question:
// - What is Int? When do I start, when do I finish?
//  - We can ask for min/max as part of the input and just range between them
//  - We estimate the 0-zoom level and count back from that
// - When do I stop "pyramiding"
//  - When we have "some resolution"
//  - When we have a single tile in a pyramid

/** Additional issue:
How do you perform resampling on categorial layer when pyramiding?
- You can not do Bilinear because that will not respect classes
- NearestNeighbor is not deterministic because all source pixels are equidistant

Possibly: List[Int] => Int is a resample method we are able to specify.
*/
```
The current `Pyramid.up` and `Pyramid.upLevels` worker methods can remain for advanced usages. Ultimately the above API will have to use them with restricted parameter list. 

Also on the docket is the addition of `Pyramid` type:
```scala
trait Pyramid[K, V] extends Seq[(Int, RDD[(K, V)]]

object Pyramid {
    def apply(levels: Map[Int, TilelayerRDD[K, V]]) = ??? // is there sanity checking? No.
}

val pyramid: Pyramid[SpatialKey, Tile] = 
    layer.pyramid(max = 15, min = 10, resampleMethod = ???)

pyramid.persist(MEMORY_AND_DISK_SER)
val minZoom = pyramid.minZoom
val maxZoom = pyramid.maxZoom
val layer: TileLayerRDD[SpatialKey] = pyramid.level(0)
val layer: TileLayerRDD[SpatialKey] = pyramid(0)
val tile: Tile = pyramid.lookup(zoom = 8, key = SpatialKey(23,23))

// Lets never do this:
val otherPyramid = pyramid + 1
// Maybe?
val otherPyramid = pyramid.mapTiles(_ + 1)

/** Danger: 
Pyramid RDDs are "recursive", asking for them in the wrong order can be suboptimal
We can use the type to take care of this common case:
*/
pyramid.write("my-layer", layerWriter)
// ^ Iterate over the pyramid in right way and generate LayerIds 
pyramid.foreach { case (zoom, rdd) =>
    layerWriter(LayerId("my-layer", zoom), rdd)
}

/** Bonus:
We can abstract over "saved" pyramids and "built" pyramids.
Also Pyramid could be the type that abstracts over COG vs Avro pyramid layers
*/

val otherPyramid = Pyramid.read(layerReader, "my-other-layer", max = 15, min = 10)
// ^ layer reader to READ the RDDs into the layer
// Why, now we can use built and saved pyramids in the same context

val cogPyramid = Pyramid.read(cogLayerReader, "my-other-layer", max = 15, min = 10)

// Pick the closest zoom level and interplate from that to desired CellSize
val layer = cogPyramid.resample(CellSize(4,4))

def doWork(a: Pyramid[K, V], b: Pyramid[K, V]) = ???
doWork(pyramid, otherPyramid)
```

jpolchlo:
## Overview

GeoTrellis creates, as one of its core functions, multi-resolution, hierarchical level-of-detail structures that we commonly call pyramids.  Despite their frequent appearance in our workflows, we have been representing them as `List[(Int, RDD[(K, V)] with Metadata[M])]`.  This PR aims to grant pyramids first-class citizenship by creating a wrapper class that streamlines common operations on those structures.

Closes #2654 

### Checklist

- [x] ~`docs/CHANGELOG.rst` updated, if necessary~
- [x] ~`docs` guides update, if necessary~
- [x] ~New user API has useful Scaladoc strings~

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/pyramid/Pyramid.scala
+++ b/spark/src/main/scala/geotrellis/spark/pyramid/Pyramid.scala
@@ -17,6 +17,10 @@
 package geotrellis.spark.pyramid
 
 import geotrellis.spark._
+import geotrellis.spark.io._
+import geotrellis.spark.io.avro._
+import geotrellis.spark.io.index.KeyIndexMethod
+import geotrellis.spark.io.json._
 import geotrellis.spark.tiling._
 import geotrellis.raster._
 import geotrellis.raster.merge._
@@ -27,8 +31,27 @@ import geotrellis.vector.Extent
 
 import org.apache.spark.Partitioner
 import org.apache.spark.rdd._
+import org.apache.spark.storage.StorageLevel
+import spray.json._
 
 import scala.reflect.ClassTag
+
+case class Pyramid[
+    K: SpatialComponent: ClassTag,
+    V <: CellGrid: ClassTag: ? => TilePrototypeMethods[V]: ? => TileMergeMethods[V],
+    M: Component[?, LayoutDefinition]: Component[?, Bounds[K]]
+](levels: Map[Int, RDD[(K, V)] with Metadata[M]]) {
+  def apply(level: Int): RDD[(K, V)] with Metadata[M] = levels(level)
+
+  def level(l: Int): RDD[(K, V)] with Metadata[M] = levels(l)
+
+  def lookup(zoom: Int, key: K): Seq[V] = levels(zoom).lookup(key)
+
+  def minZoom = levels.keys.min
+  def maxZoom = levels.keys.max
+
+  def persist(storageLevel: StorageLevel) = levels.mapValues{ _.persist(storageLevel) }
+}
 
 object Pyramid extends LazyLogging {
   case class Options(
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/geotrellis/spark/pyramid/Pyramid.scala
```diff
@@ -42,6 +65,69 @@ object Pyramid extends LazyLogging {
...
+    V <: CellGrid: ? => TilePrototypeMethods[V]: ? => TileMergeMethods[V]: AvroRecordCodec: ClassTag,
+    M: JsonFormat: Component[?, Bounds[K]]: Component[?, LayoutDefinition]
+  ](layerName: String, maxZoom: Int, minZoom: Int, layerReader: LayerReader[LayerId]): Pyramid[K, V, M] = {
+    val seq = for (z <- maxZoom to minZoom by -1) yield {
+      (z, layerReader.read[K, V, M](LayerId(layerName, z)))
+    }
+    Pyramid[K, V, M](seq.toMap)
+  }
+
+  def write[
```
echeipesh: Tempting to put this on `FilteringLayerReader` trait if we're adding this type as first class citizen.

spark/src/main/scala/geotrellis/spark/pyramid/Pyramid.scala
```diff
@@ -42,6 +65,69 @@ object Pyramid extends LazyLogging {
...
+  }
+
+  def apply[
+    K: SpatialComponent: ClassTag,
+    V <: CellGrid: ClassTag: ? => TilePrototypeMethods[V]: ? => TileMergeMethods[V],
+    M: Component[?, LayoutDefinition]: Component[?, Bounds[K]]
+  ](rdd: RDD[(K, V)] with Metadata[M],
+    startZoom: Int,
+    endZoom: Int,
+    options: Options
```
echeipesh: I would suggest unrolling `Options` into `ResampleMethod` and `Partitioner`.
Having the `Options` object solves the problem of bringing a lot of configurations from   boring to spark context but in spark if we're only adding `Partitioner` its more clear to have that as its own option.

I've followed this intuition in rasterize API for reference. What do you think of that approach?

spark/src/main/scala/geotrellis/spark/pyramid/Pyramid.scala
```diff
@@ -42,6 +65,69 @@ object Pyramid extends LazyLogging {
...
+    K: AvroRecordCodec: Boundable: JsonFormat: ClassTag: SpatialComponent,
+    V <: CellGrid: ? => TilePrototypeMethods[V]: ? => TileMergeMethods[V]: AvroRecordCodec: ClassTag,
+    M: JsonFormat: Component[?, Bounds[K]]: Component[?, LayoutDefinition]
+  ](pyramid: Pyramid[K, V, M], layerName: String, writer: LayerWriter[LayerId], keyIndexMethod: KeyIndexMethod[K]) = {
+    for (z <- pyramid.maxZoom to pyramid.minZoom by -1) {
+      writer.write[K, V, M](LayerId(layerName, z), pyramid(z), keyIndexMethod)
+    }
+  }
+
+  def apply[
```
echeipesh: Suggestion for name `fromLayerRDD`, like https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/io/cog/COGLayer.scala#L48

Trying to minimize non-obvious applies.

spark/src/main/scala/geotrellis/spark/pyramid/Pyramid.scala
```diff
@@ -42,6 +65,69 @@ object Pyramid extends LazyLogging {
     implicit def methodToOptions(m: ResampleMethod): Options = Options(resampleMethod = m)
   }
 
+  def read[
```
echeipesh: Suggestion: `fromLayerReader`

spark/src/main/scala/geotrellis/spark/pyramid/Pyramid.scala
```diff
@@ -42,6 +65,69 @@ object Pyramid extends LazyLogging {
...
+    options: Options
+  ): Pyramid[K, V, M] = {
+    val gridBounds = rdd.metadata.getComponent[Bounds[K]].asInstanceOf[KeyBounds[K]].toGridBounds
+    val maxDim = math.max(gridBounds.width, gridBounds.height).toDouble
+    val levels = math.ceil(math.log(maxDim)/math.log(2)).toInt
+
+    apply(rdd, 0, levels, options)
+  }
+
+  def apply[
```
echeipesh: I think these overloads can be actually collapsed into a single method with defaults.

spark/src/main/scala/geotrellis/spark/pyramid/Pyramid.scala
```diff
@@ -42,6 +81,40 @@ object Pyramid extends LazyLogging {
...
+    V <: CellGrid: ClassTag: ? => TilePrototypeMethods[V]: ? => TileMergeMethods[V],
+    M: Component[?, LayoutDefinition]: Component[?, Bounds[K]]
+  ](rdd: RDD[(K, V)] with Metadata[M],
+    thisZoom: Option[Int] = None,
+    endZoom: Option[Int] = None,
+    resampleMethod: ResampleMethod = Average,
+    partitioner: Option[Partitioner] = None
+  ): Pyramid[K, V, M] = {
+    val opts = Options(resampleMethod, partitioner)
+    val gridBounds = rdd.metadata.getComponent[Bounds[K]].asInstanceOf[KeyBounds[K]].toGridBounds
```
pomadchin: Bad cast, the solution here is to use pattern match

spark/src/main/scala/geotrellis/spark/pyramid/Pyramid.scala
```diff
@@ -27,9 +31,44 @@ import geotrellis.vector.Extent
...
+  def maxZoom = levels.keys.max
+
+  def persist(storageLevel: StorageLevel) = levels.mapValues{ _.persist(storageLevel) }
+
+  def write(
+    layerName: String,
+    writer: LayerWriter[LayerId],
+    keyIndexMethod: KeyIndexMethod[K]
+  )(
+    implicit arcK: AvroRecordCodec[K],
```
pomadchin: Interesting how would it work in the client code.

jpolchlo: Not sure I understand what is meant by this comment...

jpolchlo:
Question.  And maybe @echeipesh can weigh in on this, but should Pyramid live in `geotrellis.spark.pyramid` package?  Dependency hell and all?

spark/src/main/scala/geotrellis/spark/pyramid/Pyramid.scala
```diff
@@ -42,6 +81,43 @@ object Pyramid extends LazyLogging {
     implicit def methodToOptions(m: ResampleMethod): Options = Options(resampleMethod = m)
   }
 
+  def fromLayerReader[
+    K: AvroRecordCodec: Boundable: JsonFormat: ClassTag: SpatialComponent,
+    V <: CellGrid: ? => TilePrototypeMethods[V]: ? => TileMergeMethods[V]: AvroRecordCodec: ClassTag,
+    M: JsonFormat: Component[?, Bounds[K]]: Component[?, LayoutDefinition]
+  ](layerName: String, maxZoom: Int, minZoom: Int, layerReader: LayerReader[LayerId]): Pyramid[K, V, M] = {
```
jpolchlo: I don't like this interface.  Ideally, we can specify min and max zoom as `Option`s.  `None` values mean to pull the list of available zooms from the attribute store ...... except that `LayerReader`'s interface doesn't specify `val attributeStore`.  Happy to make that change to clean up this method.  Take the onus off the user to know the range of zooms.

jpolchlo:
For the sake of expediency, the documentation requirements on this PR are being deferred into the next RC of 2.0.

spark/src/test/scala/geotrellis/spark/pyramid/PyramidSpec.scala
```diff
@@ -200,5 +201,39 @@ class PyramidSpec extends FunSpec with Matchers with TestEnvironment {
...
+          3, 3,  3, 3,   4, 4,  4, 4,
+
+          3, 3,  3, 3,   4, 4,  4, 4,
+          3, 3,  3, 3,   4, 4,  4, 4
+        ) , 8, 8)
+
+      val baseLayer = createTileLayerRDD(tile, tileLayout)
+
+      // Build pyramid using Average resampling
+      val pyramid = Pyramid.fromLayerRdd(baseLayer)
```
pomadchin: Looks sweet...

PR #2662, opened 2018-05-08
Improved stitching facilities

### LINKED ISSUE #2549: stitch the croped TileLayerRDD crashed
I'm trying to use the TileLayerRDD's crop method and stitch the crop result to write a single TIFF result.
This is the test code:
```
val tileRDD = sc.hadoopMultibandGeoTiffRDD(path)
val (_, metadata) = tileRDD.
      collectMetadata[SpatialKey](FloatingLayoutScheme(256))
val tiledRdd: RDD[(SpatialKey, MultibandTile)] = tileRDD.tileToLayout[SpatialKey](metadata)
val tileLayerRDD: MultibandTileLayerRDD[SpatialKey] = MultibandTileLayerRDD(tiledRdd, metadata)
val cropedRDD = tileLayerRDD.crop(Extent(531500.0, 7735000.0, 531600.0, 7735200.0))
val stitched = cropedRDD.stitch()
val geotiff = GeoTiff(stitched, cropedRDD.metadata.crs)
val pathHadoop = new Path(pathR)
geotiff.write(pathHadoop, sc.hadoopConfiguration)
```
and got crash like this:
```
20160
java.lang.ArrayIndexOutOfBoundsException: 20160
    at geotrellis.raster.UByteRawArrayTile.update(UByteArrayTile.scala:68)
    at geotrellis.raster.MutableArrayTile$class.set(MutableArrayTile.scala:56)
    at geotrellis.raster.UByteArrayTile.set(UByteArrayTile.scala:28)
    at geotrellis.raster.MutableArrayTile$class.update(MutableArrayTile.scala:89)
    at geotrellis.raster.UByteArrayTile.update(UByteArrayTile.scala:28)
    at geotrellis.raster.stitch.Stitcher$MultibandTileStitcher$$anonfun$stitch$4$$anonfun$apply$1.apply$mcVI$sp(Stitcher.scala:102)
    at scala.collection.immutable.Range.foreach$mVc$sp(Range.scala:160)
    at geotrellis.raster.stitch.Stitcher$MultibandTileStitcher$$anonfun$stitch$4.apply(Stitcher.scala:101)
    at geotrellis.raster.stitch.Stitcher$MultibandTileStitcher$$anonfun$stitch$4.apply(Stitcher.scala:100)
    at scala.collection.TraversableLike$WithFilter$$anonfun$foreach$1.apply(TraversableLike.scala:733)
    at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
    at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
    at scala.collection.TraversableLike$WithFilter.foreach(TraversableLike.scala:732)
    at geotrellis.raster.stitch.Stitcher$MultibandTileStitcher$.stitch(Stitcher.scala:100)
    at geotrellis.raster.stitch.Stitcher$MultibandTileStitcher$.stitch(Stitcher.scala:80)
    at geotrellis.spark.stitch.TileLayoutStitcher$.stitch(StitchRDDMethods.scala:49)
    at geotrellis.spark.stitch.SpatialTileLayoutRDDStitchMethods.stitch(StitchRDDMethods.scala:58)
```
I fount note "Makes the assumption that all tiles are of equal size such that they can be placed in a grid layout." in StitchRDDMethods.scala, and i read the code found it uses the first tile's col/row count to calculate, but after crop the tile's size isn't same any more.

I hope the TileLayerRDD' stitch method can stitch tiles with different size, as long as they are in one layout.Then, my code above will work.

btw: 
as @pomadchin's advice, i tried "clamp = false" crop and stitch, and the result is quite wrong, a screenshot here alt.

I also using @pomadchin's filter code tested "filter by extent intersection, stitch and crop the result stitch" , this works good.
but i still prefer to use TileLayerRDD's accurate crop and stitch if it's possible.
I suppose, maybe it's performance will be better than "filter by extent intersection, stitch and crop the result stitch" ?

jpolchlo:
## Overview

Currently, we are able to stitch collections/layouts of tiles only when those tiles are all of the same dimension.  In some cases, such as after a call to `crop`, the tiles in a layout are not all of equal size, but we may still wish to stitch them without having to pad the pieces out to some common size.  This PR provides this ability, so long as all tiles obey a simple rule: tiles in a given column (row) have the same width (height).

Fixes #2549 

### Checklist


jpolchlo:
@frophie: I think this PR will resolve the issue you've been having.  Can you test your code using the `.genericStitch()` method extension (rather than just a plain `stitch()`)?

spark/src/main/scala/geotrellis/spark/stitch/StitchRDDMethods.scala
```diff
@@ -49,6 +52,42 @@ object TileLayoutStitcher {
...
+    *
+    * @param    tiles       A traversable collection of (key, tile) pairs
+    * @return               The stitched tile
+    */
+  def genericStitch[
+    V <: CellGrid: ? => TilePrototypeMethods[V]: ? => TileMergeMethods[V]
+  ](tiles: Traversable[(Product2[Int, Int], V)]) = {
+    assert(tiles.size > 0, "Cannot stitch empty collection")
+
+    val colWidths = MuMap.empty[Int, Int]
```
pomadchin: usually its called `mutable.Map`

raster/src/main/scala/geotrellis/raster/merge/SinglebandTileMergeMethods.scala
```diff
@@ -27,65 +27,49 @@ import spire.syntax.cfor._
   * single-band [[Tile]]s.
   */
 trait SinglebandTileMergeMethods extends TileMergeMethods[Tile] {
-  /** Merges this tile with another tile.
```
echeipesh: Feels like we should add a delegate with `baseCol = 0, baseRow = 0` to be a little more conservative with the interface.

jpolchlo: Is [this](https://github.com/jpolchlo/geotrellis/blob/3b799a6b3774243a3608c4ae70a56af74db33f57/raster/src/main/scala/geotrellis/raster/merge/TileMergeMethods.scala#L38) not sufficient?

echeipesh: Huh, missed that in the diff somehow, that's totally sufficient.

Frophie:
Thanks for your work first~
I found `genericStitch` works good with default option `clamp = true`, but got an AssertionError with `clamp = false`, maybe this is another bug about `crop`.
```
java.lang.AssertionError: assertion failed: Tiles in a layout column must have the same width
  at scala.Predef$.assert(Predef.scala:170)
  at geotrellis.spark.stitch.TileLayoutStitcher$$anonfun$genericStitch$2.apply(StitchRDDMethods.scala:72)
  at geotrellis.spark.stitch.TileLayoutStitcher$$anonfun$genericStitch$2.apply(StitchRDDMethods.scala:70)
  at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
  at scala.collection.mutable.WrappedArray.foreach(WrappedArray.scala:35)
  at geotrellis.spark.stitch.TileLayoutStitcher$.genericStitch(StitchRDDMethods.scala:70)
  at geotrellis.spark.stitch.SpatialTileRDDStitchMethods.genericStitch(StitchRDDMethods.scala:111)
```

Another thing is `stitch` returns a `Raster`, but `genericStitch` returns a `Tile`.

jpolchlo:
Thanks @Frophie.  You're right, the problem appears to be in `crop`.  This PR does its job, though.  I'll start a new PR when I figure out how to fix `crop`.

@pomadchin: can I get a review and merge on this?

spark/src/main/scala/geotrellis/spark/stitch/StitchRDDMethods.scala
```diff
@@ -49,6 +52,42 @@ object TileLayoutStitcher {
...
+  /**
+    * Stitches a collection of spatially-keyed tiles into a single tile.  Assumes that all
+    * tiles in a given column (row) have the same width (height).
+    *
+    * @param    tiles       A traversable collection of (key, tile) pairs
+    * @return               The stitched tile
+    */
+  def genericStitch[
+    V <: CellGrid: ? => TilePrototypeMethods[V]: ? => TileMergeMethods[V]
+  ](tiles: Traversable[(Product2[Int, Int], V)]) = {
```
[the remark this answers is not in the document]
pomadchin: @jpolchlo can you add here a return type?

spark/src/main/scala/geotrellis/spark/stitch/StitchCollectionMethods.scala
```diff
@@ -25,17 +27,27 @@ import geotrellis.util._
...
-    Raster(tile, mapTransform(bounds))
+    val (tile, (kx, ky), (offsx, offsy)) = TileLayoutStitcher.stitch(self)
+    val layout = self.metadata.getComponent[LayoutDefinition]
+    val mapTransform = layout.mapTransform
+    val nwTileEx = mapTransform(kx, ky)
+    val base = nwTileEx.southEast
+    val (ulx, uly) = (base.x - offsx.toDouble * layout.cellwidth, base.y + offsy * layout.cellheight)
+    Raster(tile, Extent(ulx, uly - tile.rows * layout.cellheight, ulx + tile.cols * layout.cellwidth, uly))
+
+    // val (tile, bounds) = TileLayoutStitcher.stitch(self)
```
pomadchin: Do we need this code, and do we need `Stitcher` type class at the moment?

jpolchlo: `Stitcher` is fairly well embedded in the code base and has a bit different of a usage pattern (specifying pixel locations of tiles) and so is more general.  In fact, the argument could be made that I should have used that object's methods to implement this PR (which wouldn't change the signature).  I can still do that, and probably should, actually.

PR #2667, opened 2018-05-14
Move GeoTrellis GeoMesa to use pureconfig

pomadchin:
## Overview

We depend on `pureconfig`, so we can standardize over the method to access params.
Update for the vector package is in #2619.


metasim:
@pomadchin Can you clarify the relationship between pureconfig and GeoMesa? In [this dialog](https://gitter.im/locationtech/geomesa?at=5af9a1a6f04ce53632d760bc) the GM team wasn't aware of it.

pomadchin:
hey @metasim, it's a geotrellis thing not related to geomesa at all. i decided to standartize the way we handle config files as it was a total mess: https://github.com/locationtech/geotrellis/pull/2652

PR #2668, opened 2018-05-14
Fix problems with tile layer cropping when `clamp=false`

jpolchlo:
## Overview

Cropping of RDD tile layers with `clamp=false` caused problems because the cropping region wasn't restricted to the current spatial key, resulting in tiles that have a partial intersection with the crop region producing tiles that span the entire crop region.

Note: This PR is based on #2662 and also resolves part of the problem from issue #2549 

### Checklist


pomadchin:
Ah it looks like it should be rebased on top of #2662

pomadchin:
@Frophie cna you have a look into this PR + stitch PR for your case?

PR #2670, opened 2018-05-17
Generate missing headers

pomadchin:
## Overview

This PR adds missing headers for all new files.

- [x] headers
- [x] travis ci fix
- [ ] ~~deps inclusion hygiene~~

### DIFF
--- a/.travis/build-and-test-set-1.sh
+++ b/.travis/build-and-test-set-1.sh
@@ -1,11 +1,11 @@
 #!/bin/bash
 
-./sbt -J-Xmx2G "++$TRAVIS_SCALA_VERSION" "project doc-examples" compile  || { exit 1; }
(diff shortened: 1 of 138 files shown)


pomadchin:
Also SBT 1.x introduced a regression in loading secondary `build.sbt` files from sub projects. It looks like we need to move all settings into a single `build.sbt` file.

pomadchin:
There is also a thread: https://contributors.scala-lang.org/t/proposal-deprecation-of-other-sbt-styles-outside-of-top-level-build-sbt-file/930/13 related to thoughts to deprecate `build.sbt` in sub projects. Also it looks like it's not a popular pattern across popular projects, so probably we need to refactor and to remove it. Projects to look into - akka and circe.

pomadchin:
I'll move proposal of the project structure without `build.sbt` files into a separate PR as it probably can have some discussion.

PR #2673, opened 2018-05-18
Refactor sub projects SBT settings structure

pomadchin:
## Overview

SBT 1.x introduced a regression in loading secondary `build.sbt` files from sub projects. It looks like we need to move all settings into a single `build.sbt` file.

There is also a thread: https://contributors.scala-lang.org/t/proposal-deprecation-of-other-sbt-styles-outside-of-top-level-build-sbt-file/930/13 related to thoughts to deprecate `build.sbt` files in sub projects. Also it looks like it's not a popular pattern across popular projects, so probably we need to refactor and to remove it. Projects to look into - `akka` and `circe`.

This is an attempt to demonstrate what can happen if we'll remove all `build.sbt` files. Any comments and thoughts are appreciated.


metasim:
@pomadchin I'm surprised you had to do this, as I've not had problems with sub-project in RasterFrames. Do you have information on what the regression was?

As far as sbt styles go, despite the linked thread, I doubt we could find consensus. My gut says to avoid changing GT too much just so the team doesn't have to go through the cognitive effort of getting re-aligned. I also worry about preemptively changing structure while there's no clear winner in approach.

In RF I have an `AutoPlugin` called `ProjectPlugin` that handles everything that's common across all modules, including some custom keys for things like `geoTrellisVersion`. I also declare in the `autoImport` section `ModuleID` symbols for common libraries so all the versions are anchored. Then in the individual `build.sbt` files its usually just `libraryDependencies`, `name`, and perhaps some tweaking of another option or two. I also have a `ShadeSupportPlugin` for sub-projects that want to offer up an `sbt-assembly` version.

I'd be interested in knowing what the specific sbt error you were having was.

pomadchin:
@metasim nothing really bad, the thing is that sbt loads `build.sbt` files from subprojects really slow; the thing was that smth probably changed on the travis side (probably temp change / bug / feature) and our builds could timeout randomly; for instance one run could finish in 41/ 43 minutes; the next could go out of 49 minutes. I definitely remember that switch from sbt 0.13 slow downed builds a bit. 

The solution was not to create a separate sbt processes per a single test [link](https://github.com/locationtech/geotrellis/blob/master/.travis/build-and-test-set-2.sh).

I am open to any suggestions and this issue not urgent at the moment, as travis seems to be happy.

I'd say that we can postpone it at the moment. My general observation was that usage of `build.sbt` files in subprojects is not a popular solution.

pomadchin:
Yes, during this refactor i also noticed that i am defnitely not sure that i want to go from `build.sbt` files at least for now. I'll close this PR and probably will create an issue about it and we'll move disucssion there.

PR #2676, opened 2018-05-22
S3LayerHeader Reading Fix

jbouffard:
## Overview

This PR fixes a bug that caused some `S3LayerHeader`s to be improperly read.

### DIFF
--- a/s3/src/main/scala/geotrellis/spark/io/s3/S3LayerHeader.scala
+++ b/s3/src/main/scala/geotrellis/spark/io/s3/S3LayerHeader.scala
@@ -53,10 +53,10 @@ object S3LayerHeader {
             key,
             layerType.convertTo[LayerType]
           )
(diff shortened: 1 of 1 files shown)


s3/src/main/scala/geotrellis/spark/io/s3/S3LayerHeader.scala
```diff
@@ -53,10 +53,10 @@ object S3LayerHeader {
             key,
             layerType.convertTo[LayerType]
           )
-        case Seq(JsString(keyClass), JsString(bucket), JsString(key), JsString(layerType)) =>
+        case Seq(JsString(keyClass), JsString(valueClass), JsString(bucket), JsString(key)) =>
```
pomadchin: It won’t work, you need another .fromFields call here

PR #2684, opened 2018-05-29
Crop function works properly with Extents out of tiff bounds

pomadchin:
## Overview

Fixes GeoTiff crop by extent function behaviour.

### Checklist

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
+++ b/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
@@ -529,10 +529,13 @@ abstract class GeoTiffMultibandTile(
   /**
    * Crop this tile to given pixel region.
    *
-   * @param gridBounds Pixel bounds specifying the crop area.
+   * @param bounds Pixel bounds specifying the crop area.
    */
- def crop(gridBounds: GridBounds): ArrayMultibandTile =
-  crop(List(gridBounds)).next._2
+ def crop(bounds: GridBounds): ArrayMultibandTile = {
+   val iter = crop(List(bounds))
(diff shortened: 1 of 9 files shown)


raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
```diff
@@ -529,10 +529,13 @@ abstract class GeoTiffMultibandTile(
...
    * Crop this tile to given pixel region.
    *
-   * @param gridBounds Pixel bounds specifying the crop area.
+   * @param bounds Pixel bounds specifying the crop area.
    */
- def crop(gridBounds: GridBounds): ArrayMultibandTile =
-  crop(List(gridBounds)).next._2
+ def crop(bounds: GridBounds): ArrayMultibandTile = {
+   val iter = crop(List(bounds))
+   if (iter.isEmpty) throw new reader.MalformedGeoTiffException(s"No intersections of ${bounds} vs ${gridBounds}")
```
echeipesh: Strange that this is not `GeoAttrsError`, is it because this function may be called during initial reading of GeoTiff ?

pomadchin: Yeah and it probably makes sense, i think it was an idea to separate raster exceptions from geotiff exceptions: https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffTile.scala#L734

echeipesh: Ah, I see it was to match what `GeoTiffTile` already does. Thats a strange exception to throw in either case, it should be `GeoAttrsError` unless it breaks something truly surprising in the tests.

PR #2689, opened 2018-05-31
Spark Version 2.3.0

echeipesh:
Upgrades the spark dependency to 2.3.0.
Driving reason aside from keeping current is to use improved ORC file support from Spark 2.3.0 in https://github.com/geotrellis/vectorpipe/pull/54


project/Environment.scala
```diff
@@ -21,6 +21,6 @@ object Environment {
     Properties.envOrElse(environmentVariable, default)
 
   lazy val hadoopVersion  = either("SPARK_HADOOP_VERSION", "2.8.0")
```
pomadchin: Let's bump Hadoop up to `2.8.3`?

PR #2696, opened 2018-06-13
Prevent re-preparing already prepared statement in CassandraLayerReader

### LINKED ISSUE #2695: CassandraValueReader re-prepares queries - Cassandra Driver complains
When reading tiles from the CassandraValueReader a previously prepared statement is re-prepared.  The Datastax Cassandra driver logs a warning message as a result since it keeps a prepared statement cache and preparing statements takes a non-zero amount of CPU time. 

I believe the problematic code is here: https://github.com/locationtech/geotrellis/blob/master/cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraValueReader.scala#L47
Note how the statement is re-prepared each time `read` is called. 

Best practice is to not re-prepare the statement and instead cache it.  Generally so long as you're not switching the instance of the Session used to execute the PreparedStatement, this caching is safe.  Near as I can tell with your implementation of `CassandraInstance` it should be the same session each time used to execute the PreparedStatement even if the Session had to be created at first on a new Spark Executor.

ALPSMAC:
Signed-off-by: Andrew Polack <polack.andy@gmail.com> // <andrew.polack@navy.mil>

## Overview

Fixes issue #2695 

In short, prevents the re-preparing of an already prepared statement which is a Cassandra anti-pattern. 

### Checklist

- [N/A] `docs/CHANGELOG.rst` updated, if necessary
- [N/A] `docs` guides update, if necessary
- [N/A] New user API has useful Scaladoc strings
- [?] Unit tests added for bug-fix or new feature

I could use some help figuring out how to test this... generally the symptom is a warning that appears in the log emitted from the Cassandra Datastax Java Driver if a query is re-prepared. 

I ran the `cassandra` SBT project test suite and all tests passed so... ¯\_(ツ)_/¯

### Notes

If the Session object is somehow pooled and a different Session is provided than the one originally used to prepare the statement then this fix could cause issues.  

If this turns out to be a valid fix it would be great to have it back-ported into the released 1.2.x stuff.

Thanks Much!
-Andy

Closes #2695

### DIFF
--- a/cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraValueReader.scala
+++ b/cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraValueReader.scala
@@ -43,15 +43,17 @@ class CassandraValueReader(
     val writerSchema = attributeStore.readSchema(layerId)
     val codec = KeyValueRecordCodec[K, V]
 
-    def read(key: K): V = instance.withSession { session =>
-      val statement = session.prepare(
+    private lazy val statement = instance.withSession{ session =>
+      session.prepare(
         QueryBuilder.select("value")
           .from(header.keyspace, header.tileTable)
           .where(eqs("key", QueryBuilder.bindMarker()))
           .and(eqs("name", layerId.name))
           .and(eqs("zoom", layerId.zoom))
       )
+    }
 
+    def read(key: K): V = instance.withSession { session =>
       val row = session.execute(statement.bind(keyIndex.toIndex(key): BigInteger)).all()
       val tiles = row.map { entry =>
           AvroEncoder.fromBinary(writerSchema, entry.getBytes("value").array())(codec)


cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraValueReader.scala
```diff
@@ -43,15 +43,17 @@ class CassandraValueReader(
     val writerSchema = attributeStore.readSchema(layerId)
     val codec = KeyValueRecordCodec[K, V]
 
-    def read(key: K): V = instance.withSession { session =>
-      val statement = session.prepare(
+    private lazy val statement = instance.withSession{ session =>
```
pomadchin: Do you think it should be private?

ALPSMAC: I debated... I saw that none of the other instance fields were, but if someone accesses this `statement` from the outside and uses a different `Session` than the one originally used to prepare it then that would cause an exception.  I ended up deciding to keep the field hidden to prevent potential misuse. 
If the project standard is to mark instance fields public though I don't have a problem changing it... just wasn't sure what the standard was.

ALPSMAC:
@pomadchin - trying to suppress warnings mostly, but I've seen some evidence from Cassandra maintainers (e.g. https://datastax-oss.atlassian.net/browse/JAVA-236?focusedCommentId=14736&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14736 ) that there is good reason they issue this warning.  

I also know there's a PreparedStatement cache setting in `cassandra.yaml`: 

```yaml
# Maximum size of the native protocol prepared statement cache
#
# Valid values are either "auto" (omitting the value) or a value greater 0.
#
# Note that specifying a too large value will result in long running GCs and possbily
# out-of-memory errors. Keep the value at a small fraction of the heap.
#
# If you constantly see "prepared statements discarded in the last minute because
# cache limit reached" messages, the first step is to investigate the root cause
# of these messages and check whether prepared statements are used correctly -
# i.e. use bind markers for variable parts.
#
# Do only change the default value, if you really have more prepared statements than
# fit in the cache. In most cases it is not neccessary to change this value.
# Constantly re-preparing statements is a performance penalty.
#
# Default value ("auto") is 1/256th of the heap or 10MB, whichever is greater
prepared_statements_cache_size_mb:
```

which also seems to indicate that re-preparing statements is not a recommended thing to do. 

Thanks!

-Andy

pomadchin:
Ha, thanks for references.

pomadchin:
Sry that touching you again @ALPSMAC, but have you signed ECA? https://github.com/locationtech/geotrellis/blob/master/docs/CONTRIBUTING.rst#eclipse-contributor-agreement-eca

ALPSMAC:
Yes - but I think I may have inadvertently signed the commit with the wrong email or something.  Sorry - first time trying to contribute by signing a commit.

The email I've got registered on the machine I'm working from in my .gitconfig is my work email.  My personal email is polack.andy@gmail.com if you're looking to match it against the ECA.

ALPSMAC:
@pomadchin I'm not exactly sure how to fix this PR so that it's signed by my email address that matches the one I used in the ECA.  Can you suggest what I should do?

pomadchin:
@ALPSMAC I think that `git commit --amend` and manual fix of the email in the message would do the right thing.

ALPSMAC:
@pomadchin sorry for the delay in getting back to you - other work responsibilities caught up with me.  

Okay - I think I have amended the commit properly... hopefully that worked :-) 

Let me know if it still doesn't look appropriately signed to you.  Thanks!

pomadchin:
Thank you, it works!

PR #2712, opened 2018-06-20
[DOCS] Update core concepts guide

### LINKED ISSUE #2679: Document the Pyramid class
We need a discussion of pyramiding in general, and of the `Pyramid` class (introduced in #2660) specifically.

jpolchlo:
## Overview

The core concepts guide was somewhat messy, and omitted some important topics.  This PR fills in some of those gaps and makes other changes for style and clarity.

Closes #2679

### Checklist

### DIFF
--- a/docs/guide/core-concepts.rst
+++ b/docs/guide/core-concepts.rst
@@ -13,8 +13,8 @@ alike. If GIS is brand, brand new to you, `this
 document <https://www.gislounge.com/what-is-gis/>`__ is a useful high
 level overview.
 
-Basic Terms
-===========
+Glossary
+========
 
 -  **Tile:** A grid of numeric *cells* that represent some data on the
    Earth.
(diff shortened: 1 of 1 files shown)


docs/guide/core-concepts.rst
```diff
@@ -13,8 +13,8 @@ alike. If GIS is brand, brand new to you, `this
 document <https://www.gislounge.com/what-is-gis/>`__ is a useful high
 level overview.
 
-Basic Terms
-===========
+Glossary
+========
```
jbouffard: I think it might be helpful to have links to each of the sections where these topics are discussed. That way you don't have to spend a lot of time trying to find it.

docs/guide/core-concepts.rst
```diff
@@ -13,8 +13,8 @@ alike. If GIS is brand, brand new to you, `this
...
 
 -  **Tile:** A grid of numeric *cells* that represent some data on the
    Earth.
```
jbouffard: I can't comment on the exact line since it wasn't changed, but I think that `Layer` should be `Tile Layer` since that's how it's referred to throughout the document. With a mention that a `Tile Layer` can also be called a `Layer`

docs/guide/core-concepts.rst
```diff
@@ -43,334 +43,399 @@ These definitions are expanded upon in other sections of this document.
...
+*extent* to raster data to indicate the position and scope of the raster data.
+
+Geotrellis provides a number of operations on these basic data types.  One may
+reproject, resample, crop, merge, combine, and render raster data; vector data
+may be manipulated in some limited ways, but is mostly used to generate raster
+data.
+
+The following packages contain the relevant code:
+
+- ``geotrellis.proj4``
```
jbouffard: It might be helpful to have these linked to their respective page in the docs.

docs/guide/core-concepts.rst
```diff
@@ -43,334 +43,399 @@ These definitions are expanded upon in other sections of this document.
...
-.. code-block:: scala
+Geometries in Geotrellis are exclusively point sets and piecewise linear
+representations.  A collection of points may be connected by a chain of linear
+segments into more complex shapes, and then aggregated into collections.  All
+such classes derive from the base ``geotrellis.vector.Geometry`` class.  The
+geometry subclasses are as follows:
 
-    /* Zoom level 13 */
-    val layerId = LayerId("myLayer", 13)
+- ``geotrellis.vector.Point``
```
jbouffard: Same as the above comment about linking these to their respective docs

docs/guide/core-concepts.rst
```diff
@@ -43,334 +43,399 @@ These definitions are expanded upon in other sections of this document.
...
+-  Querying a specific tile value
+-  Rescaling, resampling, cropping
 
-   <hr>
+Working with Cell Values
+------------------------
 
-Cell Types
-==========
+Tiles contain numerical data.  These can be of the form of integers, floats,
```
jbouffard: `Tiles` should be probably be in-line quoted to match its other occurrences in the file. To in-line quote a word with something that comes directly after (like an `s`) you can do: \``Tile\``\s

docs/guide/core-concepts.rst
```diff
@@ -558,435 +576,592 @@ possible, Constant ``NoData`` values are to be preferred. For
...
+position corresponding to the upper-left corner of the extent; the `x`
+coordinate increases toward the right, and the `y` coordinate increases moving
+down (into lower latitude values, say).
+
+.. |LayoutDefinition| replace:: ``geotrellis.spark.tiling.LayoutDefinition``
+.. _LayoutDefinition: https://geotrellis.github.io/scaladocs/latest/#geotrellis.spark.tiling.LayoutDefinition
+
+Thus far, we've described how an ``RDD[(K, V)]`` plus a ``LayoutDefinition``
+can be used to represent a large, distributed raster (when ``[K:
+SpatialComponent]``).  To solidify this notion, Geotrellis has a notion of a
```
jbouffard: Like with `LayoutDefinition`, having a subsection for `Metadata[M]` might be useful.

docs/guide/core-concepts.rst
```diff
@@ -558,435 +576,592 @@ possible, Constant ``NoData`` values are to be preferred. For
...
+``GridExtent`` subclass defined with an ``Extent`` and ``CellSize``.  The
+``Extent`` is subdivided into a grid of uniform, rectangular regions.  The
+size and number of the sub-regions is determined using the ``CellSize`` of the
+``LayoutDefinition``, and then the pixel dimensions of the constituent tiles.
+The sub-regions are then assigned a ``SpatialKey`` with the ``(0, 0)``
+position corresponding to the upper-left corner of the extent; the `x`
+coordinate increases toward the right, and the `y` coordinate increases moving
+down (into lower latitude values, say).
+
+.. |LayoutDefinition| replace:: ``geotrellis.spark.tiling.LayoutDefinition``
```
jbouffard: It might be good to have `LayoutDefinition` be its own subsection since the `Layout Definition` link in the index brings the user to the `Layouts and Tile Layers` section.

docs/guide/core-concepts.rst
```diff
@@ -558,435 +576,592 @@ possible, Constant ``NoData`` values are to be preferred. For
...
 
-.. code-block:: scala
+   <hr>
 
-    val pointIntersection = (Point(1.0, 2.0) & Point(1.0, 2.0)).as[Point]
-    val pointNonIntersection = (Point(1.0, 2.0) & Point(12.0, 4.0)).as[Point]
 
-    assert(pointIntersection == Some(Point(1.0, 2.0)))  // Either some point
-    assert(pointNonIntersection == None)                // Or nothing at all
+Pyramids
```
jbouffard: It might be good to have this listed and link in the glossary.

docs/guide/core-concepts.rst
```diff
@@ -558,435 +576,592 @@ possible, Constant ``NoData`` values are to be preferred. For
...
-.. code::
+   <hr>
 
-    scala> import geotrellis.vector._
-    scala> p1 & p2 match {
-         |   case PointResult(_) => println("A Point!)
-         |   case NoResult => println("Sorry, no result.")
-         | }
-    A Point!
+Catalogs & Tile Layer IO
```
jbouffard: Like `Pyramid`, having `Catalogs` in the glossary could be helpful.

docs/guide/core-concepts.rst
```diff
@@ -1152,77 +1327,18 @@ depending on your architecture.
...
-
-In essence, a ``LayoutDefinition`` is the minimum information required
-to describe the tiling of some map's area in Geotrellis. The
-``LayoutDefinition`` class extends ``GridExtent``, and exposes methods
-for querying the sizes of the grid and grid cells. Those values are
-stored in the ``TileLayout`` (the grid description) and ``CellSize``
-classes respectively. ``LayoutDefinition``\ s are most often encountered
-in raster reprojection processes.
 
 Map Algebra
```
jbouffard: Same as above comment.

docs/guide/core-concepts.rst
```diff
@@ -13,364 +13,489 @@ alike. If GIS is brand, brand new to you, `this
...
+  region.
+- `Feature <#features>`_: A *Geometry* with some associated metadata.
+- `Cell <#working-with-cell-values>`_: A single unit of data in some
+  grid.
+- `Tile <#tiles-and-rasters>`_: A grid of numeric *cells* that represent
+  some data on the Earth.
+- `Raster <#rasters>`_: A *Tile* with an *Extent*.
+- **RDD:** "Resilient Distributed Datasets" from `Apache Spark <http://spark.apache.org/>`__. Can be thought of as a distributed Scala ``Seq``.
+- `Key <#keys>`_: Used to index a *Tile* in a grid (or cube) of them.
+- `Key Index <#key-indexes>`_: A means to transform higher-dimensional
```
jbouffard: A minor thing, but it might be helpful to have these topics listed in the order in which they appear in the document.

docs/guide/core-concepts.rst
```diff
@@ -13,364 +13,489 @@ alike. If GIS is brand, brand new to you, `this
...
+  *Keys* into a unique one-dimensional identifier.
+- `Layer <#layouts-and-tile-layers>`_ or **Tile Layer**: This is a grid
+  (or cube) of *Tiles*.
+- `Metadata <#layouts-and-tile-layers>`_ or **Layer Metadata**: A
+  descriptive structure that defines how to interpret a key-value store as a
+  *Layer*.
+- `Layout Definition <#layouts-and-tile-layers>`_: A structure that
+  relates *Keys* to geographic locations and vice versa.
```
jbouffard: Having `Vector Tiles` and `Type classes` listed in the glossary should be done as well. I'm not sure if the `GeoTiff` section needs to be added, but it wouldn't hurt.

docs/guide/core-concepts.rst
jpolchlo: Eh.  `Metadata` is kind of a nothing class that just means that an implementation has a `metadata` field of type `M`.  I don't want this document to balloon up by having sections for every little thing.

docs/guide/core-concepts.rst
jpolchlo: I think we should recall that this is a glossary and not a TOC.  The TOC will be in the sidebar on readthedocs.  I want to minimize the load of terms, so I'm restricting the glossary to ideas that are really fundamental to GT, and you couldn't just google for them.  Vector tiles are well-known, and easily discovered on the web.  I'd argue that, if anything, they aren't "core concepts", and maybe shouldn't be covered in these docs at all.

Type classes shouldn't be in the glossary because, again, it's a glossary, not a TOC.  The common GT typeclasses are not fundamental to understanding the system, and pretty inside baseball.  They could be pulled out into a separate doc, but this is probably as good a place for them as any.

GeoTiff could be in the glossary because of its importance, though it is an easily discoverable idea not endemic to GT.

I'm aiming to really put a cap on the amount of terms in the glossary.  I'd say the Key Index isn't that important and could actually be removed.

docs/guide/core-concepts.rst
jpolchlo: The `Layouts and Tile Layers` section is a fairly succinct description of LayoutDefinition and TileLayerMetadata, and it would, I think, unnecessarily bloat up the delivery of the concept to break it up into sections so that each entry of a glossary (which is not an index, and links are provided as more of a courtesy) can link to a unique section.  If there's an argument for restructuring that section into smaller sections to aid clarity, and it happens that the resulting structure makes for a 1-to-1 correspondence with glossary items, I'm happy to update the links, but I don't think having unique links is a good enough reason.

docs/guide/core-concepts.rst
jpolchlo: In this case, I feel like map algebra isn't fundamental enough or unique enough to GT that I want to clutter up the glossary, which should ultimately be a very quick way to get used to our terminology.

jbouffard: That's true. I can see how map algebra isn't that crucial or unique enough to list.

docs/guide/core-concepts.rst
jbouffard: I see your point. I think that section is pretty good as it is, so there probably is no reason to rewrite it at the moment. I'm not sure if it's possible, but I think that if you can link to the middle part of a section that would be best. Though, it's not really necessary.

moradology:
Looking good

Issue #1751, opened 2016-10-30
Cross compile/publish to 2.12


lossyrob:
We need to compile and publish to 2.12.

We should use this opportunity of having multiple travis builds also to break out Travis tests into more than 2 build sets, since build times are exceeding the time limit sometimes. (Seems like SET 1 is the culprit)

pomadchin:
Spark 2.12 support issue: https://issues.apache.org/jira/browse/SPARK-14220

lossyrob:
Ah, I thought it was already building on 2.12...

lossyrob:
Pushing to 1.1, as 2.12 is too fresh for now...

pomadchin:
A small update: `Spark 2.3.0` is expected to support `Scala 2.12`

pomadchin:
Update: hopefully `Scala 2.12` support would be by the end of this year.

metasim:
Interested in sharing lessons learned, as we go through this process.

pomadchin:
@metasim no real story would be behind it; GT already can be compiled against 2.12; the only bug / feature i faced is: https://github.com/locationtech/geotrellis/pull/2727/files#diff-b79f2573bbab4c81247de9ca82d2d6f8R31

looking forward to see spark 2.4 to create a proper PR, i expect some some possible changes in runtime.

pomadchin:
https://amplab.cs.berkeley.edu/jenkins/view/Spark%20QA%20Test/job/spark-master-test-maven-hadoop-2.7-ubuntu-scala-2.12/ :tada: first Spark 2.12 green builds!

pomadchin:
Release! http://central.maven.org/maven2/org/apache/spark/spark-core_2.12/2.4.0/  :tada:

PR #2727, opened 2018-06-28
Scala 2.12 support

pomadchin:
## Overview

This PR introduces a cross compilation Scala 2.12 support. Note, that it won't work on Spark, however would allow to use GeoTrellis in 2.12 libraries without having spark calls in the code. 

A better approach would be to have `spark-core` published against scala 2.12.

@jisantuc can you check, would it work for you?

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -45,7 +45,7 @@ object Dependencies {
   val fs2Core             = "co.fs2"                     %% "fs2-core"                 % "0.10.4"
   val fs2Io               = "co.fs2"                     %% "fs2-io"                   % "0.10.4"
 
-  val sparkCore           = "org.apache.spark"           %% "spark-core"               % Version.spark
+  val sparkCore           = "org.apache.spark"            % "spark-core_2.11"          % Version.spark
   val hadoopClient        = "org.apache.hadoop"           % "hadoop-client"            % Version.hadoop
 
   val avro                = "org.apache.avro"             % "avro"                     % "1.8.2"
--- a/vector/src/main/scala/geotrellis/vector/package.scala
+++ b/vector/src/main/scala/geotrellis/vector/package.scala
@@ -28,13 +28,13 @@ package object vector extends SeqMethods
     with triangulation.Implicits
     with voronoi.Implicits {
 
-  type PointFeature[D] = Feature[Point, D]
-  type LineFeature[D] = Feature[Line, D]
-  type PolygonFeature[D] = Feature[Polygon, D]
-  type MultiPointFeature[D] = Feature[MultiPoint, D]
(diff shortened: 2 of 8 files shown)


vector/src/main/scala/geotrellis/vector/package.scala
```diff
@@ -28,13 +28,13 @@ package object vector extends SeqMethods
...
     with voronoi.Implicits {
 
-  type PointFeature[D] = Feature[Point, D]
-  type LineFeature[D] = Feature[Line, D]
-  type PolygonFeature[D] = Feature[Polygon, D]
-  type MultiPointFeature[D] = Feature[MultiPoint, D]
-  type MultiLineFeature[D] = Feature[MultiLine, D]
-  type MultiPolygonFeature[D] = Feature[MultiPolygon, D]
-  type GeometryCollectionFeature[D] = Feature[GeometryCollection, D]
+  type PointFeature[+D] = Feature[Point, D]
```
echeipesh: @pomadchin Didn't see this hidden in here. This change actually makes sense. Curious what problem inspired it.

pomadchin: @echeipesh I think it’s 2.11 compiler bug / feature. The only change that needs to be done to support 2.12; mb it makes sense to create a separate PR with this changes in order to include it into 2.0

project/Dependencies.scala
```diff
@@ -45,7 +45,7 @@ object Dependencies {
   val fs2Core             = "co.fs2"                     %% "fs2-core"                 % "0.10.4"
   val fs2Io               = "co.fs2"                     %% "fs2-io"                   % "0.10.4"
 
-  val sparkCore           = "org.apache.spark"           %% "spark-core"               % Version.spark
+  val sparkCore           = "org.apache.spark"            % "spark-core_2.11"          % Version.spark
```
echeipesh: How could this produces a correct artifact for `geotrellis-spark`? Is it just succeeding in compiling something that won't actually run?

pomadchin: @echeipesh yea, the motivation of this PR was from the one hand - to check how gt at least compiled against 2.12, from the other to help RF team with 2.12 tests. From this PR we can see that all non sparky geotrellis parts have all tests passed.

pomadchin: So this PR allows to publish 2.12 gt artifacts, and all non spark part would work as expected. Dirty but working and tested. My vote would be to backport covariance fixes into master and to wait for spark 2.4 artifacts.

pomadchin:
Scala 2.12 compatibility was resolved via #2782, the rest of the work would be done after Spark 2.4 release in terms of #1751

PR #2757, opened 2018-07-06
readSubsetBands for COGValueReader and OverzoomingCOGValueReader

### LINKED ISSUE #2706: Structured COG band subsetting
Current multiband COG readers still produce all of the bands from the GeoTiff.
However, its possible that the user knows that useful information exists only in subset of the bands and wishes to read exclusively those layers. GeoTiffs that use band-interleave naturally lend themselves to this optimization.

The main concerns is how to deal with invalid band selection. Currently you get "all the bands", whatever that is, so there is no opportunity for errors. Given band subset of [1,2,100] its not desirable to fail the whole query if some of the tiffs do not have band 100. Two choices are fill missing bands with `NODATA` tiles or return `Array[Option[Tile]]`. The latter is more explicit and more efficient because avoids needing to call `.isNoDataTile` to verify query results. However its not consistent with current API. Part of the issue here is to think through the implications and possibly establish a new convention for when we querying for band subsets.

Although we are already in 2.0.0-RC1 this should be developed with intent to back-port the addition to 2.1 branch.

jbouffard:
## Overview

This PR adds the `readSubsetBands` method to `COGValueReader` and `OverzoomingCOGValueReader`. This method will allow users to read in only certain bands from a COG layer in whatever order they want.

### Checklist

### Demo

```scala
val reader: COGReader[SpatialKey, MultibandTile] = ???
val key: SpatialKey = ???

// Tries to read bands 2, 0, 1, 10 and return them in that order
reader.readSubsetBands(key, 2, 0, 1, 10)
```

Related to #2706


spark/src/main/scala/geotrellis/spark/io/cog/COGValueReader.scala
```diff
@@ -71,13 +72,69 @@ trait COGValueReader[ID] {
...
+      val targetBandsWithIndex: Array[(Int, Int)] =
+        bands
+          .zipWithIndex
+          .filter { case (band, _) =>
+            band >= 0 && band < sourceMultibandTile.bandCount
+          }
+          .toArray
+
+      val targetBands: Array[Int] = targetBandsWithIndex.map { _._1 }
+      val targetBandsIndexes: Array[Int] = targetBandsWithIndex.map { _._2 }
```
pomadchin: You can make it in one command: 

```scala
val (targetBands, targetBandsIndexes) = targetBandsWithIndex.unzip
```

spark/src/main/scala/geotrellis/spark/io/cog/COGValueReader.scala
```diff
@@ -71,13 +72,69 @@ trait COGValueReader[ID] {
...
+      val maxWidth = Index.digits(baseKeyIndex.toIndex(baseKeyIndex.keyBounds.maxKey))
+      val uri = fullPath(keyPath(key.setComponent(spatialKey), maxWidth, baseKeyIndex, zoomRange))
+
+      val sourceGeoTiff =
+        GeoTiffReader.readMultiband(uri, streaming = true)
+
+      // Have to convert the sourceGeoTiff to a GeoTiffMultibandTile
+      // in order to access the desired crop method.
+      val sourceMultibandTile =
+        GeoTiffMultibandTile(
```
pomadchin: I suppose It's possible to use pattern matching here, but this solution is also good.

spark/src/main/scala/geotrellis/spark/io/cog/COGValueReader.scala
```diff
@@ -71,13 +72,69 @@ trait COGValueReader[ID] {
       } catch {
         case th: Throwable => exceptionHandler(key)(th)
       }
+
+    }
+
+    def readSubsetBands(key: K, bands: Seq[Int]): Array[Option[Tile]] = {
```
pomadchin: wondering about how we can express `read` using this `readSubsetBands` method.

spark/src/test/resources/cog-layer/attributes/multiband-cog-layer__.__0__.__metadata.json
```diff
@@ -0,0 +1 @@
...keyIndexes":[[{"minZoom":0,"maxZoom":0},{"type":"zorder","properties":{"keyBounds":{"minKey":{"col":0,"row":0},"maxKey":{"col":0,"row":0}}}}],[{"minZoom":7,"maxZoom":7},{"type":"zorder","properties":{"keyBounds":{"minKey":{"col":92,"row":59},"maxKey":{"col":92,"row":60}}}}],[{"minZoom":6,"maxZoom":6},{"type":"zorder","properties":{"keyBounds":{"minKey":{"col":46,"row":29},"maxKey":{"col":46,"row":30}}}}],[{"minZoom":4,"maxZoom":4},{"type":"zorder","properties":{"keyBounds":{"minKey":{"col":11,"row":7},"maxKey":{"col":11,"row":7}}}}],[{"minZoom":8,"maxZoom":11},{"type":"zorder","properties":{"keyBounds":{"minKey":{"col":184,"row":118},"maxKey":{"col":185,"row":120}}}}],[{"minZoom":1,"maxZoom":1},{"type":"zorder","properties":{"keyBounds":{"minKey":{"col":1,"row":0},"maxKey":{"col":1,"row":0}}}}],[{"minZoom":2,"maxZoom":2},{"type":"zorder","properties":{"keyBounds":{"minKey":{"col":2,"row":1},"maxKey":{"col":2,"row":1}}}}],[{"minZoom":5,"maxZoom":5},{"type":"zorder","properties":{"keyBounds":{"minKey":{"col":23,"row":14},"maxKey":{"col":23,"row":15}}}}],[{"minZoom":3,"maxZoom":3},{"type":"zorder","properties":{"keyBounds":{"minKey":{"col":5,"row":3},"maxKey":{"col":5,"row":3}}}}]]}}]
```
pomadchin: So you decided not to zip this layer?

spark/src/main/scala/geotrellis/spark/io/cog/COGValueReader.scala
jbouffard: Oh, I didn't know about that method. That does make things easier to read

spark/src/main/scala/geotrellis/spark/io/cog/COGValueReader.scala
jbouffard: I was thinking something like:

```scala
def read(key: K, bands: Option[Seq[Int]] = None): K
```

Where `None` specifies all bands to be read in. That's still a rough idea, though, so I'd actually need to see if it can be implemented better.

spark/src/test/resources/cog-layer/attributes/multiband-cog-layer__.__0__.__metadata.json
jbouffard: Yeah, but only because that's what we seemed to be doing with all of the other test files in that directory

spark/src/main/scala/geotrellis/spark/io/cog/COGValueReader.scala
pomadchin: Yea, I'm thinking that probably there should be an overload that would convert this `Array[Option[Tile]]` into `V` type.

raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala
```diff
@@ -537,6 +537,20 @@ abstract class GeoTiffMultibandTile(
...
+   *
+   * @param bounds Pixel bounds specifying the crop area.
+   * @param  bandIndices       An array of band indexes.
+   *
+   */
+ def crop(bounds: GridBounds, bandIndices: Array[Int]): ArrayMultibandTile = {
+   val iter = crop(List(bounds), bandIndices)
+   if (iter.isEmpty) throw GeoAttrsError(s"No intersections of ${bounds} vs ${gridBounds}")
+   else iter.next._2
+ }
```
lossyrob: We can DRY out this  code, by having the above crop call this overload. That's the preferred pattern through the  library.

spark/src/main/scala/geotrellis/spark/io/cog/OverzoomingCOGValueReader.scala
```diff
@@ -45,16 +45,39 @@ trait OverzoomingCOGValueReader extends COGValueReader[LayerId] {
     lazy val baseReader = reader[K, V](layerId)
     lazy val maxReader = reader[K, V](LayerId(layerName, maxAvailableZoom))
 
-    def read(key: K): V =
+    private def deriveMaxKey(key: K): K = {
```
lossyrob: no need for private here, it's inside a method

spark/src/test/scala/geotrellis/spark/io/file/cog/COGFileSpatialSpec.scala
```diff
@@ -68,4 +68,27 @@ class COGFileSpatialSpec
...
+      val expected = {
+        val tile = valueReader.read(key)
+        Array(Some(tile.band(1)), Some(tile.band(0)), None, None)
+      }
+
+      for ((exp, act) <- expected.zip(actual)) {
+        exp should be (act)
+      }
+    }
+  }
```
lossyrob: Are we testing both pixel  and band interleave tiles?

jbouffard: That is more of general test to make sure that it works. I believe the `crop` logic has been tested with both band and pixel interleave though.

PR #2762, opened 2018-07-17
Added JPEG decompression

lossyrob:
## Overview

This is a WIP PR around adding JPEG decompression. I did some tests around implementing some techniques found in TwelveMonkeys, specifically https://github.com/haraldk/TwelveMonkeys/blob/master/imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/TIFFImageReader.java#L1113 .

I was able to read and render some OpenAerialMap COGs with this code, however unit tests against checking exact values have failed.

Putting this up here as to make folk aware of this work, and maybe  soliciting  some help?


metasim:
@lossyrob Did you want to commit your tests against this too?

metasim:
Here's what the absolute difference between the two first band bands looks like (inset is zoom).

screen shot 2018-07-22 at 1 53 54 pm

If it were just a little more regular I'd think it was some sort of offset error... but suspect something with high-contrast areas is getting processed differently than in GDAL.

How much could basic numerical rounding error be to blame?

metasim:
Band 1 has much more corse-grained differences. Histogram shows that it's not just +/- a value here or there.

screen shot 2018-07-22 at 2 02 59 pm

metasim:
Additional data: Got the same results when creating a test to take YCbCr -> RGB out of the equation.

First, created new test file to read:

```bash
gdal_translate jpeg-test.tif -co PHOTOMETRIC=RGB -co COMPRESS=JPEG jpeg-test-rgb.tif
```

Then read this file and compared to the `jpeg-test-deflate.tif` file. The delta image when created as above was a bit different, but don't think that tells us much. 

(this is band 0)

screen shot 2018-07-22 at 5 02 25 pm

metasim:
My outstanding questions are:

* Is the difference (error) significant? What does that mean with lossy compression?
* Is the difference traceable to the ImageIO JPEG decompressor, or are we sure the problem is in this usage of it?  
* Are we sure the GDAL decompressor is "authoritative"?
* Can we compare against a third implementation somehow?
* How many knobs & dials might we tweak in the ImageIO decompressor (such as `setSourceSubsampling`), that could lead to a result closer to GDAL?
* Is the comparison test a fair/proper one in this context?

metasim:
An aside, but interesting:
https://blogs.msdn.microsoft.com/devdev/2006/04/12/how-does-jpeg-actually-work/

echeipesh:
@metasim Whats your opinion on this? There may have been out of band conversation.
For the record I'm on board with the view that JPEG is intended for visual tiles mostly and the small discrepancies here are acceptable.

lossyrob:
To toss in an opinion - I think we should include this, with the experimental tag, and with the documented caveat that it's not expected to match GDAL. Then we eventually replace with GDAL-IO.

metasim:
I ditto @lossyrob on this. My gut is that the error is *not* in the color space conversion, but in the built-in decompressor. Tag it as experimental, throw it out there, and see if it has legs.

metasim:
PS: First need to fix the build by committing the test file.

PR #2763, opened 2018-07-18
readSubsetBands Performance Improvement

### LINKED ISSUE #2761: Reading Many Bands with readSubsetBands Casues OutOfMemoryError
Reading many bands with `readSubsetBands` causes a `java.lang.OutOfMemoryError` (see attached image). This was on a cluster of two workers at zoom level 13.

imageedit_3_9637209415

jbouffard:
## Overview

This PR improves the performance of the `readSubsetBands` method by changing how cropping operation is performed. Before, the input GeoTiff was converted to a `GeoTiffMultibandTile` and then cropped, but now it will be pattern matched instead and crop instead. This removes unneeded memory allocation which improves performance with certain backends, mainly S3.

### Checklist

Closes #2761

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/crop/MultibandTileCropMethods.scala
+++ b/raster/src/main/scala/geotrellis/raster/crop/MultibandTileCropMethods.scala
@@ -28,25 +28,39 @@ trait MultibandTileCropMethods extends TileCropMethods[MultibandTile] {
   import Crop.Options
 
   /**
-    * Given a [[GridBounds]] and some cropping options, crop the
-    * [[MultibandTile]] and return a new MultibandTile.
+    * Given a [[GridBounds]], a sequence of bands indexes,  and some cropping options, crop the
+    * [[MultibandTile]] and return a new MultibandTile that contains the target area and bands.
     */
-  def crop(gb: GridBounds, options: Options): MultibandTile = {
+  def cropBands(gb: GridBounds, targetBands: Seq[Int], options: Options): MultibandTile = {
     if (!gb.intersects(self.gridBounds)) throw GeoAttrsError(s"Grid bounds do not intersect: ${self.gridBounds} crop $gb")
     self match {
       case geotiffTile: GeoTiffMultibandTile =>
         val cropBounds =
           if (options.clamp) gb.intersection(self).get
           else gb
-        geotiffTile.crop(cropBounds)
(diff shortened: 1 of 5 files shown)


raster/src/main/scala/geotrellis/raster/crop/MultibandTileCropMethods.scala
```diff
@@ -28,26 +28,40 @@ trait MultibandTileCropMethods extends TileCropMethods[MultibandTile] {
   import Crop.Options
 
   /**
-    * Given a [[GridBounds]] and some cropping options, crop the
-    * [[MultibandTile]] and return a new MultibandTile.
+    * Given a [[GridBounds]], a sequence of bands indexes,  and some cropping options, crop the
+    * [[MultibandTile]] and return a new MultibandTile that contains the target area and bands.
     */
-  def crop(gb: GridBounds, options: Options): MultibandTile = {
+  def crop(gb: GridBounds, targetBands: Array[Int], options: Options): MultibandTile = {
```
echeipesh: I suggest calling this `cropBands` to keep it consistent with `readSubsetBands` and avoid the possibly confusing overload of `tiff.crop(bounds, Array(1, 2))`

Also `targetBands` should be relaxed to `Seq[Int]` to allow using this method if `List` and `Range`. Also thats what we use in `subsetBands`

spark/src/main/scala/geotrellis/spark/io/cog/COGValueReader.scala
```diff
@@ -102,18 +95,18 @@ trait COGValueReader[ID] {
...
 
       val (targetBands, targetBandsIndexes) = targetBandsWithIndex.unzip
 
       val croppedTiles: Array[Tile] =
         try {
-          sourceMultibandTile
-            .crop(gridBounds, targetBands)
-            .bands
-            .toArray
+          sourceTile match {
```
echeipesh: So this change is no longer required because the match should be happening above in `crop` method?

jbouffard: Yeah, with the implicit class imported it can access the `crop` method directly. https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/crop/MultibandTileCropMethods.scala#L34-L49

PR #2771, opened 2018-07-27
Remove Raster[T] to T implicit conversion

### LINKED ISSUE #2768: Implicit conversions from Raster to Tile create confusing semantics
The existence of this automatic conversion unknowingly sent me down a rabbit hole:

https://github.com/locationtech/geotrellis/blob/0af8404c3cdb22ee98102c4d18963b08567833c8/raster/src/main/scala/geotrellis/raster/Raster.scala#L49

One result of this is that you can call `(_: ProjectedRaster[Tile]).interpretAs(someCellType)`, but what is returned is a `Tile` and not a `ProjectedRaster[Tile]`. Implied in the name `interpretAs` (at least to me), is that you're going to get back the same type as the method receiver, as you would with a case class `copy` method. I think there are many other methods that end up being like this (e.g. `mapTiles`).

I'm in the camp that considers implicit type conversions like this to be bad practice. I know it will break code, but I personally think it should be removed (along with as many other implicit conversions you can purge).

echeipesh:
In particular this conversion results in data loss and surprising behavior.

Other remaining implicit conversion, from `Raster[T]` to `PolygonFeature[T]` at least preserve the extent information.

### Checklist

- [-] `docs` guides update, if necessary
- [-] New user API has useful Scaladoc strings

### Demo

Optional. Screenshots/REPL

### Notes

Resolves #2768

### DIFF
--- a/doc-examples/src/main/scala/geotrellis/doc/examples/spark/COGSparkExamples.scala
+++ b/doc-examples/src/main/scala/geotrellis/doc/examples/spark/COGSparkExamples.scala
@@ -78,8 +78,7 @@ object COGSparkExamples {
     val layer: TileLayerRDD[SpatialKey] = reader.read[SpatialKey, Tile](LayerId("example_cog_layer", zoom))
 
     // Let's stitch the layer into tile
-    val tile: Tile = layer.stitch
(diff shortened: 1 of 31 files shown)


echeipesh:
HA, this change actually revealed a bug in `GeoTiff.crop` that was being ignored in the unit tests ^.

PR #2793, opened 2018-08-23
Fixes #2792.

### LINKED ISSUE #2792: ArrayTile combine methods perform non-exhaustive matching
The following methods make incorrect assumptions about the `Tile` type-space. Because `Tile` inheritance is not restricted to the `geotrellis` namespace, type matching without a default handler should be considered a violation of the type contract.

https://github.com/locationtech/geotrellis/blob/bdc0ce22f657d87ef3f4ae9bdb953e7c9abd2564/raster/src/main/scala/geotrellis/raster/ArrayTile.scala#L224-L234

https://github.com/locationtech/geotrellis/blob/bdc0ce22f657d87ef3f4ae9bdb953e7c9abd2564/raster/src/main/scala/geotrellis/raster/ArrayTile.scala#L269-L278

I'm not sure I understand the purpose of these special cases (performance related?), but a default `case t: Tile => ???` should be implemented if possible.

metasim:
Also:
Adds new utility `DelegatingTile`.
Fixes compiler warning in CellTypeSpec.

Signed-off-by: Simeon H.K. Fitch <fitch@astraea.io>

Closes #2792.

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/DelegatingTile.scala
+++ b/raster/src/main/scala/geotrellis/raster/DelegatingTile.scala
@@ -1,0 +1,101 @@
+/*
+ * This software is licensed under the Apache 2 license, quoted below.
+ *
+ * Copyright 2018 Astraea, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *     [http://www.apache.org/licenses/LICENSE-2.0]
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ */
+
+package geotrellis.raster
+
+/**
+ * A tile that wraps another tile. Originally intended for delayed reading, but useful in other special use cases.
+ *
+ * @since 8/22/18
+ */
+trait DelegatingTile extends Tile {
+  protected def delegate: Tile
+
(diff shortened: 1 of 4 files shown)


raster/src/main/scala/geotrellis/raster/DelegatingTile.scala
```diff
@@ -0,0 +1,101 @@
...
+ */
+
+package geotrellis.raster
+
+/**
+ * A tile that wraps another tile. Originally intended for delayed reading, but useful in other special use cases.
+ *
+ * @since 8/22/18
+ */
+trait DelegatingTile extends Tile {
```
pomadchin: What's the use case of having it in the `gt` codebase? Just wondering. Mb it makes sense to add some doc strings for it? If it's only useful for tests it makes sense to move it into the test folder.

raster/src/test/scala/geotrellis/raster/ArrayTileSpec.scala
```diff
@@ -164,5 +164,29 @@ class ArrayTileSpec extends FunSpec
...
+          override protected def delegate: Tile = DoubleConstantTile(0.0, 100, 100)
+        }
+
+        assertEqual(at.combineDouble(fauxTile)((z1, z2) => z1 + z2), at)
+      }
+    }
   }
 }
+
+object ArrayTileSpec {
```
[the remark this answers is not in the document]
metasim: Cruft... I'll remove. Originally had `DelegatingTile` in there

raster/src/main/scala/geotrellis/raster/DelegatingTile.scala
metasim: I wrote it for RasterFrames core, and have found it useful for a few use cases and thought it was general purpose enough for cases where users' might want to have a `Tile` with slightly different behavior by overriding selected methods. If it's in GT core I'll use that version in RF. If it's in `test` then I'll have to have my own copy of it.

pomadchin:
I'm good with it; think we need to merge it.

PR #2794, opened 2018-08-28
GeoTiff CellType Conversion Fix

### LINKED ISSUE #2789: GeoTiffSegment convert improper behavior
It looks like we have a bug in converting `GeoTiffs` into `Unsigned` types: https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffSegment.scala#L48

For instance this code already looks like a not valid code: 

```scala
case ByteCellType | UByteCellType =>
  val arr = Array.ofDim[Byte](size)
  cfor(0)(_ < size, _ + 1) { i => getInt(i).toByte }
  arr
case ShortCellType | UShortCellType =>
  val arr = Array.ofDim[Short](size)
  cfor(0)(_ < size, _ + 1) { i => arr(i) = getInt(i).toShort }
  arr.toArrayByte()
```

We need to cover conversions to the unsigned types with tests and double check that it works as expected.

_NOTE:_ conversion works for `ArrayTile` types. In 2.0 we removed a `GeoTiff` default conversion to `ArrayTile` inside our `GeoTiffReader`, that's probably why we never faced it.

jbouffard:
## Overview

This PR fixes the logic in the `CellType` conversion for `GeoTiff`s so that they can convert to the `UByte` and `UShort` `CellType`s correctly.

### Checklist

### Notes

I didn't update the `CHANGELOG` because I wasn't sure what the next version of GT we'd be releasing.

Closes #2789

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffSegment.scala
+++ b/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffSegment.scala
@@ -51,13 +51,21 @@ trait GeoTiffSegment {
         val bs = new BitSet(size)
         cfor(0)(_ < size, _ + 1) { i => if ((getInt(i) & 1) == 0) { bs.set(i) } }
         bs.toByteArray()
-      case ByteCellType | UByteCellType =>
+      case ByteCellType =>
         val arr = Array.ofDim[Byte](size)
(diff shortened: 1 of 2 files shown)


echeipesh:
Back-ported to `2.1` with: 6d2694a8438b859d80daefde6f1506d1143984c2

PR #2804, opened 2018-09-13
Expose getClosestOverview into public

### LINKED ISSUE #2801: GeoTiff.getClosestOverview should be public
In the current master branch we have [this](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiff.scala#L105) function being marked as private. This, or some similar interface should be exposed into user API, to prevent [such things](https://github.com/raster-foundry/raster-foundry/blob/develop/app-backend/common/src/main/scala/utils/GTHack.scala).

pomadchin:
Closes #2801


echeipesh:
Backported to `2.1` with c3d1f325bf547d8ef762e95a8be092a390c32b02

PR #2807, opened 2018-09-24
Add 2.12 geotrellis cross build support

pomadchin:
## Overview

This PR adds geotrellis cross build support. 
WARN: all spark dependent packages won't work in runtime.


project/Dependencies.scala
```diff
@@ -45,7 +45,8 @@ object Dependencies {
   val fs2Core             = "co.fs2"                     %% "fs2-core"                 % "0.10.4"
   val fs2Io               = "co.fs2"                     %% "fs2-io"                   % "0.10.4"
 
-  val sparkCore           = "org.apache.spark"           %% "spark-core"               % Version.spark
+  // TODO: make it through %% once Spark 2.4 artifacts would be published
+  val sparkCore           = "org.apache.spark"            % "spark-core_2.11"          % Version.spark
```
metasim: If it were me (just food for thought, don't take it as a need), would be to figure out how to change this in `build.sbt`:
```scala
lazy val root = Project("geotrellis", file(".")).
  aggregate(
    accumulo,
    cassandra,
    `doc-examples`,
    geomesa,
    geotools,
    geowave,
    hbase,
    macros,
    proj4,
    raster,
    `raster-testkit`,
    s3,
    `s3-testkit`,
    shapefile,
    slick,
    spark,
    `spark-etl`,
    `spark-pipeline`,
    `spark-testkit`,
    util,
    vector,
    `vector-testkit`,
    vectortile
  ).
...
```

to not include anything spark dependent if `scalaVersion` >= 2.12. That way you don't end up with artifacts that exist but are bad/useless.

metasim: This would be to obviate possible downstream confusion.

Issue #2805, opened 2018-09-17
Ingesting Sentinel 1 B images


brunifrancesco:
I'm experimenting Geotrellis feature to ingest a downloaded Sentinel 1 image downloaded from ESA scihub; specifically the image is the one tagged with the following ID: 

`S1B_IW_SLC__1SDV_20180905T013930_20180905T013957_012573_01732C_B94E`

I used  *gdal_translate* script to convert the complex data type to a floating one (since Geotrellis is not able to process complex data types); after that I ran the ingestion procedure but I got this error:

```
Caused by: geotrellis.vector.ExtentRangeError: Invalid Extent: xmin must be less than xmax (xmin=73.44075377325863, xmax=71.70715627378026)
    at geotrellis.vector.Extent.<init>(Extent.scala:117)
    at geotrellis.raster.io.geotiff.tags.TiffTags.tiePointsModelSpace(TiffTags.scala:449)
    at geotrellis.raster.io.geotiff.tags.TiffTags.extent(TiffTags.scala:259)
    at geotrellis.raster.io.geotiff.reader.GeoTiffReader$.geotrellis$raster$io$geotiff$reader$GeoTiffReader$$getGeoTiffInfo$1(GeoTiffReader.scala:455)
    at geotrellis.raster.io.geotiff.reader.GeoTiffReader$.readGeoTiffInfo(GeoTiffReader.scala:484)
    at geotrellis.spark.io.hadoop.HadoopGeoTiffInfoReader.getGeoTiffInfo(HadoopGeoTiffInfoReader.scala:53)
    at geotrellis.spark.io.GeoTiffInfoReader$$anonfun$1.apply(GeoTiffInfoReader.scala:81)
    at geotrellis.spark.io.GeoTiffInfoReader$$anonfun$1.apply(GeoTiffInfoReader.scala:79)
    at scala.collection.Iterator$$anon$12.nextCur(Iterator.scala:435)
    at scala.collection.Iterator$$anon$12.hasNext(Iterator.scala:441)
    at org.apache.spark.storage.memory.MemoryStore.putIteratorAsValues(MemoryStore.scala:216)
    at org.apache.spark.storage.BlockManager$$anonfun$doPutIterator$1.apply(BlockManager.scala:1092)
    at org.apache.spark.storage.BlockManager$$anonfun$doPutIterator$1.apply(BlockManager.scala:1083)
    at org.apache.spark.storage.BlockManager.doPut(BlockManager.scala:1018)
[...]

PR #2812, opened 2018-10-01
Fix Inverted Extent

pomadchin:
## Overview

This PR allows to handle properly such kinds of Extents:

```
Corner Coordinates:
Upper Left  (    0.0,    0.0)
Lower Left  (    0.0,13545.0)
Upper Right (21995.0,    0.0)
Lower Right (21995.0,13545.0)
Center      (10997.5, 6772.5)
```

Closes https://github.com/locationtech/geotrellis/issues/2805

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/io/geotiff/tags/TiffTags.scala
+++ b/raster/src/main/scala/geotrellis/raster/io/geotiff/tags/TiffTags.scala
@@ -446,7 +446,13 @@ case class TiffTags(
           i += 1
         }
 
-        Extent(minX, minY, maxX, maxY)
+        // fix an inverted extent, to behave more like GDAL
+        Extent(
+          math.min(minX, maxX),
+          math.min(minY, maxY),
+          math.max(minX, maxX),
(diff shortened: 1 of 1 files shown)


raster/src/main/scala/geotrellis/raster/io/geotiff/tags/TiffTags.scala
```diff
@@ -446,7 +446,13 @@ case class TiffTags(
           i += 1
         }
 
-        Extent(minX, minY, maxX, maxY)
+        // fix an inverted extent, to behave more like GDAL
```
pomadchin: do we really want to perform a silent transform or mb we can throw at least a warning?

echeipesh: I'm surprised that GDAL allows this but seems like a safe conversion here.

echeipesh:
Backport to `2.1`: f0700d625390815698e121741289081d863ddc72

PR #2814, opened 2018-10-03
Use Timer as necessary for IO

moradology:
This brings cats-effect to 1.0, which should be faster and have a more stable API going forward.

### DIFF
--- a/cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraRDDWriter.scala
+++ b/cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraRDDWriter.scala
@@ -116,6 +116,7 @@ object CassandraRDDWriter {
 
               val pool = Executors.newFixedThreadPool(threads)
               implicit val ec = ExecutionContext.fromExecutor(pool)
+              implicit val cs = IO.contextShift(ec)
 
               def elaborateRow(row: (BigInt, Vector[(K,V)])): fs2.Stream[IO, (BigInt, Vector[(K,V)])] = {
                 fs2.Stream eval IO.shift(ec) *> IO ({
@@ -148,8 +149,11 @@ object CassandraRDDWriter {
                 })
               }
 
-              val results = (rows flatMap elaborateRow flatMap rowToBytes map retire)
-                .join(threads)
+              val results = rows
+                .flatMap(elaborateRow)
+                .flatMap(rowToBytes)
+                .map(retire)
+                .parJoin(threads)
                 .onComplete {
                   fs2.Stream eval IO.shift(ec) *> IO {
                     session.closeAsync()
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -39,11 +39,11 @@ object Dependencies {
 
   val awsSdkS3            = "com.amazonaws"               % "aws-java-sdk-s3"          % "1.11.319"
 
-  val catsCore            = "org.typelevel"              %% "cats-core"                % "1.3.1"
+  val catsCore            = "org.typelevel"              %% "cats-core"                % "1.4.0"
   val catsEffect          = "org.typelevel"              %% "cats-effect"              % "1.0.0"
 
-  val fs2Core             = "co.fs2"                     %% "fs2-core"                 % "1.0.0-RC2"
-  val fs2Io               = "co.fs2"                     %% "fs2-io"                   % "1.0.0-RC2"
+  val fs2Core             = "co.fs2"                     %% "fs2-core"                 % "1.0.0"
+  val fs2Io               = "co.fs2"                     %% "fs2-io"                   % "1.0.0"
 
   val sparkCore           = "org.apache.spark"           %% "spark-core"               % Version.spark
   val sparkSQL            = "org.apache.spark"           %% "spark-sql"                % Version.spark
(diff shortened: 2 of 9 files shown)


moradology:
To get things working with a minima of pain/annoyance, fs2 stream `map` calls followed by `join` (`flatten` in the current fs2 api: https://github.com/functional-streams-for-scala/fs2/blob/series/1.0/core/shared/src/main/scala/fs2/Stream.scala#L949-L951) calls have been simplified into a single `flatMap` call.

cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraRDDWriter.scala
```diff
@@ -148,8 +148,7 @@ object CassandraRDDWriter {
                 })
               }
 
-              val results = (rows flatMap elaborateRow flatMap rowToBytes map retire)
-                .join(threads)
+              val results = (rows flatMap elaborateRow flatMap rowToBytes flatMap retire)
```
pomadchin: would this be a `join(threads)` equivalent? i was not sure, that's why decided to make it clear, to keep things simple.

pomadchin: Or probably it's a `1.0` feature for sure.

project/Dependencies.scala
```diff
@@ -39,11 +39,11 @@ object Dependencies {
 
   val awsSdkS3            = "com.amazonaws"               % "aws-java-sdk-s3"          % "1.11.319"
 
-  val catsCore            = "org.typelevel"              %% "cats-core"                % "1.1.0"
-  val catsEffect          = "org.typelevel"              %% "cats-effect"              % "0.10.1"
+  val catsCore            = "org.typelevel"              %% "cats-core"                % "1.4.0"
```
pomadchin: fs2 `1.0.0-RC2` depends on `cats-core 1.3.0`

cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraRDDWriter.scala
moradology: Semantically equivalent, I believe so, but we lose some parallelism here

pomadchin: We loose some parallelism? What do you mean? You mean that the behaviour would differ significantly?

pomadchin: @moradology don't you want to use explicit `.parJoin(threads)` instead of a flatMap on the last step?

pomadchin:
```scala
implicit val cs: ContextShift[IO] = IO.contextShift(ec)
(index map readRecord)
        .parJoin(threads)
        .compile
        .toVector
        .unsafeRunSync.flatten
```

pomadchin:
- Cats core 1.3.1 [CQ 17790](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=17790)
- Cats effect 1.0.0 [CQ 17772](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=17772)
- FS2 1.0.0-RC2 [CQ 17792](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=17792)

cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraRDDWriter.scala
pomadchin: @moradology no, it is a question for you, as you introduced this change, i was just asking :D

cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraRDDWriter.scala
```diff
@@ -148,8 +149,13 @@ object CassandraRDDWriter {
                 })
               }
 
-              val results = (rows flatMap elaborateRow flatMap rowToBytes map retire)
-                .join(threads)
+              val results = rows
+                .map(elaborateRow)
+                .parJoin(threads)
```
pomadchin: w0000t, i think 

```scala
(rows flatMap elaborateRow flatMap rowToBytes map retire).parJoin(threads)
```

Would be enough! It's already tested and benchmarked long ago, so probably you can just replace all `join`s with `parJoin`s.

spark/src/main/scala/geotrellis/spark/io/AsyncWriter.scala
```diff
@@ -74,8 +77,13 @@ abstract class AsyncWriter[Client, V, E](threads: Int) extends Serializable {
       fs2.Stream eval IO.shift(ec) *> retryFunc.fold(writeTask)(writeTask.retryEBO(_))
     }
 
-    (rows flatMap elaborateRow flatMap encode map retire)
-      .join(threads)
+    rows
+      .map(elaborateRow)
+      .parJoin(threads)
```
pomadchin: The same here, no needs in so many `parJoin`s :D

cassandra/src/main/scala/geotrellis/spark/io/cassandra/CassandraRDDWriter.scala
pomadchin: It makes no sense to do parJoin on any map step, i believe the result would be the same as with a single join.

pomadchin: Nvm, lets leave it as you did

PR #2817, opened 2018-10-12
Use sun.misc.IOUtils.readFully

echeipesh:
A more performant way to turn `InputStream` to `Array[Byte]` that minimizes use of a transient dependency `commons-io` which we're picking up from one of the `hadoop` packages.


pomadchin:
I'd say before better java versions we have to use `commons-io` ):

PR #2823, opened 2018-11-02
Bump Spark version up to 2.4

### LINKED ISSUE #1751: Cross compile/publish to 2.12

pomadchin:
## Overview

This PR updates Apache Spark dependency.

The following projects are excluded from the 2.12 builds: 

* Slick (as it's deprecated, and requires some code changes to make it work)
* GeoMesa (as it doesn't support Scala 2.12)
* GeoWave (as it doesn't support Scala 2.12)

Closes #1751

### DIFF
--- a/.travis/build-set-1.sh
+++ b/.travis/build-set-1.sh
@@ -5,9 +5,9 @@
   "project geotools" test \
   "project shapefile" test \
   "project vector" test \
-  "project vectortile" test || { exit 1; }
-  # "project slick" test \
-  # "project geowave" compile test:compile \
-  # "project hbase" test \
-  # "project geomesa" test \
-  # "project cassandra" test || { exit 1; }
+  "project vectortile" test \
+  "project hbase" test \
+  "project cassandra" test || { exit 1; }
+  # "project slick" test
+  # "project geomesa" test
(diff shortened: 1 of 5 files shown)


vertexclique:
Hi! Will you roll a release containing 2.12 updates?

pomadchin:
Hey @vertexclique ; that would be GT 3.0, which would be released pretty soon I suppose and would contain mostly dep updates.

vertexclique:
What is your timeline? I want to include this library to my project. Do you have specific release timeline?

echeipesh:
@vertexclique `SNAPSHOT` publishing looks busted, going to fix that post-haste. 3.0, mostly scala 2.12 and JTS package rename, should be out by December 11th. I'm going to start the ball rolling on that in couple of days.

pomadchin:
Hey @vertexclique we've published 3.0 scala 2.12 binaries, check it out there: https://repo.locationtech.org/content/groups/snapshots/org/locationtech/geotrellis/

PR #2825, opened 2018-11-17
Fix error in `ReprojectRasterExtent`

jpolchlo:
## Overview

While investigating a reprojection issue in geotrellis-contrib, it came to my attention that the raster extents produced by `ReprojectRasterExtent` are off.  Consider the following: using https://github.com/geotrellis/geotrellis-contrib/blob/master/vlm/src/test/resources/img/aspect-tiled.tif as the base of this example, the following definitions apply:
```scala
val tiff = GeoTiffReader.readMultiband(getByteReader("aspect-tiled.tif"), streaming = true)
val targetRE = ReprojectRasterExtent(tiff.rasterExtent, tiff.crs, WebMercator)
val region = tiff.projectedExtent.reprojectAsPolygon(WebMercator, 0.001)
```
We then note that `targetRE.extent` and `region.envelope` diverge:
```scala
targetRE.extent: geotrellis.vector.Extent = Extent(-8769150.640916323, 4257706.9716441, -8750638.524926396, 4274455.441673627)
region.envelope: geotrellis.vector.Extent = Extent(-8769150.640916323, 4257707.61418559, -8750636.281752571, 4274455.441673627)
```
Note that `ymin` and `xmax` differ, and that the tolerance value of 0.001 has no effect, since the corners of the region are on the extent boundary.

I believe that the `region` is the authoritative source of truth.  So, the reprojected raster extent should tightly bound this polygon.

### Checklist

### DIFF
--- a/.locationtech/deploy-211.sh
+++ b/.locationtech/deploy-211.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+
+ set -e
+ set -x
+
+ ./sbt "project macros" publish -no-colors \
+   && ./sbt "project vector" publish -no-colors \
+   && ./sbt "project proj4" publish -no-colors \
+   && ./sbt "project raster" publish -no-colors \
+   && ./sbt "project spark" publish -no-colors \
+   && ./sbt "project s3" publish -no-colors \
+   && ./sbt "project accumulo" publish -no-colors \
+   && ./sbt "project cassandra" publish -no-colors \
+   && ./sbt "project hbase" publish -no-colors \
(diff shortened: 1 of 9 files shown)


jpolchlo:
As of now, some of the original tests for `ReprojectRasterExtent` are failing due to small variations introduced by this PR.  I do believe that the old implementation was incorrect and needed to be fixed, but this is leading to a slight divergence between GDAL and GT, which is what the failing tests check against.  We'll need some discussion as to whether the new results are precisely what we intend them to be.

PR #2835, opened 2018-12-03
Spire Shade Rules for spark-etl and spark-pipeline

jbouffard:
## Overview

We've recently updated the version of our `spire` dependencies to 0.16.0. However, this can cause problems as Spark still uses older an older version of `spire`. With these shaded rules, it'll now be possible to run code in `spark-etl` and/or `spark-pipeline` without having to worry about binary incompatibility.


pomadchin:
Replaced by https://github.com/locationtech/geotrellis/pull/2836

PR #2839, opened 2018-12-05
Fix vectorizer bug

### LINKED ISSUE #2833: java.lang.ArrayIndexOutOfBoundsException: 1936 when using toVector() to vectorize a tile
I am trying to read in an GeoTiff (black/white) image, and vectorize it in order to count the number of polygons in it. The code is simple:

```scala
def main(args: Array[String]) = {
    val img = SinglebandGeoTiff("/path/to/sgl_tile.tif")
    try {
      val polygons = img.tile.toVector(img.extent)
      println(polygons.length)
    } catch {
      case ex: java.lang.ArrayIndexOutOfBoundsException => println(ex)
    }
  }
```

The problem is: it worked for some images (I got the correct count), failed for others (it threw the exception: `java.lang.ArrayIndexOutOfBoundsException: 1936`)

Here is a sample image that it failed to vectorize
sgl_tile_0

Cannot figure out why by googling, any advice would be appreciated!

jpolchlo:
## Overview

A user reported a condition when the rasterizer fails.  This corresponds to the case when a new cluster begins at the lower right pixel.  A bounds check wasn't quite right.  This contribution fixes that issue and includes a new test to guard against it.

### Checklist

Closes #2833


pomadchin:
Btw, here is a test file attached: https://github.com/locationtech/geotrellis/issues/2833 would not it be too hard / expensive to create a unit test for this file?

echeipesh:
Will need backported to `2.2` and `2.0`

PR #2844, opened 2018-12-07
Fixed elliptical radius check in IDW algorithm.

### LINKED ISSUE #2840: IDW interpolation radius check is incorrect
The [radius check](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/interpolation/InverseDistanceWeighted.scala#L95) in the current IDW code looks like this

```
// radiusX and radiusY come from the IDW options
val radiusXY = radiusX * radiusY

for (point <- points) {
  val dX = x - point.geom.x
  val dY = y - point.geom.y
  if (radiusY * dX2 * dX2 + radiusX * dY2 * dY2 <= radiusXY ) {
```

That last line of code has two mistakes in it
* The right had side of the equation should be squared
* The `radiusX` and `radiusY` terms on the left hand side should also be squared

The check should be `(radiusY * radiusY * dX2 * dX2 + radiusX * radiusX * dY2 * dY2 <= radiusXY * radiusXY )`

Spot checking the current code demonstrates that it is wrong. Suppose radius X and radius Y equal each other. The exponents on the left are cubic while the rhs is squared. A concrete example really shows the problem. Suppose radiusX and radiusY are both 10, and that dX equals 10 and dy is 0. The check simplifies to
```
10*10*10 <= 10*10  // false, wanted true
```
The equation for an ellipse centered on the origin shows how the code ended up flawed. 

```
x^2           y^2
----     +   -----   = 1
rx^2         ry^2

which simplifies to 
(ry^2)*(x^2) + (rx^2) * (y^2) = (rx^2) * (ry^2)
```

That last line is what the radius check should be doing. Substituting our example values results in a check of 10^4 <= 10^4 which passes now.

For reference, you can see that the [GDAL code](https://svn.osgeo.org/gdal/trunk/gdal/alg/gdalgrid.cpp) that inspired geotrellis does this correctly. The right hand side of the check is a variable called `df12` which was initialized to the equivalent of `radiusX^2 * radiusY^2`. They also squared the radius values on the left hand side.

```
// GDAL code
const double dfRadius1 = poOptions->dfRadius1 * poOptions->dfRadius1; 
const double dfRadius2 = poOptions->dfRadius2 * poOptions->dfRadius2;
const double dfR12 = dfRadius1 * dfRadius2;
...
    // Is this point located inside the search ellipse?
    if( dfRadius2 * dfRX * dfRX + dfRadius1 * dfRY * dfRY <= dfR12 )
```

hhowe29:
## Overview

Closes #2840 

### Checklist

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/interpolation/InverseDistanceWeighted.scala
+++ b/raster/src/main/scala/geotrellis/raster/interpolation/InverseDistanceWeighted.scala
@@ -65,7 +65,9 @@ object InverseDistanceWeighted {
 
     if (!points.isEmpty) {
       val ewr2 = equalWeightRadius * equalWeightRadius
-      val radiusXY = radiusX * radiusY
+      val radiusXSquared = radiusX * radiusX
+      val radiusYSquared = radiusY * radiusY
+      val radiusXYSquared = radiusXSquared * radiusYSquared
 
       def idw(points: Traversable[PointFeature[D]], x: Double, y: Double, hasRadius: Boolean) = {
         var sum = 0.0
@@ -92,7 +94,7 @@ object InverseDistanceWeighted {
               (dX, dY)
             }
 
-          if (!hasRadius || radiusY * dX2 * dX2 + radiusX * dY2 * dY2 <= radiusXY ) {
+          if (!hasRadius || radiusYSquared * dX2 * dX2 + radiusXSquared * dY2 * dY2 <= radiusXYSquared ) {
             val data = point.data: Double
             if (d2 <= ewr2) {
               sampleSum += data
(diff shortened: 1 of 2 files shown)


hhowe29:
ci test this please

pomadchin:
Hey @hhowe29 I think that it's good to go even without CLA, but can you sign it? https://github.com/locationtech/geotrellis/blob/master/docs/CONTRIBUTING.rst#eclipse-contributor-agreement-eca 

After signing just to `git commit --amend -S && git push -f master`

pomadchin:
@hhowe29 thank you so much; merging!

PR #2847, opened 2018-12-10
This PR is for experiments with TravisCI 

pomadchin:


pomadchin:
So the reason is indeed OOM in every case. 

image

I didn't have a chance to investigate more today though, but at least it's obvious that it's true OOM.

pomadchin:
Well, let's merge it; as it's the only way to check how it would work on a master branch. All the same flags I also discovered in [this](https://github.com/holdenk/spark-testing-base/blob/master/build.sbt#L162) repo.

PR #2880, opened 2019-03-07
Fixes to RasterRegionReproject

jpolchlo:
## Overview

RasterRegionReproject is a touchy algorithm.  Specifically, there is a mock scanline generation algorithm that takes place using a pretty weird geometric intersection algorithm.  It is unfortunately useful for speeding up the computation of sampling points, so we're stuck with it.  The problem is that the determination of bounds for converting these geometric line segments into a pixel range appears to be numerically touchy.  I'm proposing a slight change to the numerical fudge factors that make the process work, but I understand that this is suspicious, and will possibly cause problems depending on the projection of the input points.

We can try to finesse this PR to do something more elegant and more correct through the review process.

### Checklist


metasim:
@jpolchlo Are there any API breaking changes in this PR... if so, having them in the release notes would be helpful for predicting level of effort in future downstream upgrades.

pomadchin:
Hey @metasim there are no API changes; precision fixes only. Considered as a ‘bad’ quick fix I suppose ._.

PR #2885, opened 2019-03-26
Easy Importing for the Vector Package

jbouffard:
## Overview

This PR simplifies the imports needed to use certain functionality in the `geotrellis.vector` package. Mainly, `vector.io.wkb.Implicits`, `vector.io.wkt.implicits`, and `vector.io.json.Implicits` are now all imported with `import geotrellis.vector._`.

### Checklist

### Demo

Before:

```scala
import geotrellis.vector._
import geotrellis.vector.io._
import geotrellis.vector.io.json.JsonFeatureCollection

val path: String = ???
val f = scala.io.Source.fromFile(path)
val collection = f.mkString.parseGeoJson[JsonFeatureCollection]
```

After:

```scala
import geotrellis.vector._
import geotrellis.vector.io.json.JsonFeatureCollection

val path: String = ???
val f = scala.io.Source.fromFile(path)
val collection = f.mkString.parseGeoJson[JsonFeatureCollection]
```

### DIFF
--- a/vector/src/main/scala/geotrellis/vector/io/package.scala
+++ b/vector/src/main/scala/geotrellis/vector/io/package.scala
@@ -19,9 +19,7 @@ package geotrellis.vector
 import geotrellis.vector.io.wkb.WKB
 import geotrellis.vector.io.wkt.WKT
 
-package object io extends io.json.Implicits
-    with io.wkb.Implicits
-    with io.wkt.Implicits {
+package object io {
 
   /** Read any WKT or WKB and return the corresponding geometry */
   def readWktOrWkb(s: String): Geometry = {
@@ -33,4 +31,3 @@ package object io extends io.json.Implicits
       WKT.read(s)
   }
 }
-
--- a/vector/src/main/scala/geotrellis/vector/package.scala
+++ b/vector/src/main/scala/geotrellis/vector/package.scala
@@ -23,7 +23,19 @@ import org.locationtech.jts.{geom => jts}
 package object vector extends SeqMethods
     with reproject.Implicits
     with triangulation.Implicits
-    with voronoi.Implicits {
+    with voronoi.Implicits
+    with io.json.Implicits
+    with io.wkt.Implicits
+    with io.wkb.Implicits {
+
+  /** The algorithms herein are all implemented in JTS, but the wrapper methods
+    * here make it straightforward to call them with geotrellis.vector classes.
+    */
+  implicit class withAnyGeometryMethods[G <: Geometry](val self: G) extends MethodExtensions[G]
+      with convexhull.ConvexHullMethods[G]
+      with densify.DensifyMethods[G]
+      with prepared.PreparedGeometryMethods[G]
+      with simplify.SimplifyMethods[G]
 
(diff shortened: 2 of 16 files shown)


vector/src/main/scala/geotrellis/vector/io/package.scala
```diff
@@ -19,9 +19,7 @@ package geotrellis.vector
 import geotrellis.vector.io.wkb.WKB
 import geotrellis.vector.io.wkt.WKT
 
-package object io extends io.json.Implicits
```
jbouffard: These extensions have to be removed, because other wise if you do something like this:

```scala
import geotrellis.vector._
import geotrellis.vector.io._
```

The compiler will throw an error about not being able to find an implicit.

vector/src/main/scala/geotrellis/vector/package.scala
```diff
@@ -23,7 +23,19 @@ import org.locationtech.jts.{geom => jts}
...
-    with voronoi.Implicits {
+    with voronoi.Implicits
+    with io.json.Implicits
+    with io.wkt.Implicits
+    with io.wkb.Implicits {
+
+  /** The algorithms herein are all implemented in JTS, but the wrapper methods
+    * here make it straightforward to call them with geotrellis.vector classes.
+    */
+  implicit class withAnyGeometryMethods[G <: Geometry](val self: G) extends MethodExtensions[G]
```
jbouffard: I moved this up just so it's easier to see what implicits will be in scope when importing `geotrellis.vector._`

moradology:
I'd personally like to see a wholesale renaming of `io` packages to something that won't make importing e.g. `circe` such a pain:
As of now, this doesn't work
```scala
import geotrellis.vector._
import io.circe._
```

Instead, this must be done:
```scala
import geotrellis.vector._
import _root_.io.circe._
```

jbouffard:
@moradology Is there an alternative name for `vector.io` package that you have? Maybe it could be `serialize`? Do you have any thoughts, @echeipesh?

echeipesh:
Not sure, `formats` sounds alright to me, `serialize` is a verb which is weird for package name. The argument also goes for other first tier packages that get imported all the time so: `geotrellis.raster.io`, `geotrellis.spark.io`. Those probably have things that can/should be split amongst multiple packages to avoid `io`. 

I think that naming conflict is bad but I'd suggest dealing with that rename last or in different PR.

echeipesh:
@jbouffard Part of the sugar in cats imports seems to be to bring important types to the root level with a [`type alias`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/package.scala#L95-L102). Does anything in `geotrellis.vector` benefit from that ?

I know that I often end up reaching for `WKB` and `WKT`.

jbouffard:
@echeipesh That makes sense. I think the renaming is something we should probably discuss as a group. So I'll hold off on renaming until later.

Yeah, I think `WKB` and `WKT` would both benefit from being aliased. I'm not sure if there's anything that really needs to included in `vector`. All of the other types that aren't aliased are for pretty niche things.

moradology:
`codec` or `serde` perhaps

PR #2886, opened 2019-03-27
Refactor for GridExtent[N: Integral]

echeipesh:
## Overview

The starting motivation of this refactor is to be able to work with raster coverages that have more cells that can be represented by `Int`.

Originally this problem was addressed by `GridExtent` trait, which `RasterExtent` extends. However because `GridExtent` did not have `cols` and `rows` fields it creates an ambiguity when `extent.width` is not evenly divided by `cellwidth` (Given floating point math this is pretty much always). When used in anger this approach produced lots of fence-post errors w.r.t which pixels are included or excluded in tiles belonging to a layout.

This PR address the problem by allowing `GridExtent` to have `cols: Long` and `rows: Long`. The knock-on consequences is that `Grid` and `CellGrid` similarly have to be parametrized by `N`.

### Checklist

### Demo

`Pending...`

### Notes

This PR leans on `spire.math.Integral` because `scala.math.Integral` was tried initially and proved to be insufficient. This is the first instance where we use spire types in a way that will leak into API and user code.


raster/src/main/scala/geotrellis/raster/GridBounds.scala
```diff
@@ -18,70 +18,20 @@ package geotrellis.raster
...
-  def width = colMax - colMin + 1
-  def height = rowMax - rowMin + 1
-
-  def size: Long = width.toLong * height.toLong
-
-  @deprecated("Use `size` instead.", "2.0")
-  def sizeLong: Long = width.toLong * height.toLong
-
-  def isEmpty = size == 0
+case class GridBounds[@specialized(Short, Int, Long) N: Integral](
```
moradology: It might be sensible just to specialize on `Int`/`Long` to avoid byte code duplication for a type that we won't likely see

PR #2892, opened 2019-04-04
Updated `Feature` class

jpolchlo:
## Overview

The current implementation of `Feature` is flawed in a few ways.  First, it is implemented as a `case class`, which means that no subclasses can be made (a problem if you want to meaningfully tag your features with an `id`, for example).  Second, though `Feature`s have functor-like characteristics, they are not implemented as such.  This small PR seeks to remedy such issues.

Fixing up `Feature` is an excuse to bring in [simulacrum](https://github.com/mpilquist/simulacrum) and begin the process of introducing typeclasses into GT.  We begin with an example of sufficient complexity that it raises some of the relevant issues that are likely to be encountered in the future.  

The change presented here introduces the notion of an object carrying geometry (useful if we ever want to have a `ProjectedGeometry`, for example).  A typeclass, `HasGeometry`, can be applied to such objects.  `HasGeometry` provides a means to retrieve the geometry of an applicable instance, and to transform the geometry of an instance.

This experimental work is meant to give us a handle on typeclasses more generally and basic patterns of interaction within a very constrained scope, while figuring out new patterns of interaction.

### Checklist

### Demo

Optional. Screenshots/REPL

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -46,6 +46,8 @@ object Dependencies {
   val fs2Core             = "co.fs2"                     %% "fs2-core"                 % "1.0.0"
   val fs2Io               = "co.fs2"                     %% "fs2-io"                   % "1.0.0"
 
+  val simulacrum          = "com.github.mpilquist"       %% "simulacrum"               % "0.15.0"
+
   val sparkCore           = "org.apache.spark"           %% "spark-core"               % Version.spark
   val sparkSQL            = "org.apache.spark"           %% "spark-sql"                % Version.spark
 
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -320,7 +320,6 @@ object Settings {
     libraryDependencies ++= Seq(
       pureconfig,
       jts,
-      catsCore,
       spire,
       monocleCore,
       monocleMacro,
@@ -528,6 +527,8 @@ object Settings {
     libraryDependencies ++= Seq(
       jts,
       pureconfig,
+      catsCore,
+      simulacrum,
       sprayJson,
       apacheMath,
       spire,
--- a/vector/src/main/scala/geotrellis/vector/Feature.scala
+++ b/vector/src/main/scala/geotrellis/vector/Feature.scala
@@ -16,6 +16,9 @@
 
 package geotrellis.vector
 
+import cats._
+import cats.implicits._
+
 /** A case class which represents a geometry with some metadata
   *
   * @tparam G A subtype of Geometry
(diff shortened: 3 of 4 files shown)


jpolchlo:
A weird thing with regard to the current state of this PR: the extremely complex methods for accessing typeclass-provided functions.  For example, getting the geometry for an instance requires the call
```scala
HasGeometry[Feature.FeatureGeometry[String]#Type].geom(f)
```
Which is less than ideal.

I suspect there's some kind-projector-fu that will make this better.  Couldn't figure out how to eliminate the `#Type` access above with that plugin.

jpolchlo:
Oh, also: This didn't work with an example subclass of `Feature`.  The feature's class needs to be passed as a second-order type to `HasGeometry`, I think.

pomadchin:
Do you have an example of what's not working?

pomadchin:
I finally looked into this PR, so the two things: 

```scala
implicit def featureHasGeometry[D] = new HasGeometry[λ[`+G <: Geometry` => Feature[G, D]]] {
    def geom[G <: Geometry](ft: Feature[G, D]): G = ft.geom
    def mapGeom[G <: Geometry, T <: Geometry](ft: Feature[G, D])(fn: G => T): Feature[T, D] = Feature(fn(ft.geom), ft.data)
  }
```

Not sure how much ```λ[`+G <: Geometry` => Feature[G, D]]``` looks easier O: 

The next thing is that `Functor` is invariant by design. However we can try to define it (since we have a function it will be a sort of derivation for each call) for subtypes: 

```scala
  implicit def featureFunctor[G <: Geometry, A, F[_] <: Feature[G, A]](implicit lens: Lens[F[_], A]): Functor[F] = new Functor[F] {
    def map[A, B](fa: F[A])(f: A => B): F[B] = {
      val geom: G = fa.geom
      val data: A = fa.data.asInstanceOf[A]
      Feature(geom, f(data)).asInstanceOf[F[B]]
    }
  }
```

I think it's obvious from that definition that it's hard to make it work. The issue is in the new `F` object creation. Even if we have `Lens` to `set` and `get` type `A`, we need also to pass the `Lens` that will know how to set value of type `B` into the `F`, it means some special kind of functor, with the map defined somehow like this: `def map[A, B: Lens[F[_], ?]](fa: F[A])(f: A => B): F[B]`. However I think we'll come up with smth smarter, or with a more straightforward solution - to implement functor instances for each specific type (I think that's how originally this issue is solved).

pomadchin:
Closing this issue, the summary is that this particular PR is not mergable, the take out is that probably it really makes sense to abstract over the `Feature` type via `Lenses`, and for instance to put properties like `id`, etc into the `data: D` field.

PR #2894, opened 2019-04-08
GridExtent regressions

echeipesh:
## Overview

- Fix bug in conversion from `GridExtent[Long]` to `GridExtent[Int]`, rely on spire checkin
- Add equality methods and toString methods
- Standardize the overloads for GridExtent

Connects https://github.com/locationtech/geotrellis/pull/2886

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/tiling/LayoutDefinition.scala
+++ b/spark/src/main/scala/geotrellis/spark/tiling/LayoutDefinition.scala
@@ -47,6 +47,9 @@ case class LayoutDefinition(override val extent: Extent, tileLayout: TileLayout)
 
     LayoutDefinition(subExtent, subLayout)
   }
+
+  override def toString: String =
+    s"""LayoutDefinition($extent, $cellSize,${layoutCols}x${layoutRows} tiles,${cols}x${rows} pixels)"""
 }
 
 object LayoutDefinition {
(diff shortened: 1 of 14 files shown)


echeipesh:
The change to `GridExtent.apply` methods is worth calling out in particular. It forces the caller to always decide on the width of the cols `Long`/`Int` at creation time. This is kind of a hassle in general.

There is a decision point that we're going to hit, do we:

1. Remove `RasterExtent` and use `GridExtent[N]` in its place to cut down on number of nouns in the library
2. Keep `RasterExtent extends GridExtent[Int]` and `LayerExtent extends GridExtent[Long]` to minimize the cognitive load of thinking about type parameters. 

At the moment I'm in favor for Option 1, but that is subject to change when we see the impact of removing `RasterExtent`.

The main point is to remember that these changes are still in draft stage and to look for more confirmation or alternatives.

spark/src/main/scala/geotrellis/spark/tiling/LayoutDefinition.scala
```diff
@@ -47,6 +47,25 @@ case class LayoutDefinition(override val extent: Extent, tileLayout: TileLayout)
 
     LayoutDefinition(subExtent, subLayout)
   }
+
+  override def toString: String =
+    s"""LayoutDefinition($extent,$cellSize,${layoutCols}x${layoutRows} tiles,${cols}x${rows} pixels)"""
+
+  override def canEqual(a: Any) = a.isInstanceOf[LayoutDefinition]
+
+  override def equals(that: Any): Boolean =
```
pomadchin: Actually it's a pretty good catch :+1:

pomadchin:
@echeipesh about nouns - we also have a bunch of type aliases (TileBounds = GridBounds for instance; mb there is smth else); mb it makes sense to walk though all our aliases and to clean them up?

echeipesh:
@pomadchin Yes, turning `alias TileBounds` into `case class TileBounds` should be a PR for 3.0 for sure. Not sure about what the other candidates are, but probably same logic applies.

PR #2898, opened 2019-04-12
Bump ScalaPB version up

pomadchin:
## Overview

This PR updates scalapb deps in the vectortile project, removes all generated files from the `src` folder, starting from this point everything would be generated from [this .proto schema](https://github.com/mapbox/vector-tile-spec/blob/master/2.1/vector_tile.proto) on each fresh compilation.

### Checklist

- [x] protobuf-java 3.7.0 # [CQ 19573](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=19573)
- [x] scalapb-runtime 0.9.0-RC1 # [CQ 19574](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=19574)
- [x] fastparse 1.0.0 # [19575](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=19575)
- [x] sourcecode 0.1.4 # [19578](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=19578)

Fixes https://github.com/locationtech/geotrellis/issues/2909


vectortile/data/vector_tile.proto
```diff
@@ -0,0 +1,78 @@
+package vector_tile;
+
+option optimize_for = LITE_RUNTIME;
+
+message Tile {
```
pomadchin: @jpolchlo the ways to change names for the generated items are:

1. Change the schema
2. Add type aliases

vectortile/data/vector_tile.proto
```diff
@@ -0,0 +1,78 @@
+package vector_tile;
```
pomadchin: This file is in a `data` folder to avoid its inclusion into package jars.

project/plugins.sbt
```diff
@@ -8,3 +8,5 @@ addSbtPlugin("org.scalastyle" %% "scalastyle-sbt-plugin" % "1.0.0")
 addSbtPlugin("com.sksamuel.scapegoat" %% "sbt-scapegoat" % "1.0.9")
 addSbtPlugin("com.github.sbt" % "sbt-jacoco" % "3.0.3")
 addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.1.18")
+addSbtPlugin("com.thesamet" % "sbt-protoc" % "0.99.20")
+libraryDependencies += "com.thesamet.scalapb" %% "compilerplugin" % "0.9.0-RC1"
```
pomadchin: Tbh we don't really depend on this version any more, we support all fresh versions now AFAIK

vectortile/src/main/scala/geotrellis/vectortile/Layer.scala
```diff
@@ -127,16 +127,16 @@ import geotrellis.util.annotations.experimental
...
     data: Map[String, Value]
-  ): vt.Tile.Feature = {
+  ): vector_tile.Tile.Feature = {
     val tags = data.toSeq.foldRight(List.empty[Int]) { case (pair, acc) =>
       /* These `Option.get` _should_ never fail */
       keys.get(pair._1).get :: values.get(pair._2).get :: acc
     }
 
-    vt.Tile.Feature(None, tags, Some(geomType), Command.uncommands(cmds))
+    vector_tile.Tile.Feature(None, tags, Some(geomType), Command.uncommands(cmds))
```
jpolchlo: This line assumes that features never have an associated `id`.  We need to have a means to transmit this information, but it may require something like the work in #2892 plus an id-carrying feature type.  Not saying that this PR must have this feature to pass, but if this PR doesn't address this, an issue should be made to make sure we get back to it at some point.

pomadchin: This PR is only ScalaPB update. I think that the thing you're talking about is covered by `#164804993` (pivotal tracker issue)

vectortile/src/main/scala/geotrellis/vectortile/VectorTile.scala
```diff
@@ -45,9 +45,9 @@ import geotrellis.util.annotations.experimental
   * @constructor This is not meant to be called directly - see this class's
   * companion object for the available helper methods.
   */
-@experimental case class VectorTile(layers: Map[String, Layer], tileExtent: Extent) {
+@experimental case class VectorTile(layers: Map[String, Layer], tileExtent: Extent, normalize: Boolean = true) {
```
CloudNiner: Hmmm, I'm wondering if calling this normalize is the best name for the API we want to expose, which is allowing the user to choose the polygon external winding order and defaulting to clockwise. I'm imagining a flag that's something like `windClockwise` or `winding = WindingDirection.Clockwise`. Since we only trivially get the operation for clockwise winding via the `normalize` method though, it might be extra effort to implement the CCW case which is annoying.

The docstring for `Geometry.normalize` in JTS states:
> Converts this Geometry to normal form (or canonical form). Normal form is a unique representation for Geometrys. It can be used to test whether two Geometrys are equal in a way that is independent of the ordering of the coordinates within them. Normal form equality is a stronger condition than topological equality, but weaker than pointwise equality. The definitions for normal form use the __standard lexicographical ordering for coordinates__.

(emphasis mine) which is the only place the docstring notes how this method will reorder coordinates, and I suppose in this case lexographical order means CW based on prior convention.

Given this, and that implementing CCW option would be more effort, and for now we really only care about the CW case, I'd be ok with simply adding a note to the docstring that states why this `normalize` parameter exists, what it's actually for in practice, and why you might want to turn it off, given that in practice for vector tiles I think you pretty much always want this new option set to `true`.

vectortile/src/main/scala/geotrellis/vectortile/package.scala
```diff
@@ -99,4 +99,11 @@ package geotrellis
   * @author cwoodbury@azavea.com
   * @version 2.1
   */
-package object vectortile {}
+
+import geotrellis.vector.Geometry
+
+package object vectortile {
+  implicit class withGeometryMethods[G <: Geometry](val self: G) extends AnyVal {
+    private[vectortile] def normalize(enabled: Boolean = true): G = { if(enabled) { self.jtsGeom.normalize() }; self }
```
CloudNiner: Ew. But this seems fine given the API provided to normalize.

vectortile/src/main/scala/geotrellis/vectortile/Layer.scala
```diff
@@ -78,7 +79,7 @@ import geotrellis.util.annotations.experimental
   }
 
   /** Encode this ProtobufLayer a mid-level Layer ready to be encoded as protobuf bytes. */
-  private[vectortile] def toProtobuf: vt.Tile.Layer = {
+  private[vectortile] def toProtobuf(normalize: Boolean = true): PBLayer = {
```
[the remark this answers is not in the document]
jpolchlo: In fact, I thought we discussed having `normalize=false` as the default.  Which would both be faster in the default case, but also wouldn't impose a behavioral change.  That may not entirely dodge the bullet for breaking the API, but will be code compatible.

vectortile/src/main/scala/geotrellis/vectortile/VectorTile.scala
jpolchlo: Something like "force mapbox polygon winding" is what we're trying to expose in the API.  Can we come up with a name along those lines that isn't as opaque as `normalize` (which is pretty inside baseball).  `forcePolygonWinding`?

vectortile/src/main/scala/geotrellis/vectortile/package.scala
jpolchlo: Lol.  But if it works.... The vectortile package is already a bit gross.

vectortile/src/main/scala/geotrellis/vectortile/VectorTile.scala
pomadchin: How about `forceWindClockwise`? 
I don't think it makes sense to create a group of sealed traits only for true / false flag, so we can go with boolean.

jpolchlo: Agreed on boolean flag.  Clockwise winding only applies to exterior rings, so maybe not entirely appropriate.

CloudNiner: 👍 to boolean. I'm maybe 60/40 for `forcePolygonWinding` over `forceWindClockwise` but either would be fine -- tradeoffs for both and I don't have a better, more concise name to throw in. Pick your poison!

vectortile/src/main/scala/geotrellis/vectortile/Layer.scala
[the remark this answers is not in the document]
CloudNiner: Huh. Maybe we did? False is safer from an API change, but requires the user to think more about the choice every time. Realistically, I'd expect the user to almost always want this to be on -- seems better in this case to prefer slower and 100% correct vs faster and sometimes broken?

vectortile/src/main/scala/geotrellis/vectortile/package.scala
pomadchin: Im going to remove this mess

vectortile/src/main/scala/geotrellis/vectortile/Layer.scala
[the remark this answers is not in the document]
pomadchin: tbh I though `= true` that's what we agreed on (and @CloudNiner just confirmed it)

vectortile/src/main/scala/geotrellis/vectortile/VectorTile.scala
pomadchin: Okay, will try have some pretty name for this param; `normalize` is itself so confusing.

CloudNiner:
👍

vectortile/src/main/scala/geotrellis/vectortile/VectorTile.scala
```diff
@@ -45,9 +45,9 @@ import geotrellis.util.annotations.experimental
   * @constructor This is not meant to be called directly - see this class's
   * companion object for the available helper methods.
   */
-@experimental case class VectorTile(layers: Map[String, Layer], tileExtent: Extent) {
+@experimental case class VectorTile(layers: Map[String, Layer], tileExtent: Extent, forceWindClockwise: Boolean = true) {
```
jpolchlo: The more I look at this, the more I prefer `forcePolygonWinding`.  First, clockwise winding isn't _exactly_ right, given the interior holes are wound anti-clockwise.  Second, it only applies to polygons, not to closed linestrings.  Last (and most nitpicky), it doesn't read as proper English (`forceClockwiseWinding` would have been better).

This is very pedantic, but I feel like we get tied down to these early nomenclature choices and then end up having to live with them.  See raster/spark reproject options for an example of this gone to the worst extreme.

Issue #2897, opened 2019-04-12
AccumuloAttributeStore has really poor performance


jmt-vanderlee:
Calling `availableZoomLevels(layerName: String)` on the `AccumuloAttributeStore` results in it reading all layers from Accumulo and performing a reduce in memory.

This happens per call without any sort of caching or filtering on Accumulo's side. This makes this function completely useless in a production system.

Execution path:
```
// level 1
def availableZoomLevels(layerName: String): Seq[Int] = layersWithZoomLevels(layerName)

// level 2
def layersWithZoomLevels: Map[String, Seq[Int]] = layerIds.groupBy(_.name).mapValues(_.map(_.zoom))

// level 3
def layerIds: Seq[LayerId] = {
  val scanner = connector.createScanner(attributeTable, new Authorizations())
  try {
    scanner.iterator.asScala.map { kv =>
      val Array(name, zoomStr) = kv.getKey.getRow.toString.split(SEP)
      LayerId(name, zoomStr.toInt)
    }
    .toSeq
    .distinct
  } finally {
    scanner.close()
  }
}
```

In our test environment this results in a download of 100Mb to produce a list of `[0, 1, ..., 22]`

pomadchin:
Thanks for reporting! If you want to contribute, we would be happy to navigate you throw the process. If not - we'll try to fix it in our next release. Would you also like this to be released as a part of `2.x` minor release?

PR #2899, opened 2019-04-13
Fix Accumulo and HBase attributeStore performance issues

pomadchin:
## Overview

Closes https://github.com/locationtech/geotrellis/issues/2897

For Cassandra we probably need to think a little bit about the attributeStore table, since partition key columns are not yet supported by SASI and `layerId` is a `partition key` in our case: https://github.com/apache/cassandra/blob/094689acf10bf133637c53aeebd3e8157832a279/src/java/org/apache/cassandra/index/sasi/SASIIndex.java#L148-L149

Mb @ALPSMAC has his opinion about Cassandra `LIKE` queries.


pomadchin:
Hey @jmt-vanderlee do you want to check this fix before merging?

jmt-vanderlee:
@pomadchin Sure. Looks like it'd do the trick 👍

Issue #2900, opened 2019-04-13
CassandraAttributeStore availableZoomLevels poor performance


pomadchin:
The problem described in https://github.com/locationtech/geotrellis/issues/2897 also exists for our Cassandra backend. 

We probably need to think a little bit about the attributeStore table, since partition key columns are not yet supported by SASI and layerId is a partition key in our case: https://github.com/apache/cassandra/blob/094689acf10bf133637c53aeebd3e8157832a279/src/java/org/apache/cassandra/index/sasi/SASIIndex.java#L148-L149

Mb @ALPSMAC has his opinion about Cassandra LIKE queries.

Related to https://github.com/locationtech/geotrellis/pull/2899

ALPSMAC:
@pomadchin apologies for the delay in getting back to you... been swamped here!  

The related PR ( https://github.com/locationtech/geotrellis/pull/2901 ) seems entirely reasonable to me.

When last I checked I don't believe the SASI indexing in Cassandra was believed to be really ready for "prime time" yet anyway.

pomadchin:
@ALPSMAC thanks! tbh I was excited by the SASI index :D also, FYI: https://github.com/locationtech/geotrellis/releases/tag/v2.3.0

PR #2901, opened 2019-04-14
Fix Cassandra AttributeStore performance issues

pomadchin:
## Overview

Closes https://github.com/locationtech/geotrellis/issues/2900

@ALPSMAC attention is required, since this PR changes the AttributeStore table.

Actually this can not be back ported into 2.x branch.


ALPSMAC:
Got it @pomadchin - thanks for the heads-up!

PR #2912, opened 2019-05-01
Spark Split: Tiling and Key API

jbouffard:
## Overview

This PR breaks the tiling and key API from the `spark` package into its own, new package: `geotrellis-tiling`. While not out yet, a RFP will be released that goes over this PR in more detail. This brief overview will describe this PR until then.

The following types have been moved to the `geotrellis-tiling` package:

**geotrellis.spark**:
   - SpatialKey
   - SpaceTimeKey
   - Boundable
   - Bounds
   - EmptyBounds
   - KeyBounds
   - TileBounds
   - TemporalKey
   - TemporalProjectedExtent
   - EmptyBoundsError
   - SpatialComponent 
   - TemporalComponent

**geotrellis.spark.tiling**:
   - LayoutDefinition
   - LayoutScheme
   - LayoutLevel
   - LocalLayoutScheme
   - FloatingLayoutScheme
   - ZoomedLayoutScheme
   - MapKeyTransform

### Checklist

### Notes

There is one test that is failing that I'm not sure the cause of. Might need a second pair of eyes.

Closes #XXX


pomadchin:
Just a minor thing (I haven't looked into the code yet); can you sign all your commits to make eclipse happy?

jbouffard:
@pomadchin Just signed off on all of them. I somehow forgot one

build.sbt
```diff
@@ -287,3 +287,8 @@ lazy val bench = project
   .dependsOn(spark)
   .settings(commonSettings)
   .settings(Settings.bench)
+
+lazy val `geotrellis-tiling` = project
```
pomadchin: We have no subprojects called `geotrellis-X` since all subprojects are `geotrellis` subprojects. Mb just to name it tiling?

geotrellis-tiling/src/main/scala/geotrellis/geotrellis-tiling/LayoutDefinition.scala
```diff
@@ -14,12 +14,11 @@
  * limitations under the License.
  */
 
-package geotrellis.spark.tiling
+package geotrellis.tiling
```
pomadchin: the package name is `geotrellis.tiling`, however the full path is `geotrellis/geotrellis-tiling/src/main/scala/geotrellis/geotrellis-tiling/`. 

Should be `geotrellis/tiling/src/main/scala/geotrellis/tiling/`. The package name should be the same as the directory name.

geotrellis-tiling/src/main/scala/geotrellis/geotrellis-tiling/package.scala
```diff
@@ -0,0 +1,36 @@
...
+
+package object tiling extends Implicits {
+  type TileBounds = GridBounds[Int]
+
+  type SpatialComponent[K] = Component[K, SpatialKey]
+  type TemporalComponent[K] = Component[K, TemporalKey]
+
+  private final val WORLD_WSG84 = Extent(-180, -89.99999, 179.99999, 89.99999)
+
+  implicit class CRSWorldExtent(crs: CRS) {
```
pomadchin: `extends AnyVal`?

project/Settings.scala
```diff
@@ -549,4 +549,21 @@ object Settings {
       scalapb.gen(flatPackage = true, grpc = false) -> (sourceManaged in Compile).value
     )
   )
+
+  lazy val `geotrellis-tiling` = Seq(
```
pomadchin: Look how all the projects are named, should be: 

```scala
lazy val tiling = Seq(
  name := "geotrellis-tiling",
  ...
)
```

spark/src/main/scala/geotrellis/spark/io/slippy/FileSlippyTileReader.scala
```diff
@@ -1,71 +0,0 @@
-/*
- * Copyright 2016 Azavea
```
pomadchin: So basically you just removed all the `slippy` readers and writers? Why?
The entire slippy package. 

It can be removed only for a good reason and removal probably means creating a project in `geotrellis-contrib`.

A bit more general question: why have you decided to remove it and to remove in terms of this PR?

jbouffard:
@pomadchin It seems like I misunderstood the current state of the `slippy` package. I remember you saying that we removed it, so I thought the files I was seeing there were the result of a bad rebase. I'll add those back in.

As for the naming, the was the original name of the package when I first did the spark split, and nobody objected, so I kept it. I do agree that just `tiling` would be better.

pomadchin:
@jbouffard `slick` is wrongly rebased here: https://github.com/locationtech/geotrellis/pull/2891/files#diff-c2720ede98ad167959c38e1f34fbfb17R17

jbouffard:
@pomadchin I see. I didn't realize that it was just a portion of the slippy support.

geotrellis-tiling/src/main/scala/geotrellis/geotrellis-tiling/package.scala
jbouffard: Would that be beneficial? We don't have that in our current code.

pomadchin: Yes https://docs.scala-lang.org/overviews/core/value-classes.html 
It won't be bad to add it, but not it's an obligatory thing in terms of this PR for sure.

build.sbt
jbouffard: :+1:

pomadchin:
Yeah, just for clarification `slick` != `slippy` at all.

jbouffard:
@pomadchin Thanks! Got the `CHANGELOG` updated and was able to get that test to pass; so it should be good to merge :+1:

spark/src/main/scala/geotrellis/spark/io/slippy/FileSlippyTileReader.scala
```diff
@@ -16,16 +16,27 @@
...
 import org.apache.spark._
 import org.apache.spark.rdd._
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.mapreduce.Job
+import org.apache.hadoop.mapreduce.{InputSplit, TaskAttemptContext}
+import org.apache.hadoop.fs.Path
 
 import java.io._
-import scala.collection.JavaConverters._
+import scala.collection.JavaConversions._
```
pomadchin: It is deprecated, can you revert your changes back here and use `scala.collection.JavaConverters._`

spark/src/main/scala/geotrellis/spark/io/slippy/FileSlippyTileReader.scala
```diff
@@ -16,16 +16,27 @@
 
 package geotrellis.spark.io.slippy
 
+import geotrellis.vector._
+import geotrellis.raster._
```
pomadchin: hm why so many unused imports appeared in all slippy files? O:

jbouffard: I'm not sure how those got there....

spark/src/main/scala/geotrellis/spark/Implicits.scala
```diff
@@ -267,4 +268,11 @@ trait Implicits
       TileLayerMetadata.fromRDD[K1, V, K2](rdd, layout)
     }
   }
+
+  implicit class withGetBoundsMethod[K: Boundable, V <: CellGrid[Int]](rdd: RDD[(K, V)]) {
```
pomadchin: Wondering is it used anywhere?

jbouffard: I don't think it is. That method used to be [on KeyBounds](https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/KeyBounds.scala#L95-L98), so I just moved it in order to retain the functionality.

echeipesh:
One place we safe the FQN of `SpatialKey` and `SpaceTimeKey` is in the `metadata` attribute for GeoTrellis Avro layers:

```json
    "header": {
      "format": "s3",
      "key": "catalog/us-census-property-value-30m-epsg3857/0",
      "keyClass": "geotrellis.spark.SpatialKey",
      "valueClass": "geotrellis.raster.Tile",
      "bucket": "azavea-datahub"
    }
```

This PR will obviously change the value written to `valueClass`. Is it going to effect the ability to read previously written layers?

jbouffard:
@echeipesh Yes, changing the class path of `SpatialKey` and `SpaceTimeKey` will cause any old layer to be unreadable. This is due to [changing the class path](https://github.com/locationtech/geotrellis/blob/cd11fae4c0b9a95ccb65bb9544c684ebfddf6265/spark/src/main/scala/geotrellis/spark/io/avro/codecs/KeyCodecs.scala#L28) in the Avro records. @pomadchin suggested that we could add legacy codecs in contrib so that users could load those in if they need them. I'm on board with this idea, as I don't really see any other way we could support both.

jbouffard:
**A Note for Posterity:** @pomadchin, @echeipesh, and I were able to determine that changing the Avro namespace for `SpatialKey` and `SpaceTimeKey` does not seem to affect the reading of old Avro and COG layers. Therefore, this PR should be good to merge.

However, if in the future there are issues with reading old layers, then this change might be worth investigating as a possible cause.

jpolchlo:
Why does this depend on raster?  Seems like we should be able to do pure vector projects that tile their outputs (vector tile production, for example) that don't rely at all on rasters, and therefore, don't require the raster dependency.

pomadchin:
@jpolchlo at least `MapTransform` object

jbouffard:
@echeipesh Understood - I'll squash now :+1: @jpolchlo As mention by @pomadchin, it's because of `MapKeyTransform` and `ZoomedLayoutScheme`. I think there needs to be another issue that covers the uncoupling of the `raster` dependency.

PR #2920, opened 2019-05-13
PolygonalSummary API refactor

### LINKED ISSUE #2914: Polygonal Summary Improvements
# Problem

Current Implementation of polygonal summary has several issues that should be addressed:

## TilePolygonalSummaryHandler handles three things
https://github.com/locationtech/geotrellis/blob/fb6f94077f932f9b60ff0a5a8c97ec66b77040dc/raster/src/main/scala/geotrellis/raster/summary/polygonal/TilePolygonalSummaryHandler.scala#L27

Conflates the logic of checking for polygon/raster intersection, generating the result from that intersection, and combining partial results when generating summary from chunked source like `RDD`

Rasterizer will match and work on any geometry type, so specificity in handler is not needed.

Also now that we have made a decision to lean on more functional constructs the combination the job of combining result type should be handled by either `Monoid[T]` or `Semigroup[T]`

## TilePolygonalSummaryHandler does not expose Rasterizer Options
Because each handler is expected to implement its own intersection logic there is no way to pass in [`Rasterizer.Options`](https://github.com/locationtech/geotrellis/blob/fb6f94077f932f9b60ff0a5a8c97ec66b77040dc/raster/src/main/scala/geotrellis/raster/rasterize/Rasterizer.scala#L42-L45) that can control for intersection type. The only way around this is to create a new subclass of handler and it becomes very verbose for the user.

Example: https://github.com/locationtech/geotrellis/blob/fb6f94077f932f9b60ff0a5a8c97ec66b77040dc/raster/src/main/scala/geotrellis/raster/summary/polygonal/MaxSummary.scala#L25-L34

# Prototype

A prototype for alternative approach has been developed here:

https://github.com/geotrellis/geotrellis-contrib/tree/master/summary/src/main/scala/geotrellis/contrib/polygonal

It revolves around idea of having the rasterizer driving around a `CellVisitor` to produce the summary result:
https://github.com/geotrellis/geotrellis-contrib/blob/a9803a0007b1e4e74969e5625613a683f98719cd/summary/src/main/scala/geotrellis/contrib/polygonal/CellVisitor.scala#L8-L10

This approach should be applied to the geotrellis-raster polygonal summary in hopes that it makes things better.

# Notes
- Polygonal summary can be seen as a form of .fold, can API express that clearly?
- `Semigroup.combine` is not required when performing summary on a single raster
- `RDD[SpatialKey, Tile].polytonalSummary(RDD[Polygon])` operations are not support and are not part of this issue, but are desirable

There are some inconsistencies in the way method extensions are implemented in the prototype vs the geotrellis core. At this point I'm not sure if they are intentional or incidental, so they should be reviewed in process of working on this issue.

CloudNiner:
## Overview

This is a first draft attempt at this refactor, with a few simple example implementations in TileVisitors.scala and MultibandTileVisitors.scala for Histogram and Max. This work was done with consult from @echeipesh, starting from the prototype implementation in geotrellis-contrib.

The entirety of the subproject won't work as there's additional methods to refactor, remove and re-implement. Example usages of the new API are in the tests, specifically `geotrellis.raster.summary.polygonal.{HistogramSpec, MaxSpec}`

Additional comments for consideration in code.

Open TODOs:
- [x] Implement remaining visitors
- [x] Finish refactoring tests once final implementation decided
- [x] Sign that CLA
- [ ] Companion PR in geotrellis-contrib to remove PolygonalSummary methods there, with a note about them being moved to core

### Checklist

### Notes

Connects #2914


spark/src/main/scala/geotrellis/spark/io/kryo/KryoRegistrator.scala
```diff
@@ -187,7 +187,8 @@ class KryoRegistrator extends SparkKryoRegistrator {
     kryo.register(classOf[geotrellis.raster.histogram.StreamingHistogram.Delta])
     kryo.register(classOf[geotrellis.raster.histogram.StreamingHistogram.Bucket])
     kryo.register(classOf[geotrellis.raster.density.KernelStamper])
-    kryo.register(classOf[geotrellis.raster.summary.polygonal.MeanResult])
+    kryo.register(classOf[geotrellis.raster.summary.polygonal.visitors.MaxVisitor.TileMaxVisitor])
```
CloudNiner: Do all of these new classes need to be added to the KryoRegistrator? This would include the concrete visitor implementation classes for:
- Min
- Max
- Sum
- Mean
- StreamingHistogram
For both Tile and MultibandTile.

CloudNiner:
@echeipesh this could use a look. The `raster` package compiles and all tests pass successfully. A fresh and critical set of eyes would be useful so that I can address any major missed issues before writing up changelog notes and moving onto the spark half. There's a short summary of the highlights of the changes so far  in the commit message for 2ab1cd341a1c589125de02a0728230f10739bf21.

The entire library does not compile ATM, as the `spark` package polygonal summary methods depend on the old implementations. So we'll probably want to refactor those in a separate branch, PR those changes here, and then merge this once the whole deal compiles and passes tests.

echeipesh:
Taking a look for review now. However, I agree that we shouldn't be committing non-compiling code, which means we'll have to wait for spark changes before merging this into master as you say...

raster/src/main/scala/geotrellis/raster/summary/polygonal/visitors/MaxVisitor.scala
```diff
@@ -0,0 +1,40 @@
...
+package geotrellis.raster.summary.polygonal.visitors
+
+import geotrellis.raster._
+import geotrellis.raster.summary.CellVisitor
+
+object MaxVisitor {
+  implicit def toTileVisitor(t: MaxVisitor.type): TileMaxVisitor =
+    new TileMaxVisitor
+  implicit def toMultibandTileVisitor(
+      t: MaxVisitor.type): MultibandTileMaxVisitor = new MultibandTileMaxVisitor
```
echeipesh: Weird indentation is weird. Super-nit-picky but what is this intentional?

raster/src/main/scala/geotrellis/raster/summary/polygonal/Implicits.scala
```diff
@@ -17,16 +17,30 @@
...
-  ) extends MultibandTilePolygonalSummaryMethods
+  implicit class withPolygonalSummaryResultMethods[A](val self: PolygonalSummaryResult[A]) {
+    def toEither: Either[NoIntersection.type, A] = {
+      self match {
+        case NoIntersection => Left(NoIntersection)
+        case Summary(value) => Right(value)
+      }
+    }
+
+    def toOption: Option[A] = {
```
echeipesh: Both `toEither` and `toOption` should be directly on `PolygonalSummaryResult`. They make perfect sense as part of the class, given a chance to convert to standard type if more detail is not helpful.

From experience it is tempting to put things that don't seem core in implicit extension methods, but you still end up with tight binding between the method and the interface due to SemVer constraints and an obscure location for the functionality to boot.

raster/src/main/scala/geotrellis/raster/summary/polygonal/PolygonalSummaryResult.scala
```diff
@@ -0,0 +1,8 @@
+package geotrellis.raster.summary.polygonal
```
echeipesh: Missing license header

raster/src/main/scala/geotrellis/raster/summary/polygonal/PolygonalSummaryResult.scala
```diff
@@ -0,0 +1,8 @@
...
+
+sealed trait PolygonalSummaryResult[+A]
```
echeipesh: Missing scala-doc

raster/src/main/scala/geotrellis/raster/summary/polygonal/PolygonalSummaryResult.scala
```diff
@@ -0,0 +1,8 @@
...
+
+case object NoIntersection extends PolygonalSummaryResult[Nothing]
+case class Summary[A](value: A) extends PolygonalSummaryResult[A]
```
echeipesh: Are there more states that this is going to represent? Just two seem to map so tightly on `Option` that it raises an eye-brow. I know there are some discussions that have been had on `Option` vs ADT so I may be missing future plans here.

raster/src/main/scala/geotrellis/raster/summary/CellVisitor.scala
```diff
@@ -0,0 +1,34 @@
...
+package geotrellis.raster.summary
+
+/** Visitor for cell values of T that should record their state of R
+  *
+  * The user should implement concrete subclasses that update the value of `result` as
+  * necessary on each call to `visit(raster: T, col: Int, row: Int)`.
+  *
+  * Be sure to handle the empty state. This could occur if no points in T are ever visited.
+  */
+trait CellVisitor[-T, R] {
```
echeipesh: To refresh my memory why did we settle on contra-variance of `T`? Might be worth while calling out in the docs as it would impact its usage.

raster/src/main/scala/geotrellis/raster/summary/polygonal/visitors/TileFoldingVisitor.scala
```diff
@@ -0,0 +1,37 @@
...
+abstract class TileFoldingVisitor
+    extends CellVisitor[Raster[Tile], Option[Double]] {
+  private var accumulator: Option[Double] = None
+
+  def result: Option[Double] = accumulator
+
+  def visit(raster: Raster[Tile], col: Int, row: Int): Unit = {
+    val newValue = raster.tile.getDouble(col, row)
+    accumulator = result match {
+      case Some(accum) if isData(newValue) => Some(fold(accum, newValue))
```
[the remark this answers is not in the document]
pomadchin: @echeipesh we probably need to create a task to update https://github.com/geotrellis/geotrellis-benchmark/tree/master/geotrellis/src/test/scala/geotrellis/raster/mapalgebra/focal and move into contrib(?)

raster/src/main/scala/geotrellis/raster/summary/polygonal/PolygonalSummaryResult.scala
pomadchin: `$ sbt project raster headerCreate` command

raster/src/main/scala/geotrellis/raster/summary/polygonal/visitors/TileFoldingVisitor.scala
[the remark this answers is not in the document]
echeipesh: Not sure, we have `bench` project in GeoTrellis. Would you suggest keeping them in separate repo?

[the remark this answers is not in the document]
pomadchin: We can do both; According to the README, we deprecated this project, since it uses `scaliper` that is deprecated and not maintained. The replacement is a bench project in geotrellis-contrib (again, according to the README).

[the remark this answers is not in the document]
echeipesh:
```
[info] Benchmark                            Mode  Cnt  Score   Error  Units
[info] PolygonalSummaryBench.cleanVisitor   avgt    3  0.075 ± 0.045   s/op
[info] PolygonalSummaryBench.frugalVisitor  avgt    3  0.017 ± 0.022   s/op
```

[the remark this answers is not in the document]
echeipesh: https://github.com/echeipesh/geotrellis/tree/bench/polygonal-summary-refactor

raster/src/main/scala/geotrellis/raster/summary/polygonal/visitors/MaxVisitor.scala
CloudNiner: Yeah this is weird. Not intentional. Must have missed one with scalafmt.

raster/src/main/scala/geotrellis/raster/summary/polygonal/visitors/TileFoldingVisitor.scala
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package geotrellis.raster.summary.polygonal.visitors
+
+import geotrellis.raster._
+import geotrellis.raster.summary.CellVisitor
+
+abstract class TileFoldingVisitor
```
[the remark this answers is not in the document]
CloudNiner: So that brings up a question. Should this be public? We could make it private if we don't want to support it.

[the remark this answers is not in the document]
CloudNiner: But otherwise yes, this needs some docs. Planning to add those for everything if we're ok with the general look/feel of the API changes here.

raster/src/main/scala/geotrellis/raster/summary/polygonal/visitors/TileFoldingVisitor.scala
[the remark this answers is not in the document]
CloudNiner: Ah. Interesting. So that benchmark is saying that frugalVisitor is something like 4x faster in terms of seconds per operation? Definitely worth a refactor.

raster/src/main/scala/geotrellis/raster/summary/polygonal/visitors/MaxVisitor.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  implicit def fold(max: Double, newValue: Double): Double =
```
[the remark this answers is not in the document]
CloudNiner: Whoops, good catch will remove.

raster/src/main/scala/geotrellis/raster/summary/polygonal/visitors/MultibandTileFoldingVisitor.scala
```diff
@@ -0,0 +1,45 @@
...
+abstract class MultibandTileFoldingVisitor
+    extends CellVisitor[Raster[MultibandTile], Array[Option[Double]]] {
+
+  private var accumulator = Array[Option[Double]]()
+
+  def result: Array[Option[Double]] = accumulator
+
+  def visit(raster: Raster[MultibandTile], col: Int, row: Int): Unit = {
+    val tiles = raster.tile.bands.toArray
+    val values: Array[Option[Double]] = result ++ Array.fill[Option[Double]](
```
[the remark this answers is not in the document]
CloudNiner: Yeah I wasn't sure how to best implement this. We need to extract the tile count but only the first time we visit. Perhaps we can use some internal state to track this better. I'll followup offline about this one either way.

raster/src/main/scala/geotrellis/raster/summary/polygonal/Implicits.scala
CloudNiner: Roger.

raster/src/main/scala/geotrellis/raster/summary/polygonal/PolygonalSummaryResult.scala
CloudNiner: No, @pomadchin @moradology and I weren't able to theorycraft additional states that this might represent in the future (but it could). The key thing that this provides is that it provides additional clarity into the result types since the NoIntersection case isn't really an error, its a valid result state. It also avoids weird types such as `Option[Option[Double]]` for the `TileFoldingVisitor` class of summaries. Instead you get `PolygonalSummaryResult[Option[Double]]` which we all though provided more clarity.

raster/src/main/scala/geotrellis/raster/summary/CellVisitor.scala
CloudNiner: Oof. Also stretching my memory here. Might have to discuss offline to build some notes for discussion in docstrings. I remember some trouble with `-CellGrid[Int]` but not specifically what it was. And users could theoretically construct a CellVisitor for something that isn't a Tile. It's likely a good idea to think a bit more about where this trait belongs and how generic we want it to be. In its current form it could be implemented for basically any two dimensional array T that contains values of R, not just tiles.

CloudNiner: Renamed `CellVisitor` to `GridVisitor` because our only constraint on T is that it contain cols and rows that can be iterated over and indexed into, which Grid provides. T is also now bounded on Grid[Int]. There are some new docstrings that attempt to explain why this is -T and +R, similar to Function1. @moradology and I figured that it made sense to include some external links here because we couldn't explain the reasoning in a more succinct way (and general theory doesn't typically have a place in these types of docstrings). Happy to update with any feedback you might have, this was a tough one to explain.

raster/src/main/scala/geotrellis/raster/summary/polygonal/visitors/TileFoldingVisitor.scala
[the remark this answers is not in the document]
echeipesh: Yes, I suspect the array case would be much bigger difference as well.

raster/src/main/scala/geotrellis/raster/summary/polygonal/visitors/TileFoldingVisitor.scala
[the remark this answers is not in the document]
echeipesh: I'm fine with making it `private`. It's a little defensive but there is not so much and such useful logic in there that it can't be easily replaced in place by user.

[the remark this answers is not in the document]
CloudNiner: Yeah that's what I went with actually. It's fairly specific actually. The only thing you could really use it for is stddev and a few other basic aggregation functions. Which if someone wants we should probably implement here like the others.

raster/src/main/scala/geotrellis/raster/summary/polygonal/visitors/TileFoldingVisitor.scala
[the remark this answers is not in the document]
CloudNiner: Alright. With some changes to the internals of the MultibandTileFoldingVisitor, I was able to get the benchmark down from ~0.5s/op to the following:
```
[info] Benchmark                               Mode  Cnt  Score   Error  Units
[info] PolygonalSummaryBench.cleanVisitor      avgt    3  0.009 ± 0.026   s/op
[info] PolygonalSummaryBench.multibandVisitor  avgt    3  0.047 ± 0.014   s/op
```
These two benchmarks run on:
- raster/data/geotiff-test-files/overviews/singleband.tif
- raster/data/geotiff-test-files/overviews/multiband.tif

Which are the same RGBA four band tif, but the singleband one only has the R band. So the two tests are as equivalent as can be. We'd expect the multiband version to be slower and scale a bit with the number of bands read, so I think 0.009 vs 0.047 is about as good as we're gonna get for now.

raster/src/main/scala/geotrellis/raster/summary/polygonal/visitors/MultibandTileFoldingVisitor.scala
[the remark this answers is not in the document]
CloudNiner: See https://github.com/locationtech/geotrellis/pull/2920#discussion_r291314572 above.

CloudNiner:
@echeipesh this is ready for another look. Improvements abound in the newest five commits on the branch, starting with c393a3fc0f24d7ec3e98193ee92a5ac659c5d96c.

I pulled in the bench project changes from the branch on your fork. I'm wondering if I should revert all the changes in https://github.com/locationtech/geotrellis/pull/2920/commits/18ee4254cda07cabcd353a413a61cd08dcf0a657 _except_ for the addition of the PolygonalSummaryBench class. It seems like fixing up the bench project is a separate task assuming we ever want to tackle it based on Grishas comments about the move to contrib.

I'm going to start thinking about the Spark polygonal summary API and just let this sit for now. Please comment with any other desired changes in the meantime.

raster/src/main/scala/geotrellis/raster/summary/GridVisitor.scala
```diff
@@ -0,0 +1,64 @@
...
+  * val DoubleVisitor: GridVisitor[CellGrid[Int], Double] = ???
+  * def usesRaster(visitor: GridVisitor[Raster[Tile]], Any]): Unit = ???
+  * usesRaster(DoubleVisitor)
+  *
+  * will compile. For an alternate explanation on covariance and contravariance that might help
+  * to clarify this, see:
+  * - https://stackoverflow.com/a/10604305
+  * - https://stackoverflow.com/a/38577878
+  */
+trait GridVisitor[-T <: Grid[Int], +R] {
```
pomadchin: It looks like it could be relaxed to `-T <: Grid[_]` (everywhere in this PR we can be not that strict, at least without any obvious reason)

raster/src/main/scala/geotrellis/raster/summary/polygonal/PolygonalSummaryResult.scala
```diff
@@ -0,0 +1,67 @@
...
+  *
+  * Provides helpers to convert to Option and Either if you don't need to match
+  * on result type.
+  */
+sealed trait PolygonalSummaryResult[+A] {
+  def toOption: Option[A]
+
+  def toEither: Either[Any, A]
+
+  implicit val noIntersectionMonad: Monad[PolygonalSummaryResult] = new Monad[PolygonalSummaryResult] {
```
pomadchin: A better place for it in `PolygonalSummaryResult` companion object.

raster/src/test/scala/geotrellis/raster/io/geotiff/reader/GeoTiffReaderSpec.scala
```diff
@@ -349,11 +351,17 @@ class GeoTiffReaderSpec extends FunSpec
     val MeanEpsilon = 1e-8
 
     def testMinMaxAndMean(min: Double, max: Double, mean: Double, file: String) {
-      val SinglebandGeoTiff(tile, extent, _, _, _, _) = SinglebandGeoTiff(file)
+      import geotrellis.raster.summary.polygonal.Implicits._
```
pomadchin: are not these implicits imported somehow netter (like via `geotrellis.raster.summary.polygonal._` or `geotrellis.raster.summary._`)?

spark/src/main/scala/geotrellis/spark/io/kryo/KryoRegistrator.scala
pomadchin: If they are serializable via Java serialization - then nope.

raster/src/main/scala/geotrellis/raster/summary/GridVisitor.scala
CloudNiner: Great point -- I believe we can. I'll make that change.

raster/src/main/scala/geotrellis/raster/summary/polygonal/PolygonalSummaryResult.scala
CloudNiner: Ha yeah I realized this later, this change is in the Spark API companion PR over at cloudniner/geotrellis#2.

raster/src/test/scala/geotrellis/raster/io/geotiff/reader/GeoTiffReaderSpec.scala
CloudNiner: I suppose I might have been thinking that importing here avoided polluting the namespace everywhere when the extra methods are only required here. But I suppose that's not really a concern. In any case I'm pretty sure the imports could use some cleanup -- it's on my list of things to tackle before finally merging this.

raster/src/main/scala/geotrellis/raster/summary/polygonal/PolygonalSummaryResult.scala
```diff
@@ -0,0 +1,82 @@
...
+  * This ADT will always return Summary if the Raster and Polygon intersect.
+  * Otherwise, NoIntersection will be returned.
+  *
+  * Provides helpers to convert to Option and Either if you don't need to match
+  * on result type.
+  */
+sealed trait PolygonalSummaryResult[+A] {
+  def toOption: Option[A]
+
+  def toEither: Either[Any, A]
```
echeipesh: `Either[Any, A]` is extremely weird. The point of `Either` is to preserve type information between two outcomes that don't have a useful upper bound. Placing `Any` on the left side basically erases any type information. Just looking at it maybe the better thing is `Either[NonIntersecting, A]` .. that `NonIntersecting` have to become `GeometryError` later if we try to capture JTS errors that result when trying to intersect invalid geometries.

If there isn't strong need for `Either` in this case we could also just choose to dump this method until there is.

raster/src/main/scala/geotrellis/raster/summary/GridVisitor.scala
```diff
@@ -0,0 +1,72 @@
...
+  * val DoubleVisitor: GridVisitor[CellGrid[Int], Double] = ???
+  * def usesRaster(visitor: GridVisitor[Raster[Tile]], Any]): Unit = ???
+  * usesRaster(DoubleVisitor)
+  *
+  * will compile. For an alternate explanation on covariance and contravariance that might help
+  * to clarify this, see:
+  * - https://stackoverflow.com/a/10604305
+  * - https://stackoverflow.com/a/38577878
+  */
+trait GridVisitor[-T <: Grid[Int], +R] extends Serializable {
```
echeipesh: Do you think its worth while bumping this up to `geotrellis.raster` ?

raster/src/main/scala/geotrellis/raster/summary/GridVisitor.scala
echeipesh: Actually the `def visit(grid: T, col: Int, row: Int): Unit` implies that it has to be `Grid[Int]` and then the Rasterizer would only work on `(Int, Int)` grid.

So far semantically I've been using:
-  `Grid[Int]`  == Something that can be held in memory and addressed
- `Grid[Long]` == Something that is much larger, distributed and just describes the overall grid

So actually I'm not sure that it would be good idea to relax the constraint and I think you're going to run into a lot of problems as soon as you do it.

spark/src/main/scala/geotrellis/spark/summary/polygonal/RDDPolygonalSummary.scala
```diff
@@ -0,0 +1,143 @@
...
+                            visitor: GridVisitor[Raster[T], R],
+                            options: Rasterizer.Options)(
+        implicit semigroupR: Semigroup[R], tagR: ClassTag[R]): RDD[Feature[Geometry, PolygonalSummaryResult[R]]] = {
+      RDDPolygonalSummary(self, geometryRdd, visitor, options)
+    }
+
+    def polygonalSummary[R](geometries: Seq[Geometry],
+                            visitor: GridVisitor[Raster[T], R],
+                            options: Rasterizer.Options)(
+        implicit semigroupR: Semigroup[R], tagR: ClassTag[R], sc: SparkContext): RDD[Feature[Geometry, PolygonalSummaryResult[R]]] = {
```
echeipesh: This signature can be simplified by removing `sc: SparkContext`
Each `RDD` instance has `sparkContext` reference. We've been using this trick more consistently and the result has been tighter function signatures, which is good all around.

spark/src/main/scala/geotrellis/spark/summary/polygonal/RDDPolygonalSummary.scala
```diff
@@ -0,0 +1,143 @@
...
+    // Geometries may have been split across SpatialKey leading to multiple UUID keys.
+    // Combine by UUID and then drop the UUID, which was added within this method
+    featuresWithSummaries
+      .reduceByKey { (a, b) => Feature(a.geom, a.data.combine(b.data)) }
+      .map { _._2 }
+  }
+
+  trait RDDPolygonalSummaryMethods[R, T <: CellGrid[Int]]
+      extends MethodExtensions[RDD[(SpatialKey, T)] with Metadata[TileLayerMetadata[SpatialKey]]] {
+    def polygonalSummary[R](geometryRdd: RDD[Geometry],
```
echeipesh: These signatures look good. They need some minimal scaladoc, possibly with `@see RDDPolygonalSummary`. These are going to be the things that show up in tools like IDEA and Metals and we can do a better job providing some hints there.

raster/src/main/scala/geotrellis/raster/summary/GridVisitor.scala
pomadchin: @echeipesh Oops I didn't notice the `def visit(grid: T, col: Int, row: Int)` notion. Thanks. If it would be required, we'll have to add an extra hole into the `GridVisitor` declaration.

echeipesh: I don't think thats worth it to add the type parameter. `Long`, `Long` addressable tiles would require a whole new addressing and paging strategy (since they're larger than memory) which doesn't exist in away that would be transparent to the `visitor`.

spark/src/main/scala/geotrellis/spark/Implicits.scala
```diff
@@ -65,7 +65,6 @@ trait Implicits
     with split.Implicits
     with stitch.Implicits
     with summary.Implicits
-    with summary.polygonal.Implicits
     with tiling.Implicits
```
echeipesh: @CloudNiner The implicit was here to allow using the functionality with `import geotrellis.spark._` and keep the number of imports required to use GeoTrellis functionality down. Is this intentional move that was talked about that just goes against that?

build.sbt
```diff
@@ -99,6 +99,7 @@ lazy val commonSettings = Seq(
           }
         }.getOrElse(newText)
       } } )),
+  scapegoatVersion in ThisBuild := "1.3.8",
```
echeipesh: `scapegoatVersion` making its way back

raster/src/main/scala/geotrellis/raster/summary/GridVisitor.scala
CloudNiner: Makes sense. So we leave as Grid[Int] for now.

spark/src/main/scala/geotrellis/spark/Implicits.scala
CloudNiner: This one was removed because there no longer is a `geotrellis.spark.summary.polygonal.Implicts`. That entire file was only used by the old polygonal summary methods so it was removed.

raster/src/test/scala/geotrellis/raster/io/geotiff/reader/GeoTiffReaderSpec.scala
CloudNiner: Updated with this line deleted. I added the Implicits to the package.scala awhile back so this import is superfluous.

spark/src/main/scala/geotrellis/spark/summary/polygonal/RDDPolygonalSummary.scala
```diff
@@ -0,0 +1,143 @@
...
+    val joinedRdd = keyedGeometryRdd.join(rasterRdd)
+    val featuresWithSummaries: RDD[(UUID, Feature[Geometry, PolygonalSummaryResult[R]])] =
+      joinedRdd.map { data => {
+        val spatialKey = data._1
+        val feature: Feature[Geometry, UUID] = data._2._1
+        val tile: T = data._2._2
+        val extent: Extent = layout.mapTransform.keyToExtent(spatialKey)
+        val raster: Raster[T] = Raster(tile, extent)
+        val result: PolygonalSummaryResult[R] =
+          raster.polygonalSummary(feature.geom, visitor.getClass.newInstance, options)
```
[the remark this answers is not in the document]
CloudNiner: So this is getting a new instance of the visitor class, not of type R. I'm not sure how else to specify the visitor class you want to use and then also be able to construct new instances of it internally. Perhaps we could add a `def reset()` method to GridVisitor that should be implemented so that we can always use the passed instance? That feels pretty icky though because then we're getting farther into the weeds of mutating the internal state of a thing in a distributed environment. 

> we can use Monoid[R].empty to get a new instance of visitor

It's unclear to me how that would work. Could you perhaps explain what you're thinking in a different way?

raster/src/main/scala/geotrellis/raster/summary/polygonal/PolygonalSummaryResult.scala
CloudNiner: Whoops. Missed this one in my cleanup pass. Updated to NoIntersection.

spark/src/main/scala/geotrellis/spark/summary/polygonal/RDDPolygonalSummary.scala
CloudNiner: TIL.

raster/src/main/scala/geotrellis/raster/summary/GridVisitor.scala
CloudNiner: So...I had it there. Then I moved it because I think I made the argument with myself that it's primary use is reducing values within the raster to single result -- a summary. 

I don't feel strongly about it either way though so if you think its more discoverable/reusable up a level in `geotrellis.raster` I'm happy to move it.

CloudNiner:
@echeipesh I resolved comment threads for everything I addressed and left comments on the two remaining items. Back to you 🏓

raster/src/main/scala/geotrellis/raster/summary/GridVisitor.scala
echeipesh: I think your internal argument is good enough for me, lets leave it in here.

spark/src/main/scala/geotrellis/spark/summary/polygonal/RDDPolygonalSummary.scala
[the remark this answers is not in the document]
echeipesh: Right .. you're totally right. Thats what I get while reviewing PRs and doing something else.

I believe this was in response to the problem of having a visitor with some initial state being passed into the the summary and skewing the results in a surprising way?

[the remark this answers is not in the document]
echeipesh: I don't know that there is a great way out of this. 

Adding `def map(v: R): R` to `GridVisitor` might be one way, that way you can get back to use the `Monoid[R].empty`. Adding `Functor[GridVisitor[Raster[T], R]` would be another, more abstract way to add that constraint.

Actually all of the `R` instances already have `Monoid[R]` so you're only a step away from `Monoid[GridVisitor[Raster[T], R]` .. which is super weird because its a functional composition over dirty mutable state. This may have pay-off that some point that is not now.

I guess the thing to do is document the requirement for zero-argument constructor in `GridVisitor` doctoring and see if it ever comes up again.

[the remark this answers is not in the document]
CloudNiner: > this was in response to the problem of having a visitor with some initial state being passed into the the summary

Yup. The visitor internal state wasn't reset on each iteration of the polygonal summary call, which we make a number of times for each combination of spatial key, raster and geometry. 

> I guess the thing to do is document the requirement for zero-argument constructor in GridVisitor doc-string and see if it ever comes up again.

Seems like the best out ATM. I'm on board with that. I like that better than some workaround that leads to weird function signatures and more complicated implementation requirements for an optimization we may never need.

raster/src/main/scala/geotrellis/raster/summary/GridVisitor.scala
CloudNiner: 👍

PR #2932, opened 2019-05-23
Use JTS geometries in vector package

### LINKED ISSUE #2930: Use JTS types directly in geotrellis.vector
The current state of the vector package is workable, but the decision to wrap JTS geometries is introducing friction as we interact more with other packages in the LocationTech ecosystem.  Notably, GeoMesa has provided JTS types for use with Spark DataFrames, which requires that we swap back and forth between GT and JTS types in its use.

It seems clear that, since we are introducing very little new functionality with GT vector types, a more complete adoption of JTS is a smart move.  The trouble is that JTS isn't always the easiest to use in an idiomatic Scala way.  Object creation relies on factory classes, which is more cumbersome than Scala object methods.  Pattern matching isn't well supported natively for operations like `intersection`, which benefit from its inclusion.

The general approach to this transition will be to provide object-level funtions and method extensions to deliver the ease-of-use improvements currently available in `geotrellis.vector` without masking the native JTS types.  The former can be used in nearly any case, while the latter should be used sparingly, so as to not cause confusion when working in a pure JTS environment.

### LINKED ISSUE #2931: Prototype using native JTS types
Prototype need to determine feasibility and scope of #2930

Some of the outstanding questions:

- [ ] What impact does this have on GeoTrellis interfaces and implementations ?
- [ ] Is it possible to use the type-safe wrappers optionally ?
- [ ] What is the mechanism for optionally using current type-safe wrappers ?
- [ ] What are likely implications on user-code, will everything become `Geometry` ?

### LINKED ISSUE #2955: Push JTS vector changes into all of GT
As of now, #2932 gives a prototype implementation for replacing GT wrappers of JTS geometry types with the raw JTS types.  This work needs to be pushed into the remaining packages of GT.

This issue exists for planning purposes.

jpolchlo:
## Overview

Interactions with other systems are making our use of wrappers around JTS geometry a bit cumbersome.  This PR takes a crack at removing our wrappers.  It favors outright deletion of easy-to-implement features, replacement of useful class methods with object methods, and sparing use of method extensions to foster ease-of-use in common pathways.

Closes #2930
Closes #2931
Closes #2955 

### Checklist


jpolchlo:
In the interest of furthering a new RFP process for major changes or new features that we're trying out, I'd like to post my plans for this work, in case there is useful feedback.  This will call out some of the salient features of the experimental code.

### Proposed deletions ###

Here is the subset of the vector package that will be removed:

```
vector/src/main/scala/geotrellis/vector/
├── Dimension.scala
├── GeometryCollectionBuilder.scala
├── GeometryCollection.scala
├── affine
│   ├── AffineTransformation.scala
│   ├── GeometryCollectionTransformationMethods.scala
│   ├── LineTransformationMethods.scala
│   ├── MultiLineTransformationMethods.scala
│   ├── MultiPointTransformationMethods.scala
│   ├── MultiPolygonTransformationMethods.scala
│   ├── PointTransformationMethods.scala
│   └── PolygonTransformationMethods.scala
├── convexhull
│   └── ConvexHullMethods.scala
├── densify
│   └── DensifyMethods.scala
├── dissolve
│   └── DissolveMethods.scala
├── prepared
│   ├── PreparedGeometryMethods.scala
│   └── PreparedGeometry.scala
└── simplify
    └── SimplifyMethods.scala
```
The sub-packages to be removed add little if anything in the way of functionality.  We will favor using JTS directly for these features.  Obviously, the basic geometry types will be removed; however, I have opted to this point to keep the companion objects to aid in object creation.  This does create some type weirdness, though, in some cases requiring importing JTS into a `jts` alias to prevent confusion.

### Proposed Preservations ###

```
vector/src/main/scala/geotrellis/vector/
├── Extent.scala
├── Feature.scala
├── GeomFactory.scala
├── package.scala
├── Projected.scala
├── Results.scala
├── RobustPredicates.scala
├── SeqMethods.scala
├── SpatialIndex.scala
├── conf
│   └── JtsConfig.scala
├── interpolation
│   ├── EmpiricalVariogram.scala
│   ├── GeoKriging.scala
│   ├── InterpolationVector.scala
│   ├── Kriging.scala
│   ├── LeastSquaresFittingNuggetProblem.scala
│   ├── LeastSquaresFittingProblem.scala
│   ├── LinearSemivariogram.scala
│   ├── NonLinearSemivariogram.scala
│   ├── OrdinaryKriging.scala
│   ├── Semivariogram.scala
│   ├── SimpleKriging.scala
│   └── UniversalKriging.scala
├── io
│   ├── json
│   │   ├── CrsFormats.scala
│   │   ├── FeatureFormats.scala
│   │   ├── GeoJson.scala
│   │   ├── GeoJsonSupport.scala
│   │   ├── GeometryFormats.scala
│   │   ├── Implicits.scala
│   │   ├── JsonCRS.scala
│   │   ├── JsonFeatureCollectionMap.scala
│   │   ├── JsonFeatureCollection.scala
│   │   └── Style.scala
│   ├── package.scala
│   ├── WKB
│   │   ├── Implicits.scala
│   │   ├── WKB.scala
│   │   └── WKBWriter.scala
│   └── WKT
│       ├── Implicits.scala
│       └── WKT.scala
├── mesh
│   ├── HalfEdge.scala
│   ├── HalfEdgeTable.scala
│   └── IndexedPointSet.scala
├── reproject
│   ├── Implicits.scala
│   └── Reproject.scala
├── summary
│   └── polygonal
│       └── PolygonalSummaryHandler.scala
├── triangulation
│   ├── BoundaryDelaunay.scala
│   ├── DelaunayStitcher.scala
│   ├── DelaunayTriangulationMethods.scala
│   ├── DelaunayTriangulation.scala
│   ├── Implicits.scala
│   ├── QuadricError.scala
│   ├── StitchedDelaunay.scala
│   ├── TriangleMap.scala
│   └── TriangulationPredicates.scala
├── util
│   └── Intersection.scala
└── voronoi
    ├── DelaunayMethods.scala
    ├── Implicits.scala
    ├── package.scala
    ├── VoronoiDiagram.scala
    └── VoronoiMethods.scala
```
Notable here is the maintenance of the `Extent` type, which is easier to work with in a scala context (e.g., pattern matching).  We will provide implicit conversion from JTS's Envelope type to facilitate usage.  We're also keeping a few classes that provide enriched functionality, like `SpatialIndex`. Additionally, we provide a pureconfig-configured GeometryFactory object.  This will be used throughout the system to build compatible geometries.  Also note that the `WKB` and `WKT` objects in `io` provide thread-safe operations, and so should be kept.  Finally, the `Intersection` class in the `util` subpackage will be moved up to the root package, and will be heavily modified to provide a type-enriched intersection facility that will be more useful for pattern matching.

vector/src/main/scala/geotrellis/vector/package.scala
```diff
@@ -19,125 +19,48 @@ package geotrellis
...
-
-  implicit def seqGeometryToGeometryCollection(gs: Seq[Geometry]): GeometryCollection =
-    GeometryCollection(gs)
-
   implicit class ProjectGeometry[G <: Geometry](g: G) {
     /** Upgrade Geometry to Projected[Geometry] */
     def withSRID(srid: Int) = Projected(g, srid)
   }
+
+  @typeclass trait MultiGeometry[G]
```
echeipesh: This isn't used anywhere but for an import in `ProtobufGeom`. Do you think this is something important or just left-over at this point?

jpolchlo: Yeah, I'd agree that it's mostly vestigial.  Not entirely clear what its value was in the first place.

echeipesh:
@jpolchlo `CHANGELOG` is the only outstanding thing, it can be `[skip ci]` tagged

vector/src/main/scala/geotrellis/vector/Point.scala
```diff
@@ -16,324 +16,19 @@
...
-  def isOnLine(l: Line): Boolean =
-    CGAlgorithms.isOnLine(jtsGeom.getCoordinate, l.jtsGeom.getCoordinates)
-
-  def distanceToSegment(a: Point, b: Point): Double =
-    CGAlgorithms.distancePointLine(jtsGeom.getCoordinate, a.toCoordinate, b.toCoordinate)
-
-  def distanceToInfiniteLine(a: Point, b: Point): Double =
-    CGAlgorithms.distancePointLinePerpendicular(jtsGeom.getCoordinate, a.toCoordinate, b.toCoordinate)
-}
+object Point extends PointConstructors
```
echeipesh: I think this is actually the last thing we talked about, keeping the `geotrellis.vector.jts.Point` is the thing that would provide duplicate imports if `org.locationtech.jts.geom._` was imported. IIRC our discussion landed on removing these constructors and using only the ones in `JTS` object.

jpolchlo: We discussed this, and I remember settling on "eh, no, maybe not".  The footprint of this "simple" change is wide, necessitating either another import (after we went through the trouble to minimize them), or requiring a clumsy prefix for every geometry creation.  It's going to be less common for a user to `import org.locationtech.jts.geom._`, so I made a note in the docs (https://github.com/locationtech/geotrellis/pull/2932/files#diff-38390c0856b2dfe9eff77f20e8a0d9d1R249) to avoid doing so in a careless fashion.  The `JTS` object feels like a kludge to get around a broken thing, not a thing that I would want to use all the time.  I would hope that one day the scala REPL would fix its problems and we can just get rid of it.

echeipesh: Ok, thats not what I recall ending on, but I accept this argument now. The docs certainly help. 👍

pomadchin:
@jpolchlo sweet PR! 🎉

Issue #2934, opened 2019-05-23
Split layer packages by Spark and Collections classes


echeipesh:
Connects: https://github.com/locationtech/geotrellis/issues/2916

This issue is to split out backend-specific packages in twain: one for spark and one for collections API

- [ ] Diagram the split and agree on package naming convention (1 point)
- [ ] Split `geotrellis-geomesa` (1 points)
- [ ] Split `geotrellis-geowave` (1 points)
- [ ] Split `geotrellis-pipeline`(1 points)
- [ ] Split `geotrellis-s3` (2 points)
- [ ] Split `geotrellis-hbase` (2 points)
- [ ] Split `geotrellis-accumulo` (2 points)
- [ ] Split `geotrellis-cassandra`(2 points)

The tests should remain in the spark version of the packages because currently their implementation is very much coupled to reading and writing layers using spark. Putting in the effort to split the tests doesn't seem worth it right now.

pomadchin:
Just a note about fixing the tests: all scripts from `.locationtech`, `.travis` and scripts dirs should be fixed, since we have more projects now and all the names were changed.

echeipesh:
Resolved by https://github.com/locationtech/geotrellis/pull/2972

PR #2944, opened 2019-05-28
Accumulo Spark Split

jbouffard:
## Overview

This PR is the continuation of #2927 where the `accumulo` package is broken into two new packages: `layers-accumulo` and `accumulo`. Both projects contain `accumulo` IO logic but the former does not use Spark while the latter does.

### Checklist

### Notes

This PR addresses the `accumulo` split in #2934

### DIFF
--- a/accumulo/src/main/resources/META-INF/services/geotrellis.layers.io.AttributeStoreProvider
+++ b/accumulo/src/main/resources/META-INF/services/geotrellis.layers.io.AttributeStoreProvider
@@ -1,1 +1,0 @@
-geotrellis.spark.io.accumulo.AccumuloLayerProvider
--- a/accumulo/src/main/resources/META-INF/services/geotrellis.layers.io.CollectionLayerReaderProvider
+++ b/accumulo/src/main/resources/META-INF/services/geotrellis.layers.io.CollectionLayerReaderProvider
@@ -1,1 +1,0 @@
-geotrellis.spark.io.accumulo.AccumuloLayerProvider
--- a/accumulo/src/main/resources/META-INF/services/geotrellis.layers.io.ValueReaderProvider
+++ b/accumulo/src/main/resources/META-INF/services/geotrellis.layers.io.ValueReaderProvider
@@ -1,1 +1,0 @@
-geotrellis.spark.io.accumulo.AccumuloLayerProvider
--- a/accumulo/src/main/resources/META-INF/services/geotrellis.spark.io.LayerReaderProvider
+++ b/accumulo/src/main/resources/META-INF/services/geotrellis.spark.io.LayerReaderProvider
@@ -1,1 +1,1 @@
-geotrellis.spark.io.accumulo.AccumuloLayerProvider
+geotrellis.spark.io.accumulo.AccumuloSparkLayerProvider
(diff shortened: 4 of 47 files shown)


jbouffard:
There is an issue that needs to be addressed before this PR is merged, and that's where `AccumuloLayerCopier` and `AccumuloLayerMover` should live. Both in the `hadoop` and `file` backends, their respective `LayerCopier` and `LayerMover` live in the `layers` package (see [file](https://github.com/jbouffard/geotrellis/tree/feature/spark-split/collections/layers/src/main/scala/geotrellis/layers/file) and [hadoop](https://github.com/jbouffard/geotrellis/tree/feature/spark-split/collections/layers/src/main/scala/geotrellis/layers/hadoop)).

However, this pattern could not be implemented in `layers-accumulo`. This is because both `AccumuloLayerCopier` and `AccumuloLayerMover` refer to `AccumuloLayerWriter` and `AccumuloLayerReader`, and neither of those live in the `layers-accumulo` package.

Should there be a refactoring done so that all backends have the types available? That makes the most sense to me, but I'm not sure how it'd be done without removing features from the `layers` backends.

jbouffard:
Another question is what we should call the `accumulo` package that uses Spark. Right now, I just left it as `accumulo`, but we might want to change it to something else (though, I think the class path for that package is good).

jbouffard:
What are your thoughts on these questions: @echeipesh @pomadchin @moradology ?

echeipesh:
It's a valid question, trying to come up with consistent naming convention as you write this.
The basic problem is that we have two overlapping levels of specialization:
layers: s3, accumulo, hbase etc. (the collection specific specializations)
spark-layers: s3, accumulo, hbase etc. ( the spark specific specialization)

One of those has to come first in the package name, its probably `spark`, but trying to lay it out to confirm it.

As far as the Accumulo LayerMovers my immediate reaction is that classes should fall to where their dependency is. So the fact that other LayerMovers can be part of non-spark packages and accumulo can't be doesn't raise a flag for me.

jbouffard:
@echeipesh Just so I'm clear, you're suggesting that we use the `spark-accumulo` convention?

echeipesh:
This is what I've been trying to work out. Its conflating the issue a little bit because it has separation of layers types during run time and storing the layers (we talked about this but its not part of current PR) but the backend specific question remains the same:

image
The project names and artifact names would be:
`geotrellis-raster.jar`
`geotrellis-layer.jar`
`geotrellis-store.jar`
`geotrellis-store-s3.jar`
`geotrellis-spark-store-s3.jar`

Note that even though the `store` depends on layer, it is flattened out here to top level concept under `geotrellis` package.

I think I prefer that to: 
image

jbouffard:
@echeipesh That makes sense to me, and I prefer the former naming pattern as well. Were we planning on doing the `store` rename before 3.0? Because if so, it might not matter what they're named right now since they'll be changed later.

echeipesh:
@jbouffard Just came around this morning but @moradology @jpolchlo @pomadchin are all on board, so let's say we will go with it for 3.0. If you have no objection lets consider it official plan as of now.

jbouffard:
@echeipesh I have no objections to it. I think that's something we should have ready for 3.0.

echeipesh:
@jbouffard Reading between the lines: you're saying the current split and package naming convention you're working on can be search/replaced onto above diagram because it is structurally equivalent so you're going to follow that and deal with renames afterwards?

jbouffard:
@jbouffard That's right. I think it'd be easier to do all of the renaming at once just to keep things organized.

PR #2946, opened 2019-05-28
Cassandra Spark Split

jbouffard:
## Overview

This PR is the continuation of #2927 where the `cassandra` package is broken into two new packages: `layers-cassandra` and `cassandra`. Both projects contain `cassandra` IO logic but the former does not use Spark while the latter does.

### Checklist

### Notes

This PR addresses the `cassandra` split in #2934

### DIFF
--- a/layers-cassandra/src/main/scala/geotrellis/layers/cassandra/CassandraCollectionReader.scala
+++ b/layers-cassandra/src/main/scala/geotrellis/layers/cassandra/CassandraCollectionReader.scala
@@ -39,8 +39,6 @@ import geotrellis.layers.LayerId
 object CassandraCollectionReader {
   final val defaultThreadCount = CassandraConfig.threads.collection.readThreads
 
-  private case class SchemaWrapper(value: Option[Schema]) extends Serializable
-
   def read[K: Boundable : AvroRecordCodec : ClassTag, V: AvroRecordCodec : ClassTag](
     instance: CassandraInstance,
     keyspace: String,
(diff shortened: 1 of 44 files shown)


layers-cassandra/src/main/scala/geotrellis/layers/cassandra/CassandraCollectionReader.scala
```diff
@@ -14,28 +14,33 @@
...
 import scala.reflect.ClassTag
+
 import java.math.BigInteger
 
 import geotrellis.layers.LayerId
 
 object CassandraCollectionReader {
   final val defaultThreadCount = CassandraConfig.threads.collection.readThreads
 
+  private case class SchemaWrapper(value: Option[Schema]) extends Serializable
```
jbouffard: I had to find a new wrapper for the `Schema` as I couldn't use the `KryoWrapper`. I ended up making this simple case class which I believe should do the same thing.

pomadchin: @jbouffard you don't need KryoWrapper for the non Spark case.

jbouffard: Ah, that's right. Will do :+1:

pomadchin: But in general `KryoWrapper` makes things Serializable with Kryo and your wrapper won't make it Serializable; Look how [KryoWrapper](https://github.com/locationtech/geotrellis/blob/f8af6e6d72749b8fb175a0c44cbc7d027a940686/spark/src/main/scala/geotrellis/spark/util/KryoWrapper.scala) works when you will have some time :)

PR #2952, opened 2019-05-29
HBase/Spark split

moradology:
## Overview

This PR is the continuation of #2927 where the hbase package is broken into two new packages: hbase-store and hbase-spark.

### Notes

This PR addresses the hbase split in #2934


hbase-store/src/main/resources/META-INF/services/geotrellis.spark.io.AttributeStoreProvider
```diff
@@ -0,0 +1 @@
+geotrellis.spark.io.hbase.HBaseLayerProvider
```
jbouffard: These file names need to be changed to the file path `geotrellis.layers.AttributeStoreProvider`.

Also, the name of the HBase provider would be: `geotrellis.store.hbase.HBaseCollectionLayerProvider`

hbase/src/main/scala/geotrellis/spark/io/hbase/package.scala
```diff
@@ -20,8 +20,4 @@ import org.apache.hadoop.hbase.TableName
 import org.apache.hadoop.hbase.util.Bytes
 
 package object hbase {
```
jbouffard: Can this file just be deleted?

build.sbt
```diff
@@ -228,7 +229,7 @@ lazy val cassandra = project
   .settings(commonSettings)
   .settings(Settings.cassandra)
 
-lazy val hbase = project
+lazy val `hbase-store` = project
```
jbouffard: Since the package path is `geotrellis.store.hbase` wouldn't it make more sense to call the project `store-hbase`?

build.sbt
```diff
@@ -237,9 +238,19 @@ lazy val hbase = project
   .settings(Settings.hbase)
   .settings(projectDependencies := { Seq((projectID in spark).value.exclude("com.google.protobuf", "protobuf-java")) })
 
-lazy val `spark-etl` = Project(id = "spark-etl", base = file("spark-etl")).
-  dependsOn(spark, s3, accumulo, cassandra, hbase).
-  settings(commonSettings)
+lazy val `hbase-spark` = project
```
jbouffard: Same as above

hbase-store/src/main/resources/META-INF/services/geotrellis.spark.io.AttributeStoreProvider
jbouffard: This pertains to all of the other service files as well.

hbase/src/main/scala/geotrellis/spark/io/hbase/package.scala
moradology: It should have been; I might have screwed something up during the rebase

build.sbt
moradology: I'm not sure I find that reason compelling, but I think `store-hbase` is likely to lend itself to better ordering of modules by default

jbouffard: That's true. I'm fine with leaving it `store-xxx`. Thought I'd just mention it.

PR #2966, opened 2019-06-05
HBase Package Cleanup Spark Split

jbouffard:
## Overview

This PR cleans up, fixes, and adds to the the HBase packages so that they work correctly.


hbase-spark/src/test/scala/geotrellis/spark/HBaseTestEnvironment.scala
```diff
@@ -14,7 +14,10 @@
  * limitations under the License.
  */
 
-package geotrellis.store.hbase
```
moradology: We should make the `TestEnvironment` traits available from our local test packages and break off spark-requiring portions to live inside `geotrellis.store.spark` tests

project/Settings.scala
```diff
@@ -296,8 +296,50 @@ object Settings {
...
-  lazy val hbase = Seq(
-    name := "geotrellis-hbase",
+  lazy val `hbase-store` = Seq(
+    name := "geotrellis-hbase-store",
+    libraryDependencies ++= Seq(
+      hbaseCommon exclude("javax.servlet", "servlet-api"),
+      hbaseClient exclude("javax.servlet", "servlet-api"),
+      hbaseMapReduce exclude("javax.servlet", "servlet-api"),
+      hbaseServer exclude("org.mortbay.jetty", "servlet-api-2.5"),
+      hbaseHadoopCompact exclude("javax.servlet", "servlet-api"),
```
moradology: we should remember to rename this as `compat` because it is a compatibility layer, not a compaction library

PR #2969, opened 2019-06-05
Remove ETL

moradology:
## Overview

This PR has 3 objectives:
1. :fire: the ETL project;
2. Banish it to https://github.com/geotrellis/etl
3. Document an alternative path to ingests

### Checklist

- [x] `docs/CHANGELOG.rst` updated, if necessary
- [x] `docs` guides update, if necessary

Closes #2959

PR #2971, opened 2019-06-06
Misc Fixes

jbouffard:
## Overview

This PR fixes various problems that were introduced during the spark split that are causing Travis to fail.

This should probably be merged before: #2969

### DIFF
--- a/layers/src/main/scala/geotrellis/layers/hadoop/HadoopAttributeStore.scala
+++ b/layers/src/main/scala/geotrellis/layers/hadoop/HadoopAttributeStore.scala
@@ -31,12 +31,12 @@ import java.io.PrintWriter
 
 
 class HadoopAttributeStore(
-  val rootPathString: String,
-  val conf: Configuration
+  rootPathString: String,
+  serConf: SerializableConfiguration // This needs to be serializable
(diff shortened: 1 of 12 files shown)


layers/src/main/scala/geotrellis/layers/hadoop/HadoopAttributeStore.scala
```diff
@@ -32,11 +32,11 @@ import java.io.PrintWriter
 
 class HadoopAttributeStore(
   val rootPathString: String,
-  val conf: Configuration
+  private val serConf: SerializableConfiguration // This needs to be serializable
```
moradology: Not sure I see what `private` is for here: if it is private, don't make it a val?

moradology: Actually, I'm not sure either constructor has much use for being a `val`

jbouffard: I just made it `private` so that the user doesn't access the `SerializableConfiguration` which is something we're trying to hide. Though you're right, they probably don't need to be `val`.

PR #2977, opened 2019-06-11
RangeReader SPI

### LINKED ISSUE #2940: RangeReader SPI
We keep writing functions like this:

```scala
    val rr =  javaURI.getScheme match {
      case null =>
        FileRangeReader(Paths.get(javaURI.toString).toFile)      
      
      case "file" =>
        FileRangeReader(Paths.get(javaURI).toFile)


      case "http" | "https" if noQueryParams =>
        HttpRangeReader(new URL(uri))


      case "http" | "https" =>
        new HttpRangeReader(new URL(uri), false)


      case "s3" =>
        val s3Uri = new AmazonS3URI(java.net.URLDecoder.decode(uri, "UTF-8"))
        val s3Client = new AmazonS3Client(AmazonS3ClientBuilder.defaultClient())
        S3RangeReader(s3Uri.getBucket, s3Uri.getKey, s3Client)
```

No one project can depend on all possible implementations of `RangeReader` so this requires an SPI that goes from `URI` to instance of `StreamingByteReader`

This is also going to be required to provide a nice abstraction for `RasterSource` API where can dispatch on `URI` to produce reader of desired type and method.

- [ ] Make SPI interface
- [ ] Implement `hdfs://`
- [ ] Implement `file://` and `/`
- [ ] Implement `s3://`
- [ ] Tests
- [ ] Changelog

Note that the SPI should be able to deal with the usual overload that `file://` could mean use HDFS FileSystem to read the file or to use `FileRangeReader`. The answer here should be considered consistent with the choice elsewhere, whatever that may be.

jbouffard:
## Overview

This PR provides an SPI API for `RangeReader` for the following backeds: `file`, `hdfs`, `http`, and `s3`.

### Checklist

### Demo

```scala
val rangeReader: FileRangeReader = RangeReader(new URI("file///tmp/catalog/file.tif"))
```

### Notes

The `HdfsRangeReader` has a problem where it will use a default `Configuration` even if another one is present. This is because there is no current way of getting a `Configuration` without it being provided explicitly or found via the `SparkContext`. `HdfsConfig` introduced in https://github.com/geotrellis/geotrellis-contrib/pull/186 would resolve this issue. 

Closes #2940


moradology:
Take a look at the provider [here](https://github.com/locationtech/geotrellis/blob/f8af6e6d72749b8fb175a0c44cbc7d027a940686/spark/src/main/scala/geotrellis/spark/io/hadoop/HadoopLayerProvider.scala#L39-L53) - `hdfs+` isn't valid outside of GT but we use it to provide clues to SPI

layers/src/main/scala/geotrellis/layers/hadoop/HadoopCollectionLayerProvider.scala
```diff
@@ -11,7 +11,7 @@ import java.net.URI
 
 
 class HadoopCollectionLayerProvider extends AttributeStoreProvider with ValueReaderProvider with CollectionLayerReaderProvider {
-  val schemes: Array[String] = Array("hdfs", "hdfs+file", "s3n", "s3a", "wasb", "wasbs")
+  import HadoopCollectionLayerProvider._
```
moradology: This seems like indirection for indirection's sake. I'm not overly offended by it, but it may be worth rethinking

layers/src/main/scala/geotrellis/layers/hadoop/util/HdfsRangeReaderProvider.scala
```diff
@@ -0,0 +1,23 @@
...
+class HdfsRangeReaderProvider extends RangeReaderProvider {
+  def canProcess(uri: URI): Boolean = uri.getScheme match {
+    case str: String => HadoopCollectionLayerProvider.schemes contains str.toLowerCase
+    case null => false
+  }
+
+  def rangeReader(uri: URI): HdfsRangeReader = {
+    val conf = new Configuration()
+
+    HdfsRangeReader(new Path(uri), conf)
```
moradology: Note that you fail to call `trim` on this URI - I think this is the cause for problems with `hdfs+`

project/Dependencies.scala
```diff
@@ -36,6 +36,7 @@ object Dependencies {
 
   val apacheIO            = "commons-io"                  % "commons-io"               % "2.6"
   val apacheMath          = "org.apache.commons"          % "commons-math3"            % "3.6.1"
+  val apacheValidator     = "commons-validator"           % "commons-validator"        % "1.4.0"
```
moradology: I'm curious about what the rationale for this lib is - it looks kind of gross and java-y to me and will require CQ. I'm not saying 'no' to it necessarily, but it does seem heavier than necessary for this PR

layers/src/main/scala/geotrellis/layers/hadoop/HadoopCollectionLayerProvider.scala
jbouffard: My main reason for moving the `schemes` to the `HadoopCollectionLayerProvider` object is because they're needed to check the URI: https://github.com/locationtech/geotrellis/blob/e66aa5a59c0419993a220398219e7345d3245207/layers/src/main/scala/geotrellis/layers/hadoop/util/HdfsRangeReaderProvider.scala#L14

layers/src/main/scala/geotrellis/layers/hadoop/util/HdfsRangeReaderProvider.scala
jbouffard: Ah, that's a good point. I'll see if that's the cause.

project/Dependencies.scala
jbouffard: My main reason for adding it was to provide an easy way to validate a `URL`. Though, now that I think on it, I guess I could just catch the error produced by `URL` instead.

layers/src/main/scala/geotrellis/layers/hadoop/HadoopCollectionLayerProvider.scala
jbouffard: I ended up moving the `SCHEMES` to the Hadoop `package` object.

s3-spark/src/test/scala/geotrellis/spark/io/s3/util/S3RangeReaderProviderSpec.scala
```diff
@@ -0,0 +1,22 @@
+package geotrellis.spark.store.s3.util
+
+import geotrellis.store.s3.util._
+import geotrellis.spark.store.s3.testkit.MockS3Client
+import geotrellis.store.s3.S3ClientProducer
+import geotrellis.util.RangeReader
+
+import org.scalatest._
+
+class S3RangeReaderProviderSpec extends FunSpec with Matchers {
```
pomadchin: Why this file is in `s3-spark` and not in `s3-store`? The implementation of the S3RangeReaderProvider is in `s3-store`.

s3-spark/src/test/scala/geotrellis/spark/io/s3/util/S3RangeReaderProviderSpec.scala
```diff
@@ -1,14 +1,18 @@
 package geotrellis.spark.store.s3.util
 
+import geotrellis.spark.store.s3.S3TestUtils
```
pomadchin: I'm wondering, should `S3TestUtils` live in the `testkit` or not? @moradology

s3-store/src/main/scala/geotrellis/store/s3/util/S3RangeReader.scala
```diff
@@ -35,26 +35,31 @@ import java.net.URI
  * @return A new instance of S3RangeReader.
  */
 class S3RangeReader(
-  request: GetObjectRequest,
+  request: => GetObjectRequest,
```
pomadchin: It is by name parameter now, since otherwise, any `S3RangeReader` creation forced an immediate metadata queries.

s3-spark/src/test/scala/geotrellis/spark/io/s3/util/S3RangeReaderProviderSpec.scala
jbouffard: There wasn't a really great reason. We just have all of our tests for `s3-store` in `s3-spark` still, so I just put it there.

s3-store/src/main/scala/geotrellis/store/s3/util/S3RangeReader.scala
pomadchin: Also a reasonable question appears, should the client be a by name parameter and cached inside?

s3-spark/src/test/scala/geotrellis/spark/io/s3/util/S3RangeReaderProviderSpec.scala
moradology: The good news is that there is no longer an s3-testkit to put it in

s3-spark/src/test/scala/geotrellis/spark/io/s3/util/S3RangeReaderProviderSpec.scala
pomadchin: @jbouffard can you look into the tests structure more?

s3-store/src/main/scala/geotrellis/store/s3/util/S3RangeReader.scala
jbouffard: That's a good question. Would that be beneficial to have? Because right now the `S3RangeReader` is more robust as it can work with more than one client.

pomadchin: @jbouffard I am not sure that I follow you, the benefit would be that there would be no instance creation until the explicit RangeReader creation, in case `S3Client` was lazily created.

s3-spark/src/test/scala/geotrellis/spark/io/s3/util/S3RangeReaderProviderSpec.scala
jbouffard: @pomadchin Actually, we can't move it to `s3-store` as it needs `TestUtils` which is in `s3-spark`.

pomadchin:
@jbouffard can you rebase this PR?

jbouffard:
Superseded by #2998

PR #2998, opened 2019-06-17
RangeReader SPI

### LINKED ISSUE #2940: RangeReader SPI

jbouffard:
## Overview

This PR supersedes #2977 and it was created in order to avoid a painful rebase.

### Checklist

Closes #2940

### DIFF
--- a/layers/src/main/scala/geotrellis/layers/hadoop/package.scala
+++ b/layers/src/main/scala/geotrellis/layers/hadoop/package.scala
@@ -1,0 +1,9 @@
+package geotrellis.layers
+
+import org.apache.hadoop.fs.Path
+
+package object hadoop extends Implicits {
+  final val SCHEMES: Array[String] = Array("hdfs", "hdfs+file", "s3n", "s3a", "wasb", "wasbs")
+
+  implicit def stringToPath(path: String): Path = new Path(path)
+}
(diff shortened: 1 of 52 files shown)


layers/src/main/scala/geotrellis/layers/hadoop/package.scala
```diff
@@ -0,0 +1,9 @@
+package geotrellis.layers
```
pomadchin: It's a shame that I didn't notice this regression. We don't have this subproject anymore. Fixed here https://github.com/locationtech/geotrellis/pull/3001

Issue #2999, opened 2019-06-17
Update GeoWave project


pomadchin:
```
rfecher @rfecher 15:08
we recently took down snapshots for old versions and left releases
I suggest using the 0.9.8 release instead of the snapshot
we're moving that to maven central - all snapshots and releases will be available through maven central now (on the sonatype OS repo) but we are keeping old releases on s3 as well
and we'll go back and make sure to add old releases to maven central as well, but for now you can use s3
```

Reference: https://gitter.im/locationtech/geowave?at=5d07e53e26477710bbc9da6f

Also as @jbouffard pointed out, we don't release `geotrellis-geowave` artifacts, this also should be fixed.

pomadchin:
Project is moved into its own repo https://github.com/geotrellis/geotrellis-geowave

PR #3000, opened 2019-06-17
Update GeoWave resolvers

pomadchin:
## Overview

Temporary exclude GeoWave subproject from tests.

Connects https://github.com/locationtech/geotrellis/issues/2999

### DIFF
--- a/project/Version.scala
+++ b/project/Version.scala
@@ -15,7 +15,7 @@
  */
 
 object Version {
-  val geotrellis  = "3.0.1-LOCAL"
+  val geotrellis  = "3.0.0" + Environment.versionSuffix
   val scala       = "2.11.12"
   val crossScala  = Seq(scala, "2.12.7")
   val geotools    = "21.0"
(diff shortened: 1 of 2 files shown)


jbouffard:
@pomadchin Do we need to update the [deploy scripts](https://github.com/locationtech/geotrellis/tree/98f4abb430027eaac69b012db439a8c635fc9f9e/.locationtech) or are they okay?

pomadchin:
@jbouffard we don't publish it for some reason for a long time O:

jbouffard:
@pomadchin That's interesting haha. In that case, this LGTM :+1:

project/Version.scala
```diff
@@ -15,7 +15,7 @@
  */
 
 object Version {
-  val geotrellis  = "3.0.1-LOCAL"
+  val geotrellis  = s"3.0.1${Environment.versionSuffix}"
```
echeipesh: I think `3.0.1` is local testing leaking. We haven't published `3.0.0` yet so I think we should be fine to have this as `3.0.0-SNAPSHOT` ?

pomadchin: @echeipesh  a good point

Issue #2872, opened 2019-02-08
Move thresholds and arbitrary numbers in the code into Constant objects


pomadchin:
We have lot's over thresholds in the code and probably other arbitrary numbers that can be found in the code. It would be nice to move them into [objects](https://github.com/locationtech/geotrellis/blob/master/util/src/main/scala/geotrellis/util/Constants.scala#L32) to introduce a little bit more semantics to them if possible.

The task is to go through the entire codebase and figure out what's possible to move into separate objects.

PR #3002, opened 2019-06-20
Use semigroup where possible

### LINKED ISSUE #2925: Replace `Mergeable` with Cats' `Semigroup`
For historical reasons, we were shy about using standard functional typeclass libraries.  We therefore have some vestigial typeclasses kicking about in the codebase.  One such example is https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/merge/Mergable.scala.  This should be replaced by Cats' [Semigroup](https://typelevel.org/cats/typeclasses/semigroup.html).

moradology:
## Overview

`Combine` methods across the library have been pulled out of their classes and defined by way of typeclasses so that methods defined in terms of `semigroupal` types can generically act upon any of them

### Checklist

Closes #2925

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/GridExtent.scala
+++ b/raster/src/main/scala/geotrellis/raster/GridExtent.scala
@@ -17,10 +17,14 @@
 package geotrellis.raster
 
 import geotrellis.vector.{Extent, Point}
+import geotrellis.util.Constants
 
-import scala.math.{min, max, ceil}
 import spire.math.{Integral}
 import spire.implicits._
+import cats.SemigroupK
+import cats.syntax.semigroup._
+
+import scala.math.{min, max, ceil}
 
 /**
   * Represents an abstract grid over geographic extent.
@@ -60,13 +64,13 @@ class GridExtent[@specialized(Int, Long) N: Integral](
   * same cellsizes).  The result is a new extent at the same
   * resolution.
   */
-  def combine (that: GridExtent[N]): GridExtent[N] = {
+  def expandToInclude (that: GridExtent[N]): GridExtent[N] = {
     if (cellwidth != that.cellwidth)
       throw GeoAttrsError(s"illegal cellwidths: $cellwidth and ${that.cellwidth}")
     if (cellheight != that.cellheight)
       throw GeoAttrsError(s"illegal cellheights: $cellheight and ${that.cellheight}")
 
-    val newExtent = extent.combine(that.extent)
+    val newExtent = extent.expandToInclude(that.extent)
     val newRows = ceil(newExtent.height / cellheight)
     val newCols = ceil(newExtent.width / cellwidth)
 
@@ -372,6 +376,12 @@ object GridExtent {
     val cw = extent.width / grid.cols.toDouble
     val ch = extent.height / grid.rows.toDouble
     new GridExtent[N](extent, cw, ch, grid.cols, grid.rows)
+  }
+
+  implicit def gridExtentMergeSemigroupK: SemigroupK[GridExtent] = new SemigroupK[GridExtent] {
+    def combineK[T](x: GridExtent[T], y: GridExtent[T]): GridExtent[T] = {
+      x expandToInclude y
+    }
   }
 
   /** RasterSource interface reads GridBounds[Long] but GridBounds[Int] abounds.
--- a/raster/src/main/scala/geotrellis/raster/RasterExtent.scala
+++ b/raster/src/main/scala/geotrellis/raster/RasterExtent.scala
@@ -17,6 +17,9 @@
 package geotrellis.raster
 
 import geotrellis.vector.{Extent, Point}
+
+import cats.Semigroup
+
 import scala.math.{min, max, ceil}
 
 
@@ -79,13 +82,13 @@ case class RasterExtent(
     * same cellsizes).  The result is a new extent at the same
     * resolution.
     */
-  def combine(that: RasterExtent): RasterExtent = {
+  def expandToInclude(that: RasterExtent): RasterExtent = {
     if (cellwidth != that.cellwidth)
       throw GeoAttrsError(s"illegal cellwidths: $cellwidth and ${that.cellwidth}")
     if (cellheight != that.cellheight)
       throw GeoAttrsError(s"illegal cellheights: $cellheight and ${that.cellheight}")
 
-    val newExtent = extent.combine(that.extent)
+    val newExtent = extent.expandToInclude(that.extent)
     val newRows = ceil(newExtent.height / cellheight).toInt
     val newCols = ceil(newExtent.width / cellwidth).toInt
 
@@ -165,4 +168,8 @@ object RasterExtent {
     */
   def apply(extent: Extent, tile: CellGrid[Int]): RasterExtent =
     apply(extent, tile.cols, tile.rows)
+
+  implicit val rasterExtentSemigroup: Semigroup[RasterExtent] = new Semigroup[RasterExtent] {
+    def combine(x: RasterExtent, y: RasterExtent): RasterExtent = x expandToInclude y
+  }
 }
--- a/store/src/main/scala/geotrellis/store/cog/COGLayerMetadata.scala
+++ b/store/src/main/scala/geotrellis/store/cog/COGLayerMetadata.scala
@@ -23,12 +23,11 @@ import geotrellis.raster.io._
 import geotrellis.layer._
 import geotrellis.util._
 
-import cats.instances.either._
-import cats.instances.stream._
-import cats.syntax.foldable._
-
 import spray.json._
 import spray.json.DefaultJsonProtocol._
+
+import cats.Semigroup
+import cats.syntax.semigroup._
 
 
 case class COGLayerMetadata[K: SpatialComponent](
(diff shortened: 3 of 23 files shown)


moradology:
@echeipesh @jpolchlo @pomadchin 
The failing tests in this PR deserve a little bit of attention because they arise from what is either a misunderstanding on my part (easy enough to fix) or a bit of inconsistency about how floating point precision is represented in GT.

In particular, [this value](https://github.com/locationtech/geotrellis/pull/3002/files#diff-6657977156e7279b43abf638e7b5a8b0L359) is not motivated by any kind of documentation but we have `Constants` for epsilon values that are used library-wide. Why should these values differ?

pomadchin:
@moradology I think you've decided to implement here a part of https://github.com/locationtech/geotrellis/issues/2872? 

I think the right way would be to revert back your changes related to constants and to resolve all random constants issue in a separate PR, as they are pretty sensitive and at least deserve a separate PR.

moradology:
Yeah, I'm going to revert this change if an issue already exists. It looked like a pretty clear oversight and I remain extremely confused by its purpose

raster/src/main/scala/geotrellis/raster/GridExtent.scala
```diff
@@ -17,10 +17,14 @@
 package geotrellis.raster
 
 import geotrellis.vector.{Extent, Point}
+import geotrellis.util.Constants
 
-import scala.math.{min, max, ceil}
 import spire.math.{Integral}
 import spire.implicits._
+import cats.SemigroupK
+import cats.implicits._
```
pomadchin: ` import cats.syntax.semigroup._` would be better for library purposes, and improves the compilation time.

raster/src/main/scala/geotrellis/raster/GridExtent.scala
```diff
@@ -174,7 +178,7 @@ class GridExtent[@specialized(Int, Long) N: Integral](
     val colMax: N = Integral[N].fromLong {
       val colMaxDouble = mapXToGridDouble(subExtent.xmax)
 
-      if (math.abs(colMaxDouble - floorWithTolerance(colMaxDouble)) < GridExtent.epsilon)
+      if (math.abs(colMaxDouble - floorWithTolerance(colMaxDouble)) < Constants.DOUBLE_EPSILON)
```
pomadchin: Actually I know what is going on here, the `DOUBLE_EPSILON` is `1.11e-16`, however, in this case we need a less precise epsilon, otherwise the logic of rounding would be broken and we'll get (or won't get) extra +1 column on the east. Look also into this comment: https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/GridExtent.scala#L382-L388

GDAL uses 1e-5, QGIS 1e-6 and we're using 1e-7. It looks like it is a heuristic in all libs.

raster/src/main/scala/geotrellis/raster/RasterExtent.scala
```diff
@@ -165,4 +168,8 @@ object RasterExtent {
     */
   def apply(extent: Extent, tile: CellGrid[Int]): RasterExtent =
     apply(extent, tile.cols, tile.rows)
+
+  implicit val rasterExtentMergeSemigroup: Semigroup[RasterExtent] = new Semigroup[RasterExtent] {
```
pomadchin: The name of the implicit is weird a bit and inconsistent. Here it is `rasterExtentMergeSemigroup` (it can be just `rasterExtentSemigroup`. And for the `Extent` it is `extentIsSemigroup` but it can be `extentSemigroup`.


I'm wondering, do we really need a semigroup here for the `RasterExtent`?

store/src/main/scala/geotrellis/store/cog/COGLayerMetadata.scala
```diff
@@ -420,4 +400,25 @@ object COGLayerMetadata {
...
+  implicit def cogLayerMetadataSemigroup[K: Boundable: SpatialComponent] = new Semigroup[COGLayerMetadata[K]] {
+    def combine(x: COGLayerMetadata[K], y: COGLayerMetadata[K]): COGLayerMetadata[K] = {
+      val combinedZoomRangeInfos =
+        (x.zoomRangeInfos ++ y.zoomRangeInfos)
+          .groupBy(_._1)
+          .map { case (key, bounds) => key -> bounds.map(_._2).reduce(_ combine _) }
+          .toVector
+          .sortBy(_._1)
+
+      val combinedExtent = x.extent.expandToInclude(y.extent)
```
pomadchin: It can be `combine` in case we really need it.

vector/src/main/scala/geotrellis/vector/Extent.scala
```diff
@@ -323,3 +278,28 @@ case class Extent(
...
+    Extent(xmin,ymin,xmax,ymax)
+  }
+
+  implicit def toPolygon(extent: Extent): Polygon =
+    extent.toPolygon
+
+  implicit def jts2Extent(env: jts.Envelope): Extent =
+    Extent(env)
+
+  implicit val extentIsSemigroup: Semigroup[Extent] = new Semigroup[Extent] {
```
pomadchin: `extentSemigroup`

vector/src/test/scala/spec/geotrellis/vector/ExtentSpec.scala
```diff
@@ -18,9 +18,12 @@ package geotrellis.vector
 
 import geotrellis.vector.io._
 import geotrellis.vector.io.json.JsonFeatureCollection
+
 import org.scalatest.FunSpec
 import org.scalatest.Matchers
 import spray.json.DefaultJsonProtocol._
+// `import cats.implicits._` causes ambiguous implicits on `===`:  https://github.com/typelevel/cats/issues/219
```
pomadchin: A good reason to use `import cats.syntax.semigroup._` everywhere instead of `implicits`.

raster/src/main/scala/geotrellis/raster/GridExtent.scala
jpolchlo: @pomadchin is right!  The epsilons in `Constants` are  good baseline to use when you don't know enough about the problem you're working on.  The predefined values are the sort of best practice values for numerical methods (think computing square roots via Newton's method or something similar).  When thresholding for edge cases, other fudge factors may be appropriate, which we set using case-by-case heuristics (_cough_ _cough_ guessing _cough_).  In those cases, it may be appropriate to comment above the definition of the buffer value that it is an intentional choice and why.

vector/src/test/scala/spec/geotrellis/vector/ExtentSpec.scala
echeipesh: @pomadchin @moradology Seems like `import cats.syntax.semigroup._` is going to be required import in lots of places. It looks like we're replacing the geotrellis imports soup with cats import soup.

When faced with this problem before I've chosen to leave the `combine` methods on objects that currently have them and add the `Semigroup` instance that uses that. That at least has the potential of using the method `.combine` when you know the explicit type without import and use the `Semigroup[K].combine` when you're writing abstract code. What do you guys think of that tactic ?

store/src/main/scala/geotrellis/store/cog/COGLayerMetadata.scala
echeipesh: @pomadchin What do you mean? I don't follow.

vector/src/test/scala/spec/geotrellis/vector/ExtentSpec.scala
pomadchin: @echeipesh I don't have a strong opinion about that, I had some thoughts about it but decided to await until the PR would be +- ready. I also wanted to discuss in general do we need `Semigroups` in places where it is not obvious. Since we don't have a code that would use it, does it make any sense to add it? Mb it is a part of a user application in the case we don't use them. Tbh I would like not to include the dead code.

store/src/main/scala/geotrellis/store/cog/COGLayerMetadata.scala
pomadchin: @echeipesh we have lots of semigroup instances for our types, but why do we need them if we don't use them? see: https://github.com/locationtech/geotrellis/pull/3002#discussion_r297680809

vector/src/test/scala/spec/geotrellis/vector/ExtentSpec.scala
pomadchin: @echeipesh eh, I said yesterday Nathan to push changes into a separate commit ): but now it's squashed... The story is that he implemented these things you're asking about in a first commit, and in the second he covered #2925... you don't see the whole story since it is a force push ):

raster/src/main/scala/geotrellis/raster/GridExtent.scala
```diff
@@ -60,13 +64,13 @@ class GridExtent[@specialized(Int, Long) N: Integral](
   * same cellsizes).  The result is a new extent at the same
   * resolution.
   */
-  def combine (that: GridExtent[N]): GridExtent[N] = {
+  def expandToInclude (that: GridExtent[N]): GridExtent[N] = {
```
echeipesh: The `combine` function on `Extent` and `RasterExtent` I'd expect to be very heavily used in user code. `expendToInclude` is a way better name but I'm wondering if the frequency with which these are used warrants going through the alias/depricate cycle on these in 3.0 release.

vector/src/test/scala/spec/geotrellis/vector/ExtentSpec.scala
CloudNiner: I ran into this for polygonal summary API. It was pretty annoying to include the `cats.syntax.semigroup._` everywhere in my concrete implementations, and its sort of an open secret that some people (read: including me until @moradology told me about it) might not know about which increases friction for GT users when there's lots of other stuff to learn. 

In my case, there wasn't any external API for polygonal summary that exposed the requirement to include the cats import soup so this concern didn't come up there, but if one of our design goals is to reduce imports and increase clarity, then taking @echeipesh 's suggestion to keep explicit combine methods that internally use cats rather than pushing that out to the user seems like a real good idea. Isn't that what abstraction is for anyways?

moradology:
This PR does too much and fundamentally doesn't address the desired problem. Closing in favor of one that does and is properly scoped

PR #3003, opened 2019-06-20
Throw errors inside IO

### LINKED ISSUE #2863: Prevent IO from Swallowing Errors
Recently, I was running a job on EMR that was failing at the `SaveToS3` step. The logs showed a "general" error, that stated something went wrong and the job failed. It gave no information about what actually happened, however, which made debugging more difficult.

The reason for this unhelpful error message is that `IO` swallows the actual error, as can be seen in `SaveToS3`:
https://github.com/locationtech/geotrellis/blob/master/s3/src/main/scala/geotrellis/spark/io/s3/SaveToS3.scala#L104-L110

To stop `IO` from swallowing the error, we should match the result of the run to see if it succeeded.

```scala
requests
        .map(write)
        .parJoin(threads)
        .compile
        .toVector
        .attempt
        .unsafeRunSync() match {
          case Right(succ) => succ
          case Left(err) => throw err
        }

      pool.shutdown()
```

moradology:
## Overview

IO will simply swallow errors unless `attempt` is called. `attempt` gives an `Either[Throwable, A]` which can be matched against to either return the value or throw. This pattern has been applied throughout the library wherever `unsafeRunSync` gets called.

Closes #2863

### DIFF
--- a/accumulo-spark/src/main/scala/geotrellis/spark/store/accumulo/AccumuloWriteStrategy.scala
+++ b/accumulo-spark/src/main/scala/geotrellis/spark/store/accumulo/AccumuloWriteStrategy.scala
@@ -32,6 +32,7 @@ import org.apache.accumulo.core.client.BatchWriterConfig
 
 import cats.effect.IO
 import cats.syntax.apply._
+import cats.syntax.either._
 
 import scala.concurrent.ExecutionContext
 
@@ -143,7 +144,9 @@ case class SocketWriteStrategy(
             .parJoin(threads)
             .compile
             .drain
+            .attempt
             .unsafeRunSync()
+            .valueOr(throw _)
         } finally {
           writer.close(); pool.shutdown()
         }
--- a/store/src/main/scala/geotrellis/store/util/IOUtils.scala
+++ b/store/src/main/scala/geotrellis/store/util/IOUtils.scala
@@ -81,8 +81,11 @@ object IOUtils {
         .parJoin(threads)
         .compile
         .toVector
-        .unsafeRunSync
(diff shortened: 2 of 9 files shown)


store/src/main/scala/geotrellis/store/util/IOUtils.scala
```diff
@@ -81,8 +81,13 @@ object IOUtils {
         .parJoin(threads)
         .compile
         .toVector
-        .unsafeRunSync
+        .attempt
+        .unsafeRunSync match {
+          case Right(succ) => succ
+          case Left(err) => throw err
+        }
         .flatten
```
pomadchin: oops, this won't work; the type had changed

moradology: The type is the same and it is the compiler that is confused! (I believe it just needs some parens)

pomadchin: @moradology :+1: makes sense

accumulo-spark/src/main/scala/geotrellis/spark/store/accumulo/AccumuloWriteStrategy.scala
```diff
@@ -143,7 +143,11 @@ case class SocketWriteStrategy(
             .parJoin(threads)
             .compile
             .drain
-            .unsafeRunSync()
+            .attempt
+            .unsafeRunSync() match {
```
pomadchin: can it be replaced with `.valueOr(throw _)`?

PR #2905, opened 2019-04-21
Replace spray-json by circe

pomadchin:
## Overview

This PR introduces Circe usage instead of spray-json everywhere.
It is a rebased version of this PR: https://github.com/locationtech/geotrellis/pull/2674

It is an initial PR without codecs structure refactoring.

Partially addresses #1772

PR #3005, opened 2019-06-24
Add Feature ids to vector and vectortile packages

### LINKED ISSUE #2884: Allow vectortile.Layer to pass a valid 'id' to internal.vector_tile.Tile.Layer
Hi!  I'm using GeoTrellis and VectorPipe to create MapBox vector tiles.  The problem I have is...  I need to be able to write a top level 'id' in each Feature (not as part of 'properties')...  and in `vectortile.Layer.unfeature()`, the arg `id: Option[Long]` passed to `vectortile.internal.vector_tile.Tile.Feature` is hard-coded in .unfeature() as `None`.

To solve this for my purposes, I forked and compiled locally with a `Map[String,Value] => Option[Long]` function hard coded in `Layer.unfeature()` to create an 'id' from 'data' to pass to `vt.Tile.Feature`.  This worked, and the output .PBFs had the expected 'id' in them.

I'm trying to think through how else users might want to create an 'id'...  maybe a random Long would be sufficient for some, but that could still be coded into a 'data' => 'id' function.

I believe this would be a common pattern, to want an 'id' and to be able to create from the metadata.  It looks like the VectorPipe API has completely changed since two days ago, but it used to be in `Collate.generically()` where a `vectortile.StrictLayer` is constructed.  If there was a way to pass in a `Map[String,Value] => Option[Long]` function to Layer/StrictLayer/LazyLayer for `Layer` to use in .unfeature() to convert 'data' to valid 'id' instead of hard-coding as None...  I think that would be a common use case.

CloudNiner:
## Overview

This PR makes two small incremental changes in the absence of much larger sweeping changes to what the meaning of `Feature`s are in GeoTrellis which, due to conflicting opinions, appears best saved for another time. 

The idea here is to include a feature id as a first class citizen for the two major types of features we handle, the generic `geotrellis.vector.Feature` which approximates, but doesn't totally conform to the GeoJson spec, and the more specific `geotrellis.vectortile.MVTFeature` which conforms to the MVT2.0 spec, when before it just used `geotrellis.vector.Feature` and did not, due to the too generic `data: D` parameter and lack of an `id: Option[Long]`. 

The change to use `id: Option[String]` in `Feature` is motivated by `String` being a more "generic" type than  others in that most things have a usable `toString` that maintains uniqueness (at least this applies for numbers, strings and UUIDs). For example, the new Feature.withRandomUuid method constructs a new uuid for uniqueness and then immediately casts it to string. The additional apply methods also _mostly_ maintain backwards compatibility in that the old `Feature(geom, data)` looks the same as `Feature(None, geom, data)` in practice since None == None is true.  

These two changes are independent, so if we wanted to keep just the MVTFeature change to fix #2884 we could. 

### Checklist

### Notes

TODO:
- [x] Fix tests
- [ ] ~~Ensure geotrellis.vector.io._ encodings work correctly with new id after circe rebase is complete~~

Closes #2884

Blocked by #2905 if we want to keep the changes to `geotrellis.vector.Feature`

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -89,6 +89,11 @@ API Changes & Project structure changes
 
     ``import geotrellis.raster._``
 
+- ``geotrellis.vectortile``
+
+  - **Add:** ``geotrellis.vectortile.MVTFeature`` which properly conforms to the MVT 2.0 spec. Specifically, MVTFeature adds support for an ``Option[Long]`` id property.
+  - **Change:** The ``geotrellis.vectortile.{Layer, VectorTile}`` interfaces now uses ``MVTFeature`` instead of ``geotrellis.vector.Feature``. This better aligns these interfaces with the Mapbox Vector Tile specification.
+
 Fixes & Updates
 ^^^^^^^^^^^^^^^
 
--- a/vectortile/src/main/scala/geotrellis/vectortile/Layer.scala
+++ b/vectortile/src/main/scala/geotrellis/vectortile/Layer.scala
@@ -54,20 +54,20 @@ import scala.collection.mutable.ListBuffer
   def resolution: Double = tileExtent.height / tileWidth
 
   /** Every Point Feature in this Layer. */
-  def points: Seq[Feature[Point, Map[String, Value]]]
+  def points: Seq[MVTFeature[Point]]
   /** Every MultiPoint Feature in this Layer. */
-  def multiPoints: Seq[Feature[MultiPoint, Map[String, Value]]]
+  def multiPoints: Seq[MVTFeature[MultiPoint]]
   /** Every Line Feature in this Layer. */
-  def lines: Seq[Feature[Line, Map[String, Value]]]
+  def lines: Seq[MVTFeature[Line]]
   /** Every MultiLine Feature in this Layer. */
-  def multiLines: Seq[Feature[MultiLine, Map[String, Value]]]
+  def multiLines: Seq[MVTFeature[MultiLine]]
   /** Every Polygon Feature in this Layer. */
-  def polygons: Seq[Feature[Polygon, Map[String, Value]]]
+  def polygons: Seq[MVTFeature[Polygon]]
   /** Every MultiPolygon Feature in this Layer. */
-  def multiPolygons: Seq[Feature[MultiPolygon, Map[String, Value]]]
+  def multiPolygons: Seq[MVTFeature[MultiPolygon]]
 
   /** All Features of Single and Multi Geometries. */
-  def features: Seq[Feature[Geometry, Map[String, Value]]] = {
+  def features: Seq[MVTFeature[Geometry]] = {
     Seq(
       points,
       multiPoints,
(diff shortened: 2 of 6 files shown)


CloudNiner:
I can also add some notes to the https://geotrellis.readthedocs.io/en/latest/guide/vectors.html section of the user guide about the new id property on feature, but otherwise the user guide seems as is, pending the circe change -- theres a big chunk of that section that needs to be rewritten since it uses the spray syntax.

jpolchlo:
So, this is awkward.  We have competing interests to support here.  Note that [the vector tile spec](https://github.com/mapbox/vector-tile-spec/blob/master/2.1/vector_tile.proto#L32) requires the ID field to be of `Long` type.  But [SimpleFeatures](https://github.com/geotools/geotools/blob/master/modules/library/opengis/src/main/java/org/opengis/feature/simple/SimpleFeature.java#L81) ask for a `String` ID.

This means that we have to pick our favorite child or parameterize the Feature trait to allow for a choice of ID type.  I know that is not exactly in line with this PR exactly, but perhaps indicates a direction that we should go.  Notably, SimpleFeature should be an admissible substitute for the GT Feature type, which we've not been able to do to this point, but should if we're going through the trouble of improving this part of the API.

moradology:
The vector tile spec doesn't require numbers afaict (from (sec 4.2)[https://github.com/mapbox/vector-tile-spec/tree/master/2.1#42-features]): "A feature MAY contain an id field. If a feature has an id field, the value of the id SHOULD be unique among the features of the parent layer."

The GeoJSON spec mandates EITHER string or number representations

jpolchlo:
That's a weird disconnect between the text of the spec and the reference implementation.  I have to assume that most people using VTs will just pull the .proto file directly, meaning that, de facto, the id field of a VT is `Long`.

moradology:
It might be best to just seal up a `Feature` trait above imagined implementations rather than adding further type parameters to this otherwise simple case class.

CloudNiner:
Awkward indeed.

> It might be best to just seal up a Feature trait above imagined implementations rather than adding further type parameters to this otherwise simple case class.

If we do this, `id` probably shouldn't be a part of it. Otherwise it would need to be parameterized so that implementations can decide on its type, since known concrete implementations already differ in type.

So, are you suggesting something like:
```
trait Feature {
  def map = ...
  def geom = ...
}
```
And then having a `Feature` that provides an `Option[String]` id for compatibility with GeoTools, and then the geotrellis vectortile package defines a separate `MVTFeature` that uses `Option[Long]` to define it's id? And then geotrellis users can define a separate Feature implementation if it suits their needs? Otherwise by default we'd swap internal GT usage to use the Feature that uses Option[String] ids where necessary.

Also not sure about the sealed bit. If it is, we lose the flexibility to define alternate implementations elsewhere, such as for MVTFeature.

moradology:
Arguably there's not much need for user-provided implementations when the provided behavior is so simple and the specs we're trying to support are well-defined. Example `Feature[G <: Geometry, Data]` implementations have been very expressive; far more so than geotools `SimpleFeature`s, for instance, which can only have a flat map of properties

echeipesh:
My first thought was that it would be nice for `MVTFeature` and `Feature` to be part of the same type hiararchy but that can't work out the way its currently defined, which leads us directly into the larger conversation we've been having about `Feature`-ness.

Having said that `MVTFeature` does its job well and it makes total sense. My vote is to ditch the first commit, that adds `id: Option[String]` to `Feature` and accept the second.

CloudNiner:
> My vote is to ditch the first commit

Deal. No issues with that. I updated, just waiting now on the build to go green.

Issue #1772, opened 2016-11-03
Streaming JSON investigation


moradology:
In cases where GeoJSON is too large to be reliably be read in in its entirety, it would be nice to have the possibility of streaming in GeoJSON features. Jawn is a JSON parsing library written by @non which might be worth investigating. It seems like it can work as an alternative backend for spray-json and it supports a few different forms of asynchronous parsing: https://github.com/non/jawn#parsing

pomadchin:
After moving to fs2 streams, we can have a look at circe https://github.com/circe/circe/tree/master/examples/sf-city-lots

PR #3006, opened 2019-06-25
Replace spray with circe

moradology:
## Overview

This PR includes the changes from #2905 cherry-picked and placed on top of the project structure changes for GT 3.0 and should be reviewed/merged in as soon as we can to avoid further need to rebase

Partially addresses #1772


raster/src/main/scala/geotrellis/raster/io/json/Implicits.scala
```diff
@@ -16,113 +16,28 @@
...
 
-    def read(value: JsValue): GridBounds[Long] =
-      value.asJsObject.getFields("colMin", "rowMin", "colMax", "rowMax") match {
-        case Seq(JsNumber(colMin), JsNumber(rowMin), JsNumber(colMax), JsNumber(rowMax)) =>
-          GridBounds(colMin.toLong, rowMin.toLong, colMax.toLong, rowMax.toLong)
-        case _ =>
-          throw new DeserializationException("GridBounds expected.")
-      }
-  }
+  implicit val gridBoundsIntEncoder: Encoder[GridBounds[Int]] = deriveEncoder
```
pomadchin: These are already in a `GridBounds` companion object

s3-spark/src/main/scala/geotrellis/spark/store/s3/geotiff/S3JsonGeoTiffAttributeStore.scala
```diff
@@ -56,9 +56,13 @@ import scala.io.Source
         .mkString(" ")
     } finally objStream.close()
 
-    json
-      .parseJson
-      .convertTo[List[GeoTiffMetadata]]
+    (for {
+      parsed <- parse(json)
+      md <- parsed.as[List[GeoTiffMetadata]]
```
pomadchin: what was wrong with `parse(json).flatMap(_.as[List[GeoTiffMetadata]]).valueOr(throw _)` ?

s3/src/main/scala/geotrellis/store/s3/S3AttributeStore.scala
```diff
@@ -79,18 +78,24 @@ class S3AttributeStore(
         s3objStream.close()
       }
 
-    json.parseJson.convertTo[(LayerId, T)]
+    (for {
```
pomadchin: The same question is here; 

```scala
parse(json).flatMap(_.as[(LayerId, T)]).valueOr(throw _)
``` 

I see no reasons to use a `for` comprehension here with a manual match, since we depend on cats anyway.

project/Dependencies.scala
```diff
@@ -36,6 +36,7 @@ object Dependencies {
   val sprayJson           = "io.spray"                   %% "spray-json"               % Version.sprayJson
```
pomadchin: Spray dep is not removed

PR #3007, opened 2019-06-25
Refactor IO thread pool usage

### LINKED ISSUE #2945: Refactor IO thread pool usage
As the result of the https://github.com/locationtech/geotrellis/issues/2936 we came to a conclusion that we need to refactor thread pool usages. 

At this point we use a fixed thread pool for:

1. Each backend and each layer reader uses its own thread pool with it's own thread pool settings.
2. We use a thread pool per partition (all have different configurations).
3. Test it on EMR using [this project](https://github.com/pomadchin/vlm-performance).

The solution is to use a single thread pool for all IO operations and to use it _per executor_.
This also will touch all configuration files, instead of snake case we need to use hyphen-separated case.

pomadchin:
## Overview

This PR refactors GeoTrellis blocking thread pool usages. Instead of having multiple thread pools for each particular backend / reader / writer there would be a single and configurable thread pool. By default it is a `FixedThreadPool` with a `geotrellis-default-io-%d` name. In addition to that, it was decided to make it configurable by user. This PR allows users to use their own thread pools for all kind of readers / writers. 

In this PR we also introduced a new way to pass objects that should not be serialized (`ExecutionContext` and `S3Client`) into objects constructors. Instead of passing a function to create an instance of a thing, it was decided to pass these objects as `by-name` parameters. Such approach allows to simplify the API and makes the API more transparent. 

This PR touches all the project configuration files, and with this PR we introduce `hyphen-separated` case usage by default instead of a `camel-case`.

- [x] Test on EMR configurable common fixed thread pool
- [x] Test on EMR execution context passed instead of a threads number

Closes #2945


store/src/main/scala/geotrellis/store/file/FileCollectionLayerProvider.scala
```diff
@@ -45,6 +45,6 @@ class FileCollectionLayerProvider extends AttributeStoreProvider with ValueReade
 
   def collectionLayerReader(uri: URI, store: AttributeStore): CollectionLayerReader[LayerId] = {
     val catalogPath = new File(uri).getCanonicalPath
-    new FileCollectionLayerReader(store, catalogPath)
+    FileCollectionLayerReader(store, catalogPath)
```
echeipesh: Why do you hate the `new` keyword?

s3/src/main/scala/geotrellis/store/s3/cog/S3COGCollectionLayerReader.scala
```diff
@@ -44,11 +44,11 @@ import scala.reflect.ClassTag
 class S3COGCollectionLayerReader(
   val attributeStore: AttributeStore,
   val getClient: () => S3Client = S3ClientProducer.get,
-  val defaultThreads: Int = S3Config.threads.collection.readThreads
+  val getExecutionContext: () => ExecutionContext = () => BlockingThreadPool.executionContext
```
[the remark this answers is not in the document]
pomadchin: @echeipesh it can also be a `by-name` parameter, I can check this as well.

spark/src/main/scala/geotrellis/spark/store/file/FileRDDReader.scala
```diff
@@ -16,34 +16,31 @@
...
-
   def read[K: AvroRecordCodec: Boundable, V: AvroRecordCodec](
     keyPath: BigInt => String,
     queryKeyBounds: Seq[KeyBounds[K]],
     decomposeBounds: KeyBounds[K] => Seq[(BigInt, BigInt)],
     filterIndexOnly: Boolean,
     writerSchema: Option[Schema] = None,
     numPartitions: Option[Int] = None,
-    threads: Int = defaultThreadCount
+    getExecutionContext: () => ExecutionContext = () => BlockingThreadPool.executionContext
```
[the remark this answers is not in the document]
pomadchin: @echeipesh I think will work; I'm dumb (: will check it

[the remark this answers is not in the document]
echeipesh: If that works than we can both feel pretty dumb.

s3/src/main/scala/geotrellis/store/s3/cog/S3COGCollectionLayerReader.scala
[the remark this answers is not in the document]
pomadchin: do you want to refactor this way all our classes?

store/src/main/scala/geotrellis/store/util/BlockingThreadPool.scala
```diff
@@ -14,12 +14,14 @@
...
+import geotrellis.store.conf.BlockingThreadPoolConfig
 
+import java.util.concurrent.{ExecutorService, Executors}
+import scala.concurrent.ExecutionContext
 
-trait CamelCaseConfig {
-  implicit def hint[T]: ProductHint[T] = ProductHint[T](ConfigFieldMapping(CamelCase, CamelCase))
+object BlockingThreadPool extends Serializable {
+  @transient lazy val pool: ExecutorService = Executors.newFixedThreadPool(BlockingThreadPoolConfig.conf.threads)
+  @transient lazy val executionContext: ExecutionContext = ExecutionContext.fromExecutor(pool)
```
echeipesh: This may be a comment to be satisfied by another PR/issue, but I think it's useful to ask: How is this thread pool going to be named to facilitate debugging? When a user connects to VM is it going to be `BlockingThreadPool` and what will be the names of threads? It would be ideal if they had names like `geotrellis-defualt-io-thread-##`  and belong to thread pool that somehow was connected to GeoTrellis project.

store/src/main/scala/geotrellis/store/file/FileCollectionLayerReader.scala
```diff
@@ -68,7 +73,7 @@ class FileCollectionLayerReader(
 
 object FileCollectionLayerReader {
   def apply(attributeStore: AttributeStore, catalogPath: String): FileCollectionLayerReader =
-    new FileCollectionLayerReader(attributeStore, catalogPath)
+    new FileCollectionLayerReader(attributeStore, catalogPath, BlockingThreadPool.executionContext)
```
echeipesh: Are you avoiding assigning `BlockingThreadPool.executionContext` as default value here intentionally? If so what is the rationale?

pomadchin: @echeipesh it is a default here; what do you mean?

store/src/main/scala/geotrellis/store/conf/BlockingThreadPoolConfig.scala
```diff
@@ -0,0 +1,39 @@
...
+        case Success(threads) => Right(BlockingThreadPoolConfig(threads))
+        case Failure(_) => Right(BlockingThreadPoolConfig())
+      }
+      case Left(_) => Right(BlockingThreadPoolConfig())
+    }
+  }
+
+  lazy val conf: BlockingThreadPoolConfig = pureconfig.loadConfigOrThrow[BlockingThreadPoolConfig]("geotrellis.blocking-thread-pool")
+  implicit def blockingThreadPoolConfigToClass(obj: BlockingThreadPoolConfig.type): BlockingThreadPoolConfig = conf
+}
```
echeipesh: Do you think it would be good form to move this to `BlockingTheadPool` companion object, maybe call it `BlockingThreadPool.Config` and have `implicit val blockingThreadPoolConfigReader`. 

I'm mostly looking for a way to consolidate Configuration and the thing being configured so we don't have to look through so many files when looking for one vs the other. Again, if this is intentional, please let me know so I can we can enter it into unofficial GT best practices.

store/src/main/scala/geotrellis/store/file/FileCollectionLayerReader.scala
echeipesh: Oh, I mean about the `FileCollectionLayerReader` constructor rather than the `apply`.

s3-spark/src/test/scala/geotrellis/spark/store/s3/S3SpaceTimeSpec.scala
```diff
@@ -47,28 +38,26 @@ class S3SpaceTimeSpec
 
   lazy val bucket = "mock-bucket"
   lazy val prefix = "catalog"
-  val client = MockS3Client()
+  val client = MockS3Client.instance
```
echeipesh: Funny that this doesn't have to be `@transient lazy val` because its used for local tests only.

store/src/main/scala/geotrellis/store/file/FileCollectionLayerReader.scala
pomadchin: @echeipesh oh, a great question! my answer would be that I didn't think about it too much and was worrying mostly about serialization and changing constructors / etc minimally since it's already a significant PR; But I see what you're saying. We probably need to review all IOReaders / Writers constructors and apply methods. Don't know why does it leave in the companion object actually.

echeipesh: This is the only one I've spotted so far

pomadchin: If its the only one then it is an oversight :+1:

store/src/main/scala/geotrellis/store/util/BlockingThreadPool.scala
pomadchin: @echeipesh `geotrellis-default-io-%d` now

store/src/main/scala/geotrellis/store/cog/COGCollectionLayerReader.scala
```diff
@@ -28,12 +28,13 @@ import io.circe._
...
 import cats.syntax.either._
 
+import scala.concurrent.ExecutionContext
 import scala.reflect._
 import java.net.URI
 import java.util.ServiceLoader
 
-
 abstract class COGCollectionLayerReader[ID] { self =>
+  implicit val ec: ExecutionContext
```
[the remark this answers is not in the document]
pomadchin: For the `COGCollectionLayerReader.read` I think it makes not a lot of sense to pass execution context explicitly in the code that can work only 'locally' and is internal API in fact.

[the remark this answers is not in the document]
pomadchin: Also do I understand correct that your 1. comment can be applied to all our abstract classes? (`CollectionLayerReader` for instance)

s3-spark/src/test/scala/geotrellis/spark/store/s3/S3SpaceTimeSpec.scala
pomadchin: Ah, that's because it is used only in the following block of code: 

```
val client = MockS3Client.instance
S3TestUtils.cleanBucket(client, bucket)
registerAfterAll { () =>
  S3TestUtils.cleanBucket(client, bucket)
}
```

PR #3009, opened 2019-06-27
Replace mergable with semigroup

### LINKED ISSUE #2925: Replace `Mergeable` with Cats' `Semigroup`

moradology:
## Overview

The `mergable` trait from GT is just cats' `Semigroup`. As we are now using cats, this PR replaces mergable with Semigroup.

### Notes

This is pretty straightforward and doesn't attempt to rework any of the patterns (which should be revisited) through which we use `mergable` (and now `Semigroup`)

Closes #2925

### DIFF
--- a/cassandra-spark/src/main/scala/geotrellis/spark/store/cassandra/CassandraLayerWriter.scala
+++ b/cassandra-spark/src/main/scala/geotrellis/spark/store/cassandra/CassandraLayerWriter.scala
@@ -27,11 +27,8 @@ import geotrellis.spark.merge._
 import geotrellis.util._
 
 import com.typesafe.scalalogging.LazyLogging
-
 import org.apache.spark.rdd.RDD
-
 import io.circe._
-
 import cats.Semigroup
 
 import scala.reflect._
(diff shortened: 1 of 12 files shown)


cassandra-spark/src/main/scala/geotrellis/spark/store/cassandra/CassandraLayerWriter.scala
```diff
@@ -22,14 +22,17 @@ import geotrellis.store.avro._
...
 import geotrellis.spark.store._
 import geotrellis.spark.merge._
 import geotrellis.util._
 
 import com.typesafe.scalalogging.LazyLogging
+
 import org.apache.spark.rdd.RDD
-import io.circe._
+
+import _root_.io.circe._
```
pomadchin: why is it a `_root_` now?

moradology: Consistency throughout the library; certainly not necessity

pomadchin: mb lets use _root_ where is necessary? originally I tired to use it only in places were packages were overlapped.

echeipesh: Looks good otherwise.

PR #3014, opened 2019-07-08
Add dynamic ZFactor for slope calculation

### LINKED ISSUE #2548: Create a Method That Calculates ZFactor
Right now, we supply a `ZFactor` of 1 for the `Slope` operation on `RDD`s https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/mapalgebra/focal/FocalTileLayerCollectionMethods.scala#L49 That does not make sense, though, as `ZFactor` should be calculated on a per `Tile` basis rather than for the entire `RDD`.  However, expecting the user to supply a unique `ZFactor` for every `Tile` is unreasonable. We should instead implement a `ZFactorCalculator` similar to what is in GeoPySpark: https://github.com/locationtech-labs/geopyspark/blob/master/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/ZFactorCalculator.scala#L13 This would allow us to calculate a `ZFactor` for each `Tile` based on layer's projection.

jbouffard:
## Overview

This PR adds the `ZFactor` class as a way of producing the correct `zFactor` when calculating `slope`. Before, the `zFactor` had a default value of 1. However, this doesn't make sense, as the `zFactor` needs to be calculated on a per-tile bases. The `ZFactor` class will help by deriving the appropriate `ZFactor` for each tile depending on its spatial position.

### Checklist

Closes #2548


docs/CHANGELOG.rst
```diff
@@ -9,10 +9,13 @@ API Changes & Project structure changes
 - ``geotrellis-raster``
 
   - **Change:** ``geotrellis.raster.summary.polygonal.[Multi]TilePolygonalSummaryHandler`` replaced with ``geotrellis.raster.summary.polygonal.PolygonalSummary. Users should expect to implement concrete subclasses of ``geotrellis.raster.summary.GridVisitor`` and pass those to the new polygonalSummary methods. There are a number of default GridVisitor implementations provided for simple operations in `geotrellis.raster.summary.visitors``
+  - **New:** Created the ``geotrellis.raster.mapalgebra.focal.ZFactorCalculator`` and the ``geotrellis.raster.mapalgebra.focal.TileUnit`` types to improve
```
pomadchin: Hm I think I should have mentioned it before, but mb it makes sense to add references to exact PRs with changes?

layer/src/main/scala/geotrellis/layer/mapalgebra/focal/FocalTileLayerCollectionMethods.scala
```diff
@@ -18,7 +18,7 @@ package geotrellis.layer.mapalgebra.focal
 
 import geotrellis.raster._
 import geotrellis.raster.Neighborhoods.Square
-import geotrellis.raster.mapalgebra.focal.{Kernel, Aspect, Sum, Max, Min, Mean, Median, Mode, StandardDeviation, Convolve, Slope}
+import geotrellis.raster.mapalgebra.focal.{TargetCell => TC, _}
```
pomadchin: Why do we need this import change here?

spark/src/main/scala/geotrellis/spark/mapalgebra/focal/FocalTileLayerRDDMethods.scala
```diff
@@ -18,7 +18,7 @@ package geotrellis.spark.mapalgebra.focal
 
 import geotrellis.raster._
 import geotrellis.raster.Neighborhoods.Square
-import geotrellis.raster.mapalgebra.focal.{Kernel, Aspect, Sum, Max, Min, Mean, Median, Mode, StandardDeviation, Convolve, Slope}
+import geotrellis.raster.mapalgebra.focal.{TargetCell => TC, _}
```
pomadchin: Here again a not required change? O:

layer/src/main/scala/geotrellis/layer/mapalgebra/focal/FocalTileLayerCollectionMethods.scala
jbouffard: Oh, that was something I meant to change back. I will do that now :+1:

jbouffard:
@pomadchin Thanks! I should've resolved the issues you pointed out. Also, thanks for reminding me about the headers. I knew I forgot something...

raster/src/main/scala/geotrellis/raster/mapalgebra/focal/ZFactorCalculator.scala
```diff
@@ -0,0 +1,88 @@
...
+import org.apache.commons.math3.analysis.interpolation._
+
+
+/** Produces a ZFactor for a given point using the prescribed
+ *  conversion method.
+ *
+ *  @param produceZFactor A function that takes a latitude point and
+ *    converts it to its corresponding ZFactor.
+ */
+class ZFactorCalculator(produceZFactor: Double => Double) extends Serializable {
```
echeipesh: Continuing on the naming kick `class ZFactorCalculator` can be `class ZFactor`, the "calculator" part is inherent in the fact that it a class that has behavior and not a single value. That also gets us `ZFactor.forLatLng` and can have an overload for constant `ZFactor(5)`

raster/src/main/scala/geotrellis/raster/mapalgebra/focal/ZFactorCalculator.scala
```diff
@@ -0,0 +1,88 @@
...
+  /** Produces the associated ZFactor for the given location.
+   *
+   *  @param extent The aread of interest. The center point will be used to determine
+   *    the ZFactor.
+   */
+  def deriveZFactor(extent: Extent): Double =
```
echeipesh: Naming, `val zfactor = ZFactorCalculator.deriveZFactor`, the method name can be more terse and meaningful.

raster/src/main/scala/geotrellis/raster/mapalgebra/focal/Unit.scala
```diff
@@ -0,0 +1,7 @@
+package geotrellis.raster.mapalgebra.focal
+
+
+trait Unit
+
+case object Feet extends Unit
+case object Meters extends Unit
```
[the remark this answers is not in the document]
pomadchin: +1 for `squants`; there are some alternatives in Java world but they are not that popular.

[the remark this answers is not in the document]
jbouffard: :+1: I wasn't aware of that library, but it looks good.

spark/src/main/scala/geotrellis/spark/mapalgebra/focal/FocalOperation.scala
```diff
@@ -55,6 +56,34 @@ object FocalOperation {
...
+  )(calc: (Tile, Option[GridBounds[Int]], Extent) => Tile): RDD[(K, Tile)] =
+      deriveSlope(rdd.bufferTiles(neighborhood.extent, layerBounds, partitioner), neighborhood, keyToExtent)(calc)
+
+  def deriveSlope[
+    K: SpatialComponent: ClassTag: GetComponent[?, SpatialKey]
+  ](rasterRDD: TileLayerRDD[K], neighborhood: Neighborhood, partitioner: Option[Partitioner], keyToExtent: SpatialKey => Extent)
+  (calc: (Tile, Option[GridBounds[Int]], Extent) => Tile): TileLayerRDD[K] =
+    rasterRDD.withContext { rdd =>
+      deriveSlope(rdd, neighborhood, rasterRDD.metadata.tileBounds, partitioner, keyToExtent)(calc)
+    }
```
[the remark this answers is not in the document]
jbouffard: Yeah, that isn't the best name. I believe that I did change the name to `slope` in the `layer` version of this file, but not this one. i agree that even `slope` isn't that great, though. I just couldn't think of a better name.

echeipesh:
I'm 👍  on `squants` after realizing there still isn't a standard Java units library.
Just checked, it needs CQ added: https://github.com/locationtech/geotrellis/blob/master/docs/CQ.rst

jbouffard:
Just created the CQ for squants: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20340#c0

spark/src/main/scala/geotrellis/spark/mapalgebra/focal/FocalOperation.scala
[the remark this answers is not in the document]
echeipesh: Wait, it's certainly not `slope`, it only calculates slope if the `calc` argument happens to be that.

Looking at the use site to see what would make the most sense..

```scala
// current
FocalOperation.slope(self, n, partitioner, keyToExtent){ (raster, bounds) => calc(raster, bounds, cellSize) }
```

So really its applying a `calc` as a focal operation, taking care of borders etc.

In fact there is already an `FocalOperation.apply` in that is similar, just lacking an __ argument

```scala
  def apply[K: SpatialComponent: ClassTag](
    rdd: RDD[(K, Tile)],
    neighborhood: Neighborhood,
    layerBounds: TileBounds,
    partitioner: Option[Partitioner])
    (calc: (Tile, Option[GridBounds[Int]]) => Tile): RDD[(K, Tile)] =
      mapOverBufferedTiles(rdd.bufferTiles(neighborhood.extent, layerBounds, partitioner), neighborhood)(calc)
``` 

So the key difference seems to be the signature of `calc` function

```scala
// current
calc: (Tile, Option[GridBounds[Int]]) => Tile

// additional
calc: (Raster[Tile], Option[GridBounds[Int]]) => Tile
```

So, this can certainly be another `apply` overload. Heads up, there might be an issue duplicate signature for the method that will need worked around. So it may have to be `applyOnRaster` or something like that.

echeipesh:
@jbouffard heads up, you attached the artifact jar instead of the source jar on that CQ request.

PR #3017, opened 2019-07-08
Add sparse stitch method to StitchCollectionMethods

### LINKED ISSUE #2903: TileLayoutStitcher.stitch sparseness assumption
The current stitch function is a 'dense' stitch function. If some keys are missing it will stitch things as close as is possible (it stitches based on key order rather than the value/true extent of keys). It assumes that every possible key in some region will exist in the iterator being stitched into a single tile.

We need to add a stitch function that will work correctly even with missing / skipped keys - a 'sparse' stitch

CloudNiner:
## Overview

Adds sparseStitch methods to `SpatialTileLayoutCollectionStitchMethods`. The method uses a different algorithm, and therefore it felt more correct to name it differently from the original `stitch(): Raster[V]`. Also added a helper so that users who want to use the collection extent don't have to know how to get it, since this is the most likely default use case.

Opened PR to get tests to run, since there's a breaking change on the type bounds of V. Initial testing in a few key places didn't raise any issues but I want the whole test suite to run. **Edit:** Full test suite passed, but upon further inspection it looks like we might not test stitch anywhere else.

### Checklist

### Notes

I want to add a unit test or two because there's nothing at all for any of these stitch functions, but that looks like its a bit of an effort since there's no easy test data constructor for the type we need to test. So submitting this for review now and moving on to a different task. I can attempt to add a test or two once we've settled on approval for the API presented here. 

Closes #2903

### DIFF
--- a/layer/src/main/scala/geotrellis/layer/stitch/StitchCollectionMethods.scala
+++ b/layer/src/main/scala/geotrellis/layer/stitch/StitchCollectionMethods.scala
@@ -17,15 +17,14 @@
 package geotrellis.layer.stitch
 
 import geotrellis.raster._
-import geotrellis.raster.prototype._
-import geotrellis.raster.merge._
 import geotrellis.raster.stitch.Stitcher
 import geotrellis.layer._
+import geotrellis.raster.prototype._
 import geotrellis.vector.Extent
 import geotrellis.util._
 
 abstract class SpatialTileLayoutCollectionStitchMethods[
-  V <: CellGrid[Int]: Stitcher,
+  V <: CellGrid[Int]: Stitcher: ? => TilePrototypeMethods[V],
   M: GetComponent[?, LayoutDefinition]
 ] extends MethodExtensions[Seq[(SpatialKey, V)] with Metadata[M]] {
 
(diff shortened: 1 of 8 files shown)


CloudNiner:
> mb it makes sense to add this method into SpatialTileLayoutRDDStitchMethods

Hmm.  I'm torn. I _think_ to add that method in that class I'd have to duplicate most of the logic and then replace the self.stitch with self.collect.stitch. Given that, it seems easier, if you have a `TileLayerRDD`, to just do `tileLayerRdd.collect.toSeq.sparseStitch(extent)` which should be possible with the current changes in this PR as long as the stitch implicits are in scope. It doesn't hide the collect (which is dangerous) and is still quite clear. 

> Only tests are required

Done. Added a few simple ones to `geotrellis.spark.stitch.CollectionStitchMethodsSpec`. It's an artifact of the spark split, but its still weird that a bunch of tests for functionality defined in layer are tested in the spark package. It would be very verbose to test those layer features without some of the spark helpers to construct the collections.

pomadchin:
@CloudNiner but we already [have this duplication](https://github.com/locationtech/geotrellis/blob/ec9a2db273917e773bf022d49a198fd95ad4bbb2/spark/src/main/scala/geotrellis/spark/stitch/StitchRDDMethods.scala) 

It is our API problem - that for the current collections API we have to duplicate everything. If you'll look more close - Collection API completely duplicates the RDD API ¯\_(ツ)_/¯ 

I dunno what is better at the moment - to keep consistency or to add ambiguity by adding such a divergence into the API.

CloudNiner:
Ah, if we're going for 100% feature parity between RDD and Collections API, then I guess I have to add it. Bummer.

pomadchin:
Yea, I don't disagree with your point about duplication ): but it looks like at least for now it is better to add to keep everything consistent

CloudNiner:
@pomadchin updated to include duplicate methods for RDDs

spark/src/main/scala/geotrellis/spark/stitch/StitchRDDMethods.scala
```diff
@@ -41,6 +41,58 @@ abstract class SpatialTileLayoutRDDStitchMethods[
...
+        .map { case (x, y) => SpatialKey(x, y) }
+        .toList
+      val actualKeys = tiles.map(_._1)
+      val missingKeys = expectedKeys diff actualKeys
+
+      val missingTiles = missingKeys.map { key =>
+        (key, tile.prototype(layoutDefinition.tileLayout.tileCols, layoutDefinition.tileLayout.tileRows))
+      }
+      val allTiles = tiles.withContext { collection =>
+        collection.toList ::: missingTiles
```
moradology: I prefer the `++` operator for concatenation. It works with non-lists and does the same thing as `:::` (falling back to it when dealing with lists according to a SO post I recall having read)

CloudNiner: I don't feel strongly about it, so I updated it. Thanks for taking a look at the PR.

layer/src/main/scala/geotrellis/layer/stitch/StitchCollectionMethods.scala
```diff
@@ -38,6 +37,52 @@ abstract class SpatialTileLayoutCollectionStitchMethods[
...
+        .map { case (x, y) => SpatialKey(x, y) }
+        .toList
+      val actualKeys = self.map(_._1)
+      val missingKeys = expectedKeys diff actualKeys
+
+      val missingTiles = missingKeys.map { key =>
+        (key, tile.prototype(layoutDefinition.tileLayout.tileCols, layoutDefinition.tileLayout.tileRows))
+      }
+      val allTiles = self.withContext { collection =>
+        collection.toList ++ missingTiles
```
pomadchin: since you're using `++` you can concat sequences without .toList conversion `collection ++ missingTiles`

CloudNiner: No worries, updated.

PR #3025, opened 2019-07-16
S3RangeReader totalLength Improvment

### LINKED ISSUE #2889: S3RangeReader will always produce HEAD request
In order to produce object metadata and its overall size the underlying S3 client will issue a `HEAD` request: 

https://github.com/locationtech/geotrellis/blob/f86ef9eb459b219a5ce8e082f7b8d1f9431a6995/s3/src/main/scala/geotrellis/spark/io/s3/util/S3RangeReader.scala#L39-L42

This is problematic for two reasons:
- Sometimes the object ACT will not allow `HEAD` requests, resulting in error `400`
- In most use cases `totalLength` is not used and is not important.

To expand on the second point, the GeoTrellis GeoTiffReader will read from the start of the file until it consumes the full header. At that point it will use the offset information from the GeoTiff header to read the data segments.

Note that `HttpRangeReader` already has logic for this case:

https://github.com/locationtech/geotrellis/blob/f86ef9eb459b219a5ce8e082f7b8d1f9431a6995/spark/src/main/scala/geotrellis/spark/io/http/util/HttpRangeReader.scala#L38-L64

At a minimum the `totalSize` should be turned into `lazy val` which would avoid the issue in most cases. S3RangeReader should also have a constructor flag to avoid head requests. This should cause re-evaluation of the `RangeReader` interface, maybe we don't need that method at all?

jbouffard:
## Overview

This PR changes how the `totalLength` value of `S3RangeReader` is computed. It will now use a `GetObjectResponse` instead of a `GetHeaderResponse`. This will allow users to read these files even if they don't have permission to access its `HEADER`.

### Checklist

- [x] Check it on a real S3 bucket with permission limits

Closes #2889

### DIFF
--- a/s3/src/main/scala/geotrellis/store/s3/util/S3RangeReader.scala
+++ b/s3/src/main/scala/geotrellis/store/s3/util/S3RangeReader.scala
@@ -39,18 +39,13 @@ class S3RangeReader(
   client: S3Client
 ) extends RangeReader {
 
-  lazy val metadata: HeadObjectResponse = {
-    val headRequest =
-      HeadObjectRequest
-        .builder()
-        .bucket(request.bucket)
-        .key(request.key)
-        .build()
+  lazy val totalLength: Long = {
+    val responseStream = client.getObject(request)
+    val length = responseStream.response.contentLength
 
-    client.headObject(headRequest)
+    responseStream.close()
+    length
   }
-
-  lazy val totalLength: Long = metadata.contentLength
 
   def readClippedRange(start: Long, length: Int): Array[Byte] = {
     val getRequest =
--- a/s3/src/main/scala/geotrellis/store/s3/util/S3RangeReader.scala
+++ b/s3/src/main/scala/geotrellis/store/s3/util/S3RangeReader.scala
@@ -36,15 +36,27 @@ import java.net.URI
  */
 class S3RangeReader(
   request: => GetObjectRequest,
-  client: S3Client
+  client: S3Client,
+  readHeader: Boolean = true
 ) extends RangeReader {
 
   lazy val totalLength: Long = {
-    val responseStream = client.getObject(request)
-    val length = responseStream.response.contentLength
+    if (readHeader) {
+      val headRequest =
+        HeadObjectRequest
+          .builder()
+          .bucket(request.bucket)
+          .key(request.key)
+          .build()
 
-    responseStream.close()
-    length
+      client.headObject(headRequest).contentLength
+    } else {
+      val responseStream = client.getObject(request)
+      val length = responseStream.response.contentLength
+
+      responseStream.close()
+      length
+    }
   }
 
   def readClippedRange(start: Long, length: Int): Array[Byte] = {
--- a/s3/src/main/scala/geotrellis/store/s3/util/S3RangeReader.scala
+++ b/s3/src/main/scala/geotrellis/store/s3/util/S3RangeReader.scala
@@ -32,12 +32,13 @@ import java.net.URI
  *
  * @param request: A [[GetObjectRequest]] of the desired GeoTiff.
  * @param client: The [[S3Client]] that retrieves the data.
+ * @param readHeader: Whether the `HEAD` of the target object should be read or not. Default is `false`.
  * @return A new instance of S3RangeReader.
  */
 class S3RangeReader(
   request: => GetObjectRequest,
   client: S3Client,
-  readHeader: Boolean = true
+  readHeader: Boolean = false
 ) extends RangeReader {
 
   lazy val totalLength: Long = {
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -113,6 +113,7 @@ Fixes & Updates
 - ArrayTile equality will now check the cols, rows, and cellType of the two tiles (`#2991 <https://github.com/locationtech/geotrellis/pull/2991>`_).
 - Fix incorrect Deflate compressor usage (`#2997 <https://github.com/locationtech/geotrellis/pull/2997>`_).
 - Refactor IO thread pool usage (`#3007 <https://github.com/locationtech/geotrellis/pull/3007>`_).
+- ``S3RangeReader`` will now use a ``GetObjectResponse`` to compute the ``totalLength`` (`#3025 <https://github.com/locationtech/geotrellis/pull/3025>`_).
 
 2.3.0
 -----
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -113,7 +113,7 @@ Fixes & Updates
 - ArrayTile equality will now check the cols, rows, and cellType of the two tiles (`#2991 <https://github.com/locationtech/geotrellis/pull/2991>`_).
 - Fix incorrect Deflate compressor usage (`#2997 <https://github.com/locationtech/geotrellis/pull/2997>`_).
 - Refactor IO thread pool usage (`#3007 <https://github.com/locationtech/geotrellis/pull/3007>`_).
-- ``S3RangeReader`` will now use a ``GetObjectResponse`` to compute the ``totalLength`` (`#3025 <https://github.com/locationtech/geotrellis/pull/3025>`_).
+- ``S3RangeReader`` will now optionally read the HEADER of an object (`#3025 <https://github.com/locationtech/geotrellis/pull/3025>`_).
 
 2.3.0
 -----


jbouffard:
I'm not really sure there are any unit tests that can be created for this PR. It might be good to test this branch on a real world example.

s3/src/main/scala/geotrellis/store/s3/util/S3RangeReader.scala
```diff
@@ -39,19 +39,14 @@ class S3RangeReader(
...
-  lazy val metadata: HeadObjectResponse = {
-    val headRequest =
-      HeadObjectRequest
-        .builder()
-        .bucket(request.bucket)
-        .key(request.key)
-        .build()
+  lazy val totalLength: Long = {
+    val responseStream = client.getObject(request)
+    val length = responseStream.response.contentLength
```
jbouffard: Since we can just use a `GET` query to produce our `GetObjectResponse`, we don't need to worry about whether or not to reader the `HEADER` of the object.

pomadchin: @jbouffard I though [here](https://github.com/locationtech/geotrellis/issues/2889#issuecomment-511799089) was mentioned that we want to replicate [HttpRangeReader](https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/store/http/util/HttpRangeReader.scala#L34) approach.

jbouffard: @pomadchin Is there a problem with using `Get` for everything? I'm not really familiar with that operation.

pomadchin: We would like to use HEAD if possible (I think). 

See the description of [HEAD](https://docs.aws.amazon.com/en_us/AmazonS3/latest/API/RESTObjectHEAD.html) vs [GET](https://docs.aws.amazon.com/en_us/AmazonS3/latest/API/RESTObjectGET.html)

jbouffard: I see. You're right that it looks like `HEAD` is preferable.

s3/src/main/scala/geotrellis/store/s3/util/S3RangeReader.scala
```diff
@@ -36,22 +36,29 @@ import java.net.URI
  */
 class S3RangeReader(
   request: => GetObjectRequest,
-  client: S3Client
+  client: S3Client,
+  readHeader: Boolean = true
```
echeipesh: This should default to `false` just like `HttpRangeHeader`. Since this problem was encountered while reading one of the AWS public datasets (which relates to most of the base use cases for GeoTrellis) being able to handle that limited permission on some of those buckets should be considered the default.

s3/src/main/scala/geotrellis/store/s3/util/S3RangeReader.scala
```diff
@@ -36,22 +36,29 @@ import java.net.URI
...
+      val headRequest =
+        HeadObjectRequest
+          .builder()
+          .bucket(request.bucket)
+          .key(request.key)
+          .build()
+
+      client.headObject(headRequest).contentLength
+    } else {
+      val responseStream = client.getObject(request)
```
echeipesh: This is a neat trick if it works. I would say given how much we use this class a test is required. 

Unit test is not necessary but loading this branch in console and pointing at the largest GeoTiff you can find with `S3RangeReader` would be enough. If both header and this non-emptied InputStream provide the content length with-in the same time this is a nice win.

jbouffard: @echeipesh Will do :+1: Is there a particular GeoTiff you'd recommend?

pomadchin: @jbouffard I think you can upload any tiff from the test files an upload them into our bucket and check that you can read it with limited bucket query permissions.

jbouffard: @echeipesh @pomadchin I did some reading from the console and it didn't _seem_ like there was a big difference in time between the `HeadRequest` and the `ResponseStream`, but I could easily be mistaken.

pomadchin: @jbouffard ah, the issue was in permissions. Usually users in RF give RF user only ability to use Get requests / GetList and nothing more. The idea of this PR was to ensure that we can work even in the case we can't perform `HEAD` requests. Can you double-check that the S3 GetRequest pulls down data from a bucket with permissions only to `Get` file and that it can't to the same with a `Head` request?

pomadchin: @jbouffard can you also double check it with this tiff? `s3://geotrellis-test/issue-1691/firstPart.tif`? It's a `3.1 GB` tiff.

jbouffard: @echeipesh @pomadchin Unfortunately, it appears that using the `ResponseStream` will cause the client to pull down the whole file. I was able to confirm this using WireShark when finding the `totalLength` of `s3://geotrellis-test/issue-1691/firstPart.tif`.

jbouffard: @echeipesh After our conversation yesterday, I was curious to see what would happen if you tried to read past the file length on S3; and it turns out that it does throw an exception:

```
scala> r2.readRange(r2.totalLength, 10)
software.amazon.awssdk.services.s3.model.S3Exception: The requested range is not satisfiable (Service: S3, Status Code: 416, Request ID: B073A86E409076E4)
  at software.amazon.awssdk.core.internal.http.pipeline.stages.HandleResponseStage.handleErrorResponse(HandleResponseStage.java:115)
  at software.amazon.awssdk.core.internal.http.pipeline.stages.HandleResponseStage.handleResponse(HandleResponseStage.java:73)
  at software.amazon.awssdk.core.internal.http.pipeline.stages.HandleResponseStage.execute(HandleResponseStage.java:58)
  at software.amazon.awssdk.core.internal.http.pipeline.stages.HandleResponseStage.execute(HandleResponseStage.java:41)
  at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
  at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallAttemptTimeoutTrackingStage.execute(ApiCallAttemptTimeoutTrackingStage.java:64)
  at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallAttemptTimeoutTrackingStage.execute(ApiCallAttemptTimeoutTrackingStage.java:36)
  at software.amazon.awssdk.core.internal.http.pipeline.stages.TimeoutExceptionHandlingStage.execute(TimeoutExceptionHandlingStage.java:77)
  at software.amazon.awssdk.core.internal.http.pipeline.stages.TimeoutExceptionHandlingStage.execute(TimeoutExceptionHandlingStage.java:39)
  at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage$RetryExecutor.doExecute(RetryableStage.java:113)
  at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage$RetryExecutor.execute(RetryableStage.java:86)
  at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage.execute(RetryableStage.java:62)
  at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage.execute(RetryableStage.java:42)
  at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
  at software.amazon.awssdk.core.internal.http.StreamManagingStage.execute(StreamManagingStage.java:57)
  at software.amazon.awssdk.core.internal.http.StreamManagingStage.execute(StreamManagingStage.java:37)
  at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.executeWithTimer(ApiCallTimeoutTrackingStage.java:80)
  at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.execute(ApiCallTimeoutTrackingStage.java:60)
  at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.execute(ApiCallTimeoutTrackingStage.java:42)
  at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
  at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
  at software.amazon.awssdk.core.internal.http.pipeline.stages.ExecutionFailureExceptionReportingStage.execute(ExecutionFailureExceptionReportingStage.java:37)
  at software.amazon.awssdk.core.internal.http.pipeline.stages.ExecutionFailureExceptionReportingStage.execute(ExecutionFailureExceptionReportingStage.java:26)
  at software.amazon.awssdk.core.internal.http.AmazonSyncHttpClient$RequestExecutionBuilderImpl.execute(AmazonSyncHttpClient.java:240)
  at software.amazon.awssdk.core.client.handler.BaseSyncClientHandler.invoke(BaseSyncClientHandler.java:96)
  at software.amazon.awssdk.core.client.handler.BaseSyncClientHandler.execute(BaseSyncClientHandler.java:120)
  at software.amazon.awssdk.core.client.handler.BaseSyncClientHandler.execute(BaseSyncClientHandler.java:61)
  at software.amazon.awssdk.core.client.handler.SdkSyncClientHandler.execute(SdkSyncClientHandler.java:51)
  at software.amazon.awssdk.awscore.client.handler.AwsSyncClientHandler.execute(AwsSyncClientHandler.java:62)
  at software.amazon.awssdk.services.s3.DefaultS3Client.getObject(DefaultS3Client.java:1597)
  at software.amazon.awssdk.services.s3.S3Client.getObject(S3Client.java:2363)
  at geotrellis.store.s3.util.S3RangeReader.readClippedRange(S3RangeReader.scala:72)
  at geotrellis.util.RangeReader$class.readRange(RangeReader.scala:41)
  at geotrellis.store.s3.util.S3RangeReader.readRange(S3RangeReader.scala:38)
  ... 40 elided
```

jbouffard:
@echeipesh @pomadchin After mentioning my results yesterday, @moradology looked into it as well and found that using a `responseStream` worked fine for finding the `contentLength`. I ran the code again this morning and observed similar speeds as Nathan. I'm not sure what happened yesterday :confused: As everything was so much slower. Regardless, this PR should be good to merge now.

PR #3032, opened 2019-07-19
Fix Circe decoders for TileLayerMetadata Schema and Metadata.layoutDefinition

### LINKED ISSUE #3029: AttributeStore metadata decoder fails on layout vs layoutDefinition
Old is `metadata.layoutDefinition` new is `metadata.layout`. Revert to old to ensure backwards compatibility continues.

This is a blocker for writing new layers with current GT 3.0 snapshot

### LINKED ISSUE #3030: AttributeStore schema renders as escaped string
Using latest Geotrellis 3.0.0 snapshot after the circe upgrade. This causes properly formatted metadata json files to fail to properly decode via `AttributeStore.getSchema(layerId)`

See example: `s3://geotrellis-test/rastersource-performance/_attributes/metadata__nlcd-v201-rastersource-avro__13.json`

Revert to it rendering an actual JSON object.

```
{
     "schema":"{\"type\":\"record\",\"name\":\"KeyValueRecord\", ...}"
}

### PR DESCRIPTION
**Pull request by CloudNiner**
So that we properly serialize/deserialize LayoutDefinition
to the "layoutDefinition" key instead of "layout" thus maintaining backwards compatibility for the metadata key of our geotrellis layer metadata.json files.

This also adds a test to help catch backwards compatibility issues
in encoding/decoding TileLayerMetadata

### Checklist

- [ ] ~`docs/CHANGELOG.rst` updated, if necessary~
- [ ] ~`docs` guides update, if necessary~
- [ ] ~New user API has useful Scaladoc strings~

### Demo

```scala
import geotrellis.layer._
import geotrellis.store._
import geotrellis.store.s3._

val bucket = "ingested-loca-data"
val prefix = "CCSM4"
val store = S3AttributeStore(bucket, prefix)
val id = LayerId("CCSM4_rcp85_r6i1p1", 0)
store.readMetadata[TileLayerMetadata[SpaceTimeKey]](id)

// Before:
io.circe.DecodingFailure$$anon$2: Attempt to decode value on failed cursor: DownField(layout),DownField(metadata)

// After
res0: geotrellis.layer.TileLayerMetadata[geotrellis.layer.SpaceTimeKey] = TileLayerMetadata(float32ud1.0000000150474662E30,LayoutDefinition(Extent(-126.0, 21.608163265306125, -62.0, 54.0),CellSize(0.0625,0.06326530612244897),4x2 tiles,1024x512 pixels),Extent(-126.0, 23.0, -66.0, 54.0),longlat-CS,KeyBounds(SpaceTimeKey(0,0,1514764800000),SpaceTimeKey(3,1,4133894400000)))
```

And:
```scala
store.readSchema(id)

// Before 
io.circe.DecodingFailure$$anon$2: Attempt to decode value on failed cursor: DownField(schema)

// After
res0: Schema = Schema(...)
```

Closes #3029 

Closes #3030

### DIFF
--- a/layer/src/test/scala/geotrellis/layer/TileLayerMetadataSpec.scala
+++ b/layer/src/test/scala/geotrellis/layer/TileLayerMetadataSpec.scala
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2019 Azavea
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package geotrellis.layer
+
+import geotrellis.raster._
+import geotrellis.proj4._
+import geotrellis.vector._
+import _root_.io.circe.syntax._
+import _root_.io.circe.parser.decode
+import geotrellis.raster
+import org.scalatest._
+
+class TileLayerMetadataSpec extends FunSpec with Matchers {
+  describe("TileLayerMetadata JSON codecs") {
+
+    val cellType = DoubleCellType
+    val layoutDefinition = LayoutDefinition(Extent(-126.0, 21.608163265306125, -62.0, 54.0),
+                                            TileLayout(4, 2, 256, 256))
+    val extent = Extent(-126.0, 23.0, -66.0, 54.0)
+    val bounds = KeyBounds(SpaceTimeKey(0,0,0), SpaceTimeKey(3,1,1000000))
+    val tileLayerMetadata: TileLayerMetadata[SpaceTimeKey] = TileLayerMetadata(
+      cellType,
+      layoutDefinition,
+      extent,
+      LatLng,
+      bounds
+    )
+
+    it("should convert TileLayerMetadata to json and back with specific JSON keys") {
+      val json = tileLayerMetadata.asJson
+      json.hcursor.downField("cellType").as[CellType] should equal(Right(cellType))
+      json.hcursor.downField("layoutDefinition").as[LayoutDefinition] should equal(Right(layoutDefinition))
+      json.hcursor.downField("extent").as[Extent] should equal(Right(extent))
+      json.hcursor.downField("crs").as[CRS] should equal(Right(LatLng))
+      json.hcursor.downField("bounds").as[Bounds[SpaceTimeKey]] should equal(Right(bounds))
+
+      json.as[TileLayerMetadata[SpaceTimeKey]].right.map(_ should equal(tileLayerMetadata))
+    }
+
+    it("should parse a GeoTrellis 1.2 TileLayerMetadata object") {
+      // NOTE: If this test breaks because the keys in TileLayerMetadata have changed, then
+      //       we've broken backwards compatibility in reading TileLayerMetadata in our layers.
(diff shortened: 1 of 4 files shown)


layer/src/test/scala/geotrellis/layer/TileLayerMetadataSpec.scala
```diff
@@ -0,0 +1,63 @@
...
+      json.hcursor.downField("crs").as[CRS] should equal(Right(LatLng))
+      json.hcursor.downField("bounds").as[Bounds[SpaceTimeKey]] should equal(Right(bounds))
+
+      json.as[TileLayerMetadata[SpaceTimeKey]].right.map(_ should equal(tileLayerMetadata))
+    }
+
+    it("should parse a GeoTrellis 1.2 TileLayerMetadata object") {
+      // NOTE: If this test breaks because the keys in TileLayerMetadata have changed, then
+      //       we've broken backwards compatibility in reading TileLayerMetadata in our layers.
+      val metadata: String = "{\"extent\":{\"xmin\":-126.0,\"ymin\":23.0,\"xmax\":-66.0,\"ymax\":54.0},\"layoutDefinition\":{\"extent\":{\"xmin\":-126.0,\"ymin\":21.608163265306125,\"xmax\":-62.0,\"ymax\":54.0},\"tileLayout\":{\"layoutCols\":4,\"layoutRows\":2,\"tileCols\":256,\"tileRows\":256}},\"bounds\":{\"minKey\":{\"col\":0,\"row\":0,\"instant\":1514764800000},\"maxKey\":{\"col\":3,\"row\":1,\"instant\":4133894400000}},\"cellType\":\"float32ud1.0000000150474662E30\",\"crs\":\"+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs \"}"
```
pomadchin: Let's fix it in this PR? 

There is a problem in [`Schema` `Encoder` and `Decoder`](https://github.com/locationtech/geotrellis/blob/master/store/src/main/scala/geotrellis/store/json/Implicits.scala#L101-L106):

```scala
// extra import required that I'm adding here as well
import _root_.io.circe.parser._

implicit val schemaEncoder: Encoder[Schema] =
    Encoder.encodeJson.contramap[Schema] { schema => parse(schema.toString).valueOr(throw _) }
  implicit val schemaDecoder: Decoder[Schema] =
    Decoder.decodeJson.emap { json =>
      Either.catchNonFatal((new Schema.Parser).parse(json.toString)).leftMap(_ => "Schema expected")
    }
```

CloudNiner: looool I had that as a separate issue because I wasn't totally sure how difficult it would be. You posted this just as I was finishing my fix:
```diff
diff --git a/store/src/main/scala/geotrellis/store/json/Implicits.scala b/store/src/main/scala/geotrellis/store/json/Implicits.scala
index de7e66880..63481b369 100644
--- a/store/src/main/scala/geotrellis/store/json/Implicits.scala
+++ b/store/src/main/scala/geotrellis/store/json/Implicits.scala
@@ -22,6 +22,7 @@ import geotrellis.vector._
 import geotrellis.vector.io.json.CrsFormats

 import _root_.io.circe._
+import _root_.io.circe.parser._
 import _root_.io.circe.syntax._
 import cats.syntax.either._
 import org.apache.avro.Schema
@@ -99,9 +100,14 @@ trait Implicits extends KeyIndexFormats with CrsFormats {
     }

   implicit val schemaEncoder: Encoder[Schema] =
-    Encoder.encodeString.contramap[Schema] { _.toString }
+    Encoder.encodeJson.contramap[Schema] { schema =>
+      parse(schema.toString) match {
+        case Left(e) => throw e
+        case Right(json) => json
+      }
+    }
   implicit val schemaDecoder: Decoder[Schema] =
-    Decoder.decodeString.emap { str =>
-      Either.catchNonFatal((new Schema.Parser).parse(str)).leftMap(_ => "Schema expected")
+    Decoder.decodeJson.emap { json =>
+      Either.catchNonFatal((new Schema.Parser).parse(json.noSpaces)).leftMap(_ => "Schema expected")
     }
 }
```
So I'll use yours and add a test for it.

CloudNiner:
Updated with the Schema fix and a test that fails if we try to decode Schema from anything other than Json.

In general, I'm of the opinion that separate issues with separate fixes should each have their own PR which was the original intent for these two issues. It's better for clarity when looking back in history.  In this case though, I just went with an extra commit and a PR that closes two issues since they're a bit related in how they manifest (failing to read AttributeStores) and you already provided the fix. I'll almost certainly strive to avoid this in the future though.

pomadchin:
@CloudNiner ah probably I read bad and didn't realize that you want to include it into a separate PR O: 
Anyway, mb it is still fine that it is a single PR ~ since it is a bug fix that fixes backwards attribute store backwards compatibility.

store/src/test/scala/geotrellis/store/json/SchemaJsonSpec.scala
```diff
@@ -0,0 +1,36 @@
...
+    it("should be able to decode Schema from generic Json, NOT String") {
+      val json: Json = parse(schemaString).valueOr(throw _)
+      val schema = json.as[Schema].valueOr(throw _)
```
pomadchin: A trick you could do is (it requires `cats.syntax.either._`):

```scala
parse(schemaString).flatMap(_.as[Schema]).valueOr(throw _)
```

PR #3036, opened 2019-07-23
VectorTile Geometry Recentering

### LINKED ISSUE #2926: Reading then writing vector tile features drops some
When reading geometries from MVTs, projected coordinates correspond to the top left of the tile "pixel":

https://github.com/locationtech/geotrellis/blob/0683298e59517fa5f25cfea984221da2eaa792bd/vectortile/src/main/scala/geotrellis/vectortile/internal/package.scala#L123-L128

This has the effect of shifting (and dropping) features that are read from MVTs and subsequently written to new ones. This is more noticeable with a small `tileWidth`, e.g. 128 (used when bucketing points).

Points on the top or left edge will be assigned and y or x value corresponding to the extent. Somewhere along the way, features that fail an `extent.contains` test will be dropped. Features that pass will have been shifted slightly up and to the left.

I _think_ that the right fix for this is to translate geometries read from MVTs by `(resolution / 2, resolution / -2)`, generating projected coordinates in the center of the tile "pixel".

jbouffard:
## Overview

This PR fixes an error that can sometimes cause geometries to be dropped and/or shifted when reading them from a `VectorTile` and then writing them back. The issue was caused by using the top left corner of the pixel of the tile as the reference for the projected coordinates; as this would cause some projected points to either move, or fall outside of the tile when converting those projected coordinates back to VectorTile coordinates.

The solution was then to change the reference point from being in the corner of the pixel to being in the center.

### Checklist

- [ ] ~~`docs/CHANGELOG.rst` updated, if necessary~~

### Notes

I think I will need to talk with someone more before I can remove the `[WIP]` from this PR. Because right now the changes introduced causes other tests to fail, and I'm not sure if that behavior is okay or not. Also, I've not been able to recreate the error this PR was intended to resolve, so a unit test needs to be created before this is ready for review.

Closes #2926

### DIFF
--- a/vectortile/src/main/scala/geotrellis/vectortile/internal/package.scala
+++ b/vectortile/src/main/scala/geotrellis/vectortile/internal/package.scala
@@ -120,12 +120,11 @@ package object internal {
     * }}}
     *
     */
-  private[vectortile] def toProjection(point: (Int, Int), topLeft: Point, resolution: Double): Point = {
+  private[vectortile] def toProjection(point: (Int, Int), topLeft: Point, resolution: Double): Point =
     Point(
       topLeft.x + (resolution * point._1),
       topLeft.y - (resolution * point._2)
     )
-  }
 
   /**
     * Translate [[Point]] coordinates within a CRS to those of a fixed
@@ -136,12 +135,11 @@ package object internal {
     * @param resolution How much of the CRS's units are covered by a single VT grid coordinate.
     * @return Grid coordinates in VectorTile space.
     */
-  private[vectortile] def fromProjection(point: Point, topLeft: Point, resolution: Double): (Int, Int) = {
+  private[vectortile] def fromProjection(point: Point, topLeft: Point, resolution: Double): (Int, Int) =
     (
       ((point.x - topLeft.x) / resolution).toInt,
       ((topLeft.y - point.y) / resolution).toInt
     )
-  }
 
   /** Instance definition of the ProtobufGeom typeclass for Points. */
   private[vectortile] implicit val protoPoint = new ProtobufGeom[Point, MultiPoint] {
--- a/vectortile/src/test/scala/geotrellis/vectortile/ProjectionSpec.scala
+++ b/vectortile/src/test/scala/geotrellis/vectortile/ProjectionSpec.scala
@@ -1,0 +1,43 @@
+package geotrellis.vectortile
+
+import geotrellis.proj4._
+import geotrellis.vector._
+import geotrellis.vectortile.internal._
+
+import org.scalatest._
+
+
+class ProjectionSpec extends FunSpec {
+  describe("VectorTile Projection Conversions") {
+    it("should read a point from a VectorTile and write it back") {
+      val p1 = Point(-61.347656249999986, 10.412183158667512)
+      val reprojected = p1.reproject(LatLng, WebMercator)
+      val tileExtent = Extent(-6887893.4928338025, 1095801.2374962866, -6809621.975869782, 1174072.7544603087)
+
+      val f = MVTFeature(reprojected, Map.empty[String, Value])
+
+      val layer = StrictLayer(
+        name = "test",
+        tileWidth = 128,
+        version = 2,
+        tileExtent = tileExtent,
+        points = Seq(f),
+        multiPoints = Seq.empty,
+        lines = Seq.empty,
+        multiLines = Seq.empty,
+        polygons = Seq.empty,
+        multiPolygons = Seq.empty
+      )
+
+      val vt = VectorTile(Map("test" -> layer), tileExtent)
+
+      val vt2 = VectorTile.fromBytes(vt.toBytes, tileExtent)
+      val p2 = vt2.layers("test").features.head.geom.reproject(WebMercator, LatLng)
+
+      val vt3 = VectorTile.fromBytes(vt2.toBytes, tileExtent)
+      val p3 = vt3.layers("test").features.head.geom.reproject(WebMercator, LatLng)
+
+      assert((p1 == p2) && (p1 == p3) && (p2 == p3))
+    }
+  }
+}


mojodna:
Yeah, this feels wrong to me...

https://github.com/locationtech/geotrellis/blob/e7ff0784b2a825f3af786860c714d9368e9baf47/vectortile/src/test/scala/geotrellis/vectortile/ProtobufGeomSpec.scala#L31-L45 makes sense. `moveTo(2, 2)` on a 4096 grid **should** be (2, 4094).

The alternative is that `topLeft` should be adjusted (by `resolution / 2`) when passed into `toProjection` (in `protoPoint` and friends?) in order to align it to a pixel center rather than a pixel corner. That seems like it would fix the test failures that cropped up.

Does GT do something similar elsewhere to align to centers?

jbouffard:
@mojodna To my knowledge, we don't do anything to recenter the geometries in GT, so I [implemented it](https://github.com/locationtech/geotrellis/blob/b610f05845cd5777c32b51a9c18104342e0c6094/vectortile/src/main/scala/geotrellis/vectortile/internal/package.scala#L124-L125) in `toProjection`. @jpolchlo do you know if we do actually do it elsewhere?

Though, it appears that the above implementation doesn't produce the exact result we're looking for. I [created a unit test](https://github.com/locationtech/geotrellis/blob/b610f05845cd5777c32b51a9c18104342e0c6094/vectortile/src/test/scala/geotrellis/vectortile/ProjectionSpec.scala#L12-L38) of the failing case, and it doesn't pass because the decimals aren't the same

```
[info] ProjectionSpec:
[info] VectorTile Projection Conversions
[info] - should read a point from a VectorTile and write it back *** FAILED ***
[info]   POINT (-61.34490966796874 10.409481792727018) did not equal POINT (-61.3422 10.4068) (ProjectionSpec.scala:38)
```

I'm not sure if that difference is okay, or if we need to implement something different.

vectortile/src/main/scala/geotrellis/vectortile/internal/package.scala
```diff
@@ -121,9 +121,12 @@ package object internal {
     *
     */
   private[vectortile] def toProjection(point: (Int, Int), topLeft: Point, resolution: Double): Point = {
+    val translatedX: Double = topLeft.x + (resolution / 2)
+    val translatedY: Double = topLeft.y + (resolution / -2)
```
mojodna: Changing the translation to a full point (from half a point here) further reduces the delta between expected and actual results. It's still ~4 meters off after that though.

mojodna: (This presumes that the input coordinate is correct, though (see below).)

vectortile/src/test/scala/geotrellis/vectortile/ProjectionSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+import geotrellis.vector._
+import geotrellis.vectortile.internal._
+
+import org.scalatest._
+
+
+class ProjectionSpec extends FunSpec {
+  describe("VectorTile Projection Conversions") {
+    it("should read a point from a VectorTile and write it back") {
+      val p = Point(-61.3422, 10.4068)
```
mojodna: This coordinate was extracted from a live MVT using [vt2geojson[(https://github.com/mapbox/vt2geojson), so it uses this projection math:

https://github.com/mapbox/vector-tile-js/blob/58df1e9344ee64f26deee84a9f54cee11fb95ef6/lib/vectortilefeature.js#L137-L145

As a result, I'm not sure what value is correct for translation.

jbouffard: I see. That's good to know. It was suggested that I do a third writing/read of the point to see if  it continues to shift. This is the code I used to run this test:

```scala
    it("should read a point from a VectorTile and write it back") {
      val p1 = Point(-61.3422, 10.4068)
      val reprojected = point.reproject(LatLng, WebMercator)
      val tileExtent =
        Extent(-6887893.4928338025, 1095801.2374962866, -6809621.975869782, 1174072.7544603087)
        //SpatialKey(168, 241).extent(ZoomedLayoutScheme(WebMercator).levelForZoom(9).layout)

      val f = MVTFeature(reprojected, Map.empty[String, Value])

      val layer = StrictLayer(
        name = "test",
        tileWidth = 128,
        version = 2,
        tileExtent = tileExtent,
        points = Seq(f),
        multiPoints = Seq.empty,
        lines = Seq.empty,
        multiLines = Seq.empty,
        polygons = Seq.empty,
        multiPolygons = Seq.empty
      )

      val vt = VectorTile(Map("test" -> layer), tileExtent)

      val vt2 = VectorTile.fromBytes(vt.toBytes, tileExtent)
      val p2 = vt2.layers("test").features.head.geom.reproject(WebMercator, LatLng)

      val vt3 = VectorTile.fromBytes(vt2.toBytes, tileExtent)
      val p3 = vt3.layers("test").features.head.geom.reproject(WebMercator, LatLng)

      println(s"\n\nThis was the initial point: $p1")
      println(s"This is the point after its been written: $p2")
      println(s"This is the point after its been written, read in, and written out again: $p3")
    }
```

And these are the results:

```
This was the initial point: POINT (-61.3422 10.4068)
This is the point after its been written: POINT (-61.33666992187498 10.401377554543565)
This is the point after its been written, read in, and written out again: POINT (-61.33117675781249 10.395974612177667)
```

So it appears that the geometry continues to be shifted to the up and left. This was using the transformation of a full point instead of a half point like you commented above. I wouldn't expect `p1` to be equal to `p2` or `p3`, but I'd expected `p2` and `p3` to be the same.

mojodna: With further exploration here, `Point(-61.347656249999986, 10.412183158667512)` is a good starting point (ha!) that round trips with half a point as the offset.

jbouffard:
So it appears that this issue has been fixed at somepoint these past few months, as the `ProjectionSpec` now passes without needing to change any of the code. Therefore, this PR just adds the spec to ensure that it continues to work in the future.

build.sbt
```diff
@@ -143,7 +143,7 @@ lazy val macros = project
   .settings(Settings.macros)
 
 lazy val vectortile = project
-  .dependsOn(vector)
+  .dependsOn(vector, layer)
```
pomadchin: btw should it still depend on the layer project?

jbouffard: Ah, I added that for the test, but you're right that we don't have to depend on it. I'll remove it now :+1:

pomadchin:
Thank you @jbouffard, merging once Travis is green.

mojodna:
@jbouffard do you have any idea which commit fixed things? i'm curious how this got fixed in places we weren't looking.

PR #3038, opened 2019-07-24
Define resampling operations in terms of target proportions

### LINKED ISSUE #2980: Break-up raster Reproject.Options
[`Reproject.Options`](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/reproject/Reproject.scala#L35-L41) couples several things that are required for reproject in the local case but are unhelpful for distributed case.

Working around this has been problematic in couple of ways:
- `TilleRDDReproject` has to selectively interpret and ignore some combinations of options
- `RasterSource.reproject` has to use [`ResampleGrid`](https://github.com/geotrellis/geotrellis-contrib/blob/master/vlm/src/main/scala/geotrellis/contrib/vlm/ResampleGrid.scala) to decouple of the definition.

Possibly the solution is to promote `ResampleGrid` to GeoTrellis core and use it throughout. In either are we should consider the local API and `TileRDDReproject` to make sure they're served well by the replacement.

Connects: https://github.com/locationtech/geotrellis/issues/2430

moradology:
## Overview

The `ResampleGrid` abstraction from gt-contrib is a convenient, declarative way to specify different strategies for resampling a given bit of imagery. This PR begins the process of rewriting resampling operations (and reprojection operations - which require resampling) in terms of this `ResampleGrid` AST and not in terms of `Reproject.Options`. 

### Checklist

### Notes

So far, this PR converts only reprojection and resampling in the `raster` package. Challenges in applying these changes consistently throughout `spark` remain. Clear documentation of this new feature and a solid set of tests are a logical next step to ensure that usage remains consistent across modules and `resampling` methods

Closes #2980


raster/src/main/scala/geotrellis/raster/ResampleTarget.scala
```diff
@@ -0,0 +1,165 @@
...
+import geotrellis.raster.reproject.Reproject
+
+import _root_.io.circe._
+import _root_.io.circe.syntax._
+
+import spire.math.Integral
+import spire.implicits._
+
+/** Represents a strategy/target for resampling */
+sealed trait ResampleTarget[N] {
```
echeipesh: My main comment was that it sucks that using this forces all the `.resample` methods to become `.resample[N]` methods. I've realized that is "easily" fixed by moving the `N` type parameter from the trait definition to `apply` on the trait. So you never have to specify the span of target during definition but may encounter run-time errors if you try to resample a `Tile` to something outside of `Int` range (which is just as well basically).

Will report more as I play with this idea.

moradology:
Superseded by #3097

PR #3067, opened 2019-09-05
Backport np helper functions to geotrellis.util.np

CloudNiner:
- linspace: https://docs.scipy.org/doc/numpy/reference/generated/numpy.linspace.html
- percentile: https://docs.scipy.org/doc/numpy/reference/generated/numpy.percentile.html

Also adds implicit helper methods for Tile.percentile

Closes azavea/geotrellis#122


util/src/main/scala/geotrellis/util/np/package.scala
```diff
@@ -0,0 +1,87 @@
...
+    * https://en.wikipedia.org/wiki/Percentile
+    *
+    * @param data
+    * @param pctBreaks
+    * @return
+    */
+  def percentile(data: Array[Double],
+                 pctBreaks: Array[Double]): Array[Double] = {
+    // TODO: Include the isData filter check in the Tile implicit method
+    // val arr = data.filter(isData(_))
```
pomadchin: Should these commented out lines be removed? Mb it makes sense to add into the comments that if it is a tile - user probably may want to filter noDataValues?

CloudNiner: Oooops yeah I'll delete. Presumably for the generic version of this method, the user would just filter the Array[Double] before passing to this method. 

I kept the nodata filter in the Tile version since that's what we had. Should we remove the filter there as well? I'm waffling on it.

pomadchin: nodata filter should be present in a Tile version, that's all good :+1: 
The comment was only about adding comments for this function and removing commented out code : D

CloudNiner: Ah sure thing. Updated with comments removed.

CloudNiner:
👍

Issue #3070, opened 2019-09-12
geotrellis.raster.CellTypeEncoding#name needlessly allocates memory


metasim:
In a large compute job it became evident that something was awry with memory allocations. During the RasterFrames process of decoding `Tile`s from catalyst memory, the `CellType` representation is converted from a `String` into a `case object`. That process involves comparing names, etc. The `CellType` `name` method actually constructs the name from it's `case object` class name.... but doesn't cache the value.

<img width="1175" alt="Screen Shot 2019-09-12 at 12 24 41 PM" src="https://user-images.githubusercontent.com/131013/64802430-a3980c80-d558-11e9-9ae8-8631c06c73b1.png">

I take the blame for this, being too clever by half in taking the DRY principle too far. Solutions include turning the method into a `val`, or just writing out the string explicitly in the instantiation of each `case object`. The later would be much more straight forward.

metasim:
## Before

_Note_: performance isn't the same thing as memory pressure, but this is an easier.

<img width="1158" alt="Screen Shot 2019-09-13 at 11 04 56 AM" src="https://user-images.githubusercontent.com/131013/64874821-d1915580-d619-11e9-9189-a23237ddcf3b.png">

[jmh-results-20190913113655.json.zip](https://github.com/locationtech/geotrellis/files/3610740/jmh-results-20190913113655.json.zip)

metasim:
## After

<img width="1147" alt="Screen Shot 2019-09-13 at 12 10 35 PM" src="https://user-images.githubusercontent.com/131013/64877471-867a4100-d61f-11e9-87a4-eb1d40d0ee5b.png">

<img width="1146" alt="Screen Shot 2019-09-13 at 12 11 12 PM" src="https://user-images.githubusercontent.com/131013/64877531-9d209800-d61f-11e9-9e2b-ef8a1e6a7fd5.png">

metasim:
It would be helpful to have this backported to GT 2.3.x.

PR #3072, opened 2019-09-13
Reworks `String` <-> `CellType` conversions to be more efficient.

metasim:
See #3070 for deets on why.


metasim:
It would be helpful to have this backported to GT 2.3.x.

raster/src/main/scala/geotrellis/raster/CellTypeEncoding.scala
```diff
@@ -24,40 +26,82 @@ import scala.util.Try
...
   }
   object float64ud extends UserDefinedNoDataEncoding {
+    override val name = "float64ud"
     override val isFloatingPoint = true
   }
 }
 
-/** Root trait for CellType encoding definitions. */
-sealed trait CellTypeEncoding {
-  def name = getClass.getName.split('$').last
```
pomadchin: marking this as `lazy val` didn't help?

metasim: I didn't bother.... for some reason I went the brute force route.... maybe because that creates a `synchronized` block.

Feel strongly it should be that way instead?

pomadchin: Nope, just thinking in a way of minimizing changes and hand code writing. I'll check my idea in a spare time. Could you please add into the issue the way we can observe the memory pressure? Probably lots of people would be interested in it, and thanks for a quick PR 💯 !

bench/archive/jmh-results-20190913120118.json
```diff
@@ -0,0 +1,328 @@
+[
```
pomadchin: Should these be a part of the bench repo? How to generate them and visualize?

metasim: IDK what the bench repo is.

raster/src/main/scala/geotrellis/raster/CellTypeEncoding.scala
metasim: From https://github.com/locationtech/geotrellis/issues/3070:

> In a large compute job it became evident that something was awry with memory allocations. During the RasterFrames process of decoding `Tile`s from catalyst memory, the `CellType` representation is converted from a `String` into a `case object`. That process involves comparing names, etc. The `CellType` `name` method actually constructs the name from it's `case object` class name.... but doesn't cache the value.

Screen Shot 2019-09-12 at 12 24 41 PM

PR #3074, opened 2019-09-16
Clean up WKT <--> EPSG conversion methods

### LINKED ISSUE #2870: `WKT.fromEpsgCode` makes unsafe call to `Option.get`
Exceptions are thrown when provided WKT doesn't have an EPSG code in it (something that's not a requirement). These methods should return `Option[String]` instead.

https://github.com/locationtech/geotrellis/blob/d83fd66db9ed1276b6b82ff4adce8e362ac56cef/proj4/src/main/scala/geotrellis/proj4/io/wkt/WKT.scala#L65

https://github.com/locationtech/geotrellis/blob/d83fd66db9ed1276b6b82ff4adce8e362ac56cef/proj4/src/main/scala/geotrellis/proj4/io/wkt/WKT.scala#L72

### LINKED ISSUE #2871: `CRS.fromWKT` doesn't handle WKT with no EPSG record.
## Input
This WKT is from calling `org.opengis.referencing.crs.CoordinateReferenceSystem#toWKT` on data read from a random shapefile via GeoTools.

```
CRS.fromWKT("""|GEOGCS["GCS_WGS_1984",
               |DATUM["D_WGS_1984",
               |SPHEROID["WGS_1984", 6378137.0, 298.257223563]],
               |PRIMEM["Greenwich", 0.0],
               |UNIT["degree", 0.017453292519943295],
               |AXIS["Longitude", EAST],
               |AXIS["Latitude", NORTH]]
               |""".stripMargin.replaceAll("\n",""))

```

## Output

```
None.get
java.util.NoSuchElementException: None.get
	at scala.None$.get(Option.scala:347)
	at scala.None$.get(Option.scala:345)
	at geotrellis.proj4.io.wkt.WKT$.getEpsgCode(WKT.scala:70)
	at geotrellis.proj4.CRS$.fromWKT(CRS.scala:87)
```

Related: #2870

CloudNiner:
## Overview

This PR cleans up the conversion methods on the `WKT` object to:
- All operations now return `Option`s. Before, we would silently unwrap results within the method calls, leading to thrown errors. It's now up to the calling code to unwrap and determine what to do in case of conversion failure, as it always should have been.
  - Methods that convert from EPSG code to WKT string are consistently prefixed with `fromEpsg`
  - Methods that convert from WKT string to EPSG code are consistently prefixed with `getEpsg`

It also updates `CRS.fromWKT` to return an `Option[CRS]` instead of just `CRS`. Same issue as above, the method would silently unwrap an optional result internally, which could lead to errors.

### Checklist

- [x] Unit tests added for bug-fix or new features

### Notes

I was a bit surprised that we don't utilize any of these methods in other GeoTrellis modules. But, according to the IntelliJ "Find Usages" feature, we don't. Guess we'll see what CI says, but I was able to get the project to compile and all of the proj4 tests to pass.

Closes #2870, Closes #2871

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -109,6 +109,11 @@ API Changes & Project structure changes
 
   - **Add:** ``geotrellis.util.np`` package which contains ``linspace`` and ``percentile``methods that match numpy functionality. An implicit class was also added to ``geotrellis.raster`` which provides the ``percentile`` method for ``geotrellis.raster.Tile``.
 
+- ``geotrellis.proj4``
+
+  - **Change:** ``WKT`` object conversion methods to and from EPSG codes have changed. All now return `Option`s rather than silently unwrap. Methods that convert from EPSG code to WKT string are now prefixed consistently with `fromEpsg` and methods that convert from WKT string to EPSG code are now prefixed consistently with `toEpsg`.
+  - **Change:** ``CRS.fromWKT`` now returns `Option[CRS]` instead of `CRS` after silently unwrapping an optional internally.
+
 Fixes & Updates
 ^^^^^^^^^^^^^^^
 
--- a/proj4/src/main/scala/geotrellis/proj4/CRS.scala
+++ b/proj4/src/main/scala/geotrellis/proj4/CRS.scala
@@ -66,10 +66,8 @@ object CRS {
     * Creates a CoordinateReferenceSystem (CRS) from a
     * well-known-text String.
     */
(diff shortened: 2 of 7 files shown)


esmeetu:
A bit complement, new line in Windows is "\r\n".  
If considering windows os compatibility, we should change code
```val cleanWkt = wktString.replaceAll("\n", "")```
to 
```val cleanWkt = wktString.replaceAll("[\r\n]", "")```

Issue #3077, opened 2019-09-23
"Task not serializable" when writing to S3 layer


echeipesh:
This is issue against master. When trying to write a layer to S3 with:

```scala
val pyramid = Pyramid.fromLayerRDD(layer, resampleMethod = Sum)
val store = AttributeStore(catalog)
val writer = LayerWriter(store, catalog)
pyramid.write(layerName, writer, ZCurveKeyIndexMethod)
```

The following exception will be thrown:

```
geotrellis.store.package$LayerWriteError: Failed to write Layer(name = "MWI", zoom = 6)
	at geotrellis.spark.store.s3.S3LayerWriter._write(S3LayerWriter.scala:125)
	at geotrellis.spark.store.s3.S3LayerWriter._write(S3LayerWriter.scala:49)
	at geotrellis.spark.store.LayerWriter$class.write(LayerWriter.scala:150)
	at geotrellis.spark.store.s3.S3LayerWriter.write(S3LayerWriter.scala:49)
	at geotrellis.spark.pyramid.Pyramid$$anonfun$write$1.apply$mcVI$sp(Pyramid.scala:68)
	at scala.collection.immutable.Range.foreach$mVc$sp(Range.scala:160)
	at geotrellis.spark.pyramid.Pyramid.write(Pyramid.scala:67)
	at geotrellis.sdg.OutputPyramid$.saveLayer(OutputPyramid.scala:53)
	at geotrellis.sdg.PopulationNearRoadsJob$$anonfun$9$$anonfun$apply$11.apply(PopulationNearRoadsJob.scala:171)
	at geotrellis.sdg.PopulationNearRoadsJob$$anonfun$9$$anonfun$apply$11.apply(PopulationNearRoadsJob.scala:148)
	at scala.concurrent.impl.Future$PromiseCompletingRunnable.liftedTree1$1(Future.scala:24)
	at scala.concurrent.impl.Future$PromiseCompletingRunnable.run(Future.scala:24)
	at scala.concurrent.impl.ExecutionContextImpl$AdaptedForkJoinTask.exec(ExecutionContextImpl.scala:121)
	at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
	at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
	at scala.concurrent.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
[...]

PR #3080, opened 2019-09-23
Make LayerProviders Serializable

pomadchin:
## Overview

Closes https://github.com/locationtech/geotrellis/issues/3077


pomadchin:
not-a-fix, but a fragile workaround.

Issue #3023, opened 2019-07-15
reprojectExtentAsPolygon() - method going into infinite loop 


rajindersaini:
When a [reproject](https://github.com/locationtech/geotrellis/blob/master/vector/src/main/scala/geotrellis/vector/reproject/Reproject.scala) method is called with specific extent it is going into infinite recursion. 
Expectation: This should return an appropriate exception.  

Example code:
```scala
val reader = GeoTiff.readMultiband("/Users/rajinder.saini/dev/geotiff-utils/test-resources/outside-utm.tif")

val extant = reader.projectedExtent.reproject(LatLng)
//The calculated extent is given below. 
val ex = Extent.fromString("0.0, 0.0, 2606.0, 2208.0")
    
// This call goes into infinite recursion. 
val tile =  reader.tile.reproject(ex, reader.crs, WebMercator)    
```

pomadchin:
hey @rajindersaini! can you upload data somewhere and to update your post with a completely compiling example? 

Also, can you provide an exception message you're seeing and what GeoTrellis version do you use?

jpolchlo:
Also quite important to know the value of `reader.crs`

rajindersaini:
Please find attached the sample here - https://climate.box.com/s/srp57j3c1e9jfyna40c0e0pge5ycmidv

rossbernet:
Does not appear that this can be accessed:

![image](https://user-images.githubusercontent.com/8103418/61329438-4b0ff180-a7eb-11e9-839b-0ecee7e2f69c.png)

rossbernet:
It looks like we may be able to access this by signing up for a free business trial, but do you have an alternative, more accessible way of sharing this?

![image](https://user-images.githubusercontent.com/8103418/61329554-9aeeb880-a7eb-11e9-9beb-4fabe6d85c52.png)

rajindersaini:
Sorry about that - I updated the sharing setting - https://climate.box.com/s/srp57j3c1e9jfyna40c0e0pge5ycmidv

jmt-vanderlee:
I might be having the same issue but only when using `HadoopLayerWriter` to write to GCS.
Writing locally with `FileLayerWriter` works fine.

Calling reproject causes the spark job to hang indefinitely. 
```
val topLeftPntLatLong = Point(rasterRawMetaData.extent.xmin, rasterRawMetaData.extent.ymax)
val currentUtmInfo = new UtmInfo {
[...]

PR #3083, opened 2019-09-24
reprojectExtentAsPolygon should be more deterministic

pomadchin:
## Overview

This PR resolves the division by zero unhandled case in the `reprojectExtentAsPolygon` function,

### Checklist

Closes https://github.com/locationtech/geotrellis/issues/3023

### DIFF
--- a/vector/src/main/scala/geotrellis/vector/reproject/Reproject.scala
+++ b/vector/src/main/scala/geotrellis/vector/reproject/Reproject.scala
@@ -100,7 +100,9 @@ object Reproject {
       val length = sqrt(pow(x0 - x1, 2) + pow(y0 - y1, 2))
 
       val p2 = m -> (x2, y2)
-      if (deflect / length < relError) {
+      if (java.lang.Double.isNaN(deflect)) {
+        throw new IllegalArgumentException(s"Encountered NaN during a refinement step: ($deflect / $length). Input $extent is likely not in source projection.")
+      } else if (deflect / length < relError) {
         List(p2)
       } else {
         refine(p0, p2) ++ (p2 :: refine(p2, p1))
--- a/vector/src/test/scala/spec/geotrellis/vector/reproject/ReprojectSpec.scala
+++ b/vector/src/test/scala/spec/geotrellis/vector/reproject/ReprojectSpec.scala
@@ -14,9 +14,10 @@
  * limitations under the License.
  */
 
-package geotrellis.vector
+package geotrellis.vector.reproject
 
 import geotrellis.proj4._
+import geotrellis.vector._
 import geotrellis.vector.testkit._
 
 import org.scalatest._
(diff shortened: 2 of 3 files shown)


vector/src/main/scala/geotrellis/vector/reproject/Reproject.scala
```diff
@@ -100,7 +100,9 @@ object Reproject {
       val length = sqrt(pow(x0 - x1, 2) + pow(y0 - y1, 2))
 
       val p2 = m -> (x2, y2)
-      if (deflect / length < relError) {
+      if (deflect.isNaN) {
```
echeipesh: I recall we did a bit of benchmarking and found that `double.isNaN` was surprisingly and significantly slower than `java.lang.Double.isNaN` which is why we chose to use that in `isNoData` macro: https://github.com/locationtech/geotrellis/blob/c5094d0437e3d36436fc2d40975a1e99ec775401/macros/src/main/scala/geotrellis/macros/NoDataMacros.scala#L44

This is looks like tight enough loop that we should use that.

vector/src/main/scala/geotrellis/vector/reproject/Reproject.scala
```diff
@@ -100,7 +100,9 @@ object Reproject {
...
+        throw new IllegalArgumentException(s"Check the input $extent CRS, it is probably not in the correct projection.")
```
echeipesh: Exception should indicate a little bit of mechanical reason why it was thrown because its super surprising every time it happens. Something like s"Encountered NaN during refinement step: ($deflect / $length). Input $extent is likely not in source projection." It would be great if we could pull out the actual CRS from the [`Transform`](https://github.com/locationtech/geotrellis/blob/master/proj4/src/main/scala/geotrellis/proj4/Transform.scala) instance here but still good enough to get a hint.

vector/src/main/scala/geotrellis/vector/reproject/Reproject.scala
pomadchin: Totally forgot about it, I even benchmarked it, you're correct.

vector/src/main/scala/geotrellis/vector/reproject/Reproject.scala
pomadchin: Yea, it is impossible to get input and output CRS from the `Transform` here, since it in this case it is a Function:

```scala
type Transform = (Double, Double) => (Double, Double)
```

vector/src/main/scala/geotrellis/vector/reproject/Reproject.scala
pomadchin: Totally forgot about it, I even benchmarked it, you're correct.

vector/src/test/scala/spec/geotrellis/vector/reproject/ReprojectSpec.scala
```diff
@@ -14,9 +14,10 @@
  * limitations under the License.
  */
 
-package geotrellis.vector
+package geotrellis.vector.reproject
```
pomadchin: I think we need to review the vector package specs structure in a separate issue @echeipesh. 
Test paths here don't correspond to package names.

Issue #3018, opened 2019-07-10
Deprecate the use of `?` in our codebase


pomadchin:
The kind-projector syntax `?` is deprecated, see the following thread:

https://github.com/typelevel/kind-projector/issues/108

It is not a change that needs to be done in terms of the 3.0 release and definitely would not be a breaking change. IDEs still don't support it (for instance intelij IDEA https://youtrack.jetbrains.net/issue/SCL-15756)

PR #3084, opened 2019-09-24
Use * instead of ?

pomadchin:
## Overview

Closes https://github.com/locationtech/geotrellis/issues/3018


CloudNiner:
Might be a dumb question, but...why? Perhaps worth a note in the commit description.

pomadchin:
@CloudNiner it is a deprecated syntax, ? Is used in Dotty => due to compatibility reasons they switched to *. Also it is a more natural ‘kinds’ symbol, for the reference you can look into this discussion: https://github.com/typelevel/kind-projector/issues/108

pomadchin:
It is more a code hygiene and https://github.com/locationtech/geotrellis/issues/3018 contains the description :0

PR #3095, opened 2019-09-30
Fix windowed reproject bug

moradology:
## Overview

Windowed raster reprojection formerly constructed its window via `extentFor`, which by default has clamp set to true. Quoting the docstring, "If true, the returned extent will be contained by this RasterExtent's extent". This is fine, in most cases, but it means we can't assume that the output `RasterExtent` it produces can meaningfully be used along with `GridBounds.height`/`width` to reconstruct a RasterExtent (as the output `RasterExtent` might not have the same number of columns/rows as the `GridBounds` from which it is derived).

We can avoid this problem by simply using `CellSize` rather than cols/rows.

### Checklist

### Notes

This change required slightly widened tolerances for comparing `CellSize` values (0.05 -> 0.08)

### DIFF
--- a/raster/src/test/scala/geotrellis/raster/reproject/ReprojectSpec.scala
+++ b/raster/src/test/scala/geotrellis/raster/reproject/ReprojectSpec.scala
@@ -141,7 +141,6 @@ class ReprojectSpec extends FunSpec
       val rightRasterExtent = RasterExtent(Extent(579120.000, 4428900.000, 594480.000, 4444260.000), 30.0, 30.0, 512, 512)
 
       val leftTile = IntArrayTile(Array.ofDim[Int](256 * 256).fill(1), 256, 256)
-
       val rightTile = IntArrayTile(Array.ofDim[Int](256 * 256).fill(2), 256, 256)
 
       // Sanity check - they don't have any missing pixels before reprojecting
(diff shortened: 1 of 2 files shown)


raster/src/test/scala/geotrellis/raster/reproject/ReprojectSpec.scala
```diff
@@ -164,11 +164,11 @@ class ReprojectSpec extends FunSpec
...
         mergedRaster.reproject(GridBounds(512, 0, 1023, 1023), srcCRS, WebMercator, Options(method = Bilinear))
 
-      val RasterExtent(_, cellwidthLeft, cellheightLeft, _, _) = RasterExtent(wmLeftExtent, wmLeftTile.cols, wmLeftTile.rows)
-      val RasterExtent(_, cellwidthRight, cellheightRight, _, _) = RasterExtent(wmRightExtent, wmRightTile.cols, wmRightTile.rows)
+      val RasterExtent(_, cellwidthLeft, cellheightLeft, _, _) = wmLeft.rasterExtent
+      val RasterExtent(_, cellwidthRight, cellheightRight, _, _) = wmRight.rasterExtent
 
-      cellwidthLeft should be (cellwidthRight +- 0.05)
-      cellheightLeft should be (cellheightRight +- 0.05)
+      cellwidthLeft should be (cellwidthRight +- 0.08)
```
pomadchin: why did the cellSize approximation became worse? 
(i'm not against it; only curious about the math problems and floating point issues)

CloudNiner: (Also noting that this is a great thing to describe in the block of a commit message, so that it lives on in the history)

moradology: This is a tricky one to answer because the approximation didn't become worse (the prior answer was just wrong) though two runs happen to be further apart. Here's what I've got:

The `GridBounds` height of 1024 in this test is *not* the height of the window being produced via the clamped crop from said `GridBounds` so when you attempt to construct a `RasterExtent` from the cropped extent and the `GridBounds` width/height, you end up getting a cellheight of 1/2 what you should:
https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/RasterExtent.scala#L146-L150
Later, these numbers are used to determine the number of cols/rows to be expected from our raster after reprojection:
https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/reproject/ReprojectRasterExtent.scala#L59-L61

To establish an independent expectation of what the correct cellsize *should* be, I wrote geotifs with of the pre-projected imagery and used `gdalwarp` to reproject to WebMercator. The cellsize of the resulting tifs were 39.24ish (each tif's cellsize varied slightly).

The former strategy (which halves the CellSize height in error) produces a reprojected cellsize of 25.09. The updated strategy, which constructs the `RasterExtent` to reproject from the CellSize (which I'm claiming we know to be correct at the call site), yields a result around 39.68.

39.68 is closer to 39.24 than 25.09 is. Does this answer satisfy?

pomadchin: @moradology how does the `0.05` was replaced by `0.08` here? Was the old test bad? Or this test should be re-done?

moradology: I believe the old test was bad, yes. I don't have a better answer for why 0.08 is the new value rather than 0.05 other than that 0.08 works for the updated (and I believe more correct as indicated above) logic.

pomadchin: Well, it looks like I still have problems with understanding why `0.08` is a better precision here than `0.05`, but can you leave a summary in comments in the test about why this precision issue happens and why this is okay here as @CloudNiner suggested [here](https://github.com/locationtech/geotrellis/pull/3095#discussion_r329759058)? 

We just have so many heuristics in the code so it makes hard to interpret tests behavior changes ):

pomadchin: And probably it would be good to go. ¯\_(ツ)_/¯

echeipesh: So to confirm, the previous test was bad because it was checking for result based on incorrect implementation.
I don't know why the new tolerance is `0.08` just like I don't know why the old one was `0.05`. After diving and tracing this with @moradology I'm pretty confident that our CellSize heuristic is being applied correctly here and just arrives at different answer and he has done further verification looks like.

Why GDAL CellSize estimates are much closer together is a different question, probably worth noting and investigating but not a blocker for this issue.

Since we're just recording a manually verified result we might as well update the test to test against constant with-in tight tolerance with a comment stating so. maybe that's more clear @CloudNiner @pomadchin ?

pomadchin: @echeipesh :+1:

pomadchin: I hope that we won't loose this conversation O:

CloudNiner: That's definitely clearer. Could we summarize this discussion in the commit message, perhaps along with a link to this thread? That would make it easier to find and regain context when we look at this again.

I realize this isn't the place to bring this up, but a short summary as to why I'm pushing for this and a more detailed discussion can happen offline -- I'm pulling this idea from the "Accepting Explanations" section of https://google.github.io/eng-practices/review/reviewer/comments.html. I very much agree with this sentiment. It's bitten me now more than a few times, the most recent being when @pomadchin and I spent a significant chunk of time digging into BitGeoTiff handling which would have been a much improved experience with even a few explanatory comments in the history.

PR #3096, opened 2019-09-30
Fix S3LayerWriter NullPointerException

### LINKED ISSUE #2857: S3LayerWriter Cannot Write to URIs that Lack a Key Prefix
When saving a layer to S3, it's possible to create a `S3LayerWriter` with `LayerWriter` via a service provider. This can be an issue, though, as the `URI` passed to `LayerWriter` may not have a key prefix which is needed to write things to S3. Thus, a `URI` without a key prefix will have that value be `null`. This, in turn, results in a `NullPointerException`.

```
scala> val writer = LayerWriter("s3://un-sdg")
writer: geotrellis.spark.io.LayerWriter[geotrellis.spark.LayerId] = geotrellis.spark.io.s3.S3LayerWriter@72810bd4

scala> writer.write[SpatialKey, Tile, TileLayerMetadata[SpatialKey]](LayerId("world-pop-epsg4326", 0), tiled, ZCurveKeyIndexMethod)
java.lang.NullPointerException
  at scala.collection.immutable.StringOps$.length$extension(StringOps.scala:47)
  at scala.collection.immutable.StringOps.length(StringOps.scala:47)
  at scala.collection.IndexedSeqOptimized$class.isEmpty(IndexedSeqOptimized.scala:27)
  at scala.collection.immutable.StringOps.isEmpty(StringOps.scala:29)
  at scala.collection.TraversableOnce$class.nonEmpty(TraversableOnce.scala:111)
  at scala.collection.immutable.StringOps.nonEmpty(StringOps.scala:29)
  at geotrellis.spark.io.s3.package$$anonfun$makePath$1.applyOrElse(package.scala:25)
  at geotrellis.spark.io.s3.package$$anonfun$makePath$1.applyOrElse(package.scala:25)
  at scala.PartialFunction$$anonfun$runWith$1.apply(PartialFunction.scala:141)
  at scala.PartialFunction$$anonfun$runWith$1.apply(PartialFunction.scala:140)
  at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
  at scala.collection.mutable.WrappedArray.foreach(WrappedArray.scala:35)
  at scala.collection.TraversableLike$class.collect(TraversableLike.scala:271)
  at scala.collection.AbstractTraversable.collect(Traversable.scala:104)
  at geotrellis.spark.io.s3.package$.makePath(package.scala:25)
  at geotrellis.spark.io.s3.S3LayerWriter._write(S3LayerWriter.scala:111)
  at geotrellis.spark.io.s3.S3LayerWriter._write(S3LayerWriter.scala:49)
  at geotrellis.spark.io.LayerWriter$class.write(LayerWriter.scala:149)
  at geotrellis.spark.io.s3.S3LayerWriter.write(S3LayerWriter.scala:49)
  ... 42 elided
```

CloudNiner:
## Overview

AmazonS3URI will have a null `key` if the user passes a uri that doesn't contain one, such as `s3://foo`. This gets passed down into the various S3 reader/writer classes in GeoTrellis, and is converted to an empty string in some cases but not others.

In some cases, an S3LayerWriter will be constructed using the attached S3AttributeStore bucket + keyPrefix, if we know we're getting an S3AttributeStore. Otherwise the S3LayerWriter will be constructed using passed in bucket/keyPath values which can come from anywhere, including the nullable `AmazonS3URI.key`.

The quickest, least disruptive, and most comprehensive method seemed to be to fixup the `AmazonS3URI.getKey` method to return `""` instead of `null` and also add an additional error check in the `S3AttributeStore` apply method that converts incoming nulls in prefix to empty string. More comprehensive error handling would take two forms. The "easier" way would be to perform the check in every apply in the downstream S3Layer[Reader|Writer|...] classes, of which there are quite a few. The "harder" way would be to abstract the bucket and prefix properties to a common trait, and introduce the null check and conversion to empty string there. I went with this intermediate approach for now as its lighter weight on the eve of the GT3.0 release cutoff.

### Checklist

### Notes

I added a new script that starts the docker container necessary for running tests that target S3 via the s3 client. 

Closes #2857

### DIFF
--- a/s3/src/main/java/geotrellis/store/s3/AmazonS3URI.java
+++ b/s3/src/main/java/geotrellis/store/s3/AmazonS3URI.java
@@ -224,7 +224,11 @@ public class AmazonS3URI {
      * @return the key parsed from the URI (or null if no key specified)
      */
     public String getKey() {
-        return key;
+        if (key != null) {
+            return key;
+        } else {
+            return "";
+        }
     }
 
     /**
--- a/s3/src/main/scala/geotrellis/store/s3/S3AttributeStore.scala
+++ b/s3/src/main/scala/geotrellis/store/s3/S3AttributeStore.scala
@@ -217,7 +217,7 @@ object S3AttributeStore {
   final val SEP = "__"
 
   def apply(bucket: String, root: String, s3Client: => S3Client = S3ClientProducer.get()) =
-    new S3AttributeStore(bucket, root, s3Client)
+    new S3AttributeStore(bucket, Option(root).getOrElse(""), s3Client)
 
   def apply(bucket: String, s3Client: => S3Client): S3AttributeStore =
     apply(bucket, "", s3Client)
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -131,9 +131,11 @@ Fixes & Updates
 - Bump proj4 version to fix multiple performance issues (`#3039 <https://github.com/locationtech/geotrellis/pull/3039>`_).
 - Update dependencies (`#3053 <https://github.com/locationtech/geotrellis/pull/3053>`_).
 - Fix HttpRangeReader swallows 404 error (`#3073 https://github.com/locationtech/geotrellis/pull/3073`_)
-- Add a ToSpatial function for the collections API (`#3082 https://github.com/locationtech/geotrellis/pull/3082`_)
-- Fix TIFFTagsReader to skip unsupported tags (`#3088 https://github.com/locationtech/geotrellis/pull/3088`_)
-- reprojectExtentAsPolygon should be more deterministic (`#3083 https://github.com/locationtech/geotrellis/pull/3083`_)
+- Add a ToSpatial function for the collections API (`#3082 https://github.com/locationtech/geotrellis/pull/3082`_).
+- Fix TIFFTagsReader to skip unsupported tags (`#3088 https://github.com/locationtech/geotrellis/pull/3088`_).
+- reprojectExtentAsPolygon should be more deterministic (`#3083 https://github.com/locationtech/geotrellis/pull/3083`_).
+- AmazonS3URI.getKey() now returns an empty string instead of null if no key was provided (`#3096 https://github.com/locationtech/geotrellis/pull/3096`_).
+- Improved handling of null prefix for S3AttributeStore. Prefer use of `S3AttributeStore.apply` to take advantage of this improved handling (`#3096 https://github.com/locationtech/geotrellis/pull/3096`_).
 
 2.3.0
 -----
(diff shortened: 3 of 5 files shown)


echeipesh:
Could always throw a test in here where you write/read a layer to catalog that has no key:
https://github.com/locationtech/geotrellis/blob/master/s3-spark/src/test/scala/geotrellis/spark/store/s3/S3SpatialSpec.scala

Its a little weird but its a normal spec that gets its tests from https://github.com/locationtech/geotrellis/blob/master/spark-testkit/src/main/scala/geotrellis/spark/testkit/io/PersistenceSpec.scala -- so you can add more `describe` and `it` blocks. The test layers and readers are available to you right there.

CloudNiner:
👍 I'll give that a shot, thanks!

s3/src/main/scala/geotrellis/store/s3/S3AttributeStore.scala
```diff
@@ -217,7 +217,7 @@ object S3AttributeStore {
   final val SEP = "__"
 
   def apply(bucket: String, root: String, s3Client: => S3Client = S3ClientProducer.get()) =
-    new S3AttributeStore(bucket, root, s3Client)
+    new S3AttributeStore(bucket, Option(root).getOrElse(""), s3Client)
```
echeipesh: I think this is fine either way, no change request but question: Is this left over from initial approach or was this a special case that needed to be handled and if so why?

echeipesh:
Should be squashed for cleaner history, atomic change otherwise.

s3/src/main/scala/geotrellis/store/s3/S3AttributeStore.scala
CloudNiner: The most likely way to get nulls in the first place is addressed by the change to AmazonS3URI, but this was just another spot where a null could exist if a user is passing strings generated by some other Java API. This is a middle ground band-aid versus a more comprehensive update to ensure that all the downstream Layer classes use cleaned bucket/prefix values on the S3AttributeStore, which sometimes doesn't happen now -- each [downstream Layer class holds copies of the bucket/prefix values](https://github.com/locationtech/geotrellis/blob/master/s3-spark/src/main/scala/geotrellis/spark/store/s3/S3LayerWriter.scala#L51-L52) that should theoretically already exist on the S3AttributeStore.

s3/src/main/java/geotrellis/store/s3/AmazonS3URI.java
```diff
@@ -224,7 +224,11 @@ public String getBucket() {
      * @return the key parsed from the URI (or null if no key specified)
      */
     public String getKey() {
-        return key;
+        if (key != null) {
```
pomadchin: @CloudNiner it was also possible to do `Option(null).getOrElse("")`

PR #3106, opened 2019-10-02
updated computeResolution in IterativeCostDistance to match Iterative Viewshed

### LINKED ISSUE #3103: Update computeResolution in cost distance
The current compute resolution in iterativeCostDistance calculates the resolution of the cell sizes as if it were at the equator. The further away from the equator the larger it estimates the cell size. Around 40 degrees latitude it will compute a 10 meter cell as being over 13 meters. However in another function there is a different computeResolution function that more correctly calculates the cell size by scaling the value based on the average latitude of the tile. This takes a 30% error at 40 degrees latitude to less than a 2% error. 

Replacing the inaccurate computeResolution with the more accurate other compute resolution will solve the issue.

jmtaysom:
The compute resolution method in Iterative Cost Distance was calcultating the
cell size based on the length of a degree at the equator without scaling for
the current latitude. This introduces an error as you move away from the
equator. The same method in Iterative Viewshed compensates for this and
maintains greater spatial accuracy.

See issue https://github.com/locationtech/geotrellis/issues/3103

Signed-off-by: jmtaysom <james.taysom@gmail.com>

## Overview

The compute resolution method in Iterative Cost Distance was calcultating the
cell size based on the length of a degree at the equator without scaling for
the current latitude. This introduces an error as you move away from the
equator. The same method in Iterative Viewshed compensates for this and
maintains greater spatial accuracy.

### Checklist

Closes #3103

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/costdistance/IterativeCostDistance.scala
+++ b/spark/src/main/scala/geotrellis/spark/costdistance/IterativeCostDistance.scala
@@ -73,6 +73,10 @@ object IterativeCostDistance {
     def value: Changes = list
   }
 
+  /**
+    * Compute the resolution (in meters per pixel) of a layer.
+    * This function also exists in IterativeViewshed.scala
+    */
   def computeResolution[K: (* => SpatialKey), V: (* => Tile)](
     friction: RDD[(K, V)] with Metadata[TileLayerMetadata[K]]
   ) = {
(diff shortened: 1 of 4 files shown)


jmtaysom:
There are two tests that failed https://github.com/locationtech/geotrellis/blob/master/spark/src/test/scala/geotrellis/spark/costdistance/IterativeCostDistanceSpec.scala#L100 and https://github.com/locationtech/geotrellis/blob/master/spark/src/test/scala/geotrellis/spark/costdistance/IterativeCostDistanceSpec.scala#L110

The two previous tests that are very similar passed 
```
[info] - Should propogate left (1 second, 176 milliseconds)
[info] - Should propogate right (1 second, 227 milliseconds)
[info] - Should propogate up *** FAILED *** (1 second, 198 milliseconds)
[info]   5.000000000000001 was not equal to 5.0 (IterativeCostDistanceSpec.scala:107)
[info] - Should propogate down *** FAILED *** (1 second, 218 milliseconds)
[info]   5.000000000000001 was not equal to 5.0 (IterativeCostDistanceSpec.scala:117)
```
It looks like a floating point rounding error at first glance but I am not sure how precise the results should be. Any suggestions on fixing this?

pomadchin:
hey @jmtaysom you can fix these tests by writing smth like `hops should be (5.0 +- 1e-10)`. 
Also, mb it makes sense to put `computeResolution` function into some `object CostDistance` that we can have in a `spark` package?

pomadchin:
Anyway, even without this code shuffling, I will :+1: the PR once travis is happy. GZ with your first contribution!

jmtaysom:
@pomadchin All the tests are passing now

spark/src/main/scala/geotrellis/spark/costdistance/IterativeCostDistance.scala
```diff
@@ -52,6 +50,27 @@ object IterativeCostDistance {
 
   val logger = Logger.getLogger(IterativeCostDistance.getClass)
 
+  /**
+    * Compute the resolution (in meters per pixel) of a layer.
+    */
+  private [spark] def computeResolution[K: (* => SpatialKey), V: (* => Tile)](
```
pomadchin: I decided to move this function into the `CostDistance` object to avoid code duplication (we had two versions of the same function implemented in two different places).

PR #3115, opened 2019-10-04
Remove usage of LazyLogging

echeipesh:
## Overview

GeoTrellis uses `com.typesafe.scalalogging` package. This is a popular library but unfortunately the `LazyLogging` trait that is used to mix-in a logger has shifted around a number of timers through the version history. This causes frequent binary incompatibility problems.

This PR removes any mention of `LazyLogging` trait and replaces it with `protected lazy val logger: Logger` fields. The `Logger` class is a little more stable.

Additionally:

- ArrayTile no longer warns on narrowing celltype conversion
- GeoWave/GeoMesa do not log errors with experimental notice

### Checklist

- ~[ ] `docs` guides update, if necessary~
- ~[ ] New user API has useful Scaladoc strings~
- ~[ ] Unit tests added for bug-fix or new feature~

### DIFF
--- a/accumulo-spark/src/main/scala/geotrellis/spark/store/accumulo/AccumuloLayerWriter.scala
+++ b/accumulo-spark/src/main/scala/geotrellis/spark/store/accumulo/AccumuloLayerWriter.scala
@@ -25,8 +25,8 @@ import geotrellis.store.index._
 import geotrellis.spark.store._
 import geotrellis.spark.merge._
 import geotrellis.util._
-
-import com.typesafe.scalalogging.LazyLogging
+import org.slf4j.LoggerFactory
+import com.typesafe.scalalogging.Logger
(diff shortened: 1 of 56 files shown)


pomadchin:
Hm, I like that we had to extent a common `LazyLogger` trait, sounds like a much better alternative to copy-pasting everywhere the same `@transient protected lazy val logger = Logger(LoggerFactory.getLogger(getClass.getName))` line.

echeipesh:
Do you feel strongly about that? I think I'm being just a bit defensive and trying to make logging as private as possible without even dependency on `geotrellis-util` package

metasim:
Thanks guys for this!! I'm also in the defensive camp, even if it's not rational.

pomadchin:
@metasim we came to a solution that satisfied both sides https://github.com/locationtech/geotrellis/pull/3116

Issue #3114, opened 2019-10-03
Create 2.x release with LazyLogger removed from Tile hierarchy.


metasim:
RasterFrames has an issue when shading is applied to SLF4J/ScalaLogging, which therefore changes the `Tile` type definition. As a result, Spark SQL `DataSet[T]` that contains `Tile` fails due to the `Encoder[T]`  API using Scala reflections `TypeTag[T]`s, which checks every type up the inheritance hierarchy between what was there at compile time vs what's there at runtime. IOW, Spark + Scala detects that the type hierarchy has changed, and throws an exception.

Once done, this is hopefully addressed:

https://github.com/locationtech/rasterframes/issues/293

metasim:
Couldn't find an existing ticket for removing LazyLogger from type hierarchy, but feel like there was one...

pomadchin:
There was no, thank you for creating this one.

metasim:
I am making a leap of faith that this does address all the issues we have with shading and DB deployment. If you'd prefer to do an "RC" release and we test against that, I'm game.

pomadchin:
@echeipesh we can recover the old [LazyLogger](https://github.com/locationtech/geotrellis/blob/2.0/util/src/main/scala/geotrellis/util/LazyLogging.scala) we had in 2.0.

Tbh I'm really curious why there are problems with shading this dependency and why it is not possible to shade it.

metasim:
@pomadchin 

> we can recover the old LazyLogger we had in 2.0. It would be a pretty straightforward shift back.

This is the not problem at hand (but may be one workaround).  It's the fact that `Tile` inherits from any kind of class that gets renamed during the shading process. This is because Spark SQL's API `Encoder[T: TypeTag]` captures the type hierarchy of `Tile` at *compile* time  (when `Encoder[Tile]` is resolved), stored in some `scala.reflection` data structure. At run time, when the function is called, `scala.reflection` tells the `Encoder[T]` api that the run time class doesn't match the compile time class, because `LazyLogger` has been shaded.

My suggestion is to change the relationship between `Tile` and `LazyLogger` such that it is a "has a" relationship instead of a "is a" one.  

Example: https://github.com/s22s/rasterframes/commit/c07ccea238b900473eb471b7b97b99549e6ceca8#diff-6d174d260f23bb8ae4ec3061b03365f8R36-R39

metasim:
Here's the exception site:

```
[...]

PR #3116, opened 2019-10-04
Switch to log4s

echeipesh:
## Overview

GeoTrellis uses `com.typesafe.scalalogging package`. This is a popular library but unfortunately the `LazyLogging` trait that is used to mix-in a logger has shifted around a number of timers through the version history. This causes frequent binary incompatibility problems.

This PR removes dependency on `typesafe-scalalogging` in favor of `log4s` and moves all logging to private members to minimize their leakage. 

Hopefully this will help us with conflicts and add some nice features to our logging.

Additionally:
GeoWave/GeoMesa do not log errors with experimental notice


echeipesh:
CQ: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20902

pomadchin:
Resolves https://github.com/locationtech/geotrellis/issues/3114 but for the master branch.

Issue #2942, opened 2019-05-24
Add getPixelValue feature


thomas-maschler:
We have a use case for which we have a bunch of point geometries and we would like to enrich point attributes based on different raster bands. We would like to add pixel values for a given point location as attributes to our points.

We are missing a feature in Geotrellis which would allow to extract a value from a raster source based on a single lat/lon location.

Something like this could do
```
val myPoint = Point(1,1)
val source: RasterSource = ???
val pixelValue = source.getPixelValue(myPoint)
```

moradology:
@echeipesh I'm wondering if the best solution here wouldn't be to promote `get` and `getDouble` (not the most descriptive names...) from `Tile` up to `CellGrid` which is already guaranteed to have a grid of things belonging to some `CellType`.

This would force us to satisfy the requirements here and potentially gets at a fairly deep representational failure.

moradology:
After digging into this for a few hours, it became clear that the API we currently have doesn't have the kind of caching we'd really need to take advantage of (on the GeoTrellis side, at least - GDAL has its own story to tell about caching) for this feature to shine. The alternative would be somewhat wasteful as we'd need to read far more data than we'd report and none of that data would be freely available for future operations

This looks like a feature we'll want to revisit as we continue to sophisticated `RasterSource`s but not one we're likely to implement as of now

PR #3117, opened 2019-10-06
Rasters FeatureExtraction

pomadchin:
## Overview

This PR adds a type class to extract `PointFeatures` from `Rasters` by a passed `Geometry`

### Checklist

### Demo

```scala
scala> val ext = Extent(0.0, 0.0, 3.0, 3.0)
ext: geotrellis.vector.Extent = Extent(0.0, 0.0, 3.0, 3.0)

scala>       val data = Array(
     |         1, 2, 3,
     |         4, 5, 6,
     |         7, 8, 9
     |       )
data: Array[Int] = Array(1, 2, 3, 4, 5, 6, 7, 8, 9)

scala> val raster = Raster(ArrayTile(data, 3, 3), ext)
raster: geotrellis.raster.Raster[geotrellis.raster.IntConstantNoDataArrayTile] = Raster(IntConstantNoDataArrayTile([I@5e8a55b6,3,3),Extent(0.0, 0.0, 3.0, 3.0))

scala> val features = raster.cellFeaturesAsPoint[Int](ext.toPolygon)
features: Iterator[geotrellis.vector.Feature[geotrellis.vector.Point,Int]] = non-empty iterator

scala> features.toList
res0: List[geotrellis.vector.Feature[geotrellis.vector.Point,Int]] = List(
Feature(POINT (0.5 2.5),1), 
Feature(POINT (1.5 2.5),2), 
Feature(POINT (2.5 2.5),3), 
Feature(POINT (0.5 1.5),4), 
Feature(POINT (1.5 1.5),5), 
Feature(POINT (2.5 1.5),6), 
Feature(POINT (0.5 0.5),7), 
Feature(POINT (1.5 0.5),8), 
Feature(POINT (2.5 0.5),9))

scala> val areaFeatures = raster.cellFeaturesAs
cellFeaturesAsArea   cellFeaturesAsPoint

scala> val areaFeatures = raster.cellFeaturesAsArea[Int](ext.toPolygon)
areaFeatures: Iterator[geotrellis.vector.Feature[geotrellis.vector.Polygon,Int]] = non-empty iterator

scala> areaFeatures.toList
res1: List[geotrellis.vector.Feature[geotrellis.vector.Polygon,Int]] = List(
Feature(POLYGON ((0 2, 0 3, 1 3, 1 2, 0 2)),1), 
Feature(POLYGON ((1 2, 1 3, 2 3, 2 2, 1 2)),2), 
Feature(POLYGON ((2 2, 2 3, 3 3, 3 2, 2 2)),3), 
Feature(POLYGON ((0 1, 0 2, 1 2, 1 1, 0 1)),4), 
Feature(POLYGON ((1 1, 1 2, 2 2, 2 1, 1 1)),5), 
Feature(POLYGON ((2 1, 2 2, 3 2, 3 1, 2 1)),6), 
Feature(POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0)),7), 
Feature(POLYGON ((1 0, 1 1, 2 1, 2 0, 1 0)),8), 
Feature(POLYGON ((2 0, 2 1, 3 1, 3 0, 2 0)),9))
```

### Notes

The user is able to "easily" provide new instances for `CellFeatures` type class for in-memory rasters by using `CellFeatures.make` and providing: `getGrid: R => GridExtent[Int]`, `cellValue: (R, Int, Int) => D`

`R` is unbounded so any kind of crazy thing that should be treated as Raster will pass mustard here.
Closes https://github.com/locationtech/geotrellis/issues/2942


raster/src/main/scala/geotrellis/raster/FeatureExtraction.scala
```diff
@@ -0,0 +1,58 @@
+package geotrellis.raster
+
+import geotrellis.raster.rasterize.Rasterizer
+import geotrellis.vector.{Feature, Geometry, Point, PointFeature}
+import spire.syntax.cfor._
+
+import scala.collection.mutable.ListBuffer
+
+trait FeatureExtraction[G <: Geometry, T <: CellGrid[Int], D] {
+  def features(geom: G, raster: Raster[T]): Array[Array[PointFeature[D]]]
```
pomadchin: @echeipesh the main question here, is this interface smth that we want? 

Also a quick notion about bands subsetting - bands subset operation should be done as a step before applying this funtion and it definitely should not be a part of this interface.

The output `Array` is for `bands`, the input `Array` is for `PointFeatures`. Do we need an interface that will allow to implement any `Features` 'extraction'?

raster/src/main/scala/geotrellis/raster/FeatureExtraction.scala
```diff
@@ -0,0 +1,75 @@
...
+
+import geotrellis.raster.rasterize.Rasterizer
+import geotrellis.vector.{Feature, Geometry, Point, PointFeature}
+
+import spire.syntax.cfor._
+
+import scala.collection.mutable.ListBuffer
+
+trait FeatureExtraction[I <: Geometry, T <: CellGrid[Int], O <: Geometry, D] {
+  def features(geom: I, raster: Raster[T]): Array[Array[Feature[O, D]]]
```
CloudNiner: @pomadchin asked for input and we discussed changing this return type to `Array[Feature[O, Array[D]]]` so that we don't duplicate point data and its clearer that the data associated with the feature is the int or double value for each band in the queried raster. He's going to attempt this change. At a glance, we don't think that this makes any particular downstream operation against the result data any more difficult than it would have been with the old type `Array[Array[Feature[O, D]]]`

pomadchin: :+1:

raster/src/main/scala/geotrellis/raster/FeatureExtraction.scala
```diff
@@ -0,0 +1,71 @@
...
+
+import geotrellis.raster.rasterize.Rasterizer
+import geotrellis.vector.{Feature, Geometry, Point, PointFeature}
+
+import spire.syntax.cfor._
+
+import scala.collection.mutable.ListBuffer
+
+trait FeatureExtraction[I <: Geometry, T <: CellGrid[Int], O <: Geometry, D] {
+  def features(geom: I, raster: Raster[T]): Array[Feature[O, Array[D]]]
```
CloudNiner: Nice, looks better 👍

raster/src/main/scala/geotrellis/raster/Implicits.scala
```diff
@@ -126,4 +124,18 @@ trait Implicits
...
   }
+
+  implicit class MultibandRasterPointFeatureExtractionMethods(val self: Raster[MultibandTile]) extends MethodExtensions[Raster[MultibandTile]] {
+    def pointFeatures[D] = new {
+      def apply[I <: Geometry: PointFeatureExtraction[*, MultibandTile, D]](geom: I): Array[PointFeature[Array[D]]] =
+        FeatureExtraction[I, MultibandTile, Point, D].features(geom, self)
+    }
+  }
+
+  implicit class rasterPointFeatureExtractionMethods(val self: Raster[Tile]) extends MethodExtensions[Raster[Tile]] {
```
CloudNiner: Capitalize `R` of first word `raster`

raster/src/main/scala/geotrellis/raster/Implicits.scala
```diff
@@ -126,4 +124,18 @@ trait Implicits
...
+    def pointFeatures[D] = new {
+      def apply[I <: Geometry: PointFeatureExtraction[*, Tile, D]](geom: I): Array[PointFeature[D]] =
+        FeatureExtraction[I, Tile, Point, D].features(geom, self).map(_.mapData(_.head))
```
CloudNiner: 👍 glad we do this automatically for single band tiles.

echeipesh:
@pomadchin this PR is ready for re-review. This is the result of our conversation last night.
I tried to keep the interface open enough to be able to implement it for `RasterSource` where the raster would be fetched and featurized in chunks but I don't want to do that here.

raster/src/main/scala/geotrellis/raster/CellFeatures.scala
```diff
@@ -0,0 +1,126 @@
...
+
+import geotrellis.raster.rasterize.Rasterizer
+import geotrellis.vector.{Feature, Geometry, Point, Polygon, Extent}
+import geotrellis.util.MethodExtensions
+
+import spire.syntax.cfor._
+
+
+/** Type class to convert a raster into features of cell geometries */
+trait CellFeatures[R, D] {
```
echeipesh: `R` is unconstrained because I want to be able to do this to `RasterSource` which doesn't have much in common with `Raster[Tile]`

raster/src/main/scala/geotrellis/raster/CellFeatures.scala
```diff
@@ -0,0 +1,126 @@
...
+
+  /** Describe cell grid of the source raster */
+  def cellGrid(raster: R): GridExtent[Long]
```
echeipesh: This very clearly is crying out to be a type class of its own. `RasterLike` or something like that. That could be a meaningful filter on the `Methods` as well. If something can't be described with a `GridExtent[*]` we don't want to be talking about cell features.

IMO we can deal with that later. This method can be removed without impacting subclasses that may spring up in the meantime.

raster/src/main/scala/geotrellis/raster/CellFeatures.scala
```diff
@@ -0,0 +1,126 @@
...
+  def cellFeatures[G <: Geometry](raster: R, geom: Geometry, options: Rasterizer.Options, cellGeom: (Long, Long) => G): Iterator[Feature[G, D]]
+}
+
+object CellFeatures {
+  def apply[R, D](implicit ev: CellFeatures[R, D]): CellFeatures[R, D] = ev
+
+  /** Produce a CellFeatures instance given functions to retrieve a values and describe the cell grid.
+    * This function is suitable for use with rasters that fit entirely in memory.
+    */
+  def make[R, D](getGrid: R => GridExtent[Int], cellValue: (R, Int, Int) => D): CellFeatures[R, D] =
```
echeipesh: `cellValue` is `Function3` and is not specialized, causing boxing for col/row here. Elsewhere in the code we would make a specific trait. However since the main purpose of this interface is to turn a single double cell value into a nested class with a whole geometry ... its fine, we'll let the boxing slide here.

pomadchin: It is itself a slow operation, would like to see someday how bad is boxing here.

raster/src/main/scala/geotrellis/raster/CellFeatures.scala
```diff
@@ -0,0 +1,126 @@
...
+    new CellFeatures[R, D] {
+      def cellGrid(raster: R) = getGrid(raster).toGridType[Long]
+      def cellFeatures[G <: Geometry](raster: R, geom: Geometry, options: Rasterizer.Options, cellGeom: (Long, Long) => G): Iterator[Feature[G, D]] = {
+        val grid = getGrid(raster)
+        val mask = BitArrayTile.empty(cols = grid.cols, rows = grid.rows)
+        Rasterizer.foreachCellByGeometry(geom, grid.toRasterExtent) { case (col, row) => mask.set(col, row, 1) }
+        for {
+          row <- Iterator.range(0, grid.rows)
+          col <- Iterator.range(0, grid.cols) if mask.get(col, row) == 1
```
pomadchin: Nice catch :+1:

PR #3121, opened 2019-10-09
Refactor ResampleGrid names and type parameter

moradology:
## Overview

We want to avoid type parameters from being propagated throughout the library when they aren't absolutely necessary. This PR hides the [N: Integral] type parameter which was formerly on
each ResampleGrid class. Additionally, it renames `ResampleGrid` to `ResampleTarget` (as the grid is what is produced rather than the case class itself) and renames `Dimensions` as `TargetDimensions` for consistency and renames the `IdentityResampleGrid` as `DefaultTarget` to better reflect its context-dependent nature (the exact operation which is "default" is dependent on which resample/reprojection function is being used.

### Checklist

### DIFF
--- a/gdal/src/test/scala/geotrellis/raster/gdal/GDALWarpOptionsSpec.scala
+++ b/gdal/src/test/scala/geotrellis/raster/gdal/GDALWarpOptionsSpec.scala
@@ -62,7 +62,7 @@ class GDALWarpOptionsSpec extends FunSpec with RasterMatchers with GivenWhenThen
           rasterExtent = GridExtent(Extent(630000.0, 215000.0, 645000.0, 228500.0), 10, 10),
           CRS.fromString("+proj=lcc +lat_1=36.16666666666666 +lat_2=34.33333333333334 +lat_0=33.75 +lon_0=-79 +x_0=609601.22 +y_0=0 +datum=NAD83 +units=m +no_defs "),
           WebMercator,
-          TargetCellSize[Long](CellSize(10, 10))
+          TargetCellSize(CellSize(10, 10))
         )
     rasterSourceFromUriOptions(uri, opts)
   }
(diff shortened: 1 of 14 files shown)


raster/src/main/scala/geotrellis/raster/ResampleTarget.scala
```diff
@@ -0,0 +1,118 @@
...
+    source.withResolution(cellSize)
+}
+
+/**
+ * The default resample target is used during reprojection.
+ * A heuristic will be used to determine the best target proportions
+ *
+ * @note If used as target of resample operation it acts as identity operation.
+ */
+case object DefaultResampleTarget extends ResampleTarget {
```
echeipesh: `DefaultResampleTarget` is a bit of a mouthful and none of the other subclasses of `ResampleTarget` have `Resample` in them. Might I suggest `DefaultTarget` as alternative ?

gdal/src/test/scala/geotrellis/raster/gdal/GDALWarpOptionsSpec.scala
```diff
@@ -62,7 +62,7 @@ class GDALWarpOptionsSpec extends FunSpec with RasterMatchers with GivenWhenThen
           rasterExtent = GridExtent(Extent(630000.0, 215000.0, 645000.0, 228500.0), 10, 10),
           CRS.fromString("+proj=lcc +lat_1=36.16666666666666 +lat_2=34.33333333333334 +lat_0=33.75 +lon_0=-79 +x_0=609601.22 +y_0=0 +datum=NAD83 +units=m +no_defs "),
           WebMercator,
-          TargetCellSize[Long](CellSize(10, 10))
+          TargetCellSize(CellSize(10, 10))
```
pomadchin: Sweet change!

raster/src/main/scala/geotrellis/raster/ResampleTarget.scala
pomadchin: `DefaultTarget` sounds much better, but the word `Target` still sounds not that good as "Grid".

raster/src/main/scala/geotrellis/raster/ResampleTarget.scala
```diff
@@ -0,0 +1,118 @@
...
+
+package geotrellis.raster
+
+import geotrellis.raster.reproject.Reproject
+
+import spire.math.Integral
+import spire.implicits._
+
+/** Represents a strategy/target for resampling */
+sealed trait ResampleTarget {
```
pomadchin: Mb these are my language problems, but `ResampleGird` is a bit more understandable for me rather than `ResampleTarget`. Is not `Target` a too generic thing? Like `Data` when we were designing `Paths` / `DataPaths` for `RasterSources`. 

I am not defending `ResampleGrid` name, I am not fan of it either.

pomadchin:
LGTM in general, there are only minor naming issues and it can be merged once all comments are answered and CI is happy.

raster/src/main/scala/geotrellis/raster/ResampleTarget.scala
moradology: Target might be too generic, but I guess i'm still a bit stuck on the fact that it isn't a ResampleGrid so much as a strategy/target by which a grid will be summoned/constructed

pomadchin: @moradology totally agree, it is not `ResampleGrid`, I don't have a better naming suggestion.

Issue #3100, opened 2019-10-01
RasterSource API feedback on resolutions member


echeipesh:
This issue is based on discussion and feedback from @jisantuc

## Background

`RasterSource` interface provides list of available resolutions:
https://github.com/locationtech/geotrellis/blob/f811449e8342d86ffac8f6037643cc153e371f83/raster/src/main/scala/geotrellis/raster/RasterMetadata.scala#L49

The intent is to allow the user of the interface to reason which CellSizes are "safe" to use for resample and at what point the implementation will have to start reading progressively larger swaths of the raster to satisfy the read/resample request.

Incidentally it is theoretically possible that levels of detail or overviews reported in the `resolutions` list do NOT share the same extent. For instance when using `GeoTrellisRasterSource` each level of detail is stored as its own layer, which has its own metadata. The extent across zoom levels may or may not match either through error or design. Exposing `resolutions` as `GridExtent[Long]` allows the user to inspect that.

## Feedback

In code:

```scala
def testTiff(url: String) = {
  import geotrellis.raster.gdal.GDALRasterSource;
  val rasterSource = GDALRasterSource(url)
  val idealResolution = rasterSource.resolutions.minBy(res => scala.math.abs(250000 - res.rows * res.cols))
  val tile = rasterSource.read(idealResolution.gridBounds).get.tile
}
```

Following the types it is easy to mis-use the `read` method.
`read(grid: GridBounds)` reads relative to pixel bounds of the raster. `GridExtent` representing resolution also has `GridBounds` instance that basically represents the pixel size of the region. It is always indexed by (minCol=0, minRow=0).

Expected result in this case might be: read me the overview at resolution I have picked (say for thumbnail)
Actual result: read upper left corner from the image of same size as the pixel size of the resolution.

## Conclusions

In idea scenario types could not be connected like this to do nonsense operations. Possible solution:
- Change `resolutions` from `List[GridExtent]` to `List[CellSize]`
- Remove `gridBounds` method from `GridExtent`

Read method overloads are a little confusing:

`read(extent)` reads based on map extent
`read(gridBounds)` reads based on pixel extent

Consider renaming all overloads to `readExtent` and `readBounds` so it is more obvious from reading the code how the read is performed.

echeipesh:
## GridExtent should not have .gridBounds

Exploring this idea further.
Semantically `GridBounds` instance has two related meanings in GT codebase.
[...]

Issue #3123, opened 2019-10-14
GDALRasterSource resolutions should not change after resample


echeipesh:
```scala
gdal > console
[info] Starting scala interpreter...
Welcome to Scala 2.11.12 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_211).
Type in expressions for evaluation. Or try :help.

scala> import geotrellis.raster.gdal._
import geotrellis.raster.gdal._

scala> import geotrellis.raster.io.geotiff.{AutoHigherResolution, OverviewStrategy}
import geotrellis.raster.io.geotiff.{AutoHigherResolution, OverviewStrategy}

scala> import geotrellis.raster.resample._
import geotrellis.raster.resample._

scala> val rs = GDALRasterSource("/Users/eugene/proj/geotrellis/raster/data/vlm/aspect-tiled.tif")
rs: geotrellis.raster.gdal.GDALRasterSource = GDALRasterSource(/Users/eugene/proj/geotrellis/raster/data/vlm/aspect-tiled.tif,GDALWarpOptions(-of VRT -ovr AUTO))

scala> rs.resolutions
res0: List[geotrellis.raster.CellSize] = List(CellSize(20.0,20.0), CellSize(40.0,39.94082840236686), CellSize(79.7872340425532,79.88165680473372))

scala> val resampled = rs.resample(TargetCellSize(CellSize(25,25)), NearestNeighbor, AutoHigherResolution)
resampled: geotrellis.raster.RasterSource = GDALRasterSource(/Users/eugene/proj/geotrellis/raster/data/vlm/aspect-tiled.tif,GDALWarpOptions(-of VRT -tap -tr 25.0 25.0 -ovr AUTO -te 630000.0 215000.0 645000.0 228500.0))

scala> resampled.resolutions
res5: List[geotrellis.raster.CellSize] = List(CellSize(50.0,50.0), CellSize(100.0,100.0))
```

The purpose of [`RasterSource.resolutions`](https://github.com/locationtech/geotrellis/blob/9c6f23ad9f56277ea2d118f059808c90bf798f93/raster/src/main/scala/geotrellis/raster/RasterMetadata.scala#L37-L49) method is to inform at what resolution it is safe to read the data to minimize IO.

[...]

PR #3124, opened 2019-10-14
RasterSource.resolutions becomes List[CellSize]

echeipesh:
## Overview

This PR is intended to address usability issues around resampling RasterSource described in https://github.com/locationtech/geotrellis/issues/3100

*RasterSource.resolutions* is not `List[CellSize]`. As discussed in the issue this removes the ability to provide `GridBounds` from `RasterSource.resolutions` to the `RasterSource.read` method of the same instance. The judgement call is that if each overview somehow has differing extent this is considered very unusual and can be diagnosed by further digging into `RasterMetadata` interface or through additional specialized API.

*Grid.gridBounds* are removed. This added a very confusing case where `GridBounds` could either be a window in some raster or describe the whole footprint of raster. Removing `Grid.gridBounds` goes a long way to remove that confusion. This is a recently added member to the class, so impact is judged to be small.

*GridBounds.intersects(dimensions)* - For those cases when we're checking for intersection with a raster we're using the new `Dimensions` class. `Dimensions` can be checked for intersection vs `GridBounds` and intersection of `GridBounds` and `Dimensions` is smaller `Option[GridBounds]`.

*Grid.dimensions* becomes `Dimensions[N]` Instead of `(N, N)` - with addition of specialized API described above. Giving this a class goes a long way to clarify its usage in API. Note that this change results in fix-ups to tests that could be be caught by compiler due to checking for equality with "==". This is likely to happen in client code but should be easily fixed. 

### Checklist

### Notes
```
gdal > console
[info] Starting scala interpreter...
Welcome to Scala 2.11.12 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_211).
Type in expressions for evaluation. Or try :help.

scala> import geotrellis.raster.gdal._
import geotrellis.raster.gdal._

scala> import geotrellis.raster.io.geotiff.{AutoHigherResolution, OverviewStrategy}
import geotrellis.raster.io.geotiff.{AutoHigherResolution, OverviewStrategy}

scala> import geotrellis.raster.resample._
import geotrellis.raster.resample._

scala> val rs = GDALRasterSource("/Users/eugene/proj/geotrellis/raster/data/vlm/aspect-tiled.tif")
rs: geotrellis.raster.gdal.GDALRasterSource = GDALRasterSource(/Users/eugene/proj/geotrellis/raster/data/vlm/aspect-tiled.tif,GDALWarpOptions(-of VRT -ovr AUTO))

scala> rs.resolutions
res0: List[geotrellis.raster.CellSize] = List(CellSize(20.0,20.0), CellSize(40.0,39.94082840236686), CellSize(79.7872340425532,79.88165680473372))
```

Closes https://github.com/locationtech/geotrellis/issues/3100

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/Dimensions.scala
+++ b/raster/src/main/scala/geotrellis/raster/Dimensions.scala
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2019 Azavea
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package geotrellis.raster
+
+import spire.math.Integral
+import spire.implicits._
+
+case class Dimensions[@specialized(Byte, Short, Int, Long) N: Integral](cols: N, rows: N) extends Product2[N, N] with Serializable {
+  def _1 = cols
+  def _2 = rows
+  def size: Long = Integral[N].toType[Long](cols) * Integral[N].toType[Long](rows)
+  override def toString = s"${cols}x${rows}"
+}
+
+object Dimensions {
+  implicit def apply[N: Integral](tup: (N, N)) = new Dimensions(tup._1, tup._2)
+}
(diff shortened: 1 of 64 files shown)


echeipesh:
While preparing this PR following issue was discovered: https://github.com/locationtech/geotrellis/issues/3123

raster/src/main/scala/geotrellis/raster/Dimensions.scala
```diff
@@ -0,0 +1,31 @@
...
+
+case class Dimensions[@specialized(Byte, Short, Int, Long) N: Integral](cols: N, rows: N) extends Product2[N, N] with Serializable {
+  def _1 = cols
+  def _2 = rows
+  def size: Long = Integral[N].toType[Long](cols) * Integral[N].toType[Long](rows)
+  override def toString = s"${cols}x${rows}"
+}
+
+object Dimensions {
+  implicit def apply[N: Integral](tup: (N, N)) = new Dimensions(tup._1, tup._2)
```
pomadchin: Just a notion: boxing happens here, but we don't care too much about it.

PR #3125, opened 2019-10-15
Resolutions should always be read from the source dataset

pomadchin:
## Overview

Forces GDALRasterSources to read resolutions from the `SOURCE` `Dataset` always.

### Checklist

Closes https://github.com/locationtech/geotrellis/issues/3123

### DIFF
--- a/gdal/src/main/scala/geotrellis/raster/gdal/GDALRasterSource.scala
+++ b/gdal/src/main/scala/geotrellis/raster/gdal/GDALRasterSource.scala
@@ -97,7 +97,7 @@ class GDALRasterSource(
     * These resolutions could represent actual overview as seen in source file
     * or overviews of VRT that was created as result of resample operations.
     */
-  lazy val resolutions: List[CellSize] = dataset.resolutions(datasetType).map(_.cellSize)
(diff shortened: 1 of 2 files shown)


gdal/src/main/scala/geotrellis/raster/gdal/GDALRasterSource.scala
```diff
@@ -97,7 +97,7 @@ class GDALRasterSource(
     * These resolutions could represent actual overview as seen in source file
     * or overviews of VRT that was created as result of resample operations.
     */
-  lazy val resolutions: List[CellSize] = dataset.resolutions(datasetType).map(_.cellSize)
+  lazy val resolutions: List[CellSize] = dataset.resolutions(GDALDataset.SOURCE).map(_.cellSize)
```
pomadchin: Moved this PR into `in-progress` state, since I need to fix comments and probably to add the old version / function that will allow to get resolutions by the `GDALDataset.TYPE`

pomadchin:
@echeipesh good catch, I'll see what is possible to do here really quick today.

pomadchin:
Was closed in https://github.com/locationtech/geotrellis/pull/3214

PR #3128, opened 2019-10-16
Add Circe codecs for GeoTiffInfo and related classes

echeipesh:
## Overview

Encoding GeoTiffInfo can be useful for caching purposes. To minimize the implicit resolution time we're tucking away codecs into companion objects and making heavy use of macro encoder/decoder derivation where we can. This is likely going to be needed again in client code and we can save people some pain by just doing it here.

Related: https://github.com/raster-foundry/raster-foundry/pull/5222/commits/0fe73d721c9526ccaa471135ea1e797176449409

This PR also: 
- breaks out `GeoTiffInfo` out of `GeoTiffReader` to minimize weirdness in the codebase
- Merges `TiffTagsReader` class into `TiffTags.read` method

### Checklist


raster/src/main/scala/geotrellis/raster/io/geotiff/StorageMethod.scala
```diff
@@ -51,3 +53,47 @@ object Striped extends StripedStorageMethod {
...
+              val rowsPerStrip =
+                c.downField("rowsPerStrip").as[Option[Int]] match {
+                  case Left(_)  => None
+                  case Right(v) => v
+                }
+              new Striped(rowsPerStrip)
+            }
+            case _ => {
+              val cols = c.downField("cols").as[Int] match {
+                case Left(_)  => 256
```
pomadchin: It is not true, such logic is fine for the application, but not for the library. A GDAL fallback behavior is the following: It is not tiled => it is striped.

raster/src/main/scala/geotrellis/raster/io/geotiff/StorageMethod.scala
```diff
@@ -51,3 +53,47 @@ object Striped extends StripedStorageMethod {
...
+                case Right(v) => v
+              }
+              Tiled(cols, rows)
+            }
+          }
+        }
+      }
+    }
+
+  implicit val storageMethodEncoder: Encoder[StorageMethod] =
```
pomadchin: A side question, why `StorageMethod` is a sealed _class_? and not trait? it looks more like an interface with no implementation.

raster/src/main/scala/geotrellis/raster/io/geotiff/StorageMethod.scala
```diff
@@ -51,3 +53,47 @@ object Striped extends StripedStorageMethod {
...
+    new Encoder[StorageMethod] {
+      final def apply(a: StorageMethod): Json = a match {
+        case _: Striped => Json.obj(("storageType", Json.fromString("striped")))
```
pomadchin: `value.asJson` is more readable rather than these `Json.fromType` // just a minor readability notion. 

requires an `io.crice.syntax._` though.

raster/src/main/scala/geotrellis/raster/io/geotiff/tags/BasicTags.scala
```diff
@@ -17,6 +17,7 @@
 package geotrellis.raster.io.geotiff.tags
 
 import monocle.macros.Lenses
+import io.circe._, io.circe.generic.semiauto._
```
pomadchin: Eh mb im too pedantic here, but let's have the same importing style everywhere,

raster/src/main/scala/geotrellis/raster/io/geotiff/tags/CmykTags.scala
```diff
@@ -17,12 +17,16 @@
...
+@JsonCodec
 case class CmykTags(
   inkSet: Option[Int] = None,
   numberOfInks: Option[Int] = None,
   inkNames: Option[String] = None,
   dotRange: Option[Array[Int]] = None,
   targetPrinters: Option[String] = None
 )
+
+object CmykTags
```
pomadchin: Hm, so if we've chosen a no `double` annotations path, then probably it should follow a usual style with manual codecs definition in the companion object.

raster/src/main/scala/geotrellis/raster/io/geotiff/package.scala
```diff
@@ -24,5 +24,4 @@ package object geotiff extends Implicits {
   val Deflate = DeflateCompression
 
   val GeoTiffReader = reader.GeoTiffReader
-  val TiffTagsReader = reader.TiffTagsReader
```
pomadchin: Hm, I thought it was a good pattern to follow, to decompose 'tags' / 'tiffs' parsing from the objects they are collected into. Since parsers are really huge and can be placed into a separate file.

PR #3132, opened 2019-10-21
Switch to scala 2.12 as build default

echeipesh:
Switching to scala 2.12 gives us 25% faster build time on this project and better compatibility with tools like metals. Developer time is precious so that's nothing to sneeze at.

There are couple of issues with this PR that are still outstanding
- [x] `geowave` version we're using is 0.9.3 and has a strict dependency on scala 2.11 dependencies
- [x] `geomesa` version has similar problem
- [x] publish scripts need to be adjusted
- [x] deprecation warnings need to be resolved

Overall the geowave and geomesa projects are being unlinked from the main build for a bit as those problems look like they'll be harder to solve and both of those projects are experimental and not in heavy use.

CQs required:
- [x] [Monocle 2.0.0](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21077)
- [x] [Circe 0.12.2](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21078)
- [x] [Cats Core 2.0.0](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21159)
- [x] [Cats Effect 2.0.0](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21160)
- [x] [FS2 2.0.1](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21161)
- [x] [ScalaURI 1.5.1](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21162)

Fixes https://github.com/locationtech/geotrellis/issues/3139


project/Dependencies.scala
```diff
@@ -16,6 +16,23 @@
...
+  val hbase       = "2.2.0"
+  val geomesa     = "2.3.1"
+  val geowave     = "0.9.3"
+  val circe       = "0.11.1"
+  val gdal        = "2.4.0"
+  val gdalWarp    = "33.61199eb"
+  val hadoop      = "2.8.5"
+  val spark       = "2.4.4"
+  val previousVersion = "2.2.0"
+}
```
echeipesh: Its a surprisingly major inconvenience to have version and dependency files separated. At minimum moving this together, perhaps even nesting it in `Dependency` object would be a good thing to do.

project/Environment.scala
```diff
@@ -1,26 +0,0 @@
...
-import scala.util.Properties
-
-object Environment {
-  def either(environmentVariable: String, default: String): String =
-    Properties.envOrElse(environmentVariable, default)
-
-  lazy val hadoopVersion  = either("SPARK_HADOOP_VERSION", "2.8.5")
-  lazy val sparkVersion   = either("SPARK_VERSION", "2.4.3")
-  lazy val versionSuffix  = either("GEOTRELLIS_VERSION_SUFFIX", "-SNAPSHOT")
-}
```
echeipesh: This has never been used AFAIK

.travis/build-and-test-set-1.sh
```diff
@@ -9,9 +9,9 @@
   "project vector" test \
   "project vectortile" test \
   "project gdal" test \
-  "project geowave" compile test:compile \
+  # "project geowave" compile test:compile \
```
pomadchin: I think the right change here is to rename all `-212` files into files without this postfix and all files wihtout postfix to rename to `-211` files and add pass everywhere an explicit scala version

pomadchin: Ouch in my experience comments like these were never interpreted properly by bash

pomadchin:
The build fails because of this https://github.com/lightbend/mima/issues/422

.travis/build-and-test-set-1.sh
pomadchin: I know that it is still in progress, but I remembered about how the scala 2.12 / 2.11 dispatch happens now: https://github.com/locationtech/geotrellis/blob/master/.travis/build-and-test.sh 

`build-and-test-set-1` runs tests against Scala 2.11
We had to implement this "manual" scala versions dispatch due to geowave / geomesa 2.12 "support".

Do you think about moving GeoWave and GeoMesa packages out of this repo?

echeipesh: Perhaps at least unlinking them from the multi-project build and having a direct dependency on a specific version of GT and its own test.

echeipesh: My agenda is to get rid of these shell files as much as possible and do everything only with SBT commands. If we can pull that off that will reduce the number of places we introduce errors.

pomadchin: this is 🔥

pomadchin:
I requested a PMC approval, but not sure would it be completed successfully or not, due to ongoing changes in the locationtech eclipse structure.

project/Dependencies.scala
```diff
@@ -16,61 +16,79 @@
...
+  private def ver(for211: String, for212: String) = Def.setting {
+    CrossVersion.partialVersion(scalaVersion.value) match {
+      case Some((2, 11)) => for211
+      case Some((2, 12)) => for212
+      case _ => sys.error("not good")
+    }
+  }
+
+  def monocle(module: String) = Def.setting {
+    "com.github.julien-truffaut" %% s"monocle-$module" % ver("1.5.1-cats", "2.0.0").value
```
pomadchin: Hm I think RasterFrames are using the syntax like this.

.travis/build-and-test-set-1.sh
```diff
@@ -10,9 +10,7 @@
   "project vectortile" test \
   "project util" test \
   "project gdal" test \
-  "project geowave" compile test:compile \
```
pomadchin: I hope we won't forget to remove the code from the codebase.

doc-examples/src/main/scala/geotrellis/doc/examples/spark/EuclideanDistanceExamples.scala
```diff
@@ -58,7 +58,7 @@ object EuclideanDistanceExamples {
     val tileRDD: RDD[(SpatialKey, Tile)] = inputRDD.euclideanDistance(ld)
 
     val maxDistance = tileRDD.map(_._2.findMinMaxDouble).collect.foldLeft(-1.0/0.0){ (max, minMax) => scala.math.max(max, minMax._2) }
-    val cm = ColorMap((0.0 to maxDistance by (maxDistance/512)).toArray, ColorRamps.BlueToRed)
+    val cm = ColorMap(Range.BigDecimal.inclusive(0.0, maxDistance, maxDistance/512).map(_.toDouble).toArray, ColorRamps.BlueToRed)
```
pomadchin: What a syntax!

doc-examples/src/main/scala/geotrellis/doc/examples/spark/VoxelKey.scala
```diff
@@ -66,6 +65,9 @@ object VoxelKey {
       (k, sk) => VoxelKey(sk.col, sk.row, k.z)
     )
   }
+
+  implicit val voxelKeyEncoder: Encoder[VoxelKey] = deriveEncoder[VoxelKey]
```
pomadchin: Hm, would the annotation here more user friendly? (less lines ~ easier to understand)

scripts/buildall.sh
```diff
@@ -5,9 +5,8 @@
 ./sbt -J-Xmx2G "project cassandra" test  || { exit 1; }
 ./sbt -J-Xmx2G "project cassandra-spark" test  || { exit 1; }
 ./sbt -J-Xmx2G "project doc-examples" compile || { exit 1; }
-./sbt -J-Xmx2G "project geomesa" test || { exit 1; }
+# ./sbt -J-Xmx2G "project geomesa" test || { exit 1; }
```
pomadchin: Should it be just removed?

spark/src/main/scala/geotrellis/spark/matching/RDDHistogramMatching.scala
```diff
@@ -94,7 +94,7 @@ object RDDHistogramMatching {
     sourceHistograms: Seq[Histogram[T1]],
     targetHistograms: Seq[Histogram[T2]]
   ): RDD[(K, MultibandTile)] =
-    rdd.map({ case (key, tile: MultibandTile) =>
+    rdd.map({ case (key, tile) =>
```
pomadchin: You don't need extra brackets here: `rdd.map { case (key, tile) =>  (key, HistogramMatching(tile, sourceHistograms, targetHistograms)) }`. This comment can be applied to all changes below btw.

version.sbt
```diff
@@ -0,0 +1 @@
+ThisBuild / version := "3.1.1-SNAPSHOT"
```
pomadchin: Is keeping the version in a separate file more convenient? With my gt-contrib and gt-server experiences I didn't notice any real wins ):

spark/src/main/scala/geotrellis/spark/matching/RDDHistogramMatching.scala
echeipesh: I'm going for minimum code change outside of configuration here, would rather not get sucked into format shuffling. Having said that once we get the guild and publishing sorted out we really should look into enforcing a code formatting standard for the library.

doc-examples/src/main/scala/geotrellis/doc/examples/spark/VoxelKey.scala
echeipesh: They didn't compile for some reason and this was the work-around. I didn't look into why they didn't because it seems like the annotation has been very fragile in other cases and using `deriveEncoder` is mostly the default in my mind.

doc-examples/src/main/scala/geotrellis/doc/examples/spark/EuclideanDistanceExamples.scala
echeipesh: Oy ... I know. I get what they're saying with this deprecation but this is the least-worst alternative. I want to keep the number of warnings down so we don't go into auto-ignore warning mode and miss something that matters though.

project/Dependencies.scala
echeipesh: That's where I lifted this approach but I've certainly seen it in other projects. What do you think? I'm in favor of collapsing the version numbers to here so there is a minimum amount of indirection in the build overall.

version.sbt
echeipesh: Opening the discussion with this. Felt like I should pull this out of `Version.scala` which always stumps me. I'd be also OK with putting this version into `build.sbt` with other build settings.

project/Dependencies.scala
pomadchin: LGTM 👍

version.sbt
pomadchin: Let’s try with a version in a file; but I’m just worrying that it would be just an extra file and the story would be the same as a Versions class in a separate file.

pomadchin: Let’s try with a version in a file; but I’m just worrying that it would be just an extra file and the story would be the same as a Versions class in a separate file.

PR #3147, opened 2019-11-06
Remove unnecessary calculations in Hillshade

esmeetu:
# Overview

remove unnecessary slope and aspect calculations.

## Checklist

## Notes

Two questions:
1. In terms of `zFactor`, should it be better to put this parameter to `SurfacePointCalculation` class?
in method `calcSurface` like this:
```scala
s.`dz/dx` = zFactor * (neValue + 2*eValue + seValue - nwValue - 2*wValue - swValue) / (8 * cellWidth)
s.`dz/dy` = zFactor * (swValue + 2*sValue + seValue - nwValue - 2*nValue - neValue) / (8 * cellHeight)
```
2. in the expression `resultTile.set(x, y, round(127.0 * max(0.0, v)).toInt)` of the commited file, why `127.0` is set here while it is `255.0` in [https://desktop.arcgis.com/en/arcmap/10.3/tools/spatial-analyst-toolbox/how-hillshade-works.htm](https://desktop.arcgis.com/en/arcmap/10.3/tools/spatial-analyst-toolbox/how-hillshade-works.htm)

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/mapalgebra/focal/hillshade/Hillshade.scala
+++ b/raster/src/main/scala/geotrellis/raster/mapalgebra/focal/hillshade/Hillshade.scala
@@ -53,9 +53,8 @@ object Hillshade {
       val sinAz = sin(azimuth)
 
       def setValue(x: Int, y: Int, s: SurfacePoint) {
-        val slope = s.slope(zFactor)
-        val aspect = s.aspect
-
+        s.`dz/dx` = s.`dz/dx` * zFactor
+        s.`dz/dy` = s.`dz/dy` * zFactor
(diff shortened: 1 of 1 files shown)


pomadchin:
Hey @esmeetu thanks for your contribution! Can you please sign an ECA? https://github.com/locationtech/geotrellis/blob/master/docs/CONTRIBUTING.rst#eclipse-contributor-agreement-eca after signing it just do `git commit --amend -s && git push origin -f master`

echeipesh:
@esmeetu 
1. zFactor being parameter to `SurfacePointCalculation` sounds like it would be cleaner, but I'm not sure it makes much difference in practice. Certainly not required for this fix.

2. `127` is max value for JVM `Byte`, not sure what the historical reasoning for it was.

PR #3053, opened 2019-08-14
Add RasterSources API

pomadchin:
## Overview

This PR moves the code from the https://github.com/geotrellis/geotrellis-contrib repository.
Adds: 
* GeoTrellisRasterSources (think of a better naming (?))
* GeoTiffRasterSources
* GDALRasterSources

All the effects functionality still remains in the contrib repo.

### Checklist

- [x] `docs/CHANGELOG.rst` updated, if necessary
- [x] Make GDAL Tests work (travis should launch tests in a docker container)

### CQs
- [x] pureconfig 0.11.1 [CQ #20734](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20734)
- [x] scala-logging 3.9.2 [CQ #20735](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20735)
- [x] opencsv 4.6 [CQ #20736](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20736)
- [x] chronoscala 0.3.2 [CQ #20737](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20737)
- [x] aws s3 sdk v2 2.7.32 [CQ #20738](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20738)
- [x] cats-core 1.6.1 [CQ #20739](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20739)
- [x] cats-effect 1.3.1 [CQ #20740](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20740)
- [x] fs2-core 1.0.5 [CQ #20741](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20741)
- [x] fs2-io 1.0.5 [CQ #20741](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20741)
- [x] scaffeine 3.1.0 [CQ #20742](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20742)
- [x] scalaj-http 2.4.2 [CQ #20743](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20743)
- [x] scalapb-runtime 0.9.0 [CQ #20744](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20744)
- [x] scalactic 3.0.8 [CQ #20745](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20745)
[...]

PR #3150, opened 2019-11-08
Remove min/max in computing CellType for GDALDataset

### LINKED ISSUE #3148: Reduce unnecessary get_band_min_max calls in computing CellType
Using GDALDataset in Spark, evaluation of various datasets' metadata is happening in parallel. Observed "a lot" of calls to [GDALWarp.get_band_min_max][1]. This is forcing a lot of data reads that are ultimately unnecessary for many values of GDALDataType.  

Issue is to reduce/remove unnecessary reads and calls.

[1]: https://github.com/locationtech/geotrellis/blob/master/gdal/src/main/scala/geotrellis/raster/gdal/GDALDataset.scala#L295

vpipkt:
# Overview

GDAL's type system is narrower than GT
[GDALDataType.GDT_Byte](https://gdal.org/api/raster_c_api.html#_CPPv412GDALDataType) maps to `UByte*` `CellType`s only depending on NoData

In GDALUtils.dataTypeToCellType, only get unsigned Byte by passing in appropriate min and max. This is to avoid breaking the existing call behavior.

Some gitter discussion on the topic [here](https://gitter.im/geotrellis/geotrellis?at=5dc5839a6ba2347d2d59af2e)

## Checklist

- [x] [Module Hierarcy](https://github.com/locationtech/geotrellis/blob/master/docs/guide/module-hierarchy.rst) updated, if necessary

Closes #3148

### DIFF
--- a/gdal/src/main/scala/geotrellis/raster/gdal/GDALDataset.scala
+++ b/gdal/src/main/scala/geotrellis/raster/gdal/GDALDataset.scala
@@ -288,23 +288,7 @@ case class GDALDataset(token: Long) extends AnyVal {
     require(acceptableDatasets contains datasetType)
     val nd = noDataValue(datasetType)
     val dt = GDALDataType.intToGDALDataType(this.dataType(datasetType))
-    val mm = {
-      val minmax = Array.ofDim[Double](2)
-      val success = Array.ofDim[Int](1)
-
-      val returnValue = GDALWarp.get_band_min_max(token, datasetType.value, numberOfAttempts, 1, true, minmax, success)
-
-      if (returnValue <= 0) {
-        val positiveValue = math.abs(returnValue)
-        throw new MalformedDataTypeException(
-          s"Unable to deterime the min/max values in order to calculate CellType. GDAL Error Code: $positiveValue",
-          positiveValue
-        )
-      }
-      if (success(0) != 0) Some(minmax(0), minmax(1))
-      else None
-    }
-    GDALUtils.dataTypeToCellType(datatype = dt, noDataValue = nd, minMaxValues = mm)
+    GDALUtils.dataTypeToCellType(datatype = dt, noDataValue = nd)
   }
 
   def readTile(gb: GridBounds[Int] = GridBounds(dimensions), band: Int, datasetType: DatasetType = GDALDataset.WARPED): Tile = {
--- a/gdal/src/main/scala/geotrellis/raster/gdal/GDALUtils.scala
+++ b/gdal/src/main/scala/geotrellis/raster/gdal/GDALUtils.scala
@@ -43,17 +43,19 @@ object GDALUtils {
           case Some(bits) if bits == 1 => BitCellType
           case _ =>
             minMaxValues match {
-              case Some((mi, ma)) if (mi.toInt >= 0 && mi <= 255) && (ma.toInt >= 0 && ma <= 255) =>
-                noDataValue match {
-                  case Some(nd) if nd.toInt > 0 && nd <= 255 => UByteUserDefinedNoDataCellType(nd.toByte)
-                  case Some(nd) if nd.toInt == 0 => UByteConstantNoDataCellType
-                  case _ => UByteCellType
-                }
-              case _ =>
+              case Some((mi, ma)) if mi >= -128 && mi < 128 && ma >= -128 && ma < 128 =>
+                // signed Byte
                 noDataValue match {
                   case Some(nd) if nd.toInt > Byte.MinValue.toInt && nd <= Byte.MaxValue.toInt => ByteUserDefinedNoDataCellType(nd.toByte)
                   case Some(nd) if nd.toInt == Byte.MinValue.toInt => ByteConstantNoDataCellType
                   case _ => ByteCellType
+                }
+              case _ ⇒
+                // unsigned Byte
+                noDataValue match {
+                  case Some(nd) if nd.toInt > 0 && nd <= 255 => UByteUserDefinedNoDataCellType(nd.toByte)
+                  case Some(nd) if nd.toInt == 0 => UByteConstantNoDataCellType
+                  case _ => UByteCellType
                 }
             }
         }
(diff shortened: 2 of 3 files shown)


vpipkt:
@pomadchin says on gitter:  can you cover it with test? so we would know that GDALRasterSource behaves the same way GeoTiffRasterSource does

vpipkt:
Unsure why the CI build is failing. Looks like one is marked as fail but logs look clean? Another has a failed dependency resolution :-(

pomadchin:
Restarted the build; also this PR would probably have to be double checked by the RasterFoundry team

vpipkt:
Feel free to assign the appropriate members from the community to review it.

pomadchin:
@vpipkt btw so what about tests? This PR in general makes sense, but I remember that we introduced these changes to make it behave like `GeoTiffRasterSource`. I would like to see smth like reading different `Byte TIFFs` with `GDALRasterSource` and `GeoTiffRasterSource` and compare the results. 

For the context: 
the original issue was https://github.com/raster-foundry/raster-foundry/issues/4442 and it was addressed here https://github.com/geotrellis/geotrellis-gdal/pull/40

I also would like to mention @notthatbreezy since he can be intrested in this change and mb will help us to test it out.

vpipkt:
@pomadchin very helpful, I am looking into the original issue and PR.

vpipkt:
So @pomadchin I am not sure where to take this next. I've looked over the unit tests in https://github.com/geotrellis/geotrellis-gdal/pull/40 and confirmed they are still in the code and passing for [this branch](https://travis-ci.org/locationtech/geotrellis/jobs/609336730#L5751-L5755). So no reason to believe that this causes a regression.

I did notice one thing in that PR. In the reader, you changed the min max to be [`lazy val` ](https://github.com/pomadchin/geotrellis-gdal/commit/1cef88d8de627aa387ef1d4dbdbd3466f1ae12e2#diff-5e19e2811cf054c87ebc2afd642ebaa3R64) and the corresponding parameter to be [by-name](https://github.com/pomadchin/geotrellis-gdal/commit/1cef88d8de627aa387ef1d4dbdbd3466f1ae12e2#diff-fe0810f15e696c5a0e0dfa64e6f3d01eR39). I had initially proposed changing [`mm`](https://github.com/locationtech/geotrellis/pull/3150/files#diff-01acc687b0a8fe9ab43a1c68d4f3c77fL291) to be a lazy val, along these lines. But I wonder if perhaps this lazy val declaration was lost along the way porting into GT with https://github.com/locationtech/geotrellis/pull/3053.

I think we are at a decision point.
1. Consider the unit tests as sufficient and merge as is?
2. Change to `lazy val mm` declaration, and add more code comments documenting why this is important.

vpipkt:
Very interesting:

https://gdal.org/api/raster_c_api.html#_CPPv420GDALDataTypeIsSigned12GDALDataType

In the GDAL API (from version 2.3), you can explicitly ask if the DataType is signed. That would be a more efficient and more appropriate way to compute the GeoTrellis CellType I believe. However would require some changes in the gdal bindings library.

pomadchin:
@vpipkt this is a great find!

pomadchin:
Returning back to this issue; exposing a sign via a GDAL method makes no sense: https://github.com/geotrellis/gdal-warp-bindings/issues/71 

Will closely review it today and hopefully it would get merged soon :+1:

gdal/src/main/scala/geotrellis/raster/gdal/GDALUtils.scala
```diff
@@ -43,18 +43,20 @@ object GDALUtils {
...
           case _ =>
             minMaxValues match {
-              case Some((mi, ma)) if (mi.toInt >= 0 && mi <= 255) && (ma.toInt >= 0 && ma <= 255) =>
-                noDataValue match {
-                  case Some(nd) if nd.toInt > 0 && nd <= 255 => UByteUserDefinedNoDataCellType(nd.toByte)
-                  case Some(nd) if nd.toInt == 0 => UByteConstantNoDataCellType
-                  case _ => UByteCellType
-                }
-              case _ =>
+              case Some((mi, ma)) if mi >= -128 && mi < 128 && ma >= -128 && ma < 128 =>
```
pomadchin, line 46: If min and max is not passed into the function, than these checks make no sense. Was it done intentionally?

gdal/src/main/scala/geotrellis/raster/gdal/GDALDataset.scala
```diff
@@ -288,23 +288,7 @@ case class GDALDataset(token: Long) extends AnyVal {
...
-        val positiveValue = math.abs(returnValue)
-        throw new MalformedDataTypeException(
-          s"Unable to deterime the min/max values in order to calculate CellType. GDAL Error Code: $positiveValue",
-          positiveValue
-        )
-      }
-      if (success(0) != 0) Some(minmax(0), minmax(1))
-      else None
-    }
-    GDALUtils.dataTypeToCellType(datatype = dt, noDataValue = nd, minMaxValues = mm)
```
pomadchin, line 307: You can look into https://github.com/locationtech/geotrellis/issues/3219 for clarifications: we'd like to keep the consistent CellTypes interpretation across RasterSources.

pomadchin:
Closing it without merging! Feel free to reopen it / continue the conversation under the new issue. ): 
Also geotrellis/gdal-warp-bindings#71 became a complete dissapointment.

pomadchin:
Just for the history; the following unit test should pass:

```scala
    describe("should derive the cellType consistently with GeoTiffRasterSource") {
      val ext = Extent(0.0, 0.0, 3.0, 3.0)
      val data = Array(
        1, 2, 3,
        4, 5, 6,
        7, 8, 9
      )

      List(BitCellType, ByteCellType, UByteCellType, ShortCellType, UShortCellType, IntCellType, FloatCellType, DoubleCellType).map { ct =>
        it(ct.getClass.getName.split("\\$").last.split("\\.").last) {
          val path = s"/tmp/gdal-$ct-test.tiff"
          val raster = Raster(ArrayTile(data, 3, 3).convert(ct), ext)
          GeoTiff(raster, LatLng).write(path)

          GDALRasterSource(path).cellType shouldBe ct
          GeoTiffRasterSource(path).cellType shouldBe ct
        }
      }
    }
```

pomadchin:
Replaced with https://github.com/locationtech/geotrellis/issues/3232

PR #3156, opened 2019-11-15
Fixes fallback implementation of `ArrayTile.combine` for `DelegatingTile`

### LINKED ISSUE #3153: `combine` when implementing a `DelegatingTile` incorrectly handles `NoData`
```scala
// Regular tile
val t1 = ArrayTile(Array.fill(4)(1), 2, 2)
// Tile with single NoData value
val t2 = {
  val d = Array.fill(4)(2)
  d(1) = geotrellis.raster.NODATA
  ArrayTile(d, 2, 2)
}

val d1 = new DelegatingTile {
  override def delegate: Tile = t1
}
val d2 = new DelegatingTile {
  override def delegate: Tile = t2
}
/** Counts the number of non-NoData cells in a tile */
case object CountData {
  def apply(t: Tile) = {
    var count: Long = 0
    t.dualForeach(
      z ⇒ if(isData(z)) count = count + 1
    ) (
      z ⇒ if(isData(z)) count = count + 1
    )
    count
  }
}

// Confirm counts
CountData(t1) should be (4L)
CountData(t2) should be (3L)
CountData(d1) should be (4L)
CountData(d2) should be (3L)

// Standard Add evaluates `x + NoData` as `NoData`
CountData(Add(t1, t2)) should be (3L)
CountData(Add(d1, d2)) should be (3L)
// Is commutative
CountData(Add(t2, t1)) should be (3L)
CountData(Add(d2, d1)) should be (3L)

/** BiasedAdd is evaluates `x + NoData` as `x` */
case object BiasedAdd extends LocalTileBinaryOp {
  def op(z1: Int, z2: Int): Int = z1 + z2
  def op(z1: Double, z2: Double): Double = z1+z2

  def combine(z1: Int, z2: Int): Int =
    if (isNoData(z1) && isNoData(z2)) raster.NODATA
    else if (isNoData(z1))
      z2
    else if (isNoData(z2))
      z1
    else
      op(z1, z2)

  def combine(z1: Double, z2: Double): Double =
    if (isNoData(z1) && isNoData(z2)) raster.doubleNODATA
    else if (isNoData(z1))
      z2
    else if (isNoData(z2))
      z1
    else
      op(z1, z2)
}

// With BiasedAdd, all cells should be data cells
CountData(BiasedAdd(t1, t2)) should be (4L) // <-- passes
CountData(BiasedAdd(d1, d2)) should be (4L) // <-- fails
// Should be commutative.
CountData(BiasedAdd(t2, t1)) should be (4L) // <-- passes
CountData(BiasedAdd(d2, d1)) should be (4L) // <-- fails
```

I think this is a reemergence of #2907, but when `DelegatingTile` is used. Basically, the fallback/default implementation of `ArrayTile.combine`, when given something that's not a recognized type, ends up providing an implementation that makes assumptions about NoData handling that should be deferred to the combine operator.

Also suspicious that these two implementations are so different:

https://github.com/locationtech/geotrellis/blob/c99dc66bdabdf74b8b9a2e874540bfcf7e00fb98/raster/src/main/scala/geotrellis/raster/ArrayTile.scala#L232-L238

https://github.com/locationtech/geotrellis/blob/c99dc66bdabdf74b8b9a2e874540bfcf7e00fb98/raster/src/main/scala/geotrellis/raster/ArrayTile.scala#L283-L284

Looks like I'm the culprit of this code being broken, so planning on doing a PR. One solution is to add yet another switch based on type, but feels like a dead end. On the other hand, I'm loath to change something as fundamental as the `combine` methods without knowing there's extensive test coverage over the data space. @echeipesh @pomadchin @lossyrob: Any advice/warnings there is appreciated.

metasim:
# Overview

As evidenced by the test, `DelegatingTile` is not able to provide commutative assurances with operations using the `combine`/`combineDouble` mechanism.

## Checklist

~- [ ] [docs/CHANGELOG.rst](https://github.com/locationtech/geotrellis/blob/master/CHANGELOG.md) updated, if necessary~
~- [ ] [Module Hierarcy](https://github.com/locationtech/geotrellis/blob/master/docs/guide/module-hierarchy.rst) updated, if necessary~

## Demo

See unit test

## Notes

Closes #3153

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/ArrayTile.scala
+++ b/raster/src/main/scala/geotrellis/raster/ArrayTile.scala
@@ -229,14 +229,7 @@ abstract class ArrayTile extends Tile with Serializable {
       case ct: CroppedTile =>
         ct.combine(this)((z1, z2) => f(z2, z1))
       case t =>
-        this.map((col, row, z) => {
-          if (isNoData(z)) z
-          else {
-            val v = t.get(col, row)
-            if (isNoData(v)) v
-            else f(z, v)
-          }
-        })
+        this.map((col, row, z) => f(z, t.get(col, row)))
   }
 
   /**
--- a/raster/src/test/scala/geotrellis/raster/ArrayTileSpec.scala
+++ b/raster/src/test/scala/geotrellis/raster/ArrayTileSpec.scala
@@ -16,6 +16,8 @@
 
 package geotrellis.raster
 
+import geotrellis.raster.ArrayTileSpec.{BiasedAdd, CountData}
+import geotrellis.raster.mapalgebra.local.{Add, LocalTileBinaryOp}
 import geotrellis.raster.testkit._
 import org.scalatest._
 
@@ -186,18 +188,19 @@ class ArrayTileSpec extends FunSpec
 
       withClue("IntArrayTile with NoData") {
         val at = injectNoData(4)(createConsecutiveTile(10))
-          .convert(DoubleUserDefinedNoDataCellType(98))
+          .convert(IntUserDefinedNoDataCellType(98))
         val twice = at * 2
         val fauxTile = new DelegatingTile {
           override protected def delegate: Tile = at
         }
 
-        assertEqual(at.combine(fauxTile)((z1, z2) => z1 + z2), twice)
+        def add(l: Int, r: Int) = if(isNoData(l) || isNoData(r)) NODATA else l + r
+        assertEqual(at.combine(fauxTile)(add), twice)
       }
 
       withClue("DoubleArrayTile with NoData") {
         val at = injectNoData(4)(createValueTile(10, 10.2))
-          .convert(DoubleUserDefinedNoDataCellType(33.2))
+
         val twice = at * 2
         val fauxTile = new DelegatingTile {
           override protected def delegate: Tile = at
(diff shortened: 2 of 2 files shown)


raster/src/main/scala/geotrellis/raster/ArrayTile.scala
```diff
@@ -229,14 +229,7 @@ abstract class ArrayTile extends Tile with Serializable {
...
       case t =>
-        this.map((col, row, z) => {
-          if (isNoData(z)) z
-          else {
-            val v = t.get(col, row)
-            if (isNoData(v)) v
-            else f(z, v)
-          }
-        })
+        this.map((col, row, z) => f(z, t.get(col, row)))
```
metasim: This is the main fix. Looks like it had already been applied to `combineDouble` (by me); must have been an oversight to not address it here too.

raster/src/test/scala/geotrellis/raster/ArrayTileSpec.scala
```diff
@@ -186,18 +188,19 @@ class ArrayTileSpec extends FunSpec
 
       withClue("IntArrayTile with NoData") {
         val at = injectNoData(4)(createConsecutiveTile(10))
-          .convert(DoubleUserDefinedNoDataCellType(98))
+          .convert(IntUserDefinedNoDataCellType(98))
```
metasim: I think this is a copy/paste mistake... we should be testing integral tiles here.

raster/src/test/scala/geotrellis/raster/ArrayTileSpec.scala
```diff
@@ -231,3 +257,42 @@ class ArrayTileSpec extends FunSpec
...
+      else if (isNoData(z1))
+        z2
+      else if (isNoData(z2))
+        z1
+      else
+        op(z1, z2)
+  }
+
+  /** Counts the number of non-NoData cells in a tile */
+  case object CountData {
```
metasim: Does something like this already exist?

raster/src/test/scala/geotrellis/raster/ArrayTileSpec.scala
```diff
@@ -231,3 +257,42 @@ class ArrayTileSpec extends FunSpec
     }
   }
 }
+
+object ArrayTileSpec {
+  /** BiasedAdd is evaluates `x + NoData` as `x` */
+  case object BiasedAdd extends LocalTileBinaryOp {
```
metasim: This is added as an example where you'd want the `LocalTileBinaryOp` to control the semantics of `NoData`

raster/src/test/scala/geotrellis/raster/ArrayTileSpec.scala
echeipesh: No, in general this is a case of https://github.com/locationtech/geotrellis/issues/779 though

PR #3180, opened 2020-02-05
Add GeoTrellisRasterSource compat with GT 1.x and 2.x layers

### LINKED ISSUE #3179: GeoTrellisRasterSources legacy and temporal layers support
At this point GeoTrellisRasterSource can not read legacy (1.x and 2.x) GeoTrellis layers due to the `Key` class name change (from `geotrellis.spark.SpatialKey` to `geotrellis.layer.SpatialKey`). 

The task is to add a compact / fallback logic to read old layers. 
This is importat to make GeoTrellis-server OGC examples work.

pomadchin:
# Overview

This PR makes possible GeoTrellisRasterSources to work with old GeoTrellis layers

## Checklist

## Demo

https://github.com/geotrellis/geotrellis-server/pull/191

Closes #3179

### DIFF
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -22,6 +22,8 @@ import sbtassembly.AssemblyPlugin.autoImport.{MergeStrategy, assemblyMergeStrate
 import sbtassembly.AssemblyKeys._
 import sbtassembly.{PathList, ShadeRule}
 import com.typesafe.tools.mima.plugin.MimaKeys._
+import de.heikoseeberger.sbtheader.{CommentCreator, CommentStyle, FileType}
+import de.heikoseeberger.sbtheader.HeaderPlugin.autoImport.{HeaderLicense, headerLicense, headerMappings}
 import sbtprotoc.ProtocPlugin.autoImport.PB
 
 object Settings {
@@ -109,6 +111,21 @@ object Settings {
       Settings.Repositories.geosolutions,
       Settings.Repositories.osgeo,
       Settings.Repositories.locationtechReleases
+    ),
+    headerLicense := Some(HeaderLicense.ALv2(java.time.Year.now.getValue.toString, "Azavea")),
+    headerMappings := Map(
+      FileType.scala -> CommentStyle.cStyleBlockComment.copy(commentCreator = new CommentCreator() {
+        val Pattern = "(?s).*?(\\d{4}(-\\d{4})?).*".r
+        def findYear(header: String): Option[String] = header match {
+          case Pattern(years, _) => Some(years)
+          case _                 => None
+        }
+        def apply(text: String, existingText: Option[String]): String = {
+          // preserve year of old headers
+          val newText = CommentStyle.cStyleBlockComment.commentCreator.apply(text, existingText)
+          existingText.flatMap(_ => existingText.map(_.trim)).getOrElse(newText)
+        }
+      })
     )
   )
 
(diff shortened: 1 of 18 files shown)


project/Settings.scala
```diff
@@ -109,6 +111,21 @@ object Settings {
...
+        def findYear(header: String): Option[String] = header match {
+          case Pattern(years, _) => Some(years)
+          case _                 => None
+        }
+        def apply(text: String, existingText: Option[String]): String = {
+          // preserve year of old headers
+          val newText = CommentStyle.cStyleBlockComment.commentCreator.apply(text, existingText)
+          existingText.flatMap(_ => existingText.map(_.trim)).getOrElse(newText)
+        }
+      })
```
echeipesh: Did moving the config here fix some issue with header generation?

store/src/main/scala/geotrellis/store/GeoTrellisRasterSource.scala
```diff
@@ -174,25 +174,43 @@ object GeoTrellisRasterSource {
...
+      (Class.forName(header.keyClass), Class.forName(header.valueClass)) match {
+        case (SpatialKeyClass, TileClass) => spatialTileRead
+        case (SpatialKeyClass, MultibandTileClass) => spatialMultibandTileRead
+        case _ =>
+          throw new Exception(s"Unable to read single or multiband tiles from file: ${(header.keyClass, header.valueClass)}")
+      }
+    } else {
+      /**
+        * Legacy GeoTrellis Layers compact
+        * TODO: remove in GT 4.0
```
echeipesh: Why is this safe to remove in 4.0 ?

project/Settings.scala
pomadchin: @echeipesh yes, it should be set per project and for some reason setting through `ThisBuild` doesn't work. I tried it in geotrellis-pointcloud & geotrellis-server.

store/src/main/scala/geotrellis/store/GeoTrellisRasterSource.scala
pomadchin: @echeipesh mb we can drop old layers? :D

pomadchin:
@echeipesh I created a new issue btw https://github.com/geotrellis/geotrellis-server/issues/219

store/src/main/scala/geotrellis/store/GeoTrellisRasterSource.scala
echeipesh: Seems like an unkind  thing to do if its not buying anybody anything.

project/Settings.scala
echeipesh: 👍 TIL

store/src/main/scala/geotrellis/store/GeoTrellisRasterSource.scala
```diff
@@ -174,25 +174,43 @@ object GeoTrellisRasterSource {
...
-          .withContext(tiles =>
-            tiles.map{ case(key, tile) => (key, tile.subsetBands(bands)) }
-          )
-      case _ =>
-        throw new Exception(s"Unable to read single or multiband tiles from file: ${(header.keyClass, header.valueClass)}")
+
+    if (!header.keyClass.contains("spark")) {
+      (Class.forName(header.keyClass), Class.forName(header.valueClass)) match {
+        case (SpatialKeyClass, TileClass) => spatialTileRead
+        case (SpatialKeyClass, MultibandTileClass) => spatialMultibandTileRead
```
echeipesh: This is mostly style comment that was generated from this line of thinking:

- We get the class class when the class loader is good and match
- If the class is old, we can't get a class class, so we have to check the strings
- And to do that, we have to do a sub-string check
- So wait, we made a class and did a sub-string check to avoid doing a string match?
- Is it slower, I don't know, probably doesn't matter...

TLDR: I think the original formulation is much clearer and potentially saves somebody thinking useless thoughts:

https://github.com/geotrellis/geotrellis-server/blob/89f297c54c12825512716bbcb7c65334a51d4014/core/src/main/scala/geotrellis/store/GeoTrellisRasterSourceLegacy.scala#L237-L244

Disregard this comment if you have a clear preference for doing it this way.

pomadchin:
Replaced by https://github.com/locationtech/geotrellis/pull/3222

PR #3181, opened 2020-02-06
Remove duplicate circe json codec

notthatbreezy:
# Overview

This removes a duplicate codec for `TileLayout` - there is already one [here](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/io/json/Implicits.scala) and it is really easy to get in a situation where both are available and hit an ambiguous implicit error.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix `PolygonRasterizer` failure on some inputs [#3160](https://github.com/locationtech/geotrellis/issues/3160)
 
+### Removed
+
+- Removed duplicate JSON codecs for `TileLayout` [#3181](https://github.com/locationtech/geotrellis/pull/3181)
+- Removed `private` restriction for TiffTag `read` overload [#3181](https://github.com/locationtech/geotrellis/pull/3181)
+
 ## [3.2.0] - 2019-11-19
 
 ### Changed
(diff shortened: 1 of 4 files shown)


raster/src/main/scala/geotrellis/raster/TileLayout.scala
```diff
@@ -46,7 +45,6 @@ object TileLayout {
  * @param tileCols   number of pixel columns in each tile, East to West
  * @param tileRows   number of pixel rows in each tile, North to South
  */
-@JsonCodec
```
pomadchin: Rather than removing codec from the companion object I would prefer deleting codecs from https://github.com/locationtech/geotrellis/blob/7871e015129ae4a1ccfd06a23544a6de9ac68c6e/raster/src/main/scala/geotrellis/raster/io/json/Implicits.scala and https://github.com/locationtech/geotrellis/blob/7eb7958dde643fa4a347c9ac5b055bb333f416fd/spark-pipeline/src/main/scala/geotrellis/spark/pipeline/json/Implicits.scala 

It is just easier to use codecs when they are stored in the companion objects, since you won’t need any extra imports to make circe work.

Can you remove them instead of removing this codecs?

pomadchin:
Ah, it looks like you also need to [sign ECA](https://github.com/locationtech/geotrellis/blob/master/docs/CONTRIBUTING.rst#eclipse-contributor-agreement-eca), could you do it if it is not hard for you? 

If you have a sign ECA you can sign your commit via `git commit --amend -s && git push origin -f eature/cmb/remove-duplicate-json-codec`

notthatbreezy:
@pomadchin Updated PR and signed ECA

I also added a commit that removed the `private` restriction on an overload for reading tiff tags -- this wasn't in previous versions of GT and RF uses this method and another project now.

notthatbreezy:
Changelog updated!

pomadchin:
@notthatbreezy perfect, thanks!

PR #3191, opened 2020-02-25
Fix GeoTrellisPath parsing

CloudNiner:
# Overview

Parsing of url authority was bad. This should fix that. This PR also opened a discussion around whether absolute and relative urls without scheme should be valid as well so there's tests for those here that currently fail. If we want to support this, I can add a fix for that here as well, where scheme-less urls assume `file://`.

Before fix, 7 of new tests fail. After, only the scheme-less tests fail.

## Checklist

- [ ] [Module Hierarcy](https://github.com/locationtech/geotrellis/blob/master/docs/guide/module-hierarchy.rst) updated, if necessary

### DIFF
--- a/store/src/test/scala/geotrellis/store/GeoTrellisPathSpec.scala
+++ b/store/src/test/scala/geotrellis/store/GeoTrellisPathSpec.scala
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2020 Azavea
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package geotrellis.store
+
+import org.scalatest._
+
+class GeoTrellisPathSpec extends FunSpec {
+  describe("GeoTrellisPathSpec") {
+
+    it("should fail to parse without a layer") {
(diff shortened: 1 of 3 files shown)


store/src/test/scala/geotrellis/store/GeoTrellisPathSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    it("should parse a local relative file path with scheme") {
+      val path = GeoTrellisPath.parse("file://relative/path?layer=baz")
+      assert(path.value == "file://relative/path")
+      assert(path.layerName == "baz")
+    }
+
+    it("should parse a local relative file path without scheme") {
+      val path = GeoTrellisPath.parse("relative/path?layer=baz")
```
pomadchin: Lets try to make this behave as a URI with a `file` scheme? If nothing would fail that would be 💯

CloudNiner: 👍 on it

CloudNiner:
With fix applied:
```
scala> val rasterSource = RasterSource("s3://azavea-datahub/catalog?layer=nlcd-2011-epsg3857&zoom=13&band_count=1")
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
log4j:WARN No appenders could be found for logger (org.apache.http.client.protocol.RequestAddCookies).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
rasterSource: geotrellis.raster.RasterSource = GeoTrellisRasterSource(s3://azavea-datahub/catalog, Layer(name = "nlcd-2011-epsg3857", zoom = 13))
scala> rasterSource.extent
res3: geotrellis.vector.Extent = Extent(-1.4497467188222097E7, 2480629.988195896, -7087967.414370252, 6960317.577134008)
```
🎉

Issue #3194, opened 2020-03-03
OverviewStrategy requires improvements


jpolchlo:
After a brief interaction with the [OverviewStrategy](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/io/geotiff/OverviewStrategy.scala) "interface", it appears to me that there are significant shortcomings with this feature.

1. This is not an extensible system for resolution selection.
2. The strategies that exist are hardly comprehensive.
3. The semantics are not set by the interface (the responsibility for interpreting these strategies falls on the RasterSource implementation, so we can have inconsistencies between implementations).

I am proposing to improve this interface in the following way: Remove the `sealed` keyword and add an `apply` method to `OverviewStrategy` which takes in the list of available resolutions, the desired resolution, and returns one of the available resolutions (or the index of the best match).

If we think of the available resolution as "stops" on the number line, then we might have
```
coarse                                                                   fine
<----------*---------*---------------*-----------------------*-------------->
           1         2         ˄     3                       4
```
where the '*'s indicate the available resolutions and the caret indicates the request.

At the moment, we can choose the finest resolution (4) with `Base` (or `Auto(1)`?) or the next highest resolution (3) with `AutoHigherResolution` (or with `Auto(0)`?).  (The semantics of `Auto(n)` are not clear from the docs, hence the question marks.)  There's not currently a way to move to coarser resolutions (due to an inexplicable requirement that `Auto(n)` have n ≥ 0), or to "round" to the closest resolution.

More than anything, the lack of set semantics for these strategies **strongly** suggests that we need to improve this API, lest different users of this interface diverge in behavior.

pomadchin:
Connected to https://github.com/locationtech/geotrellis/issues/3196

echeipesh:
We discussed this issue at length and the consensus is that this would make good improvement. 

The main consequence of changing `OverviewStrategy` to be extensible is that it will no longer be possible for `GDALRasterSource` to operate on all sub-classes of it since there is no way to ensure that each possible subclass maps to an actual GDAL parameter. In those case `GDALRasterSource` will have to throw. This is not ideal but seems "fine".


```scala
// client-code.jar
val source: GDALRasterSource = ???
source.resample(OverviewStrategy.Base) // works

case object WeirdStrategy extends OverviewStrategy {
  override def select(candidates: List[CellSize]): Integer = -1
}

source.resample(WeirdStrategy ) // throws
```

[...]

PR #3200, opened 2020-03-10
Fix gdal-warp -ovr ports

### LINKED ISSUE #3196: Conform to GDAL Warp -ovr Overview Selection Options
The `OverviewStrategy` trait and subclasses (https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/io/geotiff/OverviewStrategy.scala) exist to capture the gdal warp command line `-ovr` options (https://gdal.org/programs/gdalwarp.html#cmdoption-gdalwarp-ovr)

We currently represent `Auto-N` and `None` (`Base` as we've named things) but do not represent explicit overview level selection *or* for that matter, the default GDAL behavior of `Auto` (`AutoHigher` behaves slightly differently). The `OverviewStrategy` types should work as they are described in GDAL docs

moradology:
# Overview

This PR exposes the overview selection logic implicit in each of the `OverviewStrategy` instances by way of a `selectOverview` method. Now, instead of matching and requiring users of the sum type to implement logic themselves, a call to `selectOverview` with the available `CellSizes` and the desired `CellSize` will yield a selection of the `CellSize` corresponding to the selected `Overview`

## Checklist

## Notes

Closes #3196

### DIFF
--- a/gdal/src/main/scala/geotrellis/raster/gdal/GDALUtils.scala
+++ b/gdal/src/main/scala/geotrellis/raster/gdal/GDALUtils.scala
@@ -20,7 +20,12 @@ import geotrellis.raster._
 import geotrellis.raster.io.geotiff._
 import geotrellis.raster.resample._
 
+import org.log4s._
+
 object GDALUtils {
+
+  @transient private[this] lazy val logger = getLogger
+
   def deriveResampleMethodString(method: ResampleMethod): String =
     method match {
       case NearestNeighbor => "near"
@@ -100,8 +105,12 @@ object GDALUtils {
     }
 
   def deriveOverviewStrategyString(strategy: OverviewStrategy): String = strategy match {
+    case Auto(n) if (n == 0) => "AUTO"
     case Auto(n) => s"AUTO-$n"
-    case AutoHigherResolution => "AUTO"
+    case Level(level) => s"$level"
     case Base => "NONE"
+    case other =>
+      logger.warn(s"$other is not a valid GDALWarp -ovr argument; falling back to AUTO")
+      "AUTO"
   }
 }
--- a/layer/src/test/scala/geotrellis/raster/geotiff/GeoTiffReprojectRasterSourceSpec.scala
+++ b/layer/src/test/scala/geotrellis/raster/geotiff/GeoTiffReprojectRasterSourceSpec.scala
@@ -67,7 +67,7 @@ class GeoTiffReprojectRasterSourceSpec extends FunSpec with RasterMatchers with 
       }
 
       val twiceFuzzySource = rasterSource.reprojectToGrid(LatLng, twiceFuzzyLayout).asInstanceOf[GeoTiffReprojectRasterSource]
-      twiceFuzzySource.closestTiffOverview.cellSize shouldBe CellSize(20,20)
+      twiceFuzzySource.closestTiffOverview.cellSize shouldBe CellSize(20, 20)
 
       val thriceFuzzyLayout = {
         val CellSize(width, height) = baseReproject.cellSize
@@ -75,7 +75,8 @@ class GeoTiffReprojectRasterSourceSpec extends FunSpec with RasterMatchers with 
       }
 
       val thriceFuzzySource = rasterSource.reprojectToGrid(LatLng, thriceFuzzyLayout).asInstanceOf[GeoTiffReprojectRasterSource]
-      thriceFuzzySource.closestTiffOverview.cellSize shouldBe CellSize(20,20)
+      thriceFuzzySource.closestTiffOverview.cellSize.width shouldBe 40.0 +- 0.1
+      thriceFuzzySource.closestTiffOverview.cellSize.height shouldBe 40.0 +- 0.1
 
       val quatroFuzzyLayout = {
         val CellSize(width, height) = baseReproject.cellSize
@@ -83,6 +84,7 @@ class GeoTiffReprojectRasterSourceSpec extends FunSpec with RasterMatchers with 
       }
 
       val quatroTimesFuzzySource = rasterSource.reprojectToGrid(LatLng, quatroFuzzyLayout).asInstanceOf[GeoTiffReprojectRasterSource]
+      println("TARGET", quatroFuzzyLayout.cellSize)
       quatroTimesFuzzySource.closestTiffOverview.cellSize shouldBe CellSize(40.0,39.94082840236686)
 
     }
--- a/layer/src/test/scala/geotrellis/raster/geotiff/GeoTiffReprojectRasterSourceSpec.scala
+++ b/layer/src/test/scala/geotrellis/raster/geotiff/GeoTiffReprojectRasterSourceSpec.scala
@@ -22,6 +22,8 @@ import geotrellis.raster._
 import geotrellis.raster.io.geotiff.reader.GeoTiffReader
 import geotrellis.raster.reproject._
 import geotrellis.raster.testkit.RasterMatchers
+import geotrellis.raster.resample.NearestNeighbor
+import geotrellis.raster.io.geotiff.AutoHigherResolution
 
 import org.scalatest._
 
@@ -74,9 +76,12 @@ class GeoTiffReprojectRasterSourceSpec extends FunSpec with RasterMatchers with 
         LayoutDefinition(RasterExtent(LatLng.worldExtent, CellSize(width*3.5, height*3.5)), tileSize = 256)
       }
 
-      val thriceFuzzySource = rasterSource.reprojectToGrid(LatLng, thriceFuzzyLayout).asInstanceOf[GeoTiffReprojectRasterSource]
-      thriceFuzzySource.closestTiffOverview.cellSize.width shouldBe 40.0 +- 0.1
-      thriceFuzzySource.closestTiffOverview.cellSize.height shouldBe 40.0 +- 0.1
+      val thriceFuzzySourceAutoHigher = rasterSource.reprojectToGrid(LatLng, thriceFuzzyLayout, NearestNeighbor, AutoHigherResolution).asInstanceOf[GeoTiffReprojectRasterSource]
+      thriceFuzzySourceAutoHigher.closestTiffOverview.cellSize shouldBe CellSize(20, 20)
+
+      val thriceFuzzySourceAuto = rasterSource.reprojectToGrid(LatLng, thriceFuzzyLayout).asInstanceOf[GeoTiffReprojectRasterSource]
+      thriceFuzzySourceAuto.closestTiffOverview.cellSize.width shouldBe 40.0 +- 0.1
+      thriceFuzzySourceAuto.closestTiffOverview.cellSize.height shouldBe 40.0 +- 0.1
 
       val quatroFuzzyLayout = {
         val CellSize(width, height) = baseReproject.cellSize
@@ -84,7 +89,6 @@ class GeoTiffReprojectRasterSourceSpec extends FunSpec with RasterMatchers with 
       }
 
       val quatroTimesFuzzySource = rasterSource.reprojectToGrid(LatLng, quatroFuzzyLayout).asInstanceOf[GeoTiffReprojectRasterSource]
-      println("TARGET", quatroFuzzyLayout.cellSize)
       quatroTimesFuzzySource.closestTiffOverview.cellSize shouldBe CellSize(40.0,39.94082840236686)
 
     }
--- a/layer/src/test/scala/geotrellis/raster/geotiff/GeoTiffReprojectRasterSourceSpec.scala
+++ b/layer/src/test/scala/geotrellis/raster/geotiff/GeoTiffReprojectRasterSourceSpec.scala
@@ -22,8 +22,6 @@ import geotrellis.raster._
 import geotrellis.raster.io.geotiff.reader.GeoTiffReader
 import geotrellis.raster.reproject._
 import geotrellis.raster.testkit.RasterMatchers
-import geotrellis.raster.resample.NearestNeighbor
-import geotrellis.raster.io.geotiff.AutoHigherResolution
 
 import org.scalatest._
 
@@ -76,12 +74,8 @@ class GeoTiffReprojectRasterSourceSpec extends FunSpec with RasterMatchers with 
         LayoutDefinition(RasterExtent(LatLng.worldExtent, CellSize(width*3.5, height*3.5)), tileSize = 256)
       }
 
-      val thriceFuzzySourceAutoHigher = rasterSource.reprojectToGrid(LatLng, thriceFuzzyLayout, NearestNeighbor, AutoHigherResolution).asInstanceOf[GeoTiffReprojectRasterSource]
+      val thriceFuzzySourceAutoHigher = rasterSource.reprojectToGrid(LatLng, thriceFuzzyLayout).asInstanceOf[GeoTiffReprojectRasterSource]
       thriceFuzzySourceAutoHigher.closestTiffOverview.cellSize shouldBe CellSize(20, 20)
-
-      val thriceFuzzySourceAuto = rasterSource.reprojectToGrid(LatLng, thriceFuzzyLayout).asInstanceOf[GeoTiffReprojectRasterSource]
-      thriceFuzzySourceAuto.closestTiffOverview.cellSize.width shouldBe 40.0 +- 0.1
-      thriceFuzzySourceAuto.closestTiffOverview.cellSize.height shouldBe 40.0 +- 0.1
 
       val quatroFuzzyLayout = {
         val CellSize(width, height) = baseReproject.cellSize
--- a/raster/src/main/scala/geotrellis/raster/io/geotiff/OverviewStrategy.scala
+++ b/raster/src/main/scala/geotrellis/raster/io/geotiff/OverviewStrategy.scala
@@ -18,7 +18,11 @@ package geotrellis.raster.io.geotiff
 
 import geotrellis.raster.CellSize
 
-import scala.collection.Searching._
+import cats.syntax.foldable._
+import cats.syntax.option._
+import cats.instances.list._
+import cats.instances.either._
+import org.log4s.getLogger
 
 /**
   * Options ported from GDAL (https://gdal.org/programs/gdalwarp.html#cmdoption-gdalwarp-ovr)
@@ -27,19 +31,27 @@ import scala.collection.Searching._
 sealed trait OverviewStrategy
 
 object OverviewStrategy {
-  def DEFAULT = AutoHigherResolution
+  @transient private[this] lazy val logger = getLogger
+
+  def DEFAULT: OverviewStrategy = AutoHigherResolution
 
   /**
     * Select appropriate overview given the strategy.
     *
     * WARN: this function assumes that CellSizes are sorted. It interprets idx 0 as the position with the highest CellSize.
+    * If the input list is not sorted, this function returns idx 0.
     *
     * Unless a particular strategy suggests otherwise, this method will clamp the returned
     * index to the range of overviewCS.
-    * @param overviewCS
-    * @param desiredCS
-    * @param strategy
-    * @return
+    * @param overviewCS a sorted list of resolutions, otherwise it is not guaranteed that it would work with all the input strategies.
+    * @param desiredCS  cellSize that would be searched
+    * @param strategy   overview strategy that would be used to search for the best matching CellSize
+    * @return           index of the closest cellSize. Each strategy behaves a bit different:
+    *                   Level: selects the passed index and if it is OOB clamps the index.
+    *                   Auto(n): selects the best matching cellSize (approximately) and adds n, requires the input resolutions list to be sorted.
+    *                   Base: always returns the best matching cellSize from the given list of CellSizes.
+    *                   AutoHigherResolution: selects the best matching cellSize (approximately) always selects a higher resolution,
+    *                   requires the input resolutions list to be sorted.
     */
   def selectOverview(
     overviewCS: List[CellSize],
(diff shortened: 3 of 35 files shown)


gdal/src/main/scala/geotrellis/raster/gdal/GDALUtils.scala
```diff
@@ -100,8 +100,11 @@ object GDALUtils {
     }
 
   def deriveOverviewStrategyString(strategy: OverviewStrategy): String = strategy match {
+    case Auto(n) if (n == 0) => "AUTO"
     case Auto(n) => s"AUTO-$n"
-    case AutoHigherResolution => "AUTO"
+    case Level(level) => s"$level"
     case Base => "NONE"
+    case other =>
```
pomadchin: This would never happen, it is an exhaustive match on a sealed trait.

raster/src/main/scala/geotrellis/raster/io/geotiff/OverviewStrategy.scala
```diff
@@ -16,21 +16,78 @@
...
+import geotrellis.raster.CellSize
+
+import scala.collection.Searching._
+
+/**
+  * Options ported from GDAL (https://gdal.org/programs/gdalwarp.html#cmdoption-gdalwarp-ovr)
+  *  for automatic selection from among available overviews
+  */
+sealed trait OverviewStrategy {
+  def selectOverview(overviewCS: List[CellSize], desiredCS: CellSize): CellSize
```
pomadchin: Hm, we don't need this function in case of GDAL; do you think this is an appropriate place for selecting overview? 

This function is specific to the interpretation of the `OverviewStrategy` AST, and it is not a member of an AST that will work for all interpreters the same way. 

i.e. `GDALUtils.deriveOverviewStrategyString` lives in its own object, which is an interpreter for this AST.

What do you think about moving it in a separate object since it would be used only with GeoTrellis GeoTiffs?

raster/src/test/scala/geotrellis/raster/io/geotiff/OverviewStrategySpec.scala
```diff
@@ -0,0 +1,59 @@
+package geotrellis.raster.io.geotiff
```
pomadchin: could you generate a header for tests? via `test:headerCreate`

raster/src/test/scala/geotrellis/raster/io/geotiff/OverviewStrategySpec.scala
```diff
@@ -0,0 +1,59 @@
...
+
+import geotrellis.raster._
+
+import org.scalatest._
+
+class OverviewStrategySpec extends FunSpec with Matchers {
```
pomadchin: +1000 for Specs :+1:

gdal/src/main/scala/geotrellis/raster/gdal/GDALUtils.scala
moradology: We still have `AutoHigher`, but that has no GDAL string interpretation which makes sense. Is your proposal that it be removed so that this check is once again exhaustive?

pomadchin: Ah I forgot about that; :+1: than.

raster/src/main/scala/geotrellis/raster/io/geotiff/OverviewStrategy.scala
moradology: So far, the overview selection I've seen is pretty consistently be expressible in terms of the `selectOverview` method and this way of organizing things keeps the AST and the info necessary to make sense of it together + avoids duplication of some case-handling logic. I'm not opposed to our handling the overview selection case in a more general/appropriate fashion and therefore hiding this away for occasions where GDAL interop are required, but that seems like a different PR with a larger footprint.

gdal/src/main/scala/geotrellis/raster/gdal/GDALUtils.scala
pomadchin: Hm, do we need to throw, or warning + a fallback behavior? (default GDALWapr)

raster/src/main/scala/geotrellis/raster/io/geotiff/OverviewStrategy.scala
pomadchin: Hm, Is it correct to consider GDAL an "interop" thing? It could be an acceptable implementation if we were not using the same AST for GDAL. Moving all the specific to interpreter thing into interpreter looks pretty convenient btw (throwing here the RasterSources Query DSL). 

I'm happy to discuss it though.

raster/src/test/scala/geotrellis/raster/io/geotiff/OverviewStrategySpec.scala
pomadchin: It is broken now, created an issue https://github.com/locationtech/geotrellis/issues/3201

raster/src/main/scala/geotrellis/raster/io/geotiff/OverviewStrategy.scala
echeipesh: I think the point is somewhat academic and I'm not sure I'd have regrets in my life with either choice.

However, if I had to pick one I'd go with moving the `selectOverview` function to be on the companion object of`OverviewStrategy`. I have two soft reasons for feeling that way:

As @pomadchin mentions the purpose of this ADT is to select behavior. At the moment indication this selection to our GDAL bindings is "first class" intention of this ADT. Having a method that appears to do that on the instances of the ADT is slightly misleading because it could never be used by GDAL bindings.

This also reminds me of similar design choice made in [`ResampleTarget`](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/ResampleTarget.scala) where the `apply` function does the job of creating the target `GridExtent`. I had to write some code that interpreted it in terms of old `Reproject.Options` and I found it hard to think through correctness there because trying to fit in the use of `apply` seemed it was a good thing. In the end I had to write interpretations of `ResampleTarget` over `Reproject.Options` that ignored the implementations of the `apply` method:

https://github.com/locationtech/geotrellis/blob/3e59f26807fc9d3a2dbf47fe358a1f1dbd2205ae/raster/src/main/scala/geotrellis/raster/ResampleTarget.scala#L87-L96

So in my mind that's the case where putting methods on ADT made it worse for interpreting the semantics between two similar domains for that ADT.

So perhaps the code reuse could be served by:

```scala
// before
val sourceCS = strategy.selectOverview(overviews.map(_.cellSize), cellSize)

// after
val sourceCS = OverviewStrategy.select(overviews.map(_.cellSize), cellSize, strategy)
```

Or even perhaps:
```scala
object OverviewStrategy {
   def select[A](overviews: A, desiredCS: CellSize, strategy: OverviewStrategy)(implicit ev: A => CellSize): A = {...}
}
```

echeipesh: Reading between the lines I think this is going in the direction where we end up with two ways of selecting overviews: 

1. `GDALOverviewStrategy` which is basically what we have now
2. `OverviewStrategy` is that is JVM first in lines with #3194 (non-sealed with interface)

And there is a partial function `OverviewStrategy => GDALOverviewStrategy` that is used by `GDALRasterSource`.

moradology: "Interop" might not be the right word choice In general, I like the interpreter pattern a lot but I'm not sure I see the appeal of separating out implementation from the 'language' (which in this case is just selection from among a few overview strategies) but here I don't fully see the appeal. I think that the AST's meaning is better documented to the extent that you can track down the code which will be called in case that option is supplied.

I think we might want more opinions to weigh in

moradology: Having discussed this at length with @echeipesh and @pomadchin, I've removed `OverviewStrategy` methods in favor of a function on the companion object.

gdal/src/main/scala/geotrellis/raster/gdal/GDALUtils.scala
moradology: I've gone with your suggestion of falling back on a sensible default and warning

raster/src/main/scala/geotrellis/raster/io/geotiff/OverviewStrategy.scala
```diff
@@ -16,20 +16,75 @@
...
+    strategy: OverviewStrategy
+  ): Int = strategy match {
+    case Level(overviewIdx) =>
+      overviewIdx
+    case Auto(n) =>
+      selectIndexByProximity(overviewCS, desiredCS, 0.5) + n
+    case Base =>
+      overviewCS.indexOf(overviewCS.min)
+    case AutoHigherResolution =>
+      selectIndexByProximity(overviewCS, desiredCS, 1.0)
```
pomadchin: Waiting for Travis to be happy with this change. This behavior should match to the tested old behavior (the old one had a long testing path).

layer/src/test/scala/geotrellis/raster/geotiff/GeoTiffReprojectRasterSourceSpec.scala
```diff
@@ -59,22 +59,24 @@ class GeoTiffReprojectRasterSourceSpec extends FunSpec with RasterMatchers with
...
 
       val thriceFuzzyLayout = {
         val CellSize(width, height) = baseReproject.cellSize
         LayoutDefinition(RasterExtent(LatLng.worldExtent, CellSize(width*3.5, height*3.5)), tileSize = 256)
       }
 
       val thriceFuzzySource = rasterSource.reprojectToGrid(LatLng, thriceFuzzyLayout).asInstanceOf[GeoTiffReprojectRasterSource]
-      thriceFuzzySource.closestTiffOverview.cellSize shouldBe CellSize(20,20)
+      thriceFuzzySource.closestTiffOverview.cellSize.width shouldBe 40.0 +- 0.1
+      thriceFuzzySource.closestTiffOverview.cellSize.height shouldBe 40.0 +- 0.1
```
moradology: This change is necessary as the former expectation (always round up to higher resolution levels) was baked in and doesn't correctly track GDAL behavior

raster/src/main/scala/geotrellis/raster/io/geotiff/OverviewStrategy.scala
moradology: The tests turned out to require a bit more attention than I'd anticipated as some of them baked in the `AutoHigher` behavior rather than `Auto`

layer/src/test/scala/geotrellis/raster/geotiff/GeoTiffReprojectRasterSourceSpec.scala
pomadchin: These are AutoHigher tests, let's keep them working with AutoHigher strategy? And this one cxan be a new test (this is just a sainity check that we could preserve the old behavior), how do you feel about it?

layer/src/test/scala/geotrellis/raster/geotiff/GeoTiffReprojectRasterSourceSpec.scala
```diff
@@ -59,22 +59,24 @@ class GeoTiffReprojectRasterSourceSpec extends FunSpec with RasterMatchers with
...
 
       val quatroFuzzyLayout = {
         val CellSize(width, height) = baseReproject.cellSize
         LayoutDefinition(RasterExtent(LatLng.worldExtent, CellSize(width*4.1, height*4.1)), tileSize = 256)
       }
 
       val quatroTimesFuzzySource = rasterSource.reprojectToGrid(LatLng, quatroFuzzyLayout).asInstanceOf[GeoTiffReprojectRasterSource]
+      println("TARGET", quatroFuzzyLayout.cellSize)
```
pomadchin: Probably should not be there

moradology: Well, that's a bit sloppy.

store/src/main/scala/geotrellis/store/GeoTrellisRasterSource.scala
```diff
@@ -139,25 +139,12 @@ object GeoTrellisRasterSource {
...
-          .headOption
-          .orElse(maxResultion)
-      case Auto(n) =>
-        val sorted = grids.sortBy(v => math.abs(cellSize.resolution - f(v).resolution))
-        sorted.lift(n).orElse(sorted.lastOption) // n can be out of bounds,
-      // makes only overview lookup as overview position is important
-      case Base => maxResultion
-    }
+    strategy: OverviewStrategy = OverviewStrategy.DEFAULT
+  )(implicit f: T => CellSize): T = {
```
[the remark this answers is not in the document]
CloudNiner: Ah, should have mentioned this sooner. 

It was Option[Int] because some operations could return an out of bounds index.  It's not clear how GDAL handles this situation, so we decided that it would be simpler if OverviewStrategy.selectOverview always returned an in bounds index. Which it now does due to additional error handling there. Before this change, we were doing an `option.get` everywhere which just pushes the throw further down. In that case, I wasn't sure how we would even handle getting a bad overview. 

I can add some additional docstrings to note that selectOverview should always return a clamped index, this was apparently the intent of the code based [on a test](https://github.com/locationtech/geotrellis/pull/3200/files#diff-28fce00d2a6d9757847488e95fe46c87R194) that predates me picking this task up.

spark/src/test/scala/geotrellis/store/GeoTrellisRasterSourceSpec.scala
```diff
@@ -95,11 +95,14 @@ class GeoTrellisRasterSourceSpec extends FunSpec with RasterMatchers with GivenW
...
-          // resample to 0.9 so RasterSource picks the base layer and not an overview
+          .resample(sourceMultiband.cols.toInt, sourceMultiband.rows.toInt, NearestNeighbor)
 
-      val resampledSource =
-        sourceMultiband.resample(expected.tile.cols, expected.tile.rows, NearestNeighbor)
+      val resampledSource = sourceMultiband.resample(
+        expected.tile.cols,
+        expected.tile.rows,
+        NearestNeighbor,
+        strategy = Base
```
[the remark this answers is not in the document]
CloudNiner: It was. When Nathan and I dug into this, we'd always get a raster at the base cell size. So it seems more correct to remove the fudge factor that attempts to get a non-base resolution and just have the resample operation pick the Base overview explicitly. 

Similar issue for the reproject test. When we load a raster with the GeoTiff method and then use reproject instead of resample method, that doesn't allow us to select an overview strategy since reproject doesn't care about that fundamentally. So again, we just use Base where necessary to preserve the idea that when we load the expected raster we're always dealing with the base resolution.

Hopefully that makes sense, we can discuss offline if more detail is necessary.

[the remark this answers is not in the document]
CloudNiner: I did revert to using AutoHigher in the RasterSourceRDD specs, since in those cases we load both in a way that we're resampling anyways.

store/src/main/scala/geotrellis/store/GeoTrellisRasterSource.scala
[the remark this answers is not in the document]
pomadchin: When it can’t fine an overview the idea was to use the Base layer, since it looks like what GDAL does.

spark/src/test/scala/geotrellis/store/GeoTrellisRasterSourceSpec.scala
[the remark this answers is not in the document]
pomadchin: @CloudNiner i would like to revert all the strategies to AutoHigher; to preserve the old tests. This strategy is pretty important for RF and if we need more tests, lest create new tests without killing the old ones; how do you feel about that?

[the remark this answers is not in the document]
pomadchin: I mean just as a check that the behavior of old functions did not change; if we’ll change tests, it is hard to say, was the old behavior preserved or not.

spark/src/test/scala/geotrellis/spark/RasterSourceRDDSpec.scala
```diff
@@ -141,8 +140,9 @@ class RasterSourceRDDSpec extends FunSpec with TestEnvironment with RasterMatche
     }
 
     it("should reproduce tileToLayout followed by reproject") {
+      val reprojectedRasterSource = rasterSource.reprojectToGrid(targetCRS, layout, strategy = Base)
       val reprojectedSourceRDD: MultibandTileLayerRDD[SpatialKey] =
-        RasterSourceRDD.spatial(rasterSource.reprojectToGrid(targetCRS, layout), layout)
+        RasterSourceRDD.spatial(Seq(reprojectedRasterSource), layout, strategy = Base)
```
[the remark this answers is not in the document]
CloudNiner: Yeah reverted rastersourcerdd to AutoHigher

spark/src/test/scala/geotrellis/store/GeoTrellisRasterSourceSpec.scala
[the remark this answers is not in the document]
CloudNiner: Then theres a chunk of extra work here to get this build green, as I'm not sure how to get GeoTiffReader.readMultiband to be explicit about not choosing the Base cell size for its operations. It might be best if someone else takes that work on at this point.

store/src/main/scala/geotrellis/store/GeoTrellisRasterSource.scala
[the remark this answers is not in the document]
CloudNiner: How does that reconcile with the idea that we've decided to include the base layer in the resolutions list in all RasterSource implementations?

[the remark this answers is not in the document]
CloudNiner: Perhaps this discussion should be moved offline as well?

spark/src/test/scala/geotrellis/store/GeoTrellisRasterSourceSpec.scala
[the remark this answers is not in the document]
pomadchin: @CloudNiner how is that possible that old operations can't work as they worked before with AutoHigher? What has changed?

store/src/main/scala/geotrellis/store/GeoTrellisRasterSource.scala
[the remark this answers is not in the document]
CloudNiner: After a discussion we decided on the following:
- Revert OverviewStrategy.DEFAULT = AutoHigherResolution to make the default consistent with the old behavior for now. We'll address changing the default at a later date
- Update AutoHigherResolution implementation to match the docstring and old behavior, and verify it with tests
- OverviewStrategy.selectOverview still clamps to available indices. Each overview strategy can perform custom clamping behavior, like AutoHigher above, but in the absence of that, the index will be clamped to the nearest valid value, e.g. -1 -> 0 or list.size + 5 -> list.size - 1
- Revert tests where possible to use the AutoHigher default in order to appropriately test the old behavior.

spark/src/test/scala/geotrellis/store/GeoTrellisRasterSourceSpec.scala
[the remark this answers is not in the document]
CloudNiner: After a discussion and the adjustments made that were discussed in https://github.com/locationtech/geotrellis/pull/3200#discussion_r401672704, @pomadchin is going to dig deeper as to why a number of tests are still failing.

raster/src/main/scala/geotrellis/raster/io/geotiff/OverviewStrategy.scala
```diff
@@ -16,20 +16,95 @@
...
+    }
+    if (maybeIndex < 0) 0
+    else if (maybeIndex >= overviewCS.size) overviewCS.size - 1
+    else maybeIndex
+  }
+
+  // This method is unsafe. It's up to the selectOverview method to handle each index OOB error for
+  // each indivdual strategy
+  private def selectIndexByProximity(overviewCS: List[CellSize], desiredCS: CellSize, proximityThreshold: Double): Int =
+    overviewCS.search(desiredCS) match {
```
pomadchin, line 68: I guess this could be replaed with a [foldM](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Foldable.scala#L384) function

raster/src/main/scala/geotrellis/raster/io/geotiff/OverviewStrategy.scala
```diff
@@ -16,20 +16,97 @@
...
+  */
 sealed trait OverviewStrategy
+
+object OverviewStrategy {
+  def DEFAULT = AutoHigherResolution
+
+  /**
+    * Select appropriate overview given the strategy.
+    *
+    * WARN: this function assumes that CellSizes are sorted. It interprets idx 0 as the position with the highest CellSize.
```
pomadchin, line 35: Taking into the account that this is true, usage of the search function in the `selectIndexByProximity` function makes it really hard to interpret.

raster/src/main/scala/geotrellis/raster/io/geotiff/OverviewStrategy.scala
```diff
@@ -16,20 +16,97 @@
...
+        overviewIdx
+      case Auto(n) =>
+        selectIndexByProximity(overviewCS, desiredCS, 0.5) + n
+      case Base =>
+        overviewCS.indexOf(overviewCS.min)
+      case AutoHigherResolution =>
+        val idx = selectIndexByProximity(overviewCS, desiredCS, 1.0)
+        // AutoHigherResolution defaults to the closest level
+        // idx > overviewCS.size means, that idx = overviewCS.size (last idx + 1)
+        if (idx > overviewCS.size) idx - 1 else idx
```
pomadchin, line 60: We need to know where is the highest resolution to interpret the index the `selectIndexByProximity` function. If the input list is sorted, `idx > overviewCS.size` means the lower resolution. In case it is reversed, it means the higher resolution. If the list is not sorted, this index makes no sense for its interpretation.

raster/src/main/scala/geotrellis/raster/io/geotiff/OverviewStrategy.scala
```diff
@@ -16,20 +16,122 @@
...
+  
+  /**
+    * This method is unsafe. It's up to the selectOverview method to handle each index OOB error.
+    * @param overviewCS        a sorted list of resolutions, otherwise it is not guaranteed that it would work with all the input strategies.
+    * @param desiredCS          cellSize that would be searched  
+    * @param proximityThreshold threshold to defined the search proximity
+    * @return                   index of the closest cellSize. Returns -1 in case nothing was found or the input list was not sorted.  
+    */
+  private def selectIndexByProximity(overviewCS: List[CellSize], desiredCS: CellSize, proximityThreshold: Double): Int =
+    overviewCS.zipWithIndex.foldLeftM(Option.empty[(CellSize, Int)]) { case (acc, r @ (rightCZ, _)) =>
```
pomadchin, line 87: I am not sure that with folding it is much easier to read, however it is more explicit about sorting.

raster/src/main/scala/geotrellis/raster/io/geotiff/OverviewStrategy.scala
```diff
@@ -16,20 +16,122 @@
...
+      acc match {
+        case l @ Some((leftCZ, _)) =>
+          val proportion = 1 - (rightCZ.resolution - desiredCS.resolution) / (rightCZ.resolution - leftCZ.resolution)
+
+          // if the rightCZ > than the leftCZ it means that the list is not sorted
+          // - negative if rightCZ < leftCZ
+          // - positive if rightCZ > leftCZ
+          // - zero otherwise (if rightCZ == leftCZ)
+          if (Ordering[CellSize].compare(rightCZ, leftCZ) < 0) {
```
pomadchin, line 96: short circuiting if the list is not sorted

PR #3211, opened 2020-03-26
Make GDALRasterSource behavior consistent with other RasterSources

### LINKED ISSUE #3210: RasterSources resolutions should be consistent across implementations
At this point `GeoTiffRasterSource.resolutions` and `GeoTrellisRasterSource.resolutions` functions return the base overview `cellSize` as well. 

The task is to double check that the behavior across all the RasterSources implementation is the same and returns only _overviews_ resolutions / contain the base resolution as well.

UPD: resolutions is an external API, and [you can find its usage in the `geotrellis-server`](https://github.com/geotrellis/geotrellis-server/search?q=resolutions&unscoped_q=resolutions) where resolutions are used to walks through all the supported resolutions by the overview, and in this usecase it is really convenient to have the base reslution in this list.

### LINKED ISSUE #3212: Fix Jackson dependencies
At this point we override deps using [dependeciesOverride](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html). 

This can be inconvinet for users, since it is required for all projects that depend on geotrellis-s3 to override jackson in a similar way. 

The solution is to exclude jackson from the s3 project and include the only one of compatible version.

pomadchin:
# Overview

This PR: 

- [x] Adds the base cellSize to the list of resolutions to behave similar to other RasterSources.
- [x] Bumps GDALWarp version up to 1.0.0
- [x] Fixes Jackson dependencies

This issues makes `GDALRasterSource`, `GeoTiffRasterSource` and `DEMRasterSource` to behave more consistent.

## Checklist

Related to #3210
Closes #3212

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- RasterSources resolutions should be consistent across implementations [#3210](https://github.com/locationtech/geotrellis/issues/3210)
+- Bump gdal-warp-bindings version up to 1.0.0 [#3211](https://github.com/locationtech/geotrellis/pull/3211)
+- Fixed GDALRasterSource.resample method behavior to respect the passed resampleMethod [#3211](https://github.com/locationtech/geotrellis/pull/3211)
+
+### Changed
+
 - Fix `PolygonRasterizer` failure on some inputs [#3160](https://github.com/locationtech/geotrellis/issues/3160)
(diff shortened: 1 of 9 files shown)


pomadchin:
@echeipesh at least GeoTiffRasterSource and GDALRasterSources behave the same way. I'm wondering about the GeoTrellisRasterSource, it seems to me that instead of a base cellSize it can use _the current_.

pomadchin:
Will merge this issue once travis is happy. 

The further discussion and thoughts that I came up with are here https://github.com/locationtech/geotrellis/issues/3210#issuecomment-604715291

PR #3224, opened 2020-04-15
Add a ZSpaceTimeLargeKeyIndex into specs

pomadchin:
# Overview

This PR proofs that https://github.com/geotrellis/geotrellis-server/issues/248 can be addressed by defining a different KeyIndexMethod.


pomadchin:
This is only a POC and we decided not to include it now. However it still can be used as an example of how to define user defined indicies to workaround some of the Z3 limitations.

PR #3240, opened 2020-05-13
GeoTrellisRasterSource should return None on empty reads

pomadchin:
# Overview

This PR removes a duplicate `GeoTrellisRasterSource.sparseStitch` function that also was not safe enough and could work (incorrectly) with empty inputs.

## Checklist

## Notes

This bug was revelaed during [this issue](https://github.com/geotrellis/geotrellis-server/issues/258) investigation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix S3RDDLayerReader partitioning [#3231](https://github.com/locationtech/geotrellis/pull/3231)
 - GDALRasterSource works inconsistenly with BitCellType and ByteCellType [#3232](https://github.com/locationtech/geotrellis/issues/3232)
 - rasterizeWithValue accepts only topologically valid polygons [#3236](https://github.com/locationtech/geotrellis/pull/3236)
-- Rasterizer.rasterize should be consistent with rasterizeWithValue [#3238](https://github.com/locationtech/geotrellis/pull/3238) 
+- Rasterizer.rasterize should be consistent with rasterizeWithValue [#3238](https://github.com/locationtech/geotrellis/pull/3238)
+- GeoTrellisRasterSource should return None on empty reads [#3240](https://github.com/locationtech/geotrellis/pull/3240) 
 
 ## [3.3.0] - 2020-04-07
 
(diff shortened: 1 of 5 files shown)


store/src/main/scala/geotrellis/store/GeoTrellisRasterSource.scala
```diff
@@ -258,43 +258,16 @@ object GeoTrellisRasterSource {
...
 
-  /**
-    *  The stitch method in gtcore is unable to handle missing spatialkeys correctly.
-    *  This method works around that problem by attempting to infer any missing tiles
-    **/
-  def sparseStitch(
-    tiles: Seq[(SpatialKey, MultibandTile)] with Metadata[TileLayerMetadata[SpatialKey]],
-    extent: Extent
-  ): Option[Raster[MultibandTile]] = {
-    val md = tiles.metadata
```
pomadchin, line 279: There were no `tiles.isEmpty` check, and anyway we don't need this function anymore, see https://github.com/locationtech/geotrellis/issues/2903

CHANGELOG.md
```diff
@@ -15,7 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix S3RDDLayerReader partitioning [#3231](https://github.com/locationtech/geotrellis/pull/3231)
 - GDALRasterSource works inconsistenly with BitCellType and ByteCellType [#3232](https://github.com/locationtech/geotrellis/issues/3232)
 - rasterizeWithValue accepts only topologically valid polygons [#3236](https://github.com/locationtech/geotrellis/pull/3236)
-- Rasterizer.rasterize should be consistent with rasterizeWithValue [#3238](https://github.com/locationtech/geotrellis/pull/3238) 
+- Rasterizer.rasterize should be consistent with rasterizeWithValue [#3238](https://github.com/locationtech/geotrellis/pull/3238)
```
CloudNiner, line 18: These two changes read like test assertions. Should we add tests to cover them?

store/src/main/scala/geotrellis/store/GeoTrellisRasterSource.scala
CloudNiner: Can you comment a bit further on why we don't need this anymore? That linked issue + its links don't show where the original problem `handling missing spatialkeys` is addressed in the original sparsestitch function.

pomadchin: @CloudNiner if you will look a bit above (at the place where it was used):

`sparseStitch(tiles, extent)` we can use the inbuild function now `tiles.sparseStitch(extent)`

The issue https://github.com/locationtech/geotrellis/issues/2903 was addressed in your PR here https://github.com/locationtech/geotrellis/pull/3017 which is called "Add sparse stitch method to StitchCollectionMethods".

CHANGELOG.md
pomadchin: We can add an empty test [here](https://github.com/locationtech/geotrellis/blob/f1335c31e5408ab2bdae4cb3cd56343e459f752f/spark/src/test/scala/geotrellis/spark/stitch/CollectionStitchMethodsSpec.scala), agree.

store/src/main/scala/geotrellis/store/GeoTrellisRasterSource.scala
CloudNiner: lol embarrassing -- I do in fact remember this now. 👍

Issue #3248, opened 2020-06-02
GridExtent rounding issues


pomadchin:
In the `GridExtent` funtions we use floors, ceils and rounds pretty [inconsistently](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/GridExtent.scala#L33-L49). 

QGIS and GDAL use ceils and floors with some tolerance that helps to [handle corner cases better](https://github.com/qgis/QGIS/blob/607664c5a6b47c559ed39892e736322b64b3faa4/src/analysis/raster/qgsalignraster.cpp#L38).

This task is about doublecheking how we work with rounding floating numbers, and consider usage of the `ceilWithTolerance` as well (a long with the [floorWithTolerance](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/GridExtent.scala#L414) function we already use).

jisantuc:
:wave: I put a test tif at `s3://geotrellis-test/3248/failing-withResolutions.tif`. You can reproduce the error with this ammonite script that mimics the [process for calculating a histogram](https://github.com/raster-foundry/raster-foundry/blob/develop/app-backend/common/src/main/scala/utils/CogUtils.scala#L51-L63) in Raster Foundry

```scala
import $ivy.`org.locationtech.geotrellis::geotrellis-raster:3.3.0`
import $ivy.`org.locationtech.geotrellis::geotrellis-s3:3.3.0`
import $ivy.`org.locationtech.geotrellis::geotrellis-vector:3.3.0`

import geotrellis.raster._
import geotrellis.vector._
import geotrellis.raster.geotiff._

def viableCellSizes(
        rs: RasterSource,
        baseCellSize: CellSize,
        sizes: List[CellSize],
        target: Long
    ): List[CellSize] = {
      val width = rs.cols
      val height = rs.rows

      val isGood: CellSize => Boolean = (cs: CellSize) => {
        val cellCount = width / (cs.width / baseCellSize.width) * (height / (cs.height / baseCellSize.height))
        println(s"Cell count for $cs: $cellCount")
        cellCount < target
      }

      sizes.filter(isGood)
    }

val rsSb = GeoTiffRasterSource("trl-transformed-cog.tif")
val viableSb = viableCellSizes(rsSb, rsSb.cellSize, rsSb.resolutions, 400000)

[...]

PR #3250, opened 2020-06-08
Use [ceil|floor]WithTolerance methods in GridExtent

CloudNiner:
~~Still waiting on an edge case test that allows us to consistently reproduce this issue~~, but as part of some other bugfixing/testing I updated GridExtent to use the tolerance methods consistently everywhere. This appears to have resolved the issue in the withResolution calls that caused mismatched grid extents.

They are _not_ used in `GridExtent.alignTargetPixels` because that method provides
a GDAL specific operation copied from GDAL source.

## Notes

The tolerance in GridExtent is 1e-7, in QGIS its 1e-6 and GDAL its 1e-5. Should we update our value to match one of the other projects? This could have the consequence of subtly breaking projects that rely on the current behavior for correctness. 

## Testing

Take a look at the first commit's CI build: https://app.circleci.com/pipelines/github/locationtech/geotrellis/28/workflows/e5506f7e-9024-4d1b-83b2-2a2f78945639/jobs/233

In there, you should find a test failure for the test `should compute cols and rows in GridExtents with non integer cell sizes` (use ctrl+f to find in page).

Then take a look at the HEAD of this branch's CI build: https://app.circleci.com/pipelines/github/locationtech/geotrellis/29/workflows/18bca2a1-c6b6-47d0-963c-e9ddbab9431d/jobs/246/steps

It should succeed with the new test now passing. Can search for the same passing test in the CI output.

Connects #3248

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/GridExtent.scala
+++ b/raster/src/main/scala/geotrellis/raster/GridExtent.scala
@@ -70,8 +70,8 @@ class GridExtent[@specialized(Int, Long) N: Integral](
       throw GeoAttrsError(s"illegal cellheights: $cellheight and ${that.cellheight}")
 
     val newExtent = extent.combine(that.extent)
-    val newRows = ceil(newExtent.height / cellheight)
-    val newCols = ceil(newExtent.width / cellwidth)
+    val newRows = ceilWithTolerance(newExtent.height / cellheight)
+    val newCols = ceilWithTolerance(newExtent.width / cellwidth)
 
     new GridExtent[N](newExtent, cellwidth, cellheight,
       cols = Integral[N].fromDouble(newCols),
@@ -128,8 +128,8 @@ class GridExtent[@specialized(Int, Long) N: Integral](
   * width.
   */
   def withResolution(targetCellWidth: Double, targetCellHeight: Double): GridExtent[N] = {
-    val newCols = math.ceil((extent.xmax - extent.xmin) / targetCellWidth)
-    val newRows = math.ceil((extent.ymax - extent.ymin) / targetCellHeight)
+    val newCols = ceilWithTolerance((extent.xmax - extent.xmin) / targetCellWidth)
+    val newRows = ceilWithTolerance((extent.ymax - extent.ymin) / targetCellHeight)
     new GridExtent(extent, targetCellWidth, targetCellHeight,
       cols = Integral[N].fromDouble(newCols),
       rows = Integral[N].fromDouble(newRows))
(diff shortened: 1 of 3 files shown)


pomadchin:
@CloudNiner what about math.round calls? should they still be math.rounds or it should be smth else?

CloudNiner:
Unsure. Would probably prefer to avoid further changes until we have a test case to work against which I still haven't found.

raster/src/test/scala/geotrellis/raster/GridExtentSpec.scala
```diff
@@ -136,5 +136,28 @@ class GridExtentSpec extends FunSpec with Matchers {
...
+        val isGood: CellSize => Boolean = (cs: CellSize) => {
+          val cellCount = width / (cs.width / baseCellSize.width) * (height / (cs.height / baseCellSize.height))
+          println(s"Cell count for $cs: $cellCount")
+          cellCount < targetCellCount
+        }
+
+        sizes.filter(isGood)
+      }
+      val path = baseGeoTiffPath("failing-withResolutions.tif")
+      val rsSb = GeoTiffRasterSource(path)
```
pomadchin, line 157: Hm, I think you can extract `GridExtents` manually for the test case and you don't even need this test file to be used in tests at all.

raster/src/main/scala/geotrellis/raster/GridExtent.scala
```diff
@@ -70,8 +70,8 @@ class GridExtent[@specialized(Int, Long) N: Integral](
       throw GeoAttrsError(s"illegal cellheights: $cellheight and ${that.cellheight}")
 
     val newExtent = extent.combine(that.extent)
-    val newRows = ceil(newExtent.height / cellheight)
-    val newCols = ceil(newExtent.width / cellwidth)
+    val newRows = ceilWithTolerance(newExtent.height / cellheight)
```
pomadchin, line 73: Should not be [here](https://github.com/locationtech/geotrellis/blob/4a021962d194b7693f8dff4b0cd9e10d49683c44/raster/src/main/scala/geotrellis/raster/GridExtent.scala#L45-L57) used ceilWithTolerance as well?

It looks like we're computing cols / rows via the ceil operation but validating it via the `round` function.

Can you also check that https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/MosaicRasterSource.scala#L160-L163 would work through the `GridExtent.combine`?

^ the last one I can check later; but if [this test](https://github.com/locationtech/geotrellis/blob/cf6e8596a0b21c4d3598525080b98dc087c8f898/gdal/src/test/scala/geotrellis/raster/gdal/GDALRasterSourceSpec.scala#L202) wont fail it is a good sign.

CloudNiner: > It looks like we're computing cols / rows via the ceil operation but validating it via the round function.

In short, yes, after a discussion with @echeipesh. And GridExtent currently operates in its constructor using `round` so in order to remain consistent, we should use that elsewhere when we construct new GridExtents, like in `withResolution` and `combine` to avoid mismatches throwing exceptions that are out of user control like in #3248. **I've updated this PR to use `round` everywhere and added an additional test to verify `withResolution`.** I've also created https://github.com/locationtech/geotrellis/issues/3261 in order to line up some future work to improve `withResolution`. I'm going to add a bit of documentation in here to describe this situation before merging as well.

> Can you also check...

I think we still need the logic there that you linked to for now because GridExtent.combine  will throw an exception unless the two GridExtents to be combined have the same CellSize, which isn't a guarantee here when constructing a mosaic.

pomadchin: sounds good to me, thanks!

PR #3252, opened 2020-06-09
Fix MosaicRasterSource, GDALRasterSource and GeoTiffResampleRasterSource behavior

pomadchin:
# Overview

This PR fixes: 
* MosaicRasterSources reprojection and resample logic
* GDALRasterSource chained reprojection
* GDALWarpOptions reprojection
* GeoTiffRasterSource resample methods

## Checklist


raster/src/main/scala/geotrellis/raster/MosaicRasterSource.scala
```diff
@@ -86,13 +86,20 @@ abstract class MosaicRasterSource extends RasterSource {
     resampleTarget: ResampleTarget = DefaultTarget,
     method: ResampleMethod = ResampleMethod.DEFAULT,
     strategy: OverviewStrategy = OverviewStrategy.DEFAULT
-  ): RasterSource =
-    MosaicRasterSource(
-      sources map { _.reproject(targetCRS, resampleTarget, method, strategy) },
-      crs,
-      gridExtent.reproject(this.crs, targetCRS, Reproject.Options.DEFAULT.copy(method = method)),
```
pomadchin, line 93: This gives an incorrect CellSize

raster/src/main/scala/geotrellis/raster/MosaicRasterSource.scala
```diff
@@ -86,13 +86,20 @@ abstract class MosaicRasterSource extends RasterSource {
...
-      name
-    )
+  ): RasterSource = {
+    val reprojectedSources = sources.map(_.reproject(targetCRS, resampleTarget, method, strategy))
+    val combinedExtent = reprojectedSources.map(_.extent).toList.reduce(_ combine _)
+    val minCellSize = reprojectedSources.map(_.cellSize).toList.maxBy(_.resolution)
```
CloudNiner, line 92: Down in one of the `apply()` methods, this line appears, but it uses `minBy(_.resolution)` instead of the `maxBy(_.resolution)` used here. If maxBy is actually correct, then we should probably fix that apply method as well -- the general operations applied there look pretty consistent with what's done here.

CloudNiner: Line in apply method referenced in prev comment: https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/MosaicRasterSource.scala#L171-L173

pomadchin: `maxBy` finds the less resolute cellSize (the larges cellSize, resolution = sqrt(w * h)), minBy finds the most resolute. 

I think it is appropriate to deterimne the cellSize by the least resolute raster in the mosaic.
What do you think about it?

CloudNiner: Yeah, least resolute makes more sense than the other way around, otherwise you're faking a higher resolution than you provable have across the mosaic. 

So in that case, I guess we should update the apply method for consistency with this reproject method?

Issue #3254, opened 2020-06-11
HttpRangeReader should live outside of the Spark package


pomadchin:
At this point [HttpRangeReader](https://github.com/locationtech/geotrellis/blob/7eb7958dde643fa4a347c9ac5b055bb333f416fd/spark/src/main/scala/geotrellis/spark/store/http/util/HttpRangeReader.scala) lives in a Spark project. We can put it into some other project to remove spark dep from the GeoTrellis Server.

PR #3255, opened 2020-06-12
Move HttpRangeReader out of the spark package

pomadchin:
# Overview

This PR moves HttpRangeReader into the store pacakge.

## Checklist

Closes https://github.com/locationtech/geotrellis/issues/3254

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - ArrayTile equals method always returns true if first elements are NaN [#3242](https://github.com/locationtech/geotrellis/issues/3242)
 - Fixed resource issue with JpegDecompressor that was causing a "too many open files in the system" exception on many parallel reads of JPEG compressed GeoTiffs. [#3249](https://github.com/locationtech/geotrellis/pull/3249)
 - Fix MosaicRasterSource, GDALRasterSource and GeoTiffResampleRasterSource behavior [#3252](https://github.com/locationtech/geotrellis/pull/3252)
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - ArrayTile equals method always returns true if first elements are NaN [#3242](https://github.com/locationtech/geotrellis/issues/3242)
 - Fixed resource issue with JpegDecompressor that was causing a "too many open files in the system" exception on many parallel reads of JPEG compressed GeoTiffs. [#3249](https://github.com/locationtech/geotrellis/pull/3249)
 - Fix MosaicRasterSource, GDALRasterSource and GeoTiffResampleRasterSource behavior [#3252](https://github.com/locationtech/geotrellis/pull/3252)
+- HttpRangeReader should live outside of the Spark package [#3254](https://github.com/locationtech/geotrellis/issues/3254)
```
CloudNiner, line 24: Noting that this more than a bugfix change and would necessitate a bump of at least the minor version, yes?

pomadchin: @CloudNiner do we need to add a `Fixed` section in addition to `Added`?

CloudNiner: Eh. I'm no expert but this particular entry still feels like a `Changed`. It's just a change that will cause compiler errors.

pomadchin: Yea, it is a breaking change. We can still release it as a minor release but with a note that it breaks bin compat due to an oversight during the 3.0 release.

PR #3265, opened 2020-06-23
GeoTiffRasterSource is not threadsafe enough

pomadchin:
# Overview

Lock should happen on a tiff object in GeoTiffRasterSources

## Checklist

## Demo

GeoTreslli server STAC catalog reads only

## Notes

I could not implement a unit test that would reproduce that bad behavior.

Closes https://github.com/azavea/geotrellis/issues/272


pomadchin:
I could not reproduce this bug in tests. I'm going to merge this PR to simplify further testing with geotrellis-server PR geotrellis/geotrellis-server#284

PR #3268, opened 2020-07-10
Fix ShortArrayTileResult result ArrayTile fulfillment

pomadchin:
# Overview

This PR fixes an incorrect result array tile that is used i.e. in the FocalHillshade operation. 

## Checklist

## Demo

Before (nodata around the scene is (0, 0, 0)):

image

After (nodata around the scene is valid):

image


pomadchin:
It is a quick fix, I'll merge it once CI is green.

Issue #3203, opened 2020-03-16
renderJpg() color is  bluer than expected


jdenisgiguere:
## Current situation

With I use `rendegJpg()` on a Multiband Tille, the color is not as expected. It is bluer than expected.

`renderPng()` return an image of the expected color.

See `renderJpg()` result
![renderJpg](https://user-images.githubusercontent.com/973241/76792921-7132ca80-679a-11ea-917c-30e1b10d6c94.jpg)

and `renderPng()` result
![renderPng](https://user-images.githubusercontent.com/973241/76792882-5a8c7380-679a-11ea-9068-8d7b936724bd.png)
 

## Expected situation

`repnderPng()` and `renderJpg()` should return similar image.

pomadchin:
Hi! Thanks for reporting, it is a duplicate of #3151; could you upload an original small image that you use to generate pngs / jpgs so we can test against it?

jdenisgiguere:
Thanks for the super quick follow up!

See: 
[tiled_U-19OrthoSectorTile1_2012x2012.zip](https://github.com/locationtech/geotrellis/files/4340036/tiled_U-19OrthoSectorTile1_2012x2012.zip)

pomadchin:
@jdenisgiguere thanks!

esmeetu:
Same here.

Problem here might be `ImageIO` recieve a `ARGB` while geotrellis give us a `RGBA`.
And i found a workaround:
```scala
  def renderJpg(img: MultibandTile) = {
    // get a rgba tile
    val coloredImg = img.color()

    // begin translate from rgba to argb
    val (cols, rows) = (img.cols, img.rows)
    val bi = new BufferedImage(cols, rows, BufferedImage.TYPE_INT_ARGB)
    cfor(0)(_ < cols, _ + 1) { x =>
      cfor(0)(_ < rows, _ + 1) { y =>
        val v = coloredImg.get(x, y)
        val r = (v >> 24) & 0xFF
        val g = (v >> 16) & 0xFF
        val b = (v >> 8) & 0xFF
        val a = v & 0xFF
        bi.setRGB(x, y, a << 24 | r << 16 | g << 8 | b)
      }
    }
    val result = ImageIO.write(bi, "png", new File("img.png"))
    print(result)
  }
```
[...]

PR #3278, opened 2020-08-01
Fix NoData handling in the ColorMethods.color function for the RGB and RGBA Multiband Rasters

pomadchin:
# Overview

This PR fixes the NoData interpretation by the `ColorMethods.color` function.
With the current fix for RGBA Rasters the zero would be set only in case the alpha channel is set to 0 as well.
In the RGB case, the alpha channel would be set to zero only when all the 3 bands are filled with nodata values.

## Checklist

Fix https://github.com/locationtech/geotrellis/issues/3203


raster/src/main/scala/geotrellis/raster/render/ColorMethods.scala
```diff
@@ -48,14 +48,13 @@ trait MultibandColorMethods extends MethodExtensions[MultibandTile] {
...
+        val r = if (isData(rBand)) { transparent = false; rBand } else 0
+        val g = if (isData(gBand)) { transparent = false; gBand } else 0
+        val b = if (isData(bBand)) { transparent = false; bBand } else 0
 
-        if(r + g + b == 0) 0
-        else {
-          ((r & 0xFF) << 24) | ((g & 0xFF) << 16) | ((b & 0xFF) << 8) | 0xFF
-        }
+        if(transparent) 0
+        else ((r & 0xFF) << 24) | ((g & 0xFF) << 16) | ((b & 0xFF) << 8) | 0xFF
```
pomadchin, line 57: I wanted to make this "transparency" detection more cheap, since this function is applied to all pixels of the raster.

PR #3288, opened 2020-09-04
Fix Encoder[GeometryCollection]

### LINKED ISSUE #3167: GeometryCollection encoding issue
The following code
```
val p = Point(1, 1)
println(GeometryCollection(multiPoints = Seq(MultiPoint(p, p))).toGeoJson)
```
prints
```
{
  "type" : "GeometryCollection",
  "geometries" : [
    {"type" : "MultiPoint", "coordinates" : [[1.0, 1.0], [1.0, 1.0]]},
    {"type" : "GeometryCollection", "geometries" : [
      {"type" : "Point", "coordinates" : [1.0, 1.0]},
      {"type" : "Point", "coordinates" : [1.0, 1.0]}
    ]}
  ]
}
```
The same duplication happens with `MultiPolygons` and `MultiLines` when nested into `GeometryCollection`. It appears that this is because a `MultiPoint` IS a `GeometryCollection`, so the `Encoder` retrieves it two times - first by `getAll[MultiPoint]`, then by `getAll[GeometryCollection]`.

agrian-joel:
# Overview
Encoding a GeometryCollection with a Multipolygon/MultiPoint/MultiLineString should serialize those geometries once

Current behavior encodes them in geometries as both `type: MultiX` and `type: GeometryCollection`

## Checklist

- [X] [./CHANGELOG.md](https://github.com/locationtech/geotrellis/blob/master/CHANGELOG.md) updated, if necessary. Link to the issue if closed, otherwise the PR.

Closes #3167

### DIFF
--- a/vector/src/main/scala/geotrellis/vector/io/json/GeometryFormats.scala
+++ b/vector/src/main/scala/geotrellis/vector/io/json/GeometryFormats.scala
@@ -177,7 +177,7 @@ trait GeometryFormats {
       Json.obj(
         "type" -> "GeometryCollection".asJson,
         "geometries" -> 
-          obj.geometries.collect({
+          obj.geometries.collect {
             case geom: Point => geom.asJson
             case geom: LineString => geom.asJson
             case geom: Polygon => geom.asJson
@@ -185,7 +185,7 @@ trait GeometryFormats {
             case geom: MultiPoint => geom.asJson
             case geom: MultiLineString => geom.asJson
             case geom: GeometryCollection => geom.asJson
-          }).asJson
+          }.asJson
       )
     }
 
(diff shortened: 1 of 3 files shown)


pomadchin:
Hi @agrian-joel thanks a lot for your first contribution! I'll review it a bit later.

vector/src/main/scala/geotrellis/vector/io/json/GeometryFormats.scala
```diff
@@ -176,15 +176,16 @@ trait GeometryFormats {
...
-          obj.getAll[Point].map(_.asJson),
-          obj.getAll[LineString].map(_.asJson),
-          obj.getAll[Polygon].map(_.asJson),
-          obj.getAll[MultiPoint].map(_.asJson),
-          obj.getAll[MultiLineString].map(_.asJson),
-          obj.getAll[MultiPolygon].map(_.asJson),
-          obj.getAll[GeometryCollection].map(_.asJson)
-        ).flatten.asJson
+        "geometries" -> 
+          obj.geometries.map({
```
pomadchin, line 180: In order not to throw in case of the type mismatch, it is possible to use the `collect` function instead of `map` here (see the comment https://github.com/locationtech/geotrellis/pull/3288#discussion_r483846948)

pomadchin (approved):
Thanks!

PR #3294, opened 2020-09-16
Spark 3 & Hadoop 3 support, Scala 2.13 cross compilation

### LINKED ISSUE #2893: Scala 2.13 cross compilation
Basing on the following information

https://spark.apache.org/news/spark-2-4-1-released.html

```
In Apache Spark 2.4.1, Scala 2.12 support is GA, and it's no longer experimental. 
We will drop Scala 2.11 support in Spark 3.0, so please provide us feedback.
```

And [this JIRA ticket](https://issues.apache.org/jira/browse/SPARK-25075?jql=text%20~%20%22Scala%202.13%22), I suppose we need to move along with the Spark.

_Probably_ 2.13 support would be in Spark 3.0 release. The result of this issue would be to drop Scala 2.11 and to add Scala 2.12 _and 2.13 cross compilation (for sure if there would be Spark binaries)_.

This issue will also contain changes that would be nice to have to simplify Scala 3 migration:

### Scala 2.13
- [ ] Remove usage of `Traversable` (it is possible to replace it with `Iterable`)
- [ ] Remove usage of `TraversableOnce` (it is possible to replace it with `IterableOnce`)
- [ ] [Drop Scala 2.11](https://github.com/locationtech/geotrellis/issues/3259)

### Scala 3
- [ ] After Scala 2.13 function `def foo()` can be called only as a `.foo()` in the code and can't be invoked as `.foo` without parentheses (see https://scastie.scala-lang.org/pzjfiB1VT1C9Aljp9YkzMQ). So Scala would be more strict about it.
- [ ] `do` loops [would be removed](https://dotty.epfl.ch/docs/reference/dropped-features/do-while.html); instead, it would be possible to use a new `while` notation
- [ ] [Class shadowing](https://dotty.epfl.ch/docs/reference/dropped-features/class-shadowing.html)
- [ ] [XML literals](https://dotty.epfl.ch/docs/reference/dropped-features/xml.html)
- [ ] [Symbol literals](https://dotty.epfl.ch/docs/reference/dropped-features/symlits.html)
- [ ] [Weak Conformance](https://dotty.epfl.ch/docs/reference/dropped-features/weak-conformance.html)
- [ ] [private[this] and protected[this] modifiers](https://dotty.epfl.ch/docs/reference/dropped-features/this-qualifier.html)

### LINKED ISSUE #3218: Spark 3 & Hadoop 3 Support
https://spark.apache.org/news/spark-3.0.0-preview.html

Looks like it might be an easy upgrade. At this point is a place-holder issue leading to pushing a SNAPSHOT release with Spark `3.0.0-preview2` dependency.

### LINKED ISSUE #3259: Drop Scala 2.11 cross compilation
We don't use Scala 2.11 anymore. Also this is a requirement towards https://github.com/locationtech/geotrellis/issues/2893

However [GeoMesa](https://github.com/locationtech/geomesa) is still in Scala 2.11

### LINKED ISSUE #3358: Consider releasing a subset of packages that don't depend on Spark against 2.13
## Is your feature request related to a problem? Please describe.

Spark 2.x and 3.1.x only support Scala 2.11 and 2.12
Spark 3.2 (unreleased) will only support Scala 2.12 and 2.13

This situation couples the Spark 3 and Scala 2.13 upgrades. However, not all geotrellis packages depend on Spark, and could be published independently.

Packages that don't depend on Spark (accumulo, cassandra, gdal, geotools, hbase, layer, macros, mdoc, proj4, raster, s3, shapefile, store, util, vector, vectortile) will compile under Scala 2.11, 2.12, and 2.13 (pending a few minor fixes). There are other projects, such as stac4s, that only depend on these non-Spark components, and I've also used them on internal projects. The limitation here is that since geotrellis only supports 2.12, great libraries like stac4s can't be integrated into projects that already use 2.13 without down-versioning the project. If the non-Spark libraries were released for 2.13, stac4s could cross-compile for it, and be more available to be used by more projects.

## Describe the solution you'd like

Packages that don't depend on Spark to be cross-compiled and released for 2.13. 

It may even make sense to split the geotrellis build into two separate repos -- one with the non-Spark components that can move "ahead" more cleanly, and another with the Spark components that depends on those non-Spark versions. Obviously this would be a big move, but it's something to think about.

## Describe alternatives you've considered

1. down-versioning 2.13 projects that want to use geotrellis or stac4s to 2.12
2. building and publishing 2.13 compiled geotrellis and stac4s to a private repo
3. Copying geotrellis or stac4s code into the 2.13 project

## Additional context

none

pomadchin:
# Overview

- [x] Spark 3.0.1 / 3.2.0-SNAPSHOT update [_provided_]
- [x] Scala 2.12 / 2.13 cross compilation
- [x] Hadoop 3.2.1 [_provided_]
  - [x] woodstox-core 6.2.5 [_provided_]
  - [x] stax2-api 4.2.1 [_provided_]
  - [x] commons-configuration2 2.7 [_provided_]
  - [x] re2j 1.6 [_provided_]
- [x] HBase 2.4.1 (compat reasons with Hadoop 3.x) [#23238](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23238)
- [x] Spire 0.17.0 [#23239](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23239)
- [x] Proj4j 1.1.2 
- [x] SBT 1.5.0

TODO:

- [x] Check other libraries: mb not only spire requires an upgrade
  - https://github.com/locationtech/geotrellis/issues/3381
  - https://github.com/locationtech/geotrellis/issues/3382
- [x] Test on AWS EMR 6.2 with and without GDAL
  -   https://github.com/locationtech/geotrellis/issues/3380

## Checklist

Closes #2893
Closes #3218
Closes #3358


cassandra/src/main/scala/geotrellis/store/cassandra/CassandraInstance.scala
```diff
@@ -115,6 +115,10 @@ object BaseCassandraInstance {
       val builder =
         Cluster
           .builder()
+          // Spark 3 bring dropwizard 4.1.1
+          // https://docs.datastax.com/en/developer/java-driver/3.5/manual/metrics/#metrics-4-compatibility
+          // TODO: Upd cassandra driver up to 4.9
+          .withoutJMXReporting()
```
pomadchin, line 121: It makes sense to update up to the latest Cassandra driver in terms of this PR.

hbase-spark/src/main/scala/org/apache/spark/deploy/GTSparkHadoopUtil.scala
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package org.apache.spark.deploy
+
+import org.apache.spark.SparkConf
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs._
+
+/** A proxy object for https://github.com/apache/spark/blob/v3.0.1/core/src/main/scala/org/apache/spark/deploy/SparkHadoopUtil.scala#L425-L429 */
+object GTSparkHadoopUtil {
```
pomadchin, line 24: Probably we can live without this class. The only thing is is used for is https://github.com/apache/spark/blob/v3.0.1/core/src/main/scala/org/apache/spark/deploy/SparkHadoopUtil.scala#L425-L429

project/Dependencies.scala
```diff
@@ -18,52 +18,33 @@ import sbt._
 
 object Version {
   val geotools    = "23.2"
-  val spire       = "0.13.0"
+  val spire       = "0.17.0-M1"
```
pomadchin, line 21: Spark 3 upgrades the breeze dependency as well as the transitive spire dependency.

spark/src/test/scala/geotrellis/spark/summary/StatsTileRDDMethodsSpec.scala
```diff
@@ -113,7 +113,9 @@ class StatsTileRDDMethodsSpec extends AnyFunSpec with TestEnvironment with TestF
       hist.merge(hist2).quantileBreaks(70) should be (hist.quantileBreaks(70))
     }
 
-    it ("should be able to sample a fraction of an RDD to compute a histogram") {
+    // TODO: fix this test before merge
+    // 30000.0 was not less than or equal to 20000.0 (StatsTileRDDMethodsSpec.scala:128)
+    ignore ("should be able to sample a fraction of an RDD to compute a histogram") {
```
pomadchin, line 118: Probably this is caused by the spire upgrade.

hbase-spark/src/test/resources/log4j.properties
```diff
@@ -8,6 +8,6 @@ log4j.logger.org.apache.spark.repl.SparkILoop$SparkILoopInterpreter=WARN
 log4j.logger.org.spark-project.jetty=WARN
 org.spark-project.jetty.LEVEL=WARN
 
-log4j.logger.org.apache.zookeeper=WARN
+log4j.logger.org.apache.zookeeper=ERROR
```
pomadchin, line 11: Spark 3 updates zookeeper dep, and it became more chatty.

project/Dependencies.scala
```diff
@@ -134,8 +134,10 @@ object Dependencies {
   val hbaseMetricsApi     = "org.apache.hbase" % "hbase-metrics-api"    % Version.hbase
   val hbaseZooKeeper      = "org.apache.hbase" % "hbase-zookeeper"      % Version.hbase
 
-  val woodstoxCore        = "com.fasterxml.woodstox" % "woodstox-core" % "6.2.5"
-  val stax2Api            = "org.codehaus.woodstox" % "stax2-api" % "4.2.1"
+  val woodstoxCore          = "com.fasterxml.woodstox" % "woodstox-core"          % "6.2.5"
+  val stax2Api              = "org.codehaus.woodstox"  % "stax2-api"              % "4.2.1"
+  val commonsConfiguration2 = "org.apache.commons"     % "commons-configuration2" % "2.7"
+  val re2j                  = "com.google.re2j"        % "re2j"                   % "1.6"
```
pomadchin, line 140: This deps were decomposed out of Hadoop 3.x

spark/src/test/scala/geotrellis/spark/summary/StatsTileRDDMethodsSpec.scala
pomadchin: Lets move it into a separate issue https://github.com/locationtech/geotrellis/issues/3379

cassandra/src/main/scala/geotrellis/store/cassandra/CassandraInstance.scala
pomadchin: https://github.com/locationtech/geotrellis/issues/3382

project/Dependencies.scala
pomadchin: I'm wondering, is it compatible with `0.17.0`? (tests pass though)

pomadchin: Lets file CQs for both version and explore how much it would hurt us in terms of https://github.com/locationtech/geotrellis/issues/3380

publish/publish-to-sonatype.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-./publish/publish-to-sonatype-212.sh
```
echeipesh, line 3: 2.13 not published

raster/src/main/scala/geotrellis/raster/ArrayMultibandTile.scala
```diff
@@ -107,7 +107,7 @@ class ArrayMultibandTile(_bands: Array[Tile]) extends MultibandTile with MacroMu
     *
     * @return  The object on which the method was invoked
     */
-  def toArrayTile = this
+  def toArrayTile() = this
```
echeipesh, line 110: Is this required for 2.13? I'm not sure if this an API change for 2.12 or not really.

pomadchin: @echeipesh that is not required, and it is not really an API change. The story with this method is that we override here a method that contains `()`: https://github.com/locationtech/geotrellis/blob/2.3/raster/src/main/scala/geotrellis/raster/MultibandTile.scala#L285

publish/publish-to-sonatype.sh
pomadchin: @echeipesh hm do you think we want to publish 3.6.0 on MavenCentral with the Spark 3.2.0-SNAPSHOT dep?

echeipesh: I think we should publish 3.6.0-SNAPSHOT with spark 3.2.0-SNAPSHOT dep, which would happen. Our release process is not setup for a split release though ... maybe the problem will solve itself with release of spark 3.2.0 before we actually have to cut 3.6.0 release. I don't think there is anything forcing our release to happen soon?

raster/src/main/scala/geotrellis/raster/ArrayMultibandTile.scala
echeipesh: 🤦

publish/publish-to-sonatype.sh
pomadchin: Yea, that is why 2.13 is not published, these scripts are to publish release artifacts.

raster/src/main/scala/geotrellis/raster/render/RGBA.scala
```diff
@@ -0,0 +1,50 @@
...
+  def apply(r: Int, g: Int, b: Int): Int = ((r << 24) + (g << 16) + (b << 8)) | 0xFF
+}
+
+object RGBA {
+  def apply(i: Int): RGBA = new RGBA(i)
+
+  def fromRGBA(r: Int, g: Int, b: Int, a: Int): RGBA =
+    new RGBA((r << 24) + (g << 16) + (b << 8) + a)
+
+  def fromRGBA(r: Int, g: Int, b: Int, alphaPct: Double): RGBA = {
```
echeipesh, line 46: Safer to go for `fromRGBAPct` to avoid double to Int conversion confusion. I guess the compiler will handle this correctly but I can see the user screwing it if they use "math.floor" or something else to compute the A value.

proj4/src/test/scala/geotrellis/proj4/io/wkt/WKTTest.scala
```diff
@@ -64,9 +64,8 @@ class WKTTest extends AnyFunSpec {
     assert(comparisonCode == epsgCodeOfWKT)
   }
 
-  // https://github.com/locationtech/proj4j/issues/61
-  it("should return the EPSG code(3785) of the passed WKT string with proj4 extension") {
-    val comparisonCode = "EPSG:3785" // "EPSG:3857" // https://github.com/locationtech/proj4j/issues/61
+  it("should return the EPSG code(3857) of the passed WKT string with proj4 extension") {
+    val comparisonCode = "EPSG:3857"
```
pomadchin, line 68: This test properly works now!

echeipesh (approved):

metasim:
Nicely done guys!! Much appreciated! ❤️

PR #3295, opened 2020-09-24
Fix GeometryCollection::getAll extension method

### LINKED ISSUE #3289: GeometryCollection vector.Reproject incorrect behavior
## Describe the bug

The original issue is described here: https://github.com/locationtech/geotrellis/issues/3167#issuecomment-567999614 by @atararaksin 

## To Reproduce

```scala
scala> val p = Point(1, 1)
p: org.locationtech.jts.geom.Point = POINT (1 1)

scala> val gc = GeometryCollection(multiPoints = Seq(MultiPoint(p, p)))
gc: org.locationtech.jts.geom.GeometryCollection = GEOMETRYCOLLECTION (MULTIPOINT ((1 1), (1 1)))

scala> gc.reproject(WebMercator, LatLng)
res0: org.locationtech.jts.geom.GeometryCollection = GEOMETRYCOLLECTION (MULTIPOINT ((0.0000089831528412 0.000008983152841), (0.0000089831528412 0.000008983152841)), GEOMETRYCOLLECTION (POINT (0.0000089831528412 0.000008983152841), POINT (0.0000089831528412 0.000008983152841)))
```
## Expected behavior

The Multipoint duplication from the example above should not happen

## Environment

- GeoTrellis version: 3+

## Additional context

The problem is caused in this case by the [Reproject.apply](https://github.com/locationtech/geotrellis/blob/master/vector/src/main/scala/geotrellis/vector/reproject/Reproject.scala#L164-L184) methods behavior. 

This task is about reviewing the `gc.getAll` function usages in all places. Potentially not only the `Reproject.apply` could be affected.

jpolchlo:
# Overview

As noted in #3289, we have some problems dealing with `GeometryCollection`s.  This bug was traced to the implementation of `getAll` which used reflection to extract the contained geometries that matched a desired type.  This was intended to be used to extract specific types or subclasses (`gc.getAll[Geometry]` would be useful).  The problem was with `gc.getAll(GeometryCollection)`, since `MultiPoint`, `MultiLineString`, and `MultiPolygon` are subclasses of `GeometryCollection`.  Therefore, `gc.getAll[GeometryCollection]` would return all `Multi*` and `GeometryCollection` entities.  This caused problems for methods like `reproject`, which would duplicate `Multi*` elements.

This PR, in it's first form, is for discussion purposes.  The simple solution presented here is that `getAll` loses it's subclass semantic.  This is the most straightforward, requiring no additional changes, but it does break the semantics, and therefore, technically must wait for GT 4.  A more laborious change could be implemented, keeping the subclass semantics of the old `getAll` implementation, but reimplementing `reproject` and other functions that don't correctly differentiate types.  This latter solution is more work, but may be preferred to maintain certain functionality.  

In all cases, GT 3 broke with the model of GT 2 and down where the contents of geometry collections were explicitly compartmentalized, so we might have some latitude for bending semver rules, as this PR's content can be seen as a bugfix of a broken API rather than a change of interface.

## Checklist

## Notes

Closes #3289

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Fix `Encoder[GeometryCollection]` including subclasses of GeometryCollection twice in the json
    (MultiPolygon, Multipoint,MultiLinestring) [#3167](https://github.com/locationtech/geotrellis/issues/3167)
--Fix `LayoutTileSource` buffer should only be 1/2 a cellsize to avoid going out of bounds and creating `NODATA` values [#3302](https://github.com/locationtech/geotrellis/pull/3302)
+- Fix `LayoutTileSource` buffer should only be 1/2 a cellsize to avoid going out of bounds and creating `NODATA` values [#3302](https://github.com/locationtech/geotrellis/pull/3302)
 - Remove unused allocation from CroppedTile [#3297](https://github.com/locationtech/geotrellis/pull/3297)
+- Fix GeometryCollection::getAll extension method [#3295](https://github.com/locationtech/geotrellis/pull/3295)
 
 ## [3.5.0] - 2020-08-18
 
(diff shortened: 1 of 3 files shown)


pomadchin:
@jpolchlo could you update the changelog?

pomadchin:
Hm, I also have a feeling that changes intorduced in this PR https://github.com/locationtech/geotrellis/pull/3288 are not neccesary now. I think it makes sense to revert them to keep the code consistent?

jpolchlo:
Happy to update the changelog, but I was/am waiting for discussion as to whether this is the right solution.  If you and others are happy with what's presented here, then I'll note the changes and we can move to merge.

pomadchin (approved):
Thanks Justin, PR looks good 💯 // I rebased it to resolve CHANGELOG conflicts // merging it once Travis is happy

PR #3302, opened 2020-11-10
Buffer for resample by only half a pixel

### LINKED ISSUE #3299: Row of No Data Pixels returned from Raster Source Read
## Describe the bug

When performing a tile read from a RasterSource for a tiff that is supposed to be 256x256 sometimes the resulting tile includes a single row of no data at the bottom of the tile.

## To Reproduce

I've attached a zip file that has the steps to reproduce this bug, including a tiff that exhibits the behavior. To run this - install `ammonite`, unzip the file, and run `amm test-script-worksheet.sc`

This only seems to be happening for zoom level 19 - when running the script here is some sample output:

```
Debug info for: (307710, 257618)
Tile sum:  65280
Tile size: 65536
Debug info for: (307711, 257615)
Tile sum:  65280
Tile size: 65536
Debug info for: (307711, 257616)
Tile sum:  65280
Tile size: 65536
Debug info for: (307711, 257617)
Tile sum:  65280
Tile size: 65536
Debug info for: (307711, 257618)
Tile sum:  65280
Tile size: 65536
```

The value for the `sum` and `size` should be the same: 65536 but they are not - there are 256 nodata values present in the the tile read, though the dimensions are correct 256 x 256.

## Expected behavior

Data should be returned for all pixels for this test case.

## Screenshots

In practice this manifests as blank lines of pixels when rendering PNGs
image

## Environment

- Java version: 8
- Scala version: 2.12
- GeoTrellis version: tested on 3.4 and 3.5

[nodata-tiles.zip](https://github.com/locationtech/geotrellis/files/5453543/nodata-tiles.zip)

moradology:
# Overview

`LayoutTileSource` resamples were sometimes going out of bounds. In those cases, `Resample` short circuits to `NODATA`

## Checklist

Closes #3299


layer/src/test/scala/geotrellis/layer/LayoutTileSourceSpec.scala
```diff
@@ -244,5 +244,37 @@ class LayoutTileSourceSpec extends AnyFunSpec with RasterMatchers {
...
+      val neighborhood = for {
+        col <- ((x - 2) until (x + 2)).toList
+        row <- ((y - 2) until (y + 2)).toList
+      } yield (col, row)
+
+      def checkExtent(col: Int, row: Int, rs: RasterSource): Option[Unit] = {
+        val tile = rs
+          .tileToLayout(layout)
+          .read(SpatialKey(col, row), List(0))
+        tile map { _.band(0) } map { t =>
```
pomadchin, line 268: This test would not fail in case there is no `tile` and will silently be successfull, is it an expected behavior? I would say that for sainity having unsafe `.get` calls can be fine for tests.

layer/src/test/scala/geotrellis/layer/LayoutTileSourceSpec.scala
```diff
@@ -244,5 +244,37 @@ class LayoutTileSourceSpec extends AnyFunSpec with RasterMatchers {
...
+        }
+      }
+
+      neighborhood map {
+        case (x, y) => checkExtent(x, y, rs)
+      }
+    }
   }
-}
+}
```
pomadchin, line 280: Could you add an extra line in the end of the file?

pomadchin (changes requested):
This one looks good to me, nice PR; thx @moradology! I still have a feeling that this still may hit us one day, but that is perfect for now. 

I left a minor question about the test (`checkExtent` signature looks a bit suspicious).
And the ECA is still pending (requires a renewal), if you have troubles with signing it, feel free to reach me!

pomadchin (approved):
Thanks Nathan!

PR #3314, opened 2020-12-08
Remove explicit unused Scaffeine dependency from projects

pomadchin:
# Overview

We need to review dependecnies usage in the project modules. A great example of an unused dependency is SCaffeine which is definitely not used in `geotrellis-proj4`. 

- [x] SCaffeine

## Checklist

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -127,7 +127,7 @@ lazy val `cassandra-spark` = project
 lazy val hbase = project
   .dependsOn(store)
   .settings(Settings.hbase)
-  .settings(projectDependencies := { Seq((layer / projectID).value.exclude("com.google.protobuf", "protobuf-java")) })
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -127,7 +127,7 @@ lazy val `cassandra-spark` = project
 lazy val hbase = project
   .dependsOn(store)
   .settings(Settings.hbase)
-  .settings(projectDependencies := { Seq((layer / projectID).value.exclude("com.google.protobuf", "protobuf-java")) })
+  .settings(projectDependencies := { Seq((store / projectID).value.exclude("com.google.protobuf", "protobuf-java")) })
```
pomadchin, line 130: That worked only accidentially!

CloudNiner (approved):

PR #3317, opened 2020-12-10
Add mdoc + docusaurus build to repo and run in CI

CloudNiner:
# Overview

This PR adds the skeleton for an mdoc + docusaurus static documentation in the same vein as a number of other open source Scala projects. Never again will our doc examples be out of date 🎉. No JS or CSS knowledge is necessary to write docs.

The general idea here is to:
1. Start an mdoc watcher via sbt which listens for changes in `./docs-mdoc`
2. The mdoc watcher writes to `website/docs`
3. A dev server running in `./website` picks up changes in `./website/docs` and rerenders the docs site
4. Devs can visit http://localhost:3000 to see the docs site

More detailed instructions for these steps are in `website/README.md`, follow those to test this PR to double check correctness.

## Checklist

- [ ] ~~[./CHANGELOG.md](https://github.com/locationtech/geotrellis/blob/master/CHANGELOG.md) updated, if necessary. Link to the issue if closed, otherwise the PR.~~
- [ ] ~~[Module Hierarchy](https://github.com/locationtech/geotrellis/blob/master/docs/guide/module-hierarchy.rst) updated, if necessary~~
- [ ] ~~`docs` guides update, if necessary~~
- [ ] ~~New user API has useful Scaladoc strings~~
- [ ] ~~Unit tests added for bug-fix or new feature~~

## Demo

Is that output from a RasterSource call right there on a docs website? You better bet your bottom dollar it is.

Screen Shot 2020-12-10 at 5 53 59 PM
Screen Shot 2020-12-10 at 5 54 10 PM

## Notes

I didn't make any changes to the changelog, because there's no user facing changes yet. This is just a skeleton. Future issues will begin building out content in the `./docs-mdoc` directory.

Most of the files in `./website` are autogenerated by the docusaurus generator. I made changes to the config and sidebars files, and updated the favicon and header icon. 

Closes #3311

PR #3320, opened 2020-12-16
Fix sbt unidoc build

CloudNiner:
# Overview

Fixes an erroneous commit in #3317 that wasn't caught by the PR build because the issue was in the scaladoc `unidoc` build.

## Demo

Here's a local run of `./sbt unidoc` moving past the error in the [CI build](https://app.circleci.com/pipelines/github/locationtech/geotrellis/150/workflows/d0a3c510-45a3-4218-837a-0c22c183e413/jobs/1601) where it couldn't find the `unidoc` task:

Screen Shot 2020-12-16 at 12 58 10 PM

## Testing

Run `./sbt unidoc` on `master` -- it should fail with the error in CI. Switch to this branch and run `./sbt unidoc`. It should succeed.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -18,6 +18,7 @@ lazy val root = Project("geotrellis", file("."))
     `hbase-spark`,
     layer,
     macros,
+    mdoc,
     proj4,
     raster,
     `raster-testkit`,
@@ -41,6 +42,7 @@ lazy val root = Project("geotrellis", file("."))
 lazy val mdoc = project
   .dependsOn(raster)
   .enablePlugins(MdocPlugin)
+  .settings(publish / skip := true)
   .settings(Settings.mdoc)
 
 lazy val macros = project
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -39,7 +39,6 @@ lazy val root = Project("geotrellis", file("."))
   .settings(unidocProjectFilter in (ScalaUnidoc, unidoc) := inAnyProject)
 
 lazy val mdoc = project
```
pomadchin, line 41: Could you skip it rom publish? `publish / skip := true` 
Should not mdoc be added into the [root aggregate project](https://github.com/locationtech/geotrellis/blob/57e366756aec5dc85b01e1ed9bc07173056bc899/build.sbt#L8)?

pomadchin (changes requested):
LGTM, but I left a couple of review comments!

build.sbt
CloudNiner: Good call, done.

pomadchin (approved):
LGTM, thanks! Lets give it another try in the main branch after the CI is happy with it.

PR #3321, opened 2020-12-16
Exclude mdoc project from the unidoc generation

pomadchin:
# Overview

This PR fixes the project unidoc generation. It is interesting that the message generated by SBT is so confusing though.
It is a follow up to #3320. It looks like during the #3320 test no one could wait until the successful exit.

## Testing

`./sbt unidoc` -- it should not fail locally or in CI. It is not a fast command to test.


CloudNiner (approved):
:embarrassed: That did it, after working through coursier resolution issues on my side. 

Thanks for the assist

PR #3327, opened 2020-12-22
Fix sigmoidal contrast calculation

### LINKED ISSUE #3326: SigmoidalContrast - Imcomplete formula
## Describe the bug

In the current implementation of the Geotrellis Sigmoidal contrast, the following formula is used to calculate the contrast for a given raster value ([source](https://github.com/locationtech/geotrellis/blob/1a2ea84f7a15d790a13a75ede0fecee351ac4a7e/raster/src/main/scala/geotrellis/raster/sigmoidal/SigmoidalContrast.scala#L47)) :

```
val numer = 1/(1+math.exp(beta*(alpha-u))) - 1/(1+math.exp(beta))
val denom = 1/(1+math.exp(beta*(alpha-1))) - 1/(1+math.exp(beta*alpha))
val gu = math.max(0.0, math.min(1.0, numer / denom))
```

When you look at the sigmoidal contrast formula that is defined by [ImageMagick](https://www.imagemagick.org/Usage/color_mods/#sigmoidal), we get the following equation:

`
( 1/(1+exp(β*(α-u))) - 1/(1+exp(β*(α)) ) / ( 1/(1+exp(β*(α-1))) - 1/(1+exp(β*α)) )
`

As you can see, the *numer* part of the original equation is missing a multiplication by alpha in its second diff term.The correct calculation of the *numer* part should be:
```
val numer = 1/(1+math.exp(beta*(alpha-u))) - 1/(1+math.exp(beta*alpha))
val denom = 1/(1+math.exp(beta*(alpha-1))) - 1/(1+math.exp(beta*alpha))
val gu = math.max(0.0, math.min(1.0, numer / denom))
```

## Screenshots
The first screenshot shows the original equation, the second one the updated formula as defined by ImageMagick. As you can see, when applying a lower contrast value of beta, the higher values of the rasters will be cut off. This behaviour is less apparent when using higher beta values. However the consequence is that it makes the sigmoidal contrast function very unpredictable.

image

image

JanssenBrm:
# Overview

Fix for bug described in
[https://github.com/locationtech/geotrellis/issues/3326](https://github.com/locationtech/geotrellis/issues/3326)

Closes #3326

### DIFF
--- a/gdal-spark/src/test/resources/application.conf
+++ b/gdal-spark/src/test/resources/application.conf
@@ -1,0 +1,4 @@
+geotrellis.raster.gdal.options {
+  GDAL_DISABLE_READDIR_ON_OPEN     = "YES"
+  CPL_VSIL_CURL_ALLOWED_EXTENSIONS = ".tif"
+}
--- a/gdal-spark/src/test/scala/geotrellis/spark/gdal/GDALRasterSourceRDDSpec.scala
+++ b/gdal-spark/src/test/scala/geotrellis/spark/gdal/GDALRasterSourceRDDSpec.scala
@@ -275,21 +275,23 @@ class GDALRasterSourceRDDSpec extends AnyFunSpec with TestEnvironment with Befor
         res
       }
 
-      it(s"should not fail on parallelization with a fork join pool") {
+      it("should not fail on parallelization with a fork join pool") {
         val i = 1000
         implicit val cs = IO.contextShift(ExecutionContext.global)
 
-        val res = parallelSpec(i)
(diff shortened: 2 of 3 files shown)


pomadchin:
Hi @JanssenBrm thanks for catching that! 

Could you in addition sign [ECA](https://github.com/locationtech/geotrellis/blob/master/docs/CONTRIBUTING.rst#eclipse-contributor-agreement-eca)? And after doing so you can just sign your commit via `git commit --amend -s && git push -f origin master`

PR #3330, opened 2020-12-30
Separate monoidal part of strict layers

jisantuc:
# Overview

This PR separates the monoidal part of `StrictLayer`s -- the separate collections of geometries -- from the... I guess "metadata", you know, the _other_ part.

It's useful, e.g., if you're getting geometries from somewhere and trying to combine everything into a vector tile layer. The collections of geometries form a lawful monoid, but `StrictLayer` as it's currently implemented does not.

## Checklist

- [x] Unit tests added for bug-fix or new feature -- already tested, just a small reorganization

## Demo

I would have used this [here](https://github.com/raster-foundry/raster-foundry/pull/5538/files#diff-44d0139f6f643200b12f37254493a9953133c6b14bcea8644e7b10a10779987cR110-R148) instead of hand-rolling a special `Monoid` for strict layer for each z/x/y triple

## Notes

I'd like to prove this is a lawful monoid with `discipline` -- e.g. https://github.com/raster-foundry/raster-foundry/blob/develop/app-backend/datamodel/src/test/scala/ScopeSpec.scala#L126. It's pretty easy, but it would require new testing deps and also `Arbitrary`  instances for everything going up the dependency chain from the `MVTFeature`s, which seems like a lot for what is in practice a very fancy `Map[String, Seq[A]]`.


pomadchin:
Hey James! Thanks for the PR; we haven't had a chance to look into it though. 

One minor thing here, could you sign [ECA](https://github.com/locationtech/geotrellis/blob/master/docs/CONTRIBUTING.rst#eclipse-contributor-agreement-eca) please to make Eclipse license checker happy?

After that you'd need to add a sing-off footer into the commit via `git commit --amend -s && git push origin -f feature/js/separate-monoidal-part-of-vt`

P.S. `git commit --amend -s` adds the necessary footer into the commit message that is required for the successful license check.

vectortile/src/main/scala/geotrellis/vectortile/Layer.scala
```diff
@@ -215,13 +215,15 @@ ${sortedMeta.map({ case (k,v) => s"            ${k}: ${v}"}).mkString("\n")}
...
   version: Int,
   tileExtent: Extent,
-  points: Seq[MVTFeature[Point]],
-  multiPoints: Seq[MVTFeature[MultiPoint]],
-  lines: Seq[MVTFeature[LineString]],
-  multiLines: Seq[MVTFeature[MultiLineString]],
-  polygons: Seq[MVTFeature[Polygon]],
-  multiPolygons: Seq[MVTFeature[MultiPolygon]]
-) extends Layer
+  mvtFeatures: MVTFeatures
```
echeipesh, line 218: There needs to be a constructor overload to go from old list of args to mvtFeatures to keep this PR from being binary incompatible with previous version. Otherwise this is a solid improvement.

echeipesh (changes requested):

vectortile/src/main/scala/geotrellis/vectortile/Layer.scala
jisantuc: 9e10641

vectortile/src/main/scala/geotrellis/vectortile/MVTFeatures.scala
```diff
@@ -0,0 +1,64 @@
...
+    def combine(x: MVTFeatures, y: MVTFeatures): MVTFeatures = MVTFeatures(
+      x.points ++ y.points,
+      x.multiPoints ++ y.multiPoints,
+      x.lines ++ y.lines,
+      x.multiLines ++ y.multiLines,
+      x.polygons ++ y.polygons,
+      x.multiPolygons ++ y.multiPolygons
+    )
+  }
+}
```
pomadchin, line 64: @jisantuc could you add an extra line in your files?

jisantuc: 14fa03a

pomadchin (approved):

PR #3333, opened 2021-01-12
Add rasters S3 write methods

pomadchin:
# Overview

This PR adds a specific to s3 overload that doesn't require having a hadoop dependency in the classpath.

## Checklist

## Demo

```scala
val tiff: MultibandGeoTiff = ???
tiff.write(new AmazonS3URI("s3://test/destination.tiff"))
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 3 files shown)


scalastyle-config.xml
```diff
@@ -1,117 +0,0 @@
-<scalastyle>
```
pomadchin, line 1: It is outdated; If we'd like to use smth like that I think we need to look towards scalafmt.

echeipesh (approved):
Quality of live PR is quality 👍

PR #3342, opened 2021-03-07
replace all unicode arrow operators with equivalent ascii (deprecated in 2.13)

philvarner:
replace all unicode arrow operators with equivalent ascii (deprecated in 2.13)

# Overview

Unicode arrow operators are deprecated in 2.13, so replacing them with ascii is one small step towards 2.13 support, as there will be fewer warnings.


pomadchin:
Hi @philvarner and gz with your first contribution! Thanks for catching that. 

Also the Scala repo PR that introduces deprecation: https://github.com/scala/scala/pull/7540

pomadchin (approved):
:+1: I even forgot that we had them. Nice catch! Merging once the CI is green.

PR #3343, opened 2021-03-08
update all methods to not use procedure syntax (via scalafix ProcedureSyntax)

philvarner:
update all methods to not use procedure syntax (via scalafix ProcedureSyntax)

# Overview

Procedure syntax (a def without an `=` returning Unit) is deprecated in Scala 2, and is gone from Scala 3. This change updates all uses of this syntax with the equivalent `def f: Unit = {...}` syntax per `scalafix ProcedureSyntax`.


pomadchin (approved):
Thanks for catching it :+1:

PR #3344, opened 2021-03-08
change all long literals to use uppercase L instead of lowercase l

philvarner:

# Overview

1. change all long literals to use uppercase `L` instead of lowercase `l` . This was outputting a warning during compilation (at least in 2.13)
2. change scalastyle config to make this an error instead of a warning


pomadchin:
Hm, Is the same should be done with `Double` and `Float` literals?

philvarner:
> Hm, Is the same should be done with `Double` and `Float` literals?

I don't think those are usually changed, since `d` and `f` aren't easily confused with a number depending on your font face.

pomadchin (approved):
Thanks!

PR #3345, opened 2021-03-08
configure scalastyle to fail on error

philvarner:

# Overview

configure scalastyle to fail on error. The default is not to fail on an error, though that seems like it should be the purpose of the error level. This will also allow enforcement of error-level checks in CI.


pomadchin:
I don't use scalastyle at all tbh, mb it makes sense to move it out of the repo and to replace it with scalafmt / scalafix (later)?

pomadchin (approved):
🚀

philvarner:
I've only used scalafmt, but I didn't want to suggest the replacement. But, I do think it would be a good idea to use scalafmt.

philvarner:
Closing.

philvarner:
There's also https://github.com/sksamuel/scapegoat -- which I haven't used, but am interested in trying.

pomadchin:
@philvarner we've tried that but I am not sure that is convenient ¯\_(ツ)_/¯

PR #3346, opened 2021-03-08
WIP: Cross-compile with 2.13

philvarner:
# Overview

These changes allow cross-compilation between 2.11, 2.12, and 2.13. 

Packages that depend on 2.13 support from Spark 3.2 (not yet release) will obviously not compile (or even update). For package that do not depend on Spark (e.g., raster, vector), 2.13 does not correctly build yet, as there are ~10 minor code changes that need to be made to allow that. The only issue that is currently a blocker is in `raster`, the implicits `interpolation.Implicits`, `summary.Implicits`, and `summary.polygonal.Implicits`, can't be found.  

## Checklist

## Demo

Optional. Screenshots/REPL

## Notes

Closes #XXX

### DIFF
--- a/layer/src/main/scala/geotrellis/layer/mapalgebra/local/LocalTileCollectionMethods.scala
+++ b/layer/src/main/scala/geotrellis/layer/mapalgebra/local/LocalTileCollectionMethods.scala
@@ -131,7 +131,7 @@ trait LocalTileCollectionMethods[K] extends MethodExtensions[Seq[(K, Tile)]]
   /**
     * Negate (multiply by -1) each value in a raster.
     */
-  def unary_-() = localNegate()
+  def unary_- = localNegate()
 
   /**
     * Bitwise negation of Tile.
--- a/raster/src/main/scala/geotrellis/raster/mapalgebra/local/LocalMethods.scala
+++ b/raster/src/main/scala/geotrellis/raster/mapalgebra/local/LocalMethods.scala
@@ -80,7 +80,7 @@ trait LocalMethods extends MethodExtensions[Tile]
     Negate(self)
 
   /** Negate (multiply by -1) each value in a raster. */
-  def unary_-(): Tile = localNegate()
+  def unary_- : Tile = localNegate()
 
   /**
     * Bitwise negation of Tile.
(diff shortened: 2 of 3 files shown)


build.sbt
```diff
@@ -1,8 +1,8 @@
 import sbt.Keys._
 
-ThisBuild / scalaVersion := "2.12.13"
+ThisBuild / scalaVersion := Settings.scala212
 ThisBuild / organization := "org.locationtech.geotrellis"
-ThisBuild / crossScalaVersions := List("2.12.13", "2.11.12")
+ThisBuild / crossScalaVersions := List(Settings.scala213, Settings.scala212, Settings.scala211)
```
pomadchin, line 5: I know that it is premature, but I'd drop 2.11 and move towards Spark 3 (see https://github.com/locationtech/geotrellis/pull/3294); we have a greenlight here since GeoMesa also crosscompiled against scala 2.12 now and is able to work with Spark 3.0.x

project/Dependencies.scala
```diff
@@ -67,6 +68,15 @@ object Dependencies {
 
   def scalaReflect(version: String) = "org.scala-lang" % "scala-reflect" % version
 
+  def spire(module: String) = Def.setting {
+    CrossVersion.partialVersion(scalaVersion.value) match {
+      case Some((2, 11)) => "org.spire-math" %% "spire" % "0.13.0"
+      case Some((2, 12)) => "org.spire-math" %% "spire" % "0.13.0" // 0.17.0 exists for 2.12
```
pomadchin, line 74: btw the spire version (as well as some other libs versions) should be aligned with Spark versions. i.e. if it is Spark 3 than it should be `0.17.0-M1` (but probably 0.17.0 is still compatible)

pomadchin:
I think the blocker you're talking about is caused by the relative packages namings. I don't remember the details though.

build.sbt
philvarner: I think that's the right thing to do when the spark upgrade happens.

philvarner:
> I think the blocker you're talking about is caused by the relative packages namings. I don't remember the details though.

It seems to be because `geotrellis.raster.Implicits` has `import geotrellis.vector._`, which causes confusion between `geotrellis.vector.{interpolation,summary,summary.polygonal}.Implicits` and `geotrellis.raster.{interpolation,summary,summary.polygonal}.Implicits`

PR #3347, opened 2021-03-09
fix ambiguous relative paths for implicit imports between raster and vector

philvarner:
fix ambiguous relative paths for implicit imports between raster and vector

# Overview

This object extends many other objects named `Implicit` in the `raster` project. However, it also imported `geotrellis.vector._`, which also has objects `interpolation.Implicits`, `summary.Implicits`, and `summary.polygonal.Implicits`. In 2.11 and 2.12, these resolve to the `raster` objects in an sbt build, but still confuses IntelliJ using those versions. In 2.13, they are not resolved correctly even in the sbt build, resulting in compilation errors where the implicits are expected to exist.

This PR changes the `import geotrellis.vector._` to only import the single implicit class that's actually used from the package, so there's no ambiguous resolution of the 3 relative objects references previously mentioned.

Also, remove two unnecessary uses of `new` for case class construction.

## Checklist

## Demo

n/a

## Notes

n/a

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/Implicits.scala
+++ b/raster/src/main/scala/geotrellis/raster/Implicits.scala
@@ -17,7 +17,6 @@
 package geotrellis.raster
 
 import geotrellis.vector.Point
-import geotrellis.vector._
 import geotrellis.util.{MethodExtensions, np}
 
 object Implicits extends Implicits
@@ -60,6 +59,8 @@ trait Implicits
   implicit class withMultibandTileMethods(val self: MultibandTile) extends MethodExtensions[MultibandTile]
       with DelayedConversionMultibandTileMethods
 
+  import geotrellis.vector.methods.Implicits.withExtraPointMethods
+
   implicit class SinglebandRasterAnyRefMethods(val self: SinglebandRaster) extends AnyRef {
     def getValueAtPoint(point: Point): Int =
       getValueAtPoint(point.x, point.y)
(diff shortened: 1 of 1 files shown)


raster/src/main/scala/geotrellis/raster/Implicits.scala
```diff
@@ -60,6 +59,8 @@ trait Implicits
   implicit class withMultibandTileMethods(val self: MultibandTile) extends MethodExtensions[MultibandTile]
       with DelayedConversionMultibandTileMethods
 
+  import geotrellis.vector.methods.Implicits.withExtraPointMethods
```
pomadchin, line 62: Hmm, feel like avery ad-hock solution. What are the steps to reproduce the issue you had?

pomadchin: Is this issue present in one of your PRs with the Scala 2.13 compilation?

philvarner: I see the issue in IntelliJ syntax highlighting by just loading geotrellis master using 2.12 -- the 3 mentioned `Implicits` objects show as red, I assume b/c it's trying to resolve them from the `geotrellis.vector._` import rather than the package they're in, `geotrellis.raster`.

changing the `Implicits` object references from relative to absolute (by just prefixing them with `geotrellis.raster.`) also fixed the issue, but I thought explicitly importing only the one implicit class form `geotrellis.vector` was a better solution.  Why do you say it's ad-hoc?

This problem was the only issue I hadn't resolved in the 2.13 compilation PR.

philvarner:
Converting this to a draft -- there are a couple more Implicits related issues to resolve, and I've had some compilation problems with this change in 2.13 still.

PR #3353, opened 2021-03-10
fix several issue of non-paren'ed method overrides paren'ed method

philvarner:

# Overview

1. In 2.13, it is a compile error if a parent class method has a set of empty parens on it and an overriding method in a child class does **not** also have empty parens.  In 2.11 & 2.12, these are equivalent, so there's no change in behavior.
2. In 2.13, the `.r` syntax no longer works on StringBuilder, so an intermediate toString is required
3. Fix use of getLines without parens that gives an error or warning in 2.13
4. replace several more uses of `Unit` type as a value with `()`

## Checklist

## Demo

none

## Notes

none


pomadchin (approved):
🚀

PR #3354, opened 2021-03-11
resolve amgiguity in spark and raster Implicits classes with explicit imports shadowing local identifiers

philvarner:
# Overview

First off, this doesn't actually have anything to do with implicits -- it just happens to be a problem with spark and raster Implicits class and how it references all the other classes it extends.

The core of this is that prior to 2.13, the resolution between local identifiers (in this case, a relative reference to a class) and imported identifiers did not conform to the spec. The references to several classes resolved to the local package rather than an imported package. This was fixed in 2.13, so these two files no longer compile because the class identifiers resolve differently now.

For example, `interpolation.Implicits` resolved to the local `geotrellis.raster.interpolation.Implicits` rather than what it should have resolved to according to the spec, which is that the existence of `import geotrellis.vector._` results in it being `geotrellis.vector.interpolation.Implicits` -- this is a problem, because the package `geotrellis.vector.interpolation` exists, but not `geotrellis.vector.interpolation.Implicits`, so compilation fails. 

I thought through a few solutions to this, including explicitly importing only what's actually needed from other packages instead of `_`, but just changing this to the fqn of the Implicits classes seems the most obvious option. 

Also, fixed a couple of instances of unnecessary `new` with case classes.

## Checklist

## Demo

none

## Notes

Discussion:
https://github.com/scala/bug/issues/11593

Bug(s): 

https://github.com/scala/bug/issues/9552
https://github.com/scala/bug/issues/2458

Fix for Bug:
https://github.com/scala/scala/pull/6589

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/Implicits.scala
+++ b/raster/src/main/scala/geotrellis/raster/Implicits.scala
@@ -17,13 +17,13 @@
 package geotrellis.raster
 
 import geotrellis.vector.Point
-import geotrellis.vector._
+import geotrellis.vector.withExtraPointMethods
 import geotrellis.util.{MethodExtensions, np}
 
 object Implicits extends Implicits
 
 trait Implicits
-    extends costdistance.Implicits
+    extends geotrellis.raster.costdistance.Implicits
     with geotrellis.raster.crop.Implicits
     with geotrellis.raster.density.Implicits
     with geotrellis.raster.distance.Implicits
--- a/raster/src/main/scala/geotrellis/raster/Implicits.scala
+++ b/raster/src/main/scala/geotrellis/raster/Implicits.scala
@@ -24,33 +24,33 @@ object Implicits extends Implicits
 
 trait Implicits
     extends costdistance.Implicits
-    with crop.Implicits
-    with density.Implicits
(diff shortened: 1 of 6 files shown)


raster/src/main/scala/geotrellis/raster/Implicits.scala
```diff
@@ -24,33 +24,33 @@ object Implicits extends Implicits
...
-    with reproject.Implicits
-    with resample.Implicits
-    with sigmoidal.Implicits
-    with split.Implicits
-    with summary.Implicits
-    with summary.polygonal.Implicits
-    with transform.Implicits
-    with vectorize.Implicits
-    with viewshed.Implicits {
+    with geotrellis.raster.crop.Implicits
```
pomadchin, line 27: Yea that's what I remember doing, I'd like to review more a bit later, since I definitely played around with it but don't remember to what conclusion I came.

philvarner: I think replacing the _ imports with explicit imports is maybe a better option, though we could do both.

pomadchin (approved):
Looks good for what it does, thanks!

PR #3356, opened 2021-03-11
update scalastyle; add scapegoat and scalafmt

philvarner:

# Overview

0. refactor scala versions out into variables
1. updates scalastyle dependency to the maintained 'beautiful-scala' package instead of the abandoned one
2. adds scapegoat configuration for static analysis (similar to config in stac4s)
3. add scalafmt (off by default-- maybe at some point all the code should be reformatted in one PR). The `.scalafmt.conf` was copied from stac4s

## Checklist

## Demo

none

## Notes

none

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,4 +1,7 @@
 version = 2.7.5
-align = more // For pretty alignment.
+align.preset = more // For pretty alignment.
 maxColumn = 120
 newlines.alwaysBeforeTopLevelStatements = true
+continuationIndent.callSite = 2
+continuationIndent.defnSite = 2
+continuationIndent.ctorSite = 2
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,0 +1,4 @@
+version = 2.7.5
+align = more // For pretty alignment.
+maxColumn = 120
+newlines.alwaysBeforeTopLevelStatements = true
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,7 +1,6 @@
 version = 2.7.5
-align.preset = more // For pretty alignment.
+align.preset = some
 maxColumn = 120
-newlines.alwaysBeforeTopLevelStatements = true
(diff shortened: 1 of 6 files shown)


build.sbt
```diff
@@ -1,8 +1,9 @@
 import sbt.Keys._
 
-ThisBuild / scalaVersion := "2.12.13"
+ThisBuild / scalaVersion := Version.scala211
```
pomadchin, line 3: I'd like to see `Version.scala212` as the default here. :D 
2.11 is not a priority atm, more like a legacy crossompilation that I'd really like to drop. 
We use GT mostly with 2.12.

.scalafmt.conf
```diff
@@ -0,0 +1,4 @@
+version = 2.7.5
+align = more // For pretty alignment.
+maxColumn = 120
+newlines.alwaysBeforeTopLevelStatements = true
```
pomadchin, line 4: Does `scalafmtAll` breaks everything? :/ I'd expect it to refactor 90% of our files.

build.sbt
philvarner: sorry, that must have been a typo when I originally made this change -- I've been cross-compiling most of the time and didn't notice.

.scalafmt.conf
philvarner: It does -- I mostly just wanted to get it in here, but have it be unused. But, I think it would be better to spend a bit of time aligning the rules with the existing code before merging. Putting PR in draft.

pomadchin: There is this tool that hopefully can generate the scalafmt file that would fit our needs: https://github.com/tanishiking/scalaunfmt 

(I would not be that positive about it), scalafmt is definitely going to break our type params with lots of imports formatting ):

philvarner: Great, I'll take a look at that tool.  I think it's more important to be consistent within the codebase than adhere to some arbitrary external standard, so I'll try to write the rules that way.

PR #3360, opened 2021-03-11
replace two uses of to[T] with equivalent toT

philvarner:

# Overview

There are two uses of `to[T]` methods in the proj4 module -- a `to[List]` and a ``to[Vector]`. These methods won't compile in 2.13, as to[Collection] method was replaced by the to(Collection) method, which is in the scala-collection-compat library. However, replacing these with `toList` and `toVector` is the simpler option, so doing that instead. 

## Checklist

## Demo

none

## Notes

none


pomadchin (approved):
Thanks!

PR #3362, opened 2021-03-12
fixes for non-fqn Seq being immutable in Scala 2.13

philvarner:

# Overview

These are the changes required to have the mutable/immutable Seq change in 2.13 work with cross-compilation.

**TODO** -- the changelog needs to be updated, since a couple of these involve public method signatures

## Checklist

## Demo

Optional. Screenshots/REPL

## Notes

Closes #XXX


pomadchin (approved):
Thanks! Let's merge it as it is, and we'll have a chance to create follow up PRs to fix more of these types if necessary.

PR #3369, opened 2021-03-30
Make JTSConfig load lazy

pomadchin:
# Overview

This PR fixes the JTS config loading ordering. That could cause bad JTS factories init, since sometimes the `reference.conf` file could be loaded into memory _after_ the call to the config. That could cause following errors:

```scala
Caused by: pureconfig.error.ConfigReaderException: Cannot convert configuration to a geotrellis.vector.conf.JtsConfig. Failures are:
  at the root:
    - (merge of system properties,reference.conf @ jar:file:/.../sbt/1.4.9/ssl-config-core_2.12-0.4.0.jar!/reference.conf: 1) Key not found: 'geotrellis'.
``` 

For the context: this could only happen in runs through SBT. 
I did a little bit of tracing down the history of this bug and it turned out that it appeared in `1.4.0` and all `1.4.x` versions are affected. However, in `1.5.x` (checked with `RC2`) this issue is gone. 

## Checklist


jisantuc (approved):
Tested with local publication in the setting that produced the bug, eeverything's great 🎉

PR #3378, opened 2021-04-09
Fix AWS S3LayerDeleter

### LINKED ISSUE #3371: S3LayerDeleter cannot handle over 1000 objects to delete
## Describe the bug

I have been experiencing problems deleting large datasets in AWS S3. The errors were pretty cryptic, but I've pinpointed the problem to the `S3LayerDeleter`'s call to `s3Client.deleteObjects(deleteRequest)`. 

The nested error is `software.amazon.awssdk.services.s3.model.S3Exception: The XML you provided was not well-formed or did not validate against our published schema`. I found similar issues for other projects and the root cause was sending a number of request to the AWS API exceeding the API's arbitrary limit. I found [this link](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html) when searching for delete requests limits. 

I've also tested the hypothesis, datasets containing less than 1000 objects can be deleted successfully while datasets containing over 1000 objects are not. 

## To Reproduce

- Create an S3 dataset containing over 1000 objects
- Fetch the LayerId of the dataset
- Call the S3LayerDeleter.delete(id: LayerId)

The actual output is a cryptic error.

The expected output is void and the dataset is deleted.

## Expected behavior

Dataset is deleted.

## Environment

- Java version: 1.8
- Scala version: 2.12.10
- GeoTrellis version: 3.5.2

## Proposed Solution

The solution seems relatively straightforward and minimal and involves a single file change to `S3LayerDeleter.scala`. I believe it would be a matter of splitting the deleteRequest object into chunks and iteratively calling the `s3Client.deleteObjects` on those chunks.

zacharyDez:
# Overview

Relates to previously closed #3371. 

An error is thrown due to trying to delete the attribute information of a layerId previously deleted. The fix was simply to remove the call to the attribute store outside of the foreach loop.

## Checklist

## Notes

Probably should consider implementing an integration/unit test 

Closes #3371


pomadchin:
@zacharyDez don’t worry, I think the change good as it is, thanks for catching the small bug 👍 

Hmm, could you rebase on top of master?

pomadchin (approved):
:+1: thank you!

PR #3384, opened 2021-04-13
Cleanup HBase dependencies

pomadchin:
# Overview

This PR removes unused HBase deps


pomadchin:
Testing the GT upd, noticed an issue with HBase deps, which is addressed in this PR https://github.com/locationtech/geotrellis/pull/3384

PR #3419, opened 2021-09-20
Add Buffer Tile

pomadchin:
# Overview

This PR adds BufferTile type to handle focal ops in RF.

## Checklist

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/BufferTile.scala
+++ b/raster/src/main/scala/geotrellis/raster/BufferTile.scala
@@ -16,11 +16,7 @@
 
 package geotrellis.raster
 
-import geotrellis.vector.Extent
-
 import spire.syntax.cfor._
-import geotrellis.util.Direction
-import geotrellis.util.Direction._
 
 /**
  * When combined with another BufferTile the two tiles will be aligned on (0, 0) pixel of tile center.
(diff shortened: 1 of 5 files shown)


raster/src/main/scala/geotrellis/raster/BufferTile.scala
```diff
@@ -0,0 +1,441 @@
...
+        cfor(0)(_ < rows, _ + 1) { row =>
+          cfor(0)(_ < cols, _ + 1) { col =>
+            tile.setDouble(col, row, f(getDouble(col, row), other.getDouble(col, row)))
+          }
+        }
+        tile
+    }
+  }
+
+  def mapTile(f: Tile => Tile): BufferTile = BufferTile(f(sourceTile), gridBounds)
```
pomadchin, line 440: @echeipesh how about making all functions to work within the window, but if user wants to apply smth beayond the window they need to do it explicitly. Mb the function name could be different though.

raster/src/main/scala/geotrellis/raster/mapalgebra/focal/BufferTileFocalMethods.scala
```diff
@@ -0,0 +1,90 @@
...
+package geotrellis.raster.mapalgebra.focal
+
+import geotrellis.raster._
+import geotrellis.util.MethodExtensions
+
+trait BufferTileFocalMethods extends MethodExtensions[BufferTile] {
+
+  /** Computes the minimum value of a neighborhood */
+  def focalMin(n: Neighborhood, bounds: Option[GridBounds[Int]] = None, target: TargetCell = TargetCell.All): BufferTile =
+    self.mapTile(_.focalMin(n, bounds, target))
```
pomadchin, line 26: In this case focal ops are applied to the underlying tile cc @echeipesh

raster/src/main/scala/geotrellis/raster/BufferTile.scala
```diff
@@ -0,0 +1,441 @@
...
+ * For instance: combining a tile padded with 5 pixels on all sides with tile padded with 3 pixels on all sides will
+ * result in buffer tile with 3 pixel padding on all sides.
+ *
+ * When combined with another BufferTile the operation will be executed over the maximum shared in
+ *
+ * TODO:
+ * - What should .map do? Map the buffer pixels or not?
+ * - toString method is friendly
+ * - mutable version makes sense
+ * - toBytes needs to encode padding size?
```
pomadchin, line 33: If we're ok with the explicit work with the underlying tile:
1. map would be applied to the window only (so it is consistent with any Tile behavior, with gets and foreaches)
2. still need a friednly toString
3. yes!
4. toBytes encodes window only; we don't have any reasons to encode gridBounds to the output array 🤷, no interfaces that would support reading from it.

pomadchin:
Going to merge it! We would still have a chance to change behavior and API if necessary. I would like to get it merged to remove BufferTile from RasterFrames in terms of https://github.com/locationtech/rasterframes/pull/567

PR #3433, opened 2021-10-25
Separate test code for modules with associated testkit to eliminate dependency cycle

metasim:
# Overview

The latest version of IntelliJ no longer handles previous workarounds to handle dependency cycle between a module's test scope and it's test-kit.

For example, `sbt` sees this:

    raster / Test +-> raster-testkit / Compile --> raster / Compile
                  |                                ^
                  +--------------------------------+

but IntelliJ sees this:

    raster / {Test|Compile} --> raster-testkit / Compile --+
                       ^                                   |
                       ------------------------------------+


pomadchin:
Closing it for now, I feel like to start this PR from scratch will be an easier lift. I feel like there may be another way to solve it?

PR #3440, opened 2021-11-27
Automate release process through CI

### LINKED ISSUE #3140: Automate release process through CI
Currently the release process for GeoTrellis is pretty manual thing. This goes against the need to release frequently and for multiple version of the library.

We've started adopting the release through CI workflows through smaller ecosystem libraries (ex:https://github.com/geotrellis/geotrellis-contrib/pull/236) and now its time to do this for the main repo.

Main goals are:
- Publish from CI on tags for release and snapshot builds

Sidegoals: 
- Reduce the number of scripts we're using to run CI and lean as much as possible on `sbt`
- sign all releases, SonaType as well LT

One thing that keeps things interesting is the need to release to LocationTech as well as SonaType repos. The release to LocationTech can only be done from LT Jenkins: https://ci.eclipse.org
But at least its just another nexus instance.

pomadchin:
# Overview

This PR adds the CI support and Scala 2.13.7 cross compilation.

Closes #3140


.circleci/config.yml
```diff
@@ -70,6 +73,6 @@ workflows:
               scala-version: ["2.12.15", "2.13.6"]
           filters:
             branches:
-              only: master
+              only: /.*/
```
pomadchin, line 76: On all but tag branches it publishes snapshots to sonatype.

version.sbt
```diff
@@ -1 +0,0 @@
-ThisBuild / version := "3.6.1-SNAPSHOT"
```
pomadchin, line 1: Fetched from tags now, see https://github.com/sbt/sbt-ci-release

layer/src/main/scala/geotrellis/layer/buffer/BufferTiles.scala
```diff
@@ -25,14 +25,11 @@ import geotrellis.raster.buffer.{BufferedTile, BufferSizes}
 import geotrellis.raster.buffer.Direction._
 import geotrellis.util._
 
-import org.apache.log4j.Logger
```
pomadchin, line 28: 🤦

PR #3453, opened 2022-03-24
Blanket exclude log4j. 

### LINKED ISSUE #3451: Exclude log4j 1.x dependency
Right now we don't have an exclusion of the log4j 1.x dep from the codebase, but still use log4j 2 -> 1 bridge implying that log4j is excluded by clients, i.e. [that's how it is done in RasterFrames](https://github.com/locationtech/rasterframes/blob/develop/project/RFDependenciesPlugin.scala#L76).

This task is to exclude log4j from libraries that can potentially bring it in and can make the assembly build process harder.

metasim:
Closes #3451

### DIFF
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -93,7 +93,7 @@ object Settings {
         case _ => publishTo.value
       }
     },
-
+    excludeDependencies += "log4j" % "log4j",
     credentials ++= List(
       Path.userHome / ".ivy2" / ".credentials",
       Path.userHome / ".sbt" / ".credentials"


project/Settings.scala
```diff
@@ -93,7 +93,7 @@ object Settings {
         case _ => publishTo.value
       }
     },
-
+    excludeDependencies += "log4j" % "log4j",
```
pomadchin, line 96: @metasim thx for the PR! 

As far as I know, `excludeDependencies`would not be reflected in the published artifacts (i.e. see https://repo1.maven.org/maven2/org/locationtech/geotrellis/geotrellis-geotools_2.12/3.6.1/geotrellis-geotools_2.12-3.6.1.pom), could you double check the behavior / switch to using `ExclusionRules` instead?

pomadchin (commented):
LGTM! But I'm not sure that this change will be reflected in the published packages pom files.

pomadchin:
@metasim I think I found a way to go the other direction. GT is a library, so I just moved that log4j bridge into the `test` scope so it won't affect users.

PR #3493, opened 2023-01-03
Downgrade scaffeine to 3.1.0

echeipesh:
This is the last version that has JDK 8 support

# Overview

Brief description of what this PR does, and why it is needed.

## Checklist

## Notes

Even the [latest Databricks runtime](https://docs.databricks.com/release-notes/runtime/12.0.html) runs on JDK 8. While EMR has moved on to support 12. There was no specific reason to bump this library so keeping it at this level reduced the number integration headaches.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -95,8 +95,7 @@ object Dependencies {
   val cassandraDriverCore         = "com.datastax.oss" % "java-driver-core"          % Version.cassandra
   val cassandraDriverQueryBuilder = "com.datastax.oss" % "java-driver-query-builder" % Version.cassandra
 
-  val scaffeine = "com.github.blemale"           %% "scaffeine" % "4.1.0"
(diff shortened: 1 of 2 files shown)


echeipesh (commented):
update

project/Dependencies.scala
```diff
@@ -95,8 +95,7 @@ object Dependencies {
   val cassandraDriverCore         = "com.datastax.oss" % "java-driver-core"          % Version.cassandra
   val cassandraDriverQueryBuilder = "com.datastax.oss" % "java-driver-query-builder" % Version.cassandra
 
-  val scaffeine = "com.github.blemale"           %% "scaffeine" % "4.1.0"
-  val caffeine  = "com.github.ben-manes.caffeine" % "caffeine"  % "2.9.3"
```
pomadchin, line 99: Ha, that's weird; I thought Caffeine 3+ is JDK11 only, nope?

PR #3517, opened 2023-08-07
Regrid: force crop to avoid going out of memory

jdries:
force cropping when regrid leads to much smaller tiles, to avoid memory multiplication in serialized rdd

# Overview
Issue: https://github.com/Open-EO/openeo-geotrellis-extensions/issues/191
Regrid uses cropping to create new tiles, but by default, cropping retains the original ArrayTile. So when regrid is used to go from large tile sizes (e.g. 1024 pixels) to smaller ones (e.g. 64 pixels) we get many cropped tiles all referencing these larger original tiles. When Spark then serializes these cropped tiles, the original arrays are all separately serialized, and we get a multiplication effect on the RDD size. Deserializing then can make executors go out of memory.

## Checklist

## Demo

Optional. Screenshots/REPL

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/regrid/Regrid.scala
+++ b/spark/src/main/scala/geotrellis/spark/regrid/Regrid.scala
@@ -118,11 +118,12 @@ object Regrid {
 
               val xSpan: Interval[Long] = oldXrange intersect newXrange
               val ySpan: Interval[Long] = oldYrange intersect newYrange
+              val forceCrop = newW < 0.6*oldW || newH < 0.6*oldH
               newKey ->
                 (oldTile.crop((xSpan.start - oldXstart).toInt,
                               (ySpan.start - oldYstart).toInt,
                               (xSpan.end - oldXstart).toInt,
-                              (ySpan.end - oldYstart).toInt),
+                              (ySpan.end - oldYstart).toInt),Crop.Options(force=forceCrop)),
                  ((xSpan.start - newXrange.start).toInt, (ySpan.start - newYrange.start).toInt)
                 )
             }


spark/src/main/scala/geotrellis/spark/regrid/Regrid.scala
```diff
@@ -118,11 +118,12 @@ object Regrid {
 
               val xSpan: Interval[Long] = oldXrange intersect newXrange
               val ySpan: Interval[Long] = oldYrange intersect newYrange
+              val forceCrop = newW < 0.6*oldW || newH < 0.6*oldH
               newKey ->
                 (oldTile.crop((xSpan.start - oldXstart).toInt,
                               (ySpan.start - oldYstart).toInt,
                               (xSpan.end - oldXstart).toInt,
-                              (ySpan.end - oldYstart).toInt),
+                              (ySpan.end - oldYstart).toInt),Crop.Options(force=forceCrop)),
```
pomadchin, line 126: Oops CI feels a bit unhappy about it. (2.12)

```scala
/usr/local/src/spark/src/main/scala/geotrellis/spark/regrid/Regrid.scala:126:92: ';' expected but ',' found         
(ySpan.end - oldYstart).toInt),Crop.Options(force=forceCrop)),
```

pomadchin (changes requested):
Sorry for a late review, but it looks like some tiny changes required to make compiler happy.

spark/src/main/scala/geotrellis/spark/regrid/Regrid.scala
```diff
@@ -118,11 +118,12 @@ object Regrid {
...
                  ((xSpan.start - newXrange.start).toInt, (ySpan.start - newYrange.start).toInt)
                 )
```
pomadchin, line 128: It does not allow me to add a code change suggestion, but should be smth like:

```scala
              val forceCrop = newW < 0.6 * oldW || newH < 0.6 * oldH
              newKey ->
                (
                  oldTile.crop(
                    (xSpan.start - oldXstart).toInt,
                    (ySpan.start - oldYstart).toInt,
                    (xSpan.end - oldXstart).toInt,
                    (ySpan.end - oldYstart).toInt,
                    Crop.Options(force = forceCrop)
                  ),
                  ((xSpan.start - newXrange.start).toInt, (ySpan.start - newYrange.start).toInt)
                )
```

Crop options is an argument to the crop function.

pomadchin: I made up a https://github.com/locationtech/geotrellis/pull/3518 to try out the CI, but please feel free to commit changes as a part of this PR as well. 

U can also let me pushing into your PR to speed up the progress! 

image

Issue #3541, opened 2024-06-04
RDD reproject does not correctly apply resampleMethod


jdries:
## Describe the bug

I can consistently reproduce a case where reprojecting an RDD to a lower resolution does not compute correct pixel values. The output selects one of the pixels, rather than respecting the sample method.

It happens for any method that uses the 'cs' argument here:
https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/resample/Resample.scala#L102
So Sum, Average, min, Mode, ... are all impacted.

I believe the bug is in this line:
https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/reproject/RasterRegionReproject.scala#L268

Because cellsize is supposed to be the _target_ cellsize, while on that line, the source cellsize seems to be passed on.  Hence, any subsequent computation in the resamplemethod just select one pixel where they are supposed to work on multiple pixels.

## To Reproduce

Create an rdd, and use the 'reproject' method with any of the above mentioned resample methods.
Make sure to resample to a lower resolution (so larger cellsize), so that multiple pixels are normally selected.

## Expected behavior

resample methods should behave as advertised.

## Environment
latest geotrellis version

pomadchin:
It makes sense, do you want to make a PR to address it? 

I’m 🌴 atm, so won’t be able to assist quickly, but we can include it into the upcoming release.

pomadchin:
Closed via https://github.com/locationtech/geotrellis/pull/3542

PR #3542, opened 2024-06-05
potential fix for resamplemethod in reproject

jdries:
https://github.com/locationtech/geotrellis/issues/3541

# Overview

Proposed fix for a very important bug in reprojecting to lower resolution.
I think it's better (more correct) than previous behaviour, but not sure yet if it is 100% accurate.

It seems to me that the downsampling now happens in the original projection system, and then it is warping to the new CRS. This will work fine if the low resolution pixel is rectangular in both projection systems.

## Checklist

## Demo

Optional. Screenshots/REPL

## Notes

Closes #XXX

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/reproject/RasterRegionReproject.scala
+++ b/raster/src/main/scala/geotrellis/raster/reproject/RasterRegionReproject.scala
@@ -191,7 +191,8 @@ object RasterRegionReproject {
       errorThreshold: Double
     ): Unit = {
       val trans = Proj4Transform(dest, src)
-      val resampler = Resample(resampleMethod, raster.tile, raster.extent, CellSize(raster.rasterExtent.cellwidth, raster.rasterExtent.cellheight))
+      val targetCellSizeInSrcCRS = rasterExtent.reproject(dest,src).cellSize
+      val resampler = Resample(resampleMethod, raster.tile, raster.extent, targetCellSizeInSrcCRS))
       val rowcoords = rowCoords(region, rasterExtent, trans, errorThreshold)
 
       if (raster.cellType.isFloatingPoint) {
@@ -262,10 +263,11 @@ object RasterRegionReproject {
       val trans = Proj4Transform(dest, src)
       val rowcoords = rowCoords(region, rasterExtent, trans, errorThreshold)
       val resampler: Array[Resample] = Array.ofDim[Resample](raster.tile.bandCount)
+      val targetCellSizeInSrcCRS = rasterExtent.reproject(dest,src).cellSize
 
       for (b <- 0 until raster.tile.bandCount) {
         val band: Tile = raster.tile.bands(b)
-        resampler(b) = Resample(resampleMethod, band, raster.extent, raster.rasterExtent.cellSize)
+        resampler(b) = Resample(resampleMethod, band, raster.extent, targetCellSizeInSrcCRS)
       }
 
       if (raster.cellType.isFloatingPoint) {
--- a/raster/src/main/scala/geotrellis/raster/reproject/RasterRegionReproject.scala
+++ b/raster/src/main/scala/geotrellis/raster/reproject/RasterRegionReproject.scala
@@ -192,14 +192,14 @@ object RasterRegionReproject {
     ): Unit = {
       val trans = Proj4Transform(dest, src)
 
-      val targetCellSizeInSrcCRS =
-        try{
-          rasterExtent.reproject(dest,src).cellSize
-        }catch {
-          case e:Exception => //reprojection errors happen when going outside valid area of projection, need a better fix here
-            raster.rasterExtent.cellSize
-        }
-      val resampler = Resample(resampleMethod, raster.tile, raster.extent, targetCellSizeInSrcCRS)
+      val resampler = resampleMethod match {
+        case resampleMethod1: PointResampleMethod =>
+          Resample(resampleMethod1, raster.tile, raster.extent)
+        case _ =>
+          //throws GeoAttrsError when applied for invalid extent
+          val targetCellSizeInSrcCRS = rasterExtent.reproject(dest, src).cellSize
+          Resample(resampleMethod, raster.tile, raster.extent, targetCellSizeInSrcCRS)
+      }
       val rowcoords = rowCoords(region, rasterExtent, trans, errorThreshold)
 
       if (raster.cellType.isFloatingPoint) {
@@ -271,13 +271,8 @@ object RasterRegionReproject {
       val rowcoords = rowCoords(region, rasterExtent, trans, errorThreshold)
       val resampler: Array[Resample] = Array.ofDim[Resample](raster.tile.bandCount)
 
-      val targetCellSizeInSrcCRS =
-        try{
-          rasterExtent.reproject(dest,src).cellSize
-        }catch {
-          case e:Exception => //reprojection errors happen when going outside valid area of projection, need a better fix here
-            raster.rasterExtent.cellSize
-        }
+      val targetCellSizeInSrcCRS = rasterExtent.reproject(dest,src).cellSize
+
 
       for (b <- 0 until raster.tile.bandCount) {
         val band: Tile = raster.tile.bands(b)
--- a/raster/src/main/scala/geotrellis/raster/reproject/RasterRegionReproject.scala
+++ b/raster/src/main/scala/geotrellis/raster/reproject/RasterRegionReproject.scala
@@ -191,8 +191,15 @@ object RasterRegionReproject {
       errorThreshold: Double
     ): Unit = {
       val trans = Proj4Transform(dest, src)
-      val targetCellSizeInSrcCRS = rasterExtent.reproject(dest,src).cellSize
-      val resampler = Resample(resampleMethod, raster.tile, raster.extent, targetCellSizeInSrcCRS))
+
+      val targetCellSizeInSrcCRS =
+        try{
+          rasterExtent.reproject(dest,src).cellSize
+        }catch {
+          case e:Exception => //reprojection errors happen when going outside valid area of projection, need a better fix here
+            raster.rasterExtent.cellSize
+        }
+      val resampler = Resample(resampleMethod, raster.tile, raster.extent, targetCellSizeInSrcCRS)
       val rowcoords = rowCoords(region, rasterExtent, trans, errorThreshold)
 
       if (raster.cellType.isFloatingPoint) {
@@ -263,7 +270,14 @@ object RasterRegionReproject {
       val trans = Proj4Transform(dest, src)
       val rowcoords = rowCoords(region, rasterExtent, trans, errorThreshold)
       val resampler: Array[Resample] = Array.ofDim[Resample](raster.tile.bandCount)
-      val targetCellSizeInSrcCRS = rasterExtent.reproject(dest,src).cellSize
+
+      val targetCellSizeInSrcCRS =
+        try{
+          rasterExtent.reproject(dest,src).cellSize
+        }catch {
+          case e:Exception => //reprojection errors happen when going outside valid area of projection, need a better fix here
+            raster.rasterExtent.cellSize
+        }
 
       for (b <- 0 until raster.tile.bandCount) {
         val band: Tile = raster.tile.bands(b)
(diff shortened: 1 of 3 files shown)


jdries:
The fix works but I had to fall back to old (bad) behaviour in cases at the edge of valid target crs bounds. My method to compute target cellsize resulted in an error there.

raster/src/main/scala/geotrellis/raster/reproject/RasterRegionReproject.scala
```diff
@@ -191,7 +191,15 @@ object RasterRegionReproject {
...
       val trans = Proj4Transform(dest, src)
-      val resampler = Resample(resampleMethod, raster.tile, raster.extent, CellSize(raster.rasterExtent.cellwidth, raster.rasterExtent.cellheight))
+
+      val targetCellSizeInSrcCRS =
+        try{
+          rasterExtent.reproject(dest,src).cellSize
+        }catch {
+          case e:Exception => //reprojection errors happen when going outside valid area of projection, need a better fix here
+            raster.rasterExtent.cellSize
+        }
```
pomadchin, line 201: are there any tests that cover this case?

jdries: Not yet, but I now had some time to dig deeper.
The original stack trace was:

	```
at geotrellis.raster.GridExtent.<init>(GridExtent.scala:46)
	at geotrellis.raster.GridExtent.<init>(GridExtent.scala:58)
	at geotrellis.raster.reproject.ReprojectRasterExtent$.apply(ReprojectRasterExtent.scala:71)
	at geotrellis.raster.GridExtent$gridExtentMethods.$anonfun$reproject$2(GridExtent.scala:432)
	at scala.Option.getOrElse(Option.scala:189)
	at geotrellis.raster.GridExtent$gridExtentMethods.reproject(GridExtent.scala:432)
	at geotrellis.raster.GridExtent$gridExtentMethods.reproject(GridExtent.scala:435)
	at org.openeo.geotrellis.reproject.RasterRegionReproject$$anon$2.reprojectToBuffer(RasterRegionReproject.scala:268)
	at org.openeo.geotrellis.reproject.RasterRegionReproject$$anon$2.regionReproject(RasterRegionReproject.scala:229)
	at org.openeo.geotrellis.reproject.TileRDDReproject$.createCombiner$1(TileRDDReproject.scala:192)
	at org.openeo.geotrellis.reproject.TileRDDReproject$.$anonfun$apply$6(TileRDDReproject.scala:206)
```

and got triggered by a more complicated test case here:
https://github.com/Open-EO/openeo-geotrellis-extensions/blob/106784e1cb83a87f1bacef3beb1b9df77c5ecb65/openeo-geotrellis/src/test/scala/org/openeo/geotrellis/OpenEOProcessesSpec.scala#L582

I could reproduce it in a simpler test, getting a similar but not exactly the same stack trace:
  ```
  it("should (approximately) match a GDAL average interpolation on nlcd tile at edge of valid bounds") {

      val tile = this.createTile(Array(1,4,1,4,1,4,1,4),4,2)
      val raster = Raster(tile,Extent(-40.50000387499999, -89.99999, 40.499993875000015, -51.749994249999986))
      val tempTiff = File.createTempFile("toReproject",".tif")
      GeoTiff(raster, LatLng).write(tempTiff.getPath)

      val tiffRe = RasterExtent(Extent(-4452779.631730943, -6.325724963354129E7, 0.0, -5.839130735403811E7), CellSize(15000.0,15000.0))
      val alignment = TargetRegion(tiffRe)

      val rs = GeoTiffReprojectRasterSource(tempTiff.getPath,WebMercator, alignment, Average)
      val reprojectedFromRS = rs.read().get.tile.band(0)

      val reprojected = raster.reproject(LatLng,WebMercator, Options(method = Average, errorThreshold = 0.0,targetCellSize = Some(CellSize(1000,1000))))

      val refTile = this.createTile(Array(2.5,2.5),2226, 10363)
      //assertEqual(refTile,reprojected.tile,0.1)
      assertEqual(reprojectedFromRS,reprojected.tile,0.1)
    }
```

is giving:

```

invalid rows: 0
geotrellis.raster.GeoAttrsError: invalid rows: 0
	at geotrellis.raster.GridExtent.<init>(GridExtent.scala:46)
	at geotrellis.raster.GridExtent.<init>(GridExtent.scala:58)
	at geotrellis.raster.reproject.ReprojectRasterExtent$.apply(ReprojectRasterExtent.scala:71)
	at geotrellis.raster.reproject.ReprojectRasterExtent$.apply(ReprojectRasterExtent.scala:76)
	at geotrellis.raster.geotiff.GeoTiffReprojectRasterSource.closestTiffOverview$lzycompute(GeoTiffReprojectRasterSource.scala:86)
	at geotrellis.raster.geotiff.GeoTiffReprojectRasterSource.closestTiffOverview(GeoTiffReprojectRasterSource.scala:81)
	at geotrellis.raster.geotiff.GeoTiffReprojectRasterSource.readBounds(GeoTiffReprojectRasterSource.scala:110)
	at geotrellis.raster.geotiff.GeoTiffReprojectRasterSource.read(GeoTiffReprojectRasterSource.scala:98)
	at geotrellis.raster.geotiff.GeoTiffReprojectRasterSource.read(GeoTiffReprojectRasterSource.scala:94)
	at geotrellis.raster.RasterSource.read(RasterSource.scala:134)
	at geotrellis.raster.reproject.ReprojectSpec.$anonfun$new$5(ReprojectSpec.scala:92)
```

What's happening here is that we are going across valid bounds of webmercator. The root cause is then in these lines:
https://github.com/VitoTAP/geotrellis/blob/b071b333581ed7206ad14a06b1b70cfde6fb65a7/raster/src/main/scala/geotrellis/raster/reproject/ReprojectRasterExtent.scala#L63

The 'toLong' will round down to zero, which is not accepted later on in the code. 
When that case happens, I think it also doesn't really matter any more what cellsize is used, because we warp into an area with cols or rows == 0.

pomadchin: ic ic; the thing is that this try / catch is a bit sketchy 🤔 we definitely need to find a better way addressing it / or at least make exception more specific.

jdries: Here's a case that really isolates and exactly reproduces the issue. Probably these extents are fairly problematic to begin with:

```
 it("should (approximately) match a GDAL average interpolation on nlcd tile at edge of valid bounds") {

      val tile = this.createNoData(288,272,ByteConstantNoDataCellType)
      val raster = Raster(tile,Extent(-40.50000387499999, -89.99999, 40.499993875000015, -51.749994249999986))
      val tempTiff = File.createTempFile("toReproject",".tif")
      GeoTiff(raster, LatLng).write(tempTiff.getPath)

      val tiffRe = RasterExtent(Extent(-4452779.631730943, -6.325724963354129E7, 0.0, -5.839130735403811E7), CellSize(14744.303416327626,16112.391653984045))

      val rr = implicitly[RasterRegionReproject[Tile]]
      val reprojectedRasterRegion: Raster[Tile] = rr.regionReproject(
        raster,
        LatLng,
        WebMercator,
        tiffRe,
        tiffRe.extent.toPolygon(),
        Average,
        0.2
      )

      //succesfull if it doesn't throw an exception
    }
```

pomadchin: @jdries hm there is an issue with this last example since "rasterExtent extent and resolution of target raster"; 

`val targetCellSizeInSrcCRS = rasterExtent.reproject(dest, src).cellSize` makes not a lot of sense since it should be in the target resolution. 

I think sadly we need a more real example.

pomadchin: Hm ok I see what you need it for.

raster/src/main/scala/geotrellis/raster/reproject/RasterRegionReproject.scala
```diff
@@ -263,9 +271,17 @@ object RasterRegionReproject {
...
+
       for (b <- 0 until raster.tile.bandCount) {
         val band: Tile = raster.tile.bands(b)
-        resampler(b) = Resample(resampleMethod, band, raster.extent, raster.rasterExtent.cellSize)
+        resampler(b) = Resample(resampleMethod, band, raster.extent, targetCellSizeInSrcCRS)
```
pomadchin, line 284: should not there be a cellSize of a target?

raster/src/main/scala/geotrellis/raster/reproject/RasterRegionReproject.scala
```diff
@@ -191,7 +191,15 @@ object RasterRegionReproject {
...
-      val resampler = Resample(resampleMethod, raster.tile, raster.extent, CellSize(raster.rasterExtent.cellwidth, raster.rasterExtent.cellheight))
+
+      val targetCellSizeInSrcCRS =
+        try{
+          rasterExtent.reproject(dest,src).cellSize
+        }catch {
+          case e:Exception => //reprojection errors happen when going outside valid area of projection, need a better fix here
+            raster.rasterExtent.cellSize
+        }
+      val resampler = Resample(resampleMethod, raster.tile, raster.extent, targetCellSizeInSrcCRS)
```
pomadchin, line 202: I see, the issue is that the targetCellSize was / is in the wrong CRS.

raster/src/main/scala/geotrellis/raster/reproject/RasterRegionReproject.scala
```diff
@@ -191,7 +191,15 @@ object RasterRegionReproject {
       errorThreshold: Double
     ): Unit = {
       val trans = Proj4Transform(dest, src)
-      val resampler = Resample(resampleMethod, raster.tile, raster.extent, CellSize(raster.rasterExtent.cellwidth, raster.rasterExtent.cellheight))
+
+      val targetCellSizeInSrcCRS =
+        try{
+          rasterExtent.reproject(dest,src).cellSize
```
pomadchin, line 197: I wonder if we have a broken function.

raster/src/main/scala/geotrellis/raster/reproject/RasterRegionReproject.scala
pomadchin: I compared the behavior with the GDAL behavior. `/tmp/raster-lat-lng.tif` is `GeoTiff(raster, LatLng).write("/tmp/raster-lat-lng.tif")`

```bash
$ gdalwarp -t_srs EPSG:3857 -tr 14744.303416327626 16112.391653984045 -te -4452779.631730943 -6.325724963354129E7 0.0 -5.839130735403811E7  /tmp/raster-lat-lng.tif /tmp/raster-merc.tif`
$ gdalwarp -t_srs EPSG:4326  /tmp/raster-merc.tif /tmp/raster-merc-lat.tif
Creating output file that is 427P x 0L.
ERROR 1: Attempt to create 427x0 dataset is illegal,sizes must be larger than zero.
```

This is exactly what happens in our reproject function: `newCols: 427 newRows: 0` which throws `geotrellis.raster.GeoAttrsError: invalid rows: 0`

pomadchin: I think the change is good, sadly I don't think we can merge this workaround with throws; it simply uses incorrect resolutions as an input for the functions. 

Mb it is acceptable (it worked somehow up until now) but not sure that's what we really need to do here 🤔 The error thrown seems to be legit.

jdries: Hi @pomadchin, the error is indeed legit for sure. The biggest problem is that if we start throwing it, some existing code might break, because now it is perfectly possible to do reprojection outside the valid bounds of the coordinate system.
However, from my side, this would be fine, it's at least better than having resampling fully broken. So shall I update the PR to simply not catch the exception, so that downstream users can deal with it?
Whatever your preferred way forward is, I'd very much like to fix the resampling!

pomadchin: :+1: yes, the change is very good; the error catch is no good, it is expected to fail :+1:

pomadchin: If you could remove the error catching we can merge it in I think.

jdries: committed it, I did still came up with a clean way to keep the 'old' behaviour intact when using a resample method that does not require computing this cellsize. This will reduce upgrade issues for a lot of users.

jdries:
@pomadchin  ready to merge?

pomadchin:
@jdries I'll take a look this week!

pomadchin:
Thank you!

PR #3544, opened 2024-07-08
Fixed the issue that vrt.xml is empty when written out

dingsl-giser:
Add `writer.flush()` when writing xml to vrt, fix the problem of writing empty vrt.xml when writing out COG.

### DIFF
--- a/store/src/main/scala/geotrellis/store/cog/vrt/VRT.scala
+++ b/store/src/main/scala/geotrellis/store/cog/vrt/VRT.scala
@@ -28,22 +28,22 @@ import scala.xml.{Elem, XML}
 
 
 case class VRT(
-  gridBounds: GridBounds[Int],
-  layout: LayoutDefinition,
-  extent: Extent,
-  cellType: CellType,
-  crs: CRS,
-  bands: List[Elem]
-) {
+                gridBounds: GridBounds[Int],
+                layout: LayoutDefinition,
+                extent: Extent,
+                cellType: CellType,
+                crs: CRS,
+                bands: List[Elem]
+              ) {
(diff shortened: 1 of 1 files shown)


store/src/main/scala/geotrellis/store/cog/vrt/VRT.scala
```diff
@@ -172,6 +172,7 @@ object VRT {
       doctype = null
     )
 
+    writer.flush()
```
pomadchin, line 175: Thanks! I wonder how did it work before (:

store/src/main/scala/geotrellis/store/cog/vrt/VRT.scala
```diff
@@ -28,22 +28,22 @@ import scala.xml.{Elem, XML}
...
 
   /**
-    * Calculates GeoTransform attributes
-    *
-    * TODO: refactor, code style kept to follow GDAL: https://github.com/OSGeo/gdal/blob/9a21e8dcaf36a7e046ee87cd57c8c03812dd20ed/gdal/frmts/sde/sdedataset.cpp
-    */
+   * Calculates GeoTransform attributes
+   *
+   * TODO: refactor, code style kept to follow GDAL: https://github.com/OSGeo/gdal/blob/9a21e8dcaf36a7e046ee87cd57c8c03812dd20ed/gdal/frmts/sde/sdedataset.cpp
+   */
```
pomadchin, line 46: Could u revert these changes and preserve the original fmt?

pomadchin (approved):
LGTM! Thanks for the fix! There is a minor comment to address tho around formatting, rather than that LGTM.

PR #3545, opened 2024-07-09
Fixed the issue that the generated vrt.xml is empty when writing out cog

dingsl-giser:
Add `writer.flush()` when writing xml to vrt, fix the problem of writing empty vrt.xml when writing out cog.


dingsl-giser:
@pomadchin I resubmitted the PR, hoping it can be merged.

pomadchin (approved):
Thanks!

PR #3552, opened 2024-10-10
ConstantTile with nodata: support correct celltype conversion

### LINKED ISSUE #3525: Convert NODATA ByteConstantTile and ShortConstantCellType: unexpected result
## Describe the bug

A NODATA `ByteConstantTile` that is converted to its own `CellType`, is no longer NODATA.

## To Reproduce

```scala
import geotrellis.raster.{ByteConstantTile, byteNODATA}

val t = ByteConstantTile(byteNODATA, cols = 1, rows = 1)
assert(t.isNoDataTile)
assert(t.convert(t.cellType).isNoDataTile) // boom
```

## Expected behavior

Seeing as the `CellType` doesn't change, this should be a no-op and the tile should remain NODATA.

## Environment

- Java version: openjdk version "1.8.0_382"
- Scala version: 2.12.15
- GeoTrellis version: 3.6.0

jdries:
https://github.com/locationtech/geotrellis/issues/3525

# Overview
Test and fix for the issue above.
## Checklist

Closes #3525

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/ConstantTile.scala
+++ b/raster/src/main/scala/geotrellis/raster/ConstantTile.scala
@@ -19,8 +19,9 @@ package geotrellis.raster
 import geotrellis.vector.Extent
 
 import java.nio.ByteBuffer
-
 import spire.syntax.cfor._
+
+import java.lang
 
 
 /**
@@ -82,16 +83,21 @@ abstract class ConstantTile extends Tile {
     * @return            The new Tile
     */
   def convert(newType: CellType): Tile = {
-    newType match {
-      case BitCellType => new BitConstantTile(if (iVal == 0) false else true, cols, rows)
-      case ct: ByteCells => ByteConstantTile(iVal.toByte, cols, rows, ct)
-      case ct: UByteCells => UByteConstantTile(iVal.toByte, cols, rows, ct)
-      case ct: ShortCells => ShortConstantTile(iVal.toShort , cols, rows, ct)
-      case ct: UShortCells =>  UShortConstantTile(iVal.toShort , cols, rows, ct)
-      case ct: IntCells =>  IntConstantTile(iVal , cols, rows, ct)
-      case ct: FloatCells => FloatConstantTile(dVal.toFloat , cols, rows, ct)
-      case ct: DoubleCells => DoubleConstantTile(dVal, cols, rows, ct)
+    if(dVal.isNaN && !newType.isInstanceOf[NoNoData]) {
+      ConstantTile.emptyTile(newType, cols, rows)
+    }else {
+      newType match {
+        case BitCellType => new BitConstantTile(if (iVal == 0) false else true, cols, rows)
+        case ct: ByteCells => ByteConstantTile(iVal.toByte, cols, rows, ct)
+        case ct: UByteCells => UByteConstantTile(iVal.toByte, cols, rows, ct)
+        case ct: ShortCells => ShortConstantTile(iVal.toShort , cols, rows, ct)
+        case ct: UShortCells =>  UShortConstantTile(iVal.toShort , cols, rows, ct)
+        case ct: IntCells =>  IntConstantTile(iVal , cols, rows, ct)
+        case ct: FloatCells => FloatConstantTile(dVal.toFloat , cols, rows, ct)
+        case ct: DoubleCells => DoubleConstantTile(dVal, cols, rows, ct)
+      }
     }
+
   }
 
   def interpretAs(newCellType: CellType): Tile =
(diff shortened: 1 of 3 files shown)


raster/src/main/scala/geotrellis/raster/ConstantTile.scala
```diff
@@ -19,9 +19,10 @@ package geotrellis.raster
 import geotrellis.vector.Extent
 
 import java.nio.ByteBuffer
-
 import spire.syntax.cfor._
 
+import java.lang
```
moradology, line 24: Not sure this is necessary. Should be implicitly available already

raster/src/main/scala/geotrellis/raster/ConstantTile.scala
```diff
@@ -243,6 +249,46 @@ object ConstantTile {
...
+   * Create a [[ConstantTile]] that is set to the nodata value of the given celltype.
+   *
+   * @param   t      The [[CellType]] of the new [[ConstantTile]], it should have a NoData value.
+   * @param   cols   The number of columns that the new [[ConstantTile]] should have
+   * @param   rows   The number of rows that the new [[ConstantTile]] should have
+   * @return         The new [[ConstantTile]]
+   */
+  def emptyTile(t: CellType, cols: Int, rows: Int): Tile = {
+    if(t.isInstanceOf[NoNoData]) {
+      throw new IllegalArgumentException(s"Can not construct an empty constant tile for a celltype without nodata handling: $t")
```
pomadchin, line 263: Why is it not possible?

raster/src/main/scala/geotrellis/raster/ConstantTile.scala
```diff
@@ -243,6 +249,46 @@ object ConstantTile {
...
+    }else{
+      t match {
+        case _: BitCells => BitConstantTile(false, cols, rows)
+        case ct: ByteUserDefinedNoDataCellType => ByteConstantTile(ct.noDataValue ,cols, rows, ct)
+        case ct: ByteCells => ByteConstantTile(t.asInstanceOf[HasNoData[Byte]].noDataValue ,cols, rows, ct)
```
moradology, line 268: I think we can capture all of the cases here with e.g. [these types](https://github.com/locationtech/geotrellis/blob/2cc473aaa58c544651fd0e541ae4af0fe0c1c34a/raster/src/main/scala/geotrellis/raster/CellType.scala#L347-L350) and avoid casting altogether. Should be safe since you explicitly check above to make sure the `CellType` extends `HasNoData`

An example of the change might be something like this:
before:
```
case ct: ByteCells => ByteConstantTile(t.asInstanceOf[HasNoData[Byte]].noDataValue ,cols, rows, ct)
```

After:
```
case ct: ByteConstantNoDataCellType => ByteConstantTile(ct.noDataValue ,cols, rows, ct)
```

moradology (changes requested):
A couple of small things but I think this looks good after that

raster/src/main/scala/geotrellis/raster/ConstantTile.scala
```diff
@@ -82,16 +83,21 @@ abstract class ConstantTile extends Tile {
...
-    newType match {
-      case BitCellType => new BitConstantTile(if (iVal == 0) false else true, cols, rows)
-      case ct: ByteCells => ByteConstantTile(iVal.toByte, cols, rows, ct)
-      case ct: UByteCells => UByteConstantTile(iVal.toByte, cols, rows, ct)
-      case ct: ShortCells => ShortConstantTile(iVal.toShort , cols, rows, ct)
-      case ct: UShortCells =>  UShortConstantTile(iVal.toShort , cols, rows, ct)
-      case ct: IntCells =>  IntConstantTile(iVal , cols, rows, ct)
-      case ct: FloatCells => FloatConstantTile(dVal.toFloat , cols, rows, ct)
-      case ct: DoubleCells => DoubleConstantTile(dVal, cols, rows, ct)
+    if(dVal.isNaN && !newType.isInstanceOf[NoNoData]) {
```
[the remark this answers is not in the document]
jdries: done

raster/src/main/scala/geotrellis/raster/ConstantTile.scala
jdries: Not sure, but what would I use as nodata value?
Perhaps for Float and Double celltype, NaN is fine, but what about IntCellType? Just use 0 in that case?

raster/src/main/scala/geotrellis/raster/ConstantTile.scala
jdries: Ok, makes sense. Committed a first attempt, but had a local compile issue that I don't get yet, so to be seen if CI manages to build it.

jdries: @moradology  it also failed in CI. I created this code some time ago, and am guessing that I also resorted to the casting because I couldn't find a better solution. Can you perhaps tell if it can be solved more elegantly?

raster/src/main/scala/geotrellis/raster/ConstantTile.scala
pomadchin: Default to the CellType NoData i.e. for int its the int nodaa, for float its float nodata and for double it is the double nodata (min values)

pomadchin: https://github.com/locationtech/geotrellis/blob/24f62711f6886eb64bc1622563d6213d5ee0915e/raster/src/main/scala/geotrellis/raster/package.scala#L105-L111

raster/src/main/scala/geotrellis/raster/ConstantTile.scala
pomadchin: Also probably imports are in place, and the syntax is not quite there, these are singletons:

```scala
case ct @ ByteConstantNoDataCellType => ByteConstantTile(ct.noDataValue ,cols, rows, ct)
case ct @ UByteConstantNoDataCellType => UByteConstantTile(ct.noDataValue ,cols, rows, ct)
```

raster/src/main/scala/geotrellis/raster/ConstantTile.scala
pomadchin: I'm looking through the codebase and I think this method should be called `empty` and implementation is probably a +- copy paste of https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/ArrayTile.scala#L451-L461

raster/src/main/scala/geotrellis/raster/ConstantTile.scala
jdries: Thanks, I'm discovering new Scala things here. What eventually worked for me was using `ByteConstantNoDataCellType.type`. The `@` didn't seem to work, but I probably missed something.

raster/src/main/scala/geotrellis/raster/ConstantTile.scala
jdries: rename done, this new method is indeed very similar to the one in ArrayTile, guess it's not bad to complete the API like this.
Also removed the exception, added more cases.

raster/src/main/scala/geotrellis/raster/ConstantTile.scala
```diff
@@ -243,6 +243,45 @@ object ConstantTile {
...
+   *
+   * @param   t      The [[CellType]] of the new [[ConstantTile]].
+   * @param   cols   The number of columns that the new [[ConstantTile]] should have
+   * @param   rows   The number of rows that the new [[ConstantTile]] should have
+   * @return         The new [[ConstantTile]]
+   */
+  def empty(t: CellType, cols: Int, rows: Int): Tile = {
+    t match {
+      case _: BitCells => BitConstantTile(false, cols, rows)
+      case ct: ByteUserDefinedNoDataCellType => ByteConstantTile(ct.noDataValue ,cols, rows, ct)
```
pomadchin, line 259: nice, super minor: it looks formatting is a bit off in the entire function; that space before the comma :D 

`(ct.noDataValue ,cols, rows, ct)` => `(ct.noDataValue, cols, rows, ct)`

pomadchin:
Unfotunately pushes into this PR branches are forbidden, I forked off to cleanup the fix a bit: https://github.com/locationtech/geotrellis/pull/3553

Issue #3562, opened 2025-02-02
tile.reproject  can not use ,and so many method can not straightly use  in scala 2.13 and cross scala3


mullerhai:
## Describe the bug

polygon.reproject ()  method  not have in 3.7.1 ?  

```scala
   val collectionReader: CollectionLayerReader[LayerId] = CollectionLayerReader(attributeStore, catalogPath)

    val polygon = Polygon((-71.04764,-15.756736), (-71.04764,-15.75211), (-71.036997,-15.75211), (-71.036997,-15.756736), (-71.04764,-15.756736))

    val polygon_reprojected = polygon.reproject(LatLng, layerMetadata.crs)
```

PngRenderMethods.renderPng()  can not use 
IntArrayTile.asciiDraw（) can not straightly use ,did need  some implicit import ?

```scala
package catalog

import geotrellis.layer.SpatialKey
import geotrellis.raster.MultibandTile
import geotrellis.raster.render.ColorRamps
import geotrellis.store.AttributeStore
import geotrellis.store.{LayerId, ValueReader}

//import geotrellis.spark.store.hadoop.geotiff.AttributeStore
//import geotrellis.spark.{LayerId, SpatialKey}
//import geotrellis.spark.io.{AttributeStore, SpatialKeyFormat, ValueReader}
//import geotrellis.spark.io.avro.codecs.Implicits._
object App {
  def main(args: Array[String]): Unit = {
    val catalogPath = new java.io.File("generated/catalog").toURI
    val attributeStore: AttributeStore = AttributeStore(catalogPath)
    val valueReader: ValueReader[LayerId] = ValueReader(attributeStore, catalogPath)

    val zoom = 13
    val x = 2479
    val y = 4459

    val layerId =LayerId("landsat",zoom)
    val reader = valueReader.reader[SpatialKey, MultibandTile](layerId)
    val tile = reader.read(SpatialKey(x, y))
    tile.band(0).renderPng(ColorRamps.BlueToRed).write("generated/from_catalog.png")
  }
}
```

D:\code\scala_gis\src\main\scala\catalog\App.scala:23:72
Cannot find AvroRecordCodec for geotrellis.layer.SpatialKey. Try to import geotrellis.spark.io.avro.codecs.Implicits._
    val reader = valueReader.reader[SpatialKey, MultibandTile](layerId)

## To Reproduce

Provide as able:[geotrellis](https://github.com/slawler/geotrellis)

- Steps to reproduce the behavior
- Code example
[...]

PR #3563, opened 2025-02-02
Fix implicitNotFound error and rename RGBAMethods implicit class

pomadchin:
# Overview

This PR renames and outdated implicitNotFound error and renames RGBA implicit class Methods to avoid issues when using in Scala 3

## Checklist

Linked https://github.com/locationtech/geotrellis/issues/3562

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/render/Implicits.scala
+++ b/raster/src/main/scala/geotrellis/raster/render/Implicits.scala
@@ -22,7 +22,7 @@ import geotrellis.raster.{Tile, MultibandTile}
 object Implicits extends Implicits
 
 trait Implicits {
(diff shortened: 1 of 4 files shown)


raster/src/main/scala/geotrellis/raster/render/Implicits.scala
```diff
@@ -22,7 +22,7 @@ import geotrellis.raster.{Tile, MultibandTile}
 object Implicits extends Implicits
 
 trait Implicits {
-  implicit class RGBA(val int: Int) {
+  implicit class RGBAMethods(val int: Int) {
```
pomadchin, line 25: TODO: should be a method doing conversion into the AnyVal class.

PR #3570, opened 2025-03-20
set gdal roles for well known tags

### LINKED ISSUE #3496: Geotiff Tags not fully compatible with GDAL
## Describe the bug

It seems not possible to write a band name to geotiff metadata that is compatible with GDAL.
Geotrellis writes this:
```
<GDALMetadata>
  <Item name="DESCRIPTION" sample="0">B02</Item>
</GDALMetadata>

```

While GDAL would like to also see a "role" attribute:
```
<GDALMetadata>
  <Item name="DESCRIPTION" sample="0" role="description">B02</Item>
</GDALMetadata>
```

In the Geotrellis code, it is very clear that the role attribute is not used:
https://github.com/locationtech/geotrellis/blob/fa88aeddab0849cdb8f830d3747554bc7adceb1f/raster/src/main/scala/geotrellis/raster/io/geotiff/Tags.scala#L52

Originally reported here:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/257

## To Reproduce
Simply write a geotiff with tags, trying to set a DESCRIPTION on the bands.

## Expected behavior

Geotrellis can generate metadata compliant with GDAL, as it already tries to do that.

## Environment

Geotrellis 3.6.x

jdries:
https://github.com/locationtech/geotrellis/issues/3496

# Overview

Certain tags in GDAL are standardized, but this requires a 'role' attribute to be set.
In this PR, I take the approach of simply having a static list of such 'well-known' tags that seem to be part of the GDAL standard.

## Checklist

## Notes

Caveat: if gdal invents new roles, we need to update the list.

Closes #3496

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - ConstantTile with NoData: support idempotent CellType conversions [#3553](https://github.com/locationtech/geotrellis/pull/3553)
 - GeoTiffReader Unknown tags skip fix [#3557](https://github.com/locationtech/geotrellis/pull/3557)
 - Fix implicitNotFound error and rename RGBAMethods implicit class [#3563](https://github.com/locationtech/geotrellis/pull/3563)
+- Set 'role' attribute in GDALMetadata tag written to tiff header, to match GDAL behaviour. [#3496](https://github.com/locationtech/geotrellis/issues/3496)
 
 ## [3.7.1] - 2024-01-08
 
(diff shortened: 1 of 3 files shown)


jdries:
Ready for review, I found more tags that require the role attribute by looking at GDAL source itself:
https://github.com/OSGeo/gdal/blob/b8f2a16baa8fe866c8cd053d89120844d51ca94f/frmts/gtiff/gtiffdataset_write.cpp#L4208

We'd appreciate getting this merged, because we found out that fixing the tiff header afterwards breaks COG compatibility.

pomadchin:
@jdries I'll take a look shortly and cut a release for you; there is also a tiny Proj4J release https://github.com/locationtech/proj4j/releases/tag/v1.4.0 that we'd needto include :+1: 

Thanks!

pomadchin:
Also if you could take a look into the failing CI it would def help to get it merged faster!

jdries:
CI ok!

PR #3586, opened 2025-08-25
Do not depend on private Spark API, avoids sealing violation

### LINKED ISSUE #3585: Avoid using spark private API, and sealing violation
## Describe the bug

Packages can be 'sealed', so multiple jars can no longer define the same package. This avoids unwanted access to package-private classes. When trying to do so, this exception is thrown:

`java.lang.SecurityException: sealing violation: can't seal package org.apache.spark.rdd: already defined `

There's one case in Geotrellis where a Spark package is extended:

https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/org/apache/spark/rdd/FilteredCartesianRDD.scala

## To Reproduce

- Package Geotrellis and dependencies in a shaded jar
- Make sure Spark and Geotrellis are in separate jars.
- In META-INF/MANIFEST.MF of geotrellis jar set Sealed to true

## Expected behavior

- It should work with sealing enabled
- Avoid use of Spark private API to be future proof

## Proposed solution

Move FilteredCartesianRDD and possible depedencies into Geotrellis specific package.

## Environment

- Java version: 21+
- Scala version: 12
- GeoTrellis version: 3.8

jdries:
https://github.com/locationtech/geotrellis/issues/3585

# Overview

Removes use of Spark private API. This avoids issues when sealing is enabled.
This PR simply imports code that was used from Spark, without trying to modify or optimize it further.

## Checklist

Closes #3585

### DIFF
--- a/spark/src/main/scala/geotrellis/spark/join/FilteredCartesianRDD.scala
+++ b/spark/src/main/scala/geotrellis/spark/join/FilteredCartesianRDD.scala
@@ -19,12 +19,51 @@
   *
   * 1. https://github.com/apache/spark/blob/2f8776ccad532fbed17381ff97d302007918b8d8/core/src/main/scala/org/apache/spark/rdd/CartesianRDD.scala
   */
-package org.apache.spark.rdd
-
-
-import scala.reflect.ClassTag
+package geotrellis.spark.join
 
 import org.apache.spark._
+import org.apache.spark.rdd.RDD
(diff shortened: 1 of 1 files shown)


pomadchin:
Will take a look once I have time and we'll get it merged in!

pomadchin:
Thanks for a fix! I'll update the CHANGELOG separately!

PR #3590, opened 2025-09-21
MinResample should ignore Int NODATA values

### LINKED ISSUE #3589: MinResample returns noDataValue
## Describe the bug
When resampling an IntArrayTile with method Min, it will return the noDataValue if it is in the array. The value is computed [here](https://github.com/locationtech/geotrellis/blob/ee576022a6cd37702dc60d2035662afc5b6b1ac8/raster/src/main/scala/geotrellis/raster/resample/MinResample.scala#L33) and the reason it returns the noDataValues is because `get` will return `-2147483648` if it is the noDataValue.

## To Reproduce

```
val intArray = IntArrayTile(Array(2,4,256,16), 2, 2, noDataValue = 256)
val nodata = intArray.get(0,1)                                          // -2147483648
val resampled = intArray.resample(extent, 1, 1, method = Min)
resampled.toArray                                                       // contains 256, expect 2 as min
```

## Expected behavior

The noDataValue should be ignored and the min is than computed on the remaining values

## Environment

- Java version: 21+
- Scala version: 12
- GeoTrellis version: 3.8

pomadchin:
# Overview

This PR addresses the incorrect MinResampling for the non floating point Tiles.

Closes #3589

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Do not depend on private Spark API, avoids sealing violation [#3586](https://github.com/locationtech/geotrellis/pull/3586)
 - Add predictor 2 (integer) and predictor 3 (float) support for writing compressed GTiff files [#3588](https://github.com/locationtech/geotrellis/pull/3588)
 
+### Changed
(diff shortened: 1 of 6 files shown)


pomadchin:
@ElienVandermaesenVITO mind taking a look and trying it out?

PR #3616, opened 2026-07-20
fix BigTiff offset overflow

bossie:
# Overview

Fixes offset overflow in full res BigTiffs. Manifests itself when the first overview already tips the image size over the 2^32 threshold.

## Checklist

### DIFF
--- a/raster/src/main/scala/geotrellis/raster/io/geotiff/writer/GeoTiffWriter.scala
+++ b/raster/src/main/scala/geotrellis/raster/io/geotiff/writer/GeoTiffWriter.scala
@@ -93,7 +93,7 @@ class GeoTiffWriter(geoTiff: GeoTiffData, dos: DataOutputStream) {
     val (fieldValues, offsetFieldValueBuilder) = TiffTagFieldValue.collect(geoTiff)
     val segments = geoTiff.imageData.segmentBytes
     val segmentCount = segments.size
-    val segmentBytesCount = (0 until segmentCount).map(segments.getSegmentByteCount).foldLeft(0L) {_ + _}
+    val segmentBytesCount = (0 until segmentCount).map(segments.getSegmentByteCount).foldLeft(0L)(_ + _)
 
     // footprint of tags themselves
     val tagFieldSize = if (isBigTiff) 20 else 12
@@ -212,7 +212,7 @@ class GeoTiffWriter(geoTiff: GeoTiffData, dos: DataOutputStream) {
       val (fieldValues, offsetFieldValueBuilder) = TiffTagFieldValue.collect(geoTiff)
       val segments = geoTiff.imageData.segmentBytes
       val segmentCount = segments.size
-      val segmentBytesCount = (0 until segmentCount).map(segments.getSegmentByteCount).foldLeft(0L) {_ + _}
+      val segmentBytesCount = (0 until segmentCount).map(segments.getSegmentByteCount).foldLeft(0L)(_ + _)
 
       val tagFieldSize = if (isBigTiff) 20 else 12
       val tagFieldByteCount = (fieldValues.length + 1) * tagFieldSize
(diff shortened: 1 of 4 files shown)


raster/src/test/scala/geotrellis/raster/io/geotiff/BigTiffSpec.scala
```diff
@@ -141,5 +144,25 @@ class BigTiffSpec extends AnyFunSpec with RasterMatchers with GeoTiffTestUtils w
         actual.getOverviewsCount should be (5)
       }
     }
+
+    it("should handle offsets greater than 2^32 without overflowing") {
```
pomadchin, line 151: Unfoturnately this won't get through the CI (OOO); I'm fine with ignoring this one for now to make sure CI is happy; or if there is a better way to test the changes that'd be great ofc.

raster/src/main/scala/geotrellis/raster/io/geotiff/writer/GeoTiffWriter.scala
```diff
@@ -93,7 +93,7 @@ class GeoTiffWriter(geoTiff: GeoTiffData, dos: DataOutputStream) {
     val (fieldValues, offsetFieldValueBuilder) = TiffTagFieldValue.collect(geoTiff)
     val segments = geoTiff.imageData.segmentBytes
     val segmentCount = segments.size
-    val segmentBytesCount = (0 until segmentCount).map(segments.getSegmentByteCount).sum
+    val segmentBytesCount = (0 until segmentCount).map(segments.getSegmentByteCount).foldLeft(0L) {_ + _}
```
pomadchin: minor: `{_ + _}` => `(_ + _)`

raster/src/main/scala/geotrellis/raster/io/geotiff/writer/GeoTiffWriter.scala
```diff
@@ -212,7 +212,7 @@ class GeoTiffWriter(geoTiff: GeoTiffData, dos: DataOutputStream) {
       val (fieldValues, offsetFieldValueBuilder) = TiffTagFieldValue.collect(geoTiff)
       val segments = geoTiff.imageData.segmentBytes
       val segmentCount = segments.size
-      val segmentBytesCount = (0 until segmentCount).map(segments.getSegmentByteCount).sum
+      val segmentBytesCount = (0 until segmentCount).map(segments.getSegmentByteCount).foldLeft(0L) {_ + _}
```
pomadchin: minor: `{_ + _}` => `(_ + _)`

pomadchin (commented):
LGTM; I'll take another look ofc.

But also the CI won't be happy unfortuantely due to OOO errs.

raster/src/test/scala/geotrellis/raster/io/geotiff/BigTiffSpec.scala
bossie: Right, having a tile this large in memory is undesirable. I'm sure it can be improved, let me take a look.

As for the actual test: I currently do this on my machine by invoking `tiffinfo -s` on the generated BigTiff and taking a look at the actual tile offsets. The reason I do it this way is because:

- instantiating a `geotrellis.raster.io.geotiff.SinglebandGeoTiff` from it throws a `java.nio.BufferUnderflowException`;
- unless I'm mistaken, I don't think Geotrellis even exposes the actual offsets.

That being said, invoking something "independent" like `tiffinfo` to check things could actually be an advantage but this obviously needs to be available in the build container. Is it feasible to to install an additional package? I think it came with package `libtiff-tools`.

pomadchin: 🤔  hmmm, or to make it easier for you we can just make this test 'ignored' for now and that's probably better; to not introduce this complex dep on the tiff info

bossie: Well, TBH at this point the heavy lifting has been done and now it's just a matter of installing a package in the test image. I favored `tiffdump` over `tiffinfo` because its output was easier to parse; it is also in the `libtiff-tools` package.

pomadchin (approved):
:+1: but I'll comment out the test; images rebuild is needed as well.

Thank you so much!

pomadchin:
That is a big improvement, thank you! 🎉

